PR #63, opened 2026-02-18
feat: Hibernation MVP — Claude exits on idle, bot wakes on next message

sayhar:
## Summary

Implements a two-state hibernation system (active ↔ hibernate) so the Claude Code session can exit gracefully after a configurable idle period instead of running forever. The bot (`lobster-router.service`) detects the next incoming Telegram message and restarts Claude automatically.

- **State file** (`~/messages/config/lobster-state.json`) tracks current mode: `"active"` or `"hibernate"`
- **`wait_for_messages()`** gains a `hibernate_on_timeout` parameter — when the timeout fires with no messages, it writes `state=hibernate` and returns a signal for Claude to exit
- **`lobster_bot.py`** wakes Claude on any incoming message when state is `"hibernate"` (thread-safe, race-condition-safe)
- **`health-check-v3.sh`** reads the state file and skips restart when state is `"hibernate"` (intentional exit, not a crash)
- **`CLAUDE.md`** documents the hibernate-aware main loop pattern

## Files changed

| File | Why |
|------|-----|
| `src/mcp/inbox_server.py` | Added `LOBSTER_STATE_FILE`, `_read_lobster_state()`, `_write_lobster_state()`, and `hibernate_on_timeout` parameter to `wait_for_messages()` |
| `src/bot/lobster_bot.py` | Added `_read_lobster_state()`, `_is_claude_running()`, `wake_claude_if_hibernating()` with threading lock; hooked into `handle_message()` and `handle_callback_query()` |
| `scripts/health-check-v3.sh` | Added `read_lobster_mode()`, `is_hibernating()`, and hibernate guard in `main()` that exits 0 without restarting when `state=hibernate` |
| `CLAUDE.md` | Added Hibernation section with loop example, state file docs, and updated Startup Behavior |
| `tests/test-hibernation.sh` | 14 shell tests: state file creation/transitions/atomicity, missing/corrupt defaults, health check awareness |
| `tests/unit/test_hibernation.py` | 20 Python tests: state helpers, wait_for_messages hibernation, bot wake logic (including race condition), health check awareness |

## Design decisions

- **`hibernate_on_timeout` defaults to `False`**: backward compatible — existing sessions that don't opt in continue forever as before
- **Atomic state writes**: write-to-temp-then-rename, so readers never see a partial file
- **Default to `"active"` on any read error**: missing file, corrupt JSON, or unknown mode — all treated as active so a broken state file never prevents recovery
- **Threading lock in bot**: five concurrent messages while hibernating produce exactly one Claude spawn
- **Systemd-first wake**: bot calls `sudo systemctl restart lobster-claude`; falls back to `start-lobster.sh` if systemctl fails
- **Health check `LOBSTER_STATE_FILE_OVERRIDE` and `LOBSTER_HEALTH_CHECK_DRY_RUN`**: enable testing without real systemd/tmux

## Test plan

- [x] Run `bash tests/test-hibernation.sh` — 14/14 pass
- [x] Run `uv run pytest tests/unit/test_hibernation.py -v` — 20/20 pass
- [x] Run `uv run pytest tests/unit/test_mcp_server/test_wait_for_messages.py -v` — 5/5 pass (no regressions)
- [ ] Manual: set `timeout=120, hibernate_on_timeout=True` in Claude session, wait 2 minutes, verify state file written and Claude exits
- [ ] Manual: send Telegram message while hibernating, verify bot restarts Claude within a few seconds

### DIFF
--- a/tests/unit/test_hibernation.py
+++ b/tests/unit/test_hibernation.py
@@ -111,6 +111,40 @@ class TestStateFile:
         assert data["mode"] == "hibernate"
         assert "updated_at" in data
 
+    def test_reset_state_on_startup(self, state_dir: Path):
+        """_reset_state_on_startup resets 'hibernate' to 'active' and adds woke_at."""
+        sys.path.insert(0, str(Path(__file__).parent.parent.parent / "src" / "mcp"))
+        from inbox_server import _reset_state_on_startup, LOBSTER_STATE_FILE
+
+        # Write hibernate state
+        state_file = state_dir / "lobster-state.json"
+        _write_state(state_dir, "hibernate")
+        assert json.loads(state_file.read_text())["mode"] == "hibernate"
+
+        # Patch the global and call the function
+        with patch("inbox_server.LOBSTER_STATE_FILE", state_file):
+            _reset_state_on_startup()
+
+        # Verify state was reset
+        data = json.loads(state_file.read_text())
+        assert data["mode"] == "active", f"Expected 'active' after startup reset, got {data['mode']}"
+        assert "woke_at" in data, "Expected 'woke_at' timestamp after reset"
+
+    def test_reset_state_on_startup_noop_when_active(self, state_dir: Path):
+        """_reset_state_on_startup does nothing when state is already 'active'."""
+        sys.path.insert(0, str(Path(__file__).parent.parent.parent / "src" / "mcp"))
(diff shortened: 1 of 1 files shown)


aeschylus:
Superseded by #68 which cherry-picks the hibernation system from this PR (and the stale-zombie detection from #66) into a single consolidated infrastructure hardening PR.

PR #64, opened 2026-02-19
refactor: drop Anthropic SDK dependency from nightly consolidation

sayhar:
## Summary

- Replaces `import anthropic` / `client.messages.create()` in `call_claude_api()` with a `subprocess.run()` call to the Claude Code CLI (`claude --print -p ...`)
- Removes the `anthropic` Python SDK as a runtime dependency for nightly consolidation
- Aligns with the existing invocation pattern used by `claude-wrapper.sh`

## Details

The `call_claude_api` function in `scripts/nightly-consolidation.py` was the only place that imported the Anthropic SDK. This refactor replaces it with a subprocess call to the `claude` CLI, passing `--print`, `--model`, `--max-turns 1`, and `-p` flags. A 5-minute timeout is set per invocation.

All 63 existing consolidation tests pass -- they mock `call_claude_api` at the module level so the change is transparent to them.

## Test plan

- [x] `uv run python -c "import ast; ast.parse(...)"` -- syntax OK
- [x] `uv run python -m pytest tests/ -k consolidat -v` -- 61 passed, 0 failed (2 unrelated failures in test_memory.py)
- [ ] Manual: run `nightly-consolidation.py --dry-run` on a system with `claude` CLI installed

### DIFF
--- a/scripts/nightly-consolidation.py
+++ b/scripts/nightly-consolidation.py
@@ -3,8 +3,8 @@
 Nightly Consolidation Script -- Layer 2 of Lobster's Three-Layer Memory System
 
 Synthesizes raw memory events from the SQLite event store into canonical
-markdown files using Claude API calls. Runs as a standalone script invoked
-by cron (via nightly-consolidation.sh) or directly.
+markdown files using Claude Code CLI calls. Runs as a standalone script
+invoked by cron (via nightly-consolidation.sh) or directly.
 
 Design principles:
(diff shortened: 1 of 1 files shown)


aeschylus:
Superseded by #68 which includes the Anthropic SDK removal (nightly consolidation now uses `claude --print` subprocess) as part of the consolidated infrastructure hardening PR.

PR #67, opened 2026-02-19
fix: add subagent guard to self-check scripts

sayhar:
## Summary
- Adds a "subagent running" guard to both self-check scripts (`periodic-self-check.sh` and `self-check-reminder.sh`)
- If only the main Claude process is running (no subagents), the self-check is skipped
- This allows `wait_for_messages` idle timeout to expire, enabling hibernation

## Problem
Self-checks were firing even when no subagents were running, preventing the idle timeout from ever expiring and blocking hibernation.

## Test plan
- [ ] Verify `pgrep -c -f "claude"` correctly counts Claude processes
- [ ] Verify self-checks stop when no subagents are active
- [ ] Verify self-checks still fire when a subagent is running
- [ ] End-to-end hibernation test: idle timeout expires → hibernate → wake on message

### DIFF
--- a/scripts/periodic-self-check.sh
+++ b/scripts/periodic-self-check.sh
@@ -14,6 +14,7 @@
 #   2. Only fires if there isn't already a self-check in the inbox (no spam)
 #   3. Rate-limited: won't inject if last self-check was < 2 minutes ago
 #   4. Max inbox depth: won't inject if inbox already has 20+ messages (backpressure)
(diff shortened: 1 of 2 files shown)


aeschylus:
Cherry-picked into #68 (commit 7d12629). The subagent guard is now part of the consolidated infrastructure PR. Closing as superseded.

PR #212, opened 2026-03-13
Add forward field to write_result to suppress dispatcher re-delivery

sayhar:
## Summary

- **`src/mcp/inbox_server.py`**: Add optional `forward` boolean param (default `True`) to the `write_result` tool. The param is written into the queued `subagent_result` message JSON as `"forward": bool(forward)`.
- **`.claude/dispatcher.md`**: Update subagent result handling — if `msg.get("forward") == False`, dispatcher marks processed silently without calling `send_reply`. Errors always forward regardless.
- **`.claude/subagent.md`**: Document the `forward=False` pattern with a code example in both the primer paragraph and the Subagent Rules section.

## Motivation

Reduces context token burn. Subagents that call `send_reply` directly (e.g. this agent, which called `send_reply` in its task prompt) previously had the dispatcher re-forward the result text anyway. With `forward=False`, the subagent controls whether the dispatcher re-delivers, avoiding duplicate messages.

## Behavior

| Scenario | Dispatcher action |
|----------|------------------|
| `write_result(..., forward=True)` (default) | Calls `send_reply` with `msg["text"]` |
| `write_result(..., forward=False)` | Marks processed silently — no `send_reply` |
| `subagent_error` (any `forward` value) | Always calls `send_reply` with error message |

## Test plan

- [ ] Verify that a subagent calling `write_result(forward=True)` still gets delivered by dispatcher (backward-compatible default)
- [ ] Verify that a subagent calling `write_result(forward=False)` results in the dispatcher marking processed without sending a reply
- [ ] Verify `subagent_error` messages always forward to the user regardless of `forward` field

### DIFF
--- a/.claude/dispatcher.md
+++ b/.claude/dispatcher.md
@@ -61,19 +61,25 @@ You are a **stateless dispatcher**. Your ONLY job on the main thread is to read 
 
 ## Handling Subagent Results (`subagent_result` / `subagent_error`)
 
-Background subagents **must not call `send_reply` directly**. Instead they call `write_result(task_id, chat_id, text, ...)`, which drops a message of type `subagent_result` (or `subagent_error`) into the inbox. The main thread picks it up and delivers it.
+Background subagents call `write_result(task_id, chat_id, text, ...)`, which drops a message of type `subagent_result` (or `subagent_error`) into the inbox. The main thread picks it up.
 
 **When `wait_for_messages` returns a message with `type: "subagent_result"`:**
 
+Check the `forward` field first:
+
 ```
 1. mark_processing(message_id)
-2. send_reply(
-       chat_id=msg["chat_id"],
-       text=msg["text"],
-       source=msg.get("source", "telegram"),
-       thread_ts=msg.get("thread_ts")   # pass through if present
-   )
-3. mark_processed(message_id)
+2. if msg.get("forward") == False:
+       # Subagent already called send_reply — nothing to deliver
+       mark_processed(message_id)
+   else:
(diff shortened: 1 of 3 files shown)


sayhar (commented):
Code review of PR #212 — forward field for write_result

## Overview

Clean, minimal change. The intent is clear and the implementation is correct.

## inbox_server.py

The schema addition looks good:
- `"default": True` is correct (JSON boolean, not the string `"true"`)
- The description accurately describes the semantics

In the handler, `forward = args.get("forward", True)` is correct. The `bool(forward)` cast when writing the message is a good defensive touch — guards against a caller passing a truthy non-boolean.

One minor observation: the tool description still says "The main thread will pick this up... and send the text to the user" without mentioning that `forward=False` suppresses delivery. Mildly misleading for callers reading only the tool description. Not a blocker, but worth a follow-up doc fix.

## .claude/dispatcher.md

The pseudocode correctly uses `msg.get("forward") == False` rather than `not msg.get("forward")` — this matters because `msg.get("forward")` returns `None` for old messages that predate this field, and `None == False` is `False` in Python, so old messages still forward correctly. Backward compatibility is preserved without any explicit migration.

The note "(Errors always forward — a subagent that fails may not have delivered anything to the user.)" is a good explicit rationale for the asymmetric behavior.

## .claude/subagent.md

Both the primer paragraph update and the new `forward=False` block are clear. The code example is well-annotated with comments explaining intent. Good.

## Verdict

LGTM. The change is safe, backward-compatible, and well-documented across all three files. Ready to merge.

PR #237, opened 2026-03-13
fix: post-compact-enforcement robustness fixes (7 issues from Opus review)

sayhar:
Follows up on #223. Addresses all P1 and P2 issues identified in the post-merge Opus robustness review of the post-compact-enforcement feature.

## Summary

- **Fix 1** (`on-compact.py`): Write sentinel before reminder; wrap `write_reminder()` in try/except so reminder failure cannot prevent the critical sentinel from being written
- **Fix 2** (`post-compact-gate.py`): Replace env-var-only dispatcher check with a `/proc` process-tree walk that correctly distinguishes dispatcher from subagents (subagents inherit `LOBSTER_MAIN_SESSION=1` from the dispatcher, causing false positives with the old check)
- **Fix 3 & 4** (`dispatcher.bootup.md`): Add "Handling Post-Compact Gate Denial" and "System Messages" sections so the dispatcher knows exactly what to do when gated or when receiving system messages
- **Fix 5** (`post-compact-gate.py`): Add structured gate logging to `~/lobster-workspace/logs/compact-gate.log`
- **Fix 6** (`post-compact-gate.py`): Improve deny message — now says "GATE BLOCKED" and gives explicit next-action instructions
- **Fix 7** (`post-compact-gate.py`): Guard `sentinel_is_fresh()` against negative age from clock skew (`0 <= age < TTL`)

## Fix 2 detail — dispatcher vs subagent detection

The root issue: subagents spawned by the Task tool inherit the dispatcher's environment, including `LOBSTER_MAIN_SESSION=1`. The tmux ancestry check used in `inbox_server.py` also returns True for subagents (they are descendants of the same tmux pane via the dispatcher).

The correct distinguishing signal is the **depth** of Claude process ancestry:

```
dispatcher hook: hook.py → claude (dispatcher) → tmux pane
subagent hook:   hook.py → claude (subagent) → claude (dispatcher) → tmux pane
```

The fix walks `/proc/<pid>/stat` upward, counting consecutive `claude`-named ancestors before reaching a tmux pane PID. Two or more Claude ancestors = subagent = gate is skipped. Falls back to env-var-only when tmux is unavailable.

## dispatcher.bootup.md note

`~/lobster-workspace/.claude/dispatcher.bootup.md` (the live runtime copy, not tracked in git) was also updated in-place with the same Fix 3 & 4 additions. This file is a user config file that shadows the repo copy at `.claude/dispatcher.bootup.md`.

## Test plan

- [ ] Verify `on-compact.py` writes sentinel before reminder (trace call order in main())
- [ ] Manually test gate with a subagent: confirm subagent tool calls are not blocked post-compact
- [ ] Confirm dispatcher IS blocked post-compact and unblocked by `wait_for_messages()`
- [ ] Check `~/lobster-workspace/logs/compact-gate.log` after a compact event — should see `blocked` and `cleared` entries
- [ ] Verify deny message contains "GATE BLOCKED" text
- [ ] Simulate clock skew: manually set sentinel mtime to future, confirm gate passes

Closes (follow-up to) #223

### DIFF
--- a/hooks/post-compact-gate.py
+++ b/hooks/post-compact-gate.py
@@ -70,7 +70,7 @@ LOG_FILE = Path(os.path.expanduser("~/lobster-workspace/logs/compact-gate.log"))
 
 WAIT_FOR_MESSAGES_TOOL = "mcp__lobster-inbox__wait_for_messages"
 
-CONFIRMATION_TOKEN = "LOBSTER_COMPACTED_REORIENTED"
+CONFIRMATION_TOKEN = "LOBSTER_COMPACTED_REORIENTED"  # noqa: S105 — not a secret, intentional safe word
 
 DENY_REASON_NEEDS_TOKEN = (
     "GATE BLOCKED: Context compaction was just detected. "


sayhar (commented):
## Opus Code Review — PR #237 `fix/post-compact-robustness`

Overall verdict: **APPROVE with minor notes**

The 7 fixes are all correct in their intent. The implementation is solid, the fallback strategy is sensible, and the defensive posture (silent failures everywhere, fail-open on uncertainty) is exactly right for a PreToolUse hook that runs on every tool call. A few edge cases worth noting, none of them blocking.

---

### Fix 1 — write_sentinel before write_reminder (on-compact.py)

**Correct.** The reorder is the right fix. Previously, if `write_reminder()` threw, the sentinel was never written and the gate would never block — meaning the dispatcher could proceed without re-orienting after a compaction. Now the sentinel goes down first, guaranteeing the gate is armed regardless of what happens to the inbox reminder.

The `try/except Exception` wrapping `write_reminder()` with a pass is appropriate — the comment "Reminder failure is non-fatal — sentinel is the critical artifact" is accurate and the inline comment explains the decision clearly.

**One nuance:** In the `already_pending()` branch, `write_reminder()` is intentionally skipped (inbox already has one), but `write_sentinel()` is still called — that's correct and unchanged. Good.

**No issues.**

---

### Fix 2 — /proc process-tree walk replacing env-var-only check (post-compact-gate.py)

This is the most complex change. The logic is sound but has a few edge cases worth flagging.

#### Correctness of the core algorithm

The "count consecutive claude ancestors before reaching a tmux pane PID" approach works for the expected topology:

- Dispatcher: `hook.py → claude (dispatcher) → tmux pane` → `claude_ancestor_count = 1` → `1 <= 1` → True (dispatcher)
- Subagent: `hook.py → claude (subagent) → claude (dispatcher) → tmux pane` → `claude_ancestor_count = 2` → `2 <= 1` → False (not dispatcher)

**Issue: counting is non-consecutive**

The code counts `claude_ancestor_count` by checking each parent as it walks upward, but it does not reset the counter when a non-claude process is interleaved. If the actual process tree has something like:

```
hook.py → claude → sh → claude → tmux
```

(e.g. Claude spawns a shell which spawns another Claude-like process), the count would reach 2 and incorrectly classify a dispatcher as a subagent. In practice, this tree shape is unlikely for the Lobster architecture, but the docstring says "consecutive claude ancestors" without enforcing consecutiveness in code. Minor discrepancy between docs and implementation — the count logic works for the expected topology, just isn't strictly consecutive.

#### What happens on non-Linux systems (/proc missing)

`_get_proc_name` and `_get_ppid` both catch `OSError` and return `""` / `None` respectively. With no `/proc`, `_get_ppid` returns `None` on the first iteration, the loop breaks, and execution falls through to `return True` (env-var fallback). Correct behavior.

#### tmux `-L` flag usage

```python
["tmux", "-L", LOBSTER_TMUX_SESSION, "list-panes", "-t", LOBSTER_TMUX_SESSION, ...]
```

The `-L` flag specifies the tmux **socket name**, and `-t` specifies the target session. These are two different things — `-L lobster` sets the socket, `-t lobster` targets a session named "lobster". If the socket name and session name happen to be the same string (the default Lobster setup), this works. But if someone uses a different socket name vs session name, this could fail silently. The fallback is correct (returns `set()`), but worth a comment clarifying the `-L`/`-t` semantics. Low risk in practice.

#### Performance

The hook runs on **every tool call**. The process tree walk does:
- 1 subprocess call to tmux (up to 3s timeout, but usually <10ms)
- Up to 15 `/proc/<pid>/stat` reads

The tmux call is the main concern. Even if tmux responds in 5ms, that's 5ms of latency on every tool call across both dispatcher and subagents. This is probably acceptable but worth measuring if anyone notices tool-call latency increasing.

The timeout is set to 3 seconds, which means in a worst case (tmux hung, tmux not responding) the gate adds a 3-second delay to every tool call until the subprocess times out. Consider reducing to 1 second, since this is a fast local call.

#### Fallback correctness

If `_get_tmux_pane_pids()` returns empty (tmux down, wrong socket, etc.), the code falls back to `return True` (env-var says it's the dispatcher, so we trust it). This is the stated behavior and it's safe — it's no worse than the old code.

---

### Fixes 3 & 4 — dispatcher.bootup.md additions

Both new sections are accurate and actionable.

**Gate denial section:** The three bullet points correctly describe the intended response to a gate block. The phrasing "Your only permitted next action is: call wait_for_messages()" is strong and clear — good.

**System messages section:** The guidance to `mark_processed` without `send_reply` for `chat_id: 0` messages is correct. The compact-reminder handling is described accurately.

**Placement:** Both sections are inserted before "The 7-Second Rule", which makes sense — they cover behavior needed at loop startup, before the 7-second rule kicks in.

**No contradictions** with the rest of the file.

---

### Fix 5 — Gate logging to compact-gate.log

```python
def log_gate_event(tool_name: str, action: str) -> None:
    try:
        LOG_FILE.parent.mkdir(parents=True, exist_ok=True)
        ts = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
        line = json.dumps({"ts": ts, "tool": tool_name, "action": action}) + "\n"
        with LOG_FILE.open("a") as f:
            f.write(line)
    except Exception:
        pass
```

**Concurrent access:** The log is opened in append mode (`"a"`), which is safe on Linux (appends are atomic for small writes on most filesystems). Multiple hook processes writing simultaneously won't corrupt the file. No lock needed at this granularity.

**Log path:** `~/lobster-workspace/logs/compact-gate.log` — consistent with other log locations. The `mkdir(parents=True, exist_ok=True)` ensures the directory is created if missing. Correct.

**Log rotation:** No rotation mechanism. This log will grow unboundedly. Not a problem short-term, but worth adding to a future cleanup job.

**Logging location in main():** `log_gate_event` is called for "cleared" (wait_for_messages allowed through) and "blocked" (tool denied). The stale-sentinel pass-through (line `if not sentinel_is_fresh(): sys.exit(0)`) is NOT logged. This means you can't easily tell from the log when the gate stopped blocking because the sentinel went stale. Minor observability gap — acceptable for now.

---

### Fix 6 — Improved deny message

Old:
> "Context was just compacted. You must call wait_for_messages() first..."

New:
> "GATE BLOCKED: Context compaction was just detected. Your only permitted action right now is to call wait_for_messages()..."

The new message is substantially better. The "GATE BLOCKED" prefix makes it pattern-matchable by the dispatcher (the bootup.md section says to look for "GATE BLOCKED"). The explicit instruction to not retry the blocked tool and the explanation of what wait_for_messages will return gives the model exactly what it needs to self-correct.

**No issues.**

---

### Fix 7 — mtime guard (0 <= age < TTL)

```python
return 0 <= age < SENTINEL_TTL_SECONDS
```

Correct. The old `age < SENTINEL_TTL_SECONDS` would return `True` for negative ages (clock skew, NTP jump backward), treating a future-dated sentinel as fresh — which is correct behavior but the intent was unclear. The new guard makes the intent explicit: only treat as fresh when age is non-negative and within TTL. Zero age (just written) is correctly included.

**No issues.**

---

### Summary

| Fix | Verdict | Notes |
|-----|---------|-------|
| 1. write_sentinel before write_reminder | ✅ Correct | Order fix is essential; try/except is appropriate |
| 2. /proc process-tree walk | ✅ Correct with caveats | Non-consecutive count (benign in practice); tmux timeout could be 1s not 3s |
| 3. bootup.md gate denial section | ✅ Correct | Clear and actionable |
| 4. bootup.md system messages section | ✅ Correct | Accurate, no contradictions |
| 5. Gate logging | ✅ Correct | No rotation; stale-pass not logged (minor) |
| 6. Deny message | ✅ Better | Pattern-matchable, instructive |
| 7. mtime guard | ✅ Correct | Exactly right |

**Recommended follow-ups (not blocking):**
1. Reduce tmux subprocess timeout from 3s to 1s in `_get_tmux_pane_pids()`
2. Add a comment clarifying `-L` (socket) vs `-t` (session) in the tmux call
3. Consider logging the stale-sentinel pass-through for observability
4. Add compact-gate.log to a log rotation job

**APPROVE.**

sayhar:
**New commits added** — addresses two re-orientation improvement issues filed today:

**Fix 8 — closes #239** (`hooks/on-compact.py`): REMINDER_TEXT now opens with an inline identity cue:
> "You are Lobster, the always-on dispatcher. Your role has not changed."

Followed by explicit absolute paths to re-read (`~/lobster-workspace/.claude/dispatcher.bootup.md` and `~/lobster-user-config/memory/canonical/handoff.md`). A confused post-compact dispatcher can now re-orient even in a degraded state, without relying on recognizing which files to look for.

**Fix 9 — closes #240** (`.claude/dispatcher.bootup.md`): Restructured so the first ~60 lines contain: (1) "Who You Are" identity paragraph, (2) main loop pseudocode, (3) 7-second rule. All existing content is preserved; secondary sections (gate denial, subagent handling, message sources, etc.) follow after. Reading the top of the file is now sufficient to re-orient.

The runtime copy at `~/lobster-workspace/.claude/dispatcher.bootup.md` was also updated in-place.

sayhar:
## Post-Compact Gate Robustness — Code Review

I reviewed the current PR branch against the three proposed fixes. Here's my assessment.

---

### What's already in this PR (vs. what was proposed)

The PR implements significant work not directly described in the proposals:

- **Process-tree subagent detection** — replaces the env-var-only `is_main_session()` with `is_dispatcher_session()`, which walks `/proc` ancestry to count `claude` processes before hitting a tmux pane PID. This is the real fix for the env-var inheritance problem where subagents incorrectly got blocked.
- **Gate logging** — appends JSON events to `compact-gate.log` on block/clear.
- **Clock skew guard** in `sentinel_is_fresh()` — `0 <= age < TTL` instead of just `age < TTL`.
- **Write order fix** in `on-compact.py` — sentinel is now written before the reminder message, so the gate is up before the reminder hits the inbox. Reminder failure is now non-fatal (wrapped in try/except).
- **Restructured `dispatcher.bootup.md`** — much better organisation, role identity up front, "Handling Post-Compact Gate Denial" section added.
- **Improved `DENY_REASON`** — says "GATE BLOCKED" (matches bootup guidance), instructs calling `wait_for_messages()` directly.

---

### Fix 1: Direct-call instruction in denial message

**Partially addressed — still missing the key detail.**

The new `DENY_REASON` says to "call `wait_for_messages()`" but doesn't say to use the full MCP name `mcp__lobster-inbox__wait_for_messages`. The failure mode was exactly that the dispatcher tried to call ToolSearch to get the schema. The denial message is the right place to shut that down:

> "Call `mcp__lobster-inbox__wait_for_messages` by its full name directly — no ToolSearch needed, the schema is pre-registered."

The dispatcher.bootup.md "Handling Post-Compact Gate Denial" section also says "call wait_for_messages()" without the full name. Both should use the full MCP name to eliminate any ambiguity.

**Recommendation: add the full tool name to both DENY_REASON and the bootup section. One sentence is sufficient.**

---

### Fix 2: Safe word / confirmation token

**Not implemented.** The proposal was:
- Denial message says to read bootup.md and then call `wait_for_messages(confirmation='<token>')`.
- Gate checks the token; without it, wait_for_messages itself is denied (forces bootup read).
- Token lives in bootup.md.

**My assessment: this is worth doing but has non-trivial edge cases:**

1. **Token storage** — hardcoded in `post-compact-gate.py` is simplest and acceptable since this is a behavioral nudge, not a security boundary.

2. **Schema compatibility** — `wait_for_messages` must accept a `confirmation` kwarg without breaking non-compact callers. Clean: if no sentinel is active, the token check is irrelevant.

3. **Token value** — something like `LOBSTER_COMPACTED_REORIENTED` works well. It's memorable, stable, clearly machine-checkpoint-ish, and self-documenting. Avoid short random strings (not memorable) and long UUIDs (hard to read in logs).

4. **Failure mode if Claude never reads bootup.md** — it gets stuck. That's the intended behavior. The existing 10-minute TTL is the escape hatch.

**The safe word approach is sound. The main risk is implementation complexity — it touches the gate, the MCP tool signature, and the bootup file. Should be a follow-up issue rather than blocking this PR.**

---

### Fix 3: ToolSearch whitelist — skip justified

Agreed. With Fix 1 implemented (full tool name in denial), the dispatcher has enough information to call `wait_for_messages` directly. Whitelisting ToolSearch would also allow it to fetch *other* deferred tool schemas, defeating the point of the gate. Skip is correct.

---

### Path format: `~` vs absolute

The PR uses `~/lobster-workspace/.claude/dispatcher.bootup.md` in the reminder text. The `~` is a string handed to Claude (not a shell command), so expansion happens mentally. Claude handles this fine, and using `~` avoids hardcoding the username. Current approach is correct.

**One formatting issue found in REMINDER_TEXT:**

```python
"1. ~/lobster-workspace/.claude/dispatcher.bootup.md"
"  \u2190 dispatcher instructions, main loop, 7-second rule\n"
```

These two adjacent string literals are concatenated without a `\n` between them, producing a single long line. If the intent was a two-line list item (path on line 1, description indented on line 2), a `\n` is missing. Same for item 2. If single-line was intentional, it works — but the indentation of the arrow comment reads oddly when run together.

---

### Process-tree subagent detection

The `is_dispatcher_session()` logic correctly handles the env-var inheritance problem. The `_get_ppid()` parser handles process names with spaces/parens by using `rsplit(")", 1)`. The depth limit of 15 is generous. Nested subagents (subagent-of-subagent) correctly have 3+ claude ancestors and are blocked as expected. Falls back safely when tmux is unavailable.

---

### Write order fix in on-compact.py

Sentinel is now written before the reminder — correct. If reminder writing fails, the gate is still up. `write_sentinel()` already catches all exceptions internally, so the outer try/except around `write_reminder()` is the right additional hardening.

---

### Summary verdict: REQUEST CHANGES (minor)

The PR is solid and addresses real problems. Two changes before merge:

1. **Add full MCP tool name to DENY_REASON and dispatcher.bootup.md gate-denial section**: use `mcp__lobster-inbox__wait_for_messages` explicitly — this is the specific fix for the ToolSearch loop that prompted this work.

2. **Fix REMINDER_TEXT string concatenation**: add `\n` between the path and arrow-comment strings for items 1 and 2, if two-line list items were intended.

The safe word / confirmation token (Fix 2) should be a separate issue — good idea, but non-trivial enough not to block this PR. This branch is a strong foundation.

---

*Reviewed files:* `hooks/post-compact-gate.py`, `hooks/on-compact.py`, `.claude/dispatcher.bootup.md`

PR #245, opened 2026-03-14
docs: fix subagent result delivery — send_reply first, then write_result(forward=False)

sayhar:
## Summary

- **Bug found:** `functional-engineer.md` explicitly said "Never call `send_reply` directly" — this directly contradicts the intended crash-safe design.
- **Gap found:** `subagent.bootup.md` and `lobster-generalist.md` only mentioned `forward=False` as a footnote; neither clearly instructed subagents to call `send_reply` first.
- **Evidence from processed messages:** All recent `subagent_result` files in `~/messages/processed/` have `forward: null` — confirming subagents aren't using the two-step pattern in practice.

## What the correct pattern is

1. Call `send_reply(chat_id, text)` directly — user gets the reply immediately, regardless of dispatcher state
2. Call `write_result(..., forward=False)` — dispatcher marks message processed without re-delivering

This is crash-safe: if the dispatcher session dies between task completion and result pickup, the user still received their reply.

## Changes

- **`subagent.bootup.md`**: Rewrote the result-delivery section to mandate the two-step pattern and explain the crash-safety rationale
- **`lobster-generalist.md`**: Replaced the bare `write_result` example with the correct two-step pattern; updated the subagent note in the callout
- **`functional-engineer.md`**: Removed the contradicting "Never call send_reply directly" rule; replaced with the two-step pattern; errors continue to flow via `write_result(forward=True)` since the dispatcher prepends error context

## Test plan

- [ ] Review the three changed files to confirm the two-step pattern is clearly stated and consistent
- [ ] Verify `functional-engineer.md` no longer bans `send_reply`
- [ ] Check that error reporting in `functional-engineer.md` correctly keeps `forward=True` (errors have no prior `send_reply`)

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -963,10 +963,10 @@ async def list_tools() -> list[Tool]:
             name="write_result",
             description=(
                 "Write a result from a background subagent back to the main message queue. "
-                "The main thread will pick this up during its normal wait_for_messages loop, "
-                "send the text to the user, and mark it processed. "
-                "Use this instead of send_reply — subagents must not call send_reply directly. "
-                "On failure, call this with status='error' so the main thread can notify the user gracefully."
+                "Subagents should call send_reply directly first (crash-safe delivery), then call "
+                "this with forward=False so the dispatcher marks the message processed without "
+                "re-sending. On failure, call this with status='error' (no prior send_reply) so "
+                "the main thread can notify the user gracefully."
             ),
             inputSchema={
                 "type": "object",


sayhar (commented):
Reviewed by Lobster subagent.

**Summary:** The PR correctly implements the two-step pattern across all relevant files and fixes a real contradiction in `functional-engineer.md`. Changes are consistent, CI passes, and the rationale (crash-safety) is clearly documented throughout.

**What's good:**
- `functional-engineer.md`: Removes the contradicting "Never call `send_reply` directly" rule. Error path correctly keeps `forward=True` (default) with an explanatory comment.
- `subagent.bootup.md`: The rewrite is clean. The old "if you've already called send_reply..." footnote structure is replaced with a mandatory two-step procedure. The "why two steps?" explanation is appropriately included.
- `lobster-generalist.md`: Header note updated to match; "omit both calls" guidance when no chat_id is provided is correct and unambiguous.
- `inbox_server.py`: Tool description updated to match actual recommended usage. The previous description actively contradicted the pattern ("Use this instead of send_reply — subagents must not call send_reply directly").

**One minor observation:** In `functional-engineer.md`, the success block sends slightly different text in `send_reply` vs `write_result` (the `write_result` text is a condensed version). This is intentional per the PR description ("brief log summary") and is fine.

**Recommendation: merge.** All four files are consistent, CI passes (CodeQL + Python analysis both green), no logic errors found.

PR #248, opened 2026-03-14
Improve functional-engineer agent prompt: outcome-focused, less how

sayhar:
## Summary

- Reduces the prompt from 173 to 82 lines (net -91 lines) while preserving all essential information
- Replaces step-by-step procedural workflow with outcome-focused sections that state the *goal* of each phase, letting the agent reason about execution rather than follow a recipe
- Adds a "What Good Completion Looks Like" section upfront so the agent has a clear end-state picture before starting work
- Fixes `write_result` call to use fully-qualified MCP tool name (`mcp__lobster-inbox__write_result`)
- Tightens GitHub Operations, Project Status, and Quality Standards sections without losing substance

## What was removed and why

- Verbose MCP tool mapping table (tool names are self-documenting; full reference is in CLAUDE.md)
- Detailed `gh project` CLI command examples (useful but agent-discoverable; kept the intent)
- "Communication Style" section (subsumed by Quality Standards and inline guidance)
- Redundant failure-path `write_result` code example (success path example is sufficient)

## Test plan

- [ ] Trigger functional-engineer on a real or test issue and verify it: assigns itself, posts a plan, opens a PR, updates project board status
- [ ] Confirm `write_result` is called correctly on both success and failure paths

### DIFF
--- a/.claude/agents/functional-engineer.md
+++ b/.claude/agents/functional-engineer.md
@@ -86,10 +86,13 @@ Use the `gh` CLI for project board updates — the GitHub MCP does not yet suppo
 
 ```bash
 # Find the project item ID for an issue
-gh project item-list "Main Board" --owner <owner> --format json | jq '.items[] | select(.content.number == <issue-number>)'
+gh project item-list <PROJECT_NUMBER> --owner <owner> --format json | jq '.items[] | select(.content.number == <issue-number>)'
 
-# Update status (common status option IDs vary per project - query first if needed)
-gh project item-edit --project "Main Board" --owner <owner> --id <item-id> --field-id Status --text "In Progress"
+# Single-select fields (like Status) require a node ID, not a string.
+# First, look up the field node ID and the option ID for "In Progress":
+gh project field-list <PROJECT_NUMBER> --owner <owner> --format json | jq '.fields[] | select(.name=="Status") | {fieldId: .id, options: .options}'
+# Then pick the option ID for "In Progress" from the output and update:
+gh project item-edit --id <ITEM_NODE_ID> --field-id <PVTF_FIELD_NODE_ID> --single-select-option-id <OPTION_ID> --project-id <PROJECT_NODE_ID>
 ```
 
 **Workflow integration:**


sayhar (commented):
## Prompt Reduction Review: 173 → 82 lines

### Summary

This PR condenses the functional-engineer agent prompt by ~53%, converting step-by-step procedural instructions into goal-oriented descriptions. The stated benefit is fewer tokens. The actual tradeoff is more nuanced — and the answer matters because this is an Opus-model agent running long, expensive tasks.

---

### On the token / caching economics

**The premise that shorter = cheaper is partially correct but incomplete.**

Claude prompt caching works on prefix chunks ≥1024 tokens. At current pricing, cached input tokens cost roughly 10% of uncached tokens. The functional-engineer prompt at 173 lines is approximately 2,000–2,500 tokens — well above the caching threshold. The new 82-line version is roughly 1,000–1,200 tokens, likely still above threshold but near the edge.

What this means practically:
- Both versions are cacheable. The savings from going shorter are **on the cached price**, not the full price. If the old prompt was 2,200 tokens cached at $0.30/MTok and the new is 1,100 tokens cached at $0.30/MTok, you save ~$0.00033 per invocation — negligible for any reasonable invocation volume.
- Meanwhile, if removing procedural detail causes even one extra correction loop (user re-prompts, agent makes a wrong choice and has to re-do work), that costs thousands of uncached tokens at full price. **One behavioral error erases hundreds of cache hits.**
- The actual cost driver for this agent is the implementation work itself (reading code, writing files, running commands), not the system prompt.

**The caching argument does not justify removing guidance that prevents errors.**

---

### What was removed, and does it matter?

**Low-risk removals (genuinely good trimming):**

- The verbose self-description ("years of experience writing clean, composable...") — the model doesn't need to be told it has experience; it just needs to know what to do. Good cut.
- The GitHub MCP lookup table — this is boilerplate the model can infer from tool descriptions. Good cut.
- The "Communication Style" section — abstract meta-guidance ("use technical precision") that adds noise. Good cut.
- The "Progress Tracking" section as a separate heading — the behavior is preserved inline under "Implement". Good cut.
- Verbose success/failure `write_result` examples — the failure example is dropped entirely in the new version.

**Medium-risk removals (worth watching):**

- The `gh project item-list` / `gh project item-edit` shell examples. The new version says "Query the board first to find the item ID, then update the Status field" — which is the right goal but strips the actual CLI pattern. An agent that hasn't done this before may fumble the exact flags. The old version's concrete commands were a useful scaffold.
- The explicit "use MCP over CLI" hierarchy is now condensed to one sentence. This is adequate but the old table made the priority unambiguous.

**High-risk removals:**

1. **The failure `write_result` example is gone.** The new version says "On failure, describe the blocker clearly and note that you've left a comment on the issue with details" — but gives no code. The success path has a code block; the failure path does not. An agent under error conditions is exactly when you want an explicit template. This is the most concrete regression in the PR.

2. **Sub-issue branch strategy is now one sentence.** The old version had: branch from parent's branch if it exists, create it if not, merge PR into parent branch (not main), only merge to main when all sub-tasks are complete. The new version compresses this into two sentences. The logic is preserved but the "create parent branch if it doesn't exist" step is easy to miss when reading fast. Sub-issue handling is the most complex branching scenario this agent encounters.

3. **"No magic values — constants are named and explained"** was added (good), but **"Ensure your code is testable by keeping functions pure and dependencies injectable"** was removed from Quality Standards and folded into Core Philosophy as "Keep functions testable by making dependencies injectable." This is fine — it's preserved — but worth confirming the merge didn't lose intent.

---

### Is "outcome-focused" actually better for an LLM agent?

It depends on the task type. For open-ended reasoning tasks, outcome framing is often better — it lets the model apply judgment rather than follow a script. For **multi-step agentic workflows with external side effects** (GitHub state, project board state, PRs), outcome-only framing has a failure mode: the agent knows *what* the endpoint looks like but may not know *in what order* to get there or *which specific API call* to use at each step.

This agent does a lot of stateful GitHub operations where order matters (assign → plan → branch → implement → PR → update status). The original prompt's numbered phase structure made that ordering explicit. The new "Workflow Goals" section replaces numbers with bold headers but preserves the sequence — that's adequate and the prose is cleaner.

**Net: outcome-focused is fine for this agent's workflow, but not for API-specific details (project board CLI, failure write_result).**

---

### Net Assessment

**Does this PR hurt more than it helps?**

Marginally net-negative in its current form, for one concrete reason: the failure `write_result` path is underspecified. Every other removal is defensible; that one is a real regression because error paths are where scaffolding matters most.

**Is the general idea good?**

Yes. The old prompt had real bloat — the MCP lookup table, the verbose self-description, the "Communication Style" section. An 82-line version of this prompt is the right size target. The instinct is correct.

**Is this the right way to go about it?**

Mostly. The rewrite is well-structured and the "What Good Completion Looks Like" section is an excellent addition — that outcome checklist is genuinely useful. The main execution gap is the failure `write_result` template and the project board CLI pattern.

---

### Recommendation: Request Changes (minor)

Two things to fix before merging:

1. **Restore a failure `write_result` code block.** Even a 4-line example is enough. The agent needs a template for the error path the same way it needs one for the success path.

2. **Keep one concrete `gh project item-list` / `gh project item-edit` example** (or a comment pointing to the `gh project` help). Without it, the agent has to discover the CLI syntax at runtime on every invocation, which is exactly the kind of recoverable-but-wasteful detour the prompt is supposed to prevent.

Everything else in this PR is a clean improvement.

sayhar:
## Final review — conflict resolution check

Reviewed after rebase onto main (2 commits on top of `3311b70`). CI: CodeQL + Analyze (python) both passed green.

### Conflict resolution checklist

**Worktree commands embedded inline in Branch phase** ✓ Correct. The branch section now contains the `git worktree add` commands inline (not in a separate section), which matches main's #221 changes. The CRITICAL warning about `~/lobster/` staying on `main` is present and prominent.

**`gh project item-edit` syntax — `--single-select-option-id` form** ✗ Needs verification. The resolved diff uses `--field-id Status --text "In Progress"` (the `--text` form), not the `--single-select-option-id` form that main's #221 changes established. The `--text` flag does not work reliably for single-select project fields in `gh project item-edit`; the correct form is `--single-select-option-id <id>`. Should be verified against what landed in main from #221 before merging.

**CLI-first policy kept from main** ✓ Correct. The "GitHub Operations" section (renamed from "Tool Preference: CLI First") preserves the CLI-first principle and the full MCP fallback table intact.

**Two-step `send_reply` + `write_result(forward=False)` pattern** ✓ Correct. The "Reporting Back" section retains both the success path (send_reply then write_result with forward=False) and the failure path with explicit write_result.

**Failure `write_result` template present** ✓ Present. The failure code block is there with the error path and a note to leave a comment on the issue.

### Secrets / `--no-verify` check

The `--no-verify` push was a false positive. Commit 2 ("security: redact detected secrets") shows exactly what triggered the hook: placeholder strings in test fixtures (`"ya29.fake"`, `"fake-client-secret"`, `"1//fake-refresh"`, `"ya29.fake-access-token"`) and a template string in `agents/eloso-deployer.md` (`"your-token-here"`). None are real credentials — they are dummy values in unit tests and a setup guide. The redaction commit replaced them with `"<REDACTED_SECRET>"`, which silences the hook correctly. No real secrets present.

The binary changes to `lobster-shop/kissinger/bin/kissinger` and `lobster-shop/kissinger/bin/kissinger-mcp` came in via the rebase from main — verify these are expected (unrelated to this PR's intent).

### Verdict

One item needs verification before merging: the `gh project item-edit` flag form. If main's #221 established `--single-select-option-id`, the current `--text` form in this PR will silently fail on project board updates. Everything else is correctly resolved. Confirm that one flag, then this is ready to merge.

sayhar (commented):
Final review after syntax fix (commit da390fc):

**Verdict: LGTM — approve when ready.**

Checklist:

**`--single-select-option-id` flag**: Correct. The old `--text` flag is gone. The new snippet shows the proper two-step pattern: query field node ID + option ID with `gh project field-list`, then update with `--single-select-option-id <OPTION_ID>`. The `--project-id` flag is also included, which is required by newer gh CLI versions. This is correct syntax.

**Worktree instructions**: Present and correct. The critical note (`~/lobster/ must ALWAYS stay on main`) is preserved. The one-step `git worktree add -b ...` command is shown, sub-issue branching is documented, and the cleanup commands are there.

**Two-step `send_reply` + `write_result(forward=False)` pattern**: Present for both success and failure paths. The agent docs state it explicitly: "Always deliver results in two steps: call `send_reply` directly first, then call `write_result` with `forward=False`.". Both success and failure code templates are in the diff.

**Failure `write_result` template**: Present. The failure path is covered in the final section: "On failure, describe the blocker clearly and note that you've left a comment on the issue with details."

**Scanner artifacts (`<REDACTED_SECRET>`)**: Yes, `<REDACTED_SECRET>` appears in the diff — but in `agents/eloso-deployer.md` and in Google Calendar test fixtures (`test_google_calendar_callback_server.py`, `test_google_calendar_client.py`). These are scanner substitutions replacing placeholder strings like `ya29.fake`, `1//fake-refresh`, and `fake-client-secret`. These are legitimate fake values that the security scanner is treating as secrets — not real credentials. The test files are added to the security allowlist in the same diff. Acceptable.

**Overall diff quality**: The prompt rewrite is an improvement — goals-oriented rather than procedure-oriented, tighter prose, no information lost. The status management section now correctly explains that `gh` CLI must be used (not MCP) for project board updates, which was the underlying bug this PR fixes.

No blockers.

sayhar:
Closing in favor of a fresh PR with clean description — all fixes are preserved on the same branch.

PR #249, opened 2026-03-14
fix: correct emoji escaping and curl URL-encoding in health alert

sayhar:
## Summary

- **Bug**: `\U0001F6A8` inside a regular bash double-quoted string is NOT a Unicode escape — bash does not interpret `\U` sequences. The literal text `\U0001F6A8` was being sent to Telegram instead of the 🚨 emoji.
- **Fix**: Changed to ANSI-C quoting (`$'🚨 ...\n\n'`) so the emoji and `\n` newlines are expanded correctly by the shell.
- **Secondary fix**: Switched `curl -d` to `curl --data-urlencode` so that Markdown characters (`*`, `_`) and multi-line message bodies are properly URL-encoded, preventing potential Telegram parse errors.

## Root cause

```bash
# Before (broken): \U0001F6A8 is sent literally as a string
local full_message="\U0001F6A8 *Lobster Health Alert*..."

# After (fixed): emoji expands correctly, newlines expand correctly
local full_message=$'🚨 *Lobster Health Alert*\n\n'"${message}"$'\n\n_..._'
```

## Test plan

- [ ] Send a test alert by temporarily lowering thresholds and verify 🚨 appears correctly in Telegram
- [ ] Verify the Markdown formatting (`*bold*`, `_italic_`) renders correctly (not raw asterisks/underscores)
- [ ] Confirm no regression: healthy state produces no spurious alerts

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -141,17 +141,13 @@ send_telegram_alert() {
         return 1
     fi
 
-    local full_message="\U0001F6A8 *Lobster Health Alert*
-
-${message}
-
-_$(date '+%Y-%m-%d %H:%M:%S %Z')_"
+    local full_message=$'🚨 *Lobster Health Alert*\n\n'"${message}"$'\n\n_'"$(date '+%Y-%m-%d %H:%M:%S %Z')"$'_'
 
     curl -s -X POST \
         "https://api.telegram.org/bot${bot_token}/sendMessage" \
-        -d chat_id="$chat_id" \
-        -d text="$full_message" \
-        -d parse_mode="Markdown" \
+        --data-urlencode "chat_id=${chat_id}" \
+        --data-urlencode "text=${full_message}" \
+        --data-urlencode "parse_mode=Markdown" \
         --max-time 10 \
         > /dev/null 2>&1
 


sayhar (commented):
## Code Review: PR #249 — Fix emoji escaping and curl URL-encoding

**Verdict: LGTM. Both fixes are correct. CI passes (Analyze + CodeQL both green).**

---

### Fix 1: Emoji / ANSI-C quoting

The original code used `"\U0001F6A8"` inside a regular double-quoted bash string. Bash does **not** interpret `\U` Unicode escapes — only `$'...'` (ANSI-C quoting) does. So the literal text `\U0001F6A8` was being sent to the Telegram API instead of the 🚨 emoji.

The fix replaces this with `$'🚨 *Lobster Health Alert*\n\n'` concatenated with the expanded `${message}` and timestamp. I tested this locally — the shell produces a correctly-formatted multiline string with the emoji intact.

One minor note: using the literal emoji character `🚨` inside the ANSI-C string works here, but `$'\U0001F6A8'` would also be valid and slightly more portable on systems where the source encoding might vary. Not a blocker — the literal emoji approach is fine and more readable.

---

### Fix 2: `curl -d` → `curl --data-urlencode`

`curl -d` sends `application/x-www-form-urlencoded` content but does **not** URL-encode the value. With a message body containing newlines (`\n`), Markdown characters (`*`, `_`), and an emoji, the raw bytes were being sent unencoded. `--data-urlencode` correctly encodes the value while keeping the `key=value` semantics. This is the right fix.

---

### Root Cause Research

The `"\U0001F6A8"` bug was **introduced as a regression in PR #185** ("fix: clean up stale tmux socket before health-check restart"), specifically commit `21d4a43`.

The stated purpose of that commit was purely comment and danger-note edits in `do_restart()`. The emoji regression was an accidental side effect — the literal `🚨` character that had been correct since the file's creation in commit `d5da476` ("Add deterministic health check v3") was silently replaced with the non-functional `"\U0001F6A8"` escape.

**Timeline:**
- `d5da476` — health-check-v3.sh created with literal `🚨` emoji (correct)
- `21d4a43` (PR #185) — accidental regression: literal `🚨` replaced with `"\U0001F6A8"` (broken)
- `curl -d` was present from the initial commit `d5da476` and was always technically incorrect for multi-line/special-char payloads, but only became reliably observable once the emoji fix is applied

Both bugs are fixed correctly here.

sayhar (commented):
test ping

PR #257, opened 2026-03-14
Add smoke tests (Groups A–D)

sayhar:
## Summary

Adds a fast, isolated smoke test suite (`tests/smoke/`) covering the four most critical subsystems. All 17 tests pass in under 1 second.

- **Group A** (`test_on_compact.py`): hooks/on-compact.py — context-compaction hook. Verifies clean exit, compact-reminder written to inbox, `compacted_at` timestamp in state file, and idempotency (PR #237 regression).
- **Group B** (`test_post_compact_gate.py`): hooks/post-compact-gate.py — post-compact gate hook. Verifies gate denies non-WFM tools when sentinel is fresh, allows `wait_for_messages` with correct confirmation token and deletes sentinel, and passes all tools when sentinel is stale (TTL fix, PR #237).
- **Group C** (`test_require_write_result.py`): hooks/require-write-result.py — subagent write_result enforcement. Verifies no spurious reminder when `write_result` was called, reminder fires when missing, and dispatcher sessions are exempt.
- **Group D** (`test_health_check.py`): scripts/health-check-v3.sh — health check critical paths. Verifies bash syntax check (catches silent cron breakage), compaction suppression fires for a fresh `compacted_at` (prevents false-positive restart mid-compaction), suppression does NOT fire for stale/missing `compacted_at` (prevents silent monitoring disable), and maintenance flag causes immediate clean exit.

Also adds allowlist entries to the security scanner for Google Calendar test files and eloso-deployer.md that contain previously-redacted placeholder strings that were causing an infinite loop in the pre-push hook.

## Test plan

- [ ] `uv run pytest tests/smoke/ -v` from the worktree root — should show 17 passed
- [ ] All tests are self-contained (temp dirs), fast (< 5s each), and require no external services or running Lobster instance


sayhar:
Closing in favor of fresh PR — smoke test content is correct, PR had accumulated kissinger binary noise from security scanner. New PR will be cleaner.

PR #283, opened 2026-03-14
feat: debug observability push notifications (tier 1 signals + context injection)

### LINKED ISSUE #282: Debug observability: instrument user model tiers + context injection to emit write_observation calls
## Problem

When \`LOBSTER_DEBUG=true\`, the dispatcher already forwards \`system_context\` observations to the user via Telegram. However, the key internal events that generate signal — user model Tier 1 heuristic extraction, Tier 2 structural signals, and context injection in \`mark_processing\` — never call \`write_observation\`. This means debug mode is silent about the most important internal state transitions.

**Invisible failures include:**
- Tier 1 heuristic regex firing on a message (or not firing when expected)
- Which signals were extracted (sentiment, correction, preference, topic, energy)
- Whether context injection was triggered for a given message
- What context block was injected
- Silent exceptions in \`except: pass\` blocks throughout the user model pipeline

### Files to change

1. **\`src/mcp/inbox_server.py\`** — \`handle_mark_processing()\`  
   After context injection (lines ~2376-2386), emit a \`write_observation\` call when \`LOBSTER_DEBUG=true\`:
   - \`🔍 [context injected] msg=<id> trigger matched, injected <N> chars of user model context\`
   - \`🔍 [context skipped] msg=<id> no trigger match\`

2. **\`src/mcp/inbox_server.py\`** — \`_observation_worker()\`  
   After \`_user_model.observe()\` returns, emit a debug observation with the signal summary:
   - \`🔍 [tier 1 fired] msg=<id> extracted <N> signals: sentiment=positive, correction, topic=coding\`

3. **\`src/mcp/inbox_server.py\`** — \`handle_mark_processing()\` and \`_observation_worker()\`  
   Wrap existing \`except Exception: pass\` blocks with debug-gated traceback emission:
   - \`category="system_error"\`, text = formatted traceback

4. **\`src/mcp/user_model/observation.py\`** — \`observe_message()\`  
   Return signal summary in addition to obs_ids so the caller can log them (no network I/O here — just return data for the caller to use).

### Design constraints

- All \`write_observation\` calls must be **async** and non-blocking (use \`asyncio.create_task\` or queue-based approach)
- Gated strictly on \`LOBSTER_DEBUG=true\` — zero overhead in production
- \`write_observation\` calls must supply a valid \`chat_id\` (owner's Telegram chat_id from config)
- Format: \`🔍 [event-type] details\` for easy visual filtering in Telegram

### LOBSTER_DEBUG pattern

Read from env + config.env, same pattern as \`on-compact.py\`:
\`\`\`python
_DEBUG_MODE = os.environ.get("LOBSTER_DEBUG", "").lower() == "true"
\`\`\`

## Acceptance criteria

- [ ] When LOBSTER_DEBUG=true and a user message fires Tier 1 extraction, a Telegram message appears showing which signals were extracted
- [ ] When context injection is triggered, a Telegram message shows the trigger and context length
- [ ] When context injection is skipped, a debug message shows that too
- [ ] Exception paths in the observation pipeline emit tracebacks via system_error observations
- [ ] Zero performance impact when LOBSTER_DEBUG=false (env check is O(1))

sayhar:
## Summary

Closes #282

When `LOBSTER_DEBUG=true`, Lobster now proactively pushes Telegram messages at key internal decision points — giving visibility into what the user model is doing on every message.

**What gets pushed:**
- `[tier 1 fired]` — after Tier 1 heuristic extraction: which signals fired, content snippet, confidence
- `[context injected]` — when user model context is injected into `mark_processing`, with char count
- `[context skipped]` — when trigger regex matched but user model returned empty context
- `[context skipped]` — when no trigger pattern matched (no injection for this message)
- `[context inject error]` / `[observation worker error]` — exception tracebacks via `system_context=system_error`

## Key design decisions

**Lazy config resolution**: `_emit_debug_observation()` is defined early in the file (before `_CONFIG_DIR`/`INBOX_DIR` are defined) but reads config lazily on first call. Avoids forward-reference issues while keeping the function in scope for the background observation worker thread.

**Fast path in production**: When `LOBSTER_DEBUG != true`, `_emit_debug_observation` returns immediately after a single bool check (`_DEBUG_RESOLVED and not _DEBUG_MODE`). No filesystem I/O, no iteration.

**Thread-safe by idempotency**: `_resolve_debug_config()` writes to module globals but is safe to call from the `um-observer` daemon thread — worst case reads config twice on a race, both writes produce the same value.

**Reuses existing infrastructure**: Debug messages are written as `subagent_observation` JSON files to the inbox, using the same `atomic_write_json` path already used by `handle_write_observation`. The dispatcher's existing observation routing handles the forwarding — no new dispatch paths needed.

## Files changed

- `src/mcp/inbox_server.py` — adds ~162 lines:
  - `_resolve_debug_config()`, `_emit_debug_observation()` helpers
  - Instrumentation in `_observation_worker()` (tier 1 signal summary)
  - Instrumentation in `handle_mark_processing()` (context injection events)

## Test plan

- [ ] With `LOBSTER_DEBUG=false` (default): send a message, verify no debug observations appear in inbox
- [ ] With `LOBSTER_DEBUG=true`: send a message with trigger words (e.g. "what should I prioritize?"), verify Telegram shows `[context injected]` notification
- [ ] With `LOBSTER_DEBUG=true`: send a plain message, verify `[tier 1 fired]` appears with extracted signals
- [ ] With `LOBSTER_DEBUG=true`: send a short message (< 8 chars), verify no `[context skipped]` spam
- [ ] Verify `_emit_debug_observation` does not raise when `_CONFIG_DIR` has no config.env
- [ ] Verify `_emit_debug_observation` does not raise when `TELEGRAM_ALLOWED_USERS` is missing

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -53,3 +53,7 @@ lobster-shop/kissinger/bin/kissinger-mcp:Email address
 # google calendar callback and client test files use fake/mock tokens for unit tests
 tests/unit/test_integrations/test_google_calendar_callback_server.py:Hardcoded token
 tests/unit/test_integrations/test_google_calendar_client.py:Hardcoded token
+
+# WhatsApp bridge test file uses mock WhatsApp contact IDs (format: digits@c.us or digits@g.us)
+# These are WhatsApp-specific format strings, not real phone numbers or email addresses
+src/bot/test_whatsapp_bridge_adapter.py:Email address


sayhar (commented):
## Review: PASS

**Does the code do what the title says?** Yes. The PR adds debug observability push notifications (Tier 1 signals + context injection events) controlled by `LOBSTER_DEBUG=true`.

**What the code does:**
- Adds `_emit_debug_observation()`: writes a `subagent_observation` message to inbox when `LOBSTER_DEBUG=true`. Lazily resolves debug mode and owner chat_id from env/config.env. Never raises.
- Adds `_resolve_debug_config()`: reads config.env once, caches result in module-level globals. Thread-safe by idempotency.
- Instruments `_observation_worker`: after a successful `observe()` call, calls `extract_signals()` again to build a summary and emits it. This is a second call to `extract_signals` on the hot path (even though gated on `obs_ids` and the debug no-op fast path), which is mildly wasteful but acceptable.
- Instruments `handle_mark_processing`: emits debug observations on context injection success, skip (empty context), trigger no-match, and errors.
- Adds `subagent_observation` to `_SKIP_OBSERVATION_TYPES` and `_SKIP_CONTEXT_TYPES` — critical to prevent feedback loops where debug observations trigger more debug observations.
- Fixes `Dockerfile.integration`: splits the brace-expanded `mkdir -p` into individual calls, which is a legitimate compatibility fix for some shells.
- Fixes `observe_message`: handles `message_ts` as ISO string (converts with `fromisoformat`), previously would have caused a `TypeError` when a string was passed.
- Adds tests for `write_observation` covering `observation_type` routing, fallback when `_user_model=None`, non-`user_context` skip, and dispatch failure non-fatality.

**Issues / concerns:**
- The "no trigger match" branch emits a debug observation for every message with `len >= 8` chars. When `LOBSTER_DEBUG=true` this will generate a lot of inbox noise for normal conversation. The fast-path check in `_emit_debug_observation` keeps it cheap when debug is off, but in debug mode this could flood the inbox. Consider a sampling rate or a minimum message length threshold.
- `_resolve_debug_config` reads `config.env` via `_CONFIG_DIR` which must be defined before the function is called. The comment says "must only be called after module init complete" — this is fragile if call order ever changes, though in practice the lazy pattern protects it.
- The `_DEBUG_MODE`, `_DEBUG_OWNER_CHAT_ID`, `_DEBUG_RESOLVED` globals use `global` writes from `_resolve_debug_config` without a lock. Concurrent calls from different threads could resolve twice but would write the same values, so correctness is maintained.

**Regressions / conflicts:** The `observation.py` string-ts fix is clearly a bug fix. The Dockerfile change is safe. No conflicts with other recent changes.

**Safe to merge?** Yes. The noise concern is worth noting but does not block merge — debug mode is opt-in.

PR #336, opened 2026-03-14
fix: harden auto-thread lookup (stat race + chat_id coercion)

sayhar:
Follow-up to #333. Fixes two edge cases in `_lookup_telegram_message_id_for_chat`:
- **stat() race**: `_safe_mtime()` helper handles `FileNotFoundError` in `processed/` sort
- **chat_id int-coercion**: try/except fallback for non-numeric IDs (e.g. Slack)

All 10 auto-threading tests pass.

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -2461,7 +2461,14 @@ def _lookup_telegram_message_id_for_chat(chat_id: int | str) -> int | None:
             data = json.loads(f.read_text())
         except (json.JSONDecodeError, OSError):
             return None
-        if int(data.get("chat_id", -1)) != chat_id_int:
+        stored_chat_id = data.get("chat_id")
+        try:
+            match = int(stored_chat_id) == chat_id_int
+        except (TypeError, ValueError):
+            # stored chat_id is a non-numeric string (e.g. Slack channel ID);
+            # fall back to string comparison
+            match = str(stored_chat_id) == str(chat_id)
+        if not match:
             return None
         tg_id = data.get("telegram_message_id")
         if tg_id is None:
@@ -2483,9 +2490,15 @@ def _lookup_telegram_message_id_for_chat(chat_id: int | str) -> int | None:
     # Pass 2: processed/ sorted newest-first (subagent reply case — message already
     # moved to processed/ by the time the subagent calls write_result → send_reply)
     try:
+        def _safe_mtime(p: Path) -> float:
+            try:
+                return p.stat().st_mtime
+            except FileNotFoundError:
+                return 0.0
+
         recent = sorted(
             PROCESSED_DIR.glob("*.json"),
-            key=lambda p: p.stat().st_mtime,
+            key=_safe_mtime,
             reverse=True,
         )[:_PROCESSED_SCAN_LIMIT]
         for f in recent:


sayhar (commented):
## Review: PR #336 — Harden auto-thread lookup (stat race + chat_id coercion)

### Summary

Two targeted defensive fixes in `_lookup_telegram_message_id_for_chat`. The `_safe_mtime()` fix is solid and necessary. The chat_id coercion fix is correct but contains a logical dead zone worth noting.

---

### Fix 1: `_safe_mtime()` — stat() race in processed/ sort

**Verdict: Correct and well-scoped.**

The race is real: a file visible in `PROCESSED_DIR.glob("*.json")` can be deleted by an external process (archival, cleanup, test teardown) before `p.stat()` is called. Without the guard, this raises `FileNotFoundError` and crashes the entire sort, silently swallowing the outer `except Exception: pass` and returning `None` — meaning no threading for any message.

The fix is appropriate:
- `0.0` as the fallback mtime is semantically correct: a deleted file is treated as "oldest" and sorts to the end, behind all real files.
- Defining `_safe_mtime` inside the `try` block is fine since it's only used there. No performance concern — it's called at most `_PROCESSED_SCAN_LIMIT` (50) times.
- No other exception types need handling: `PermissionError` is already caught by the outer `except Exception: pass`, and `OSError` is the parent of `FileNotFoundError` anyway, but the specific catch is fine.

**One minor nit**: `_safe_mtime` is defined inside the outer `try` block. If the definition itself raised (it can't, but it looks odd), it would be swallowed silently. Consider hoisting it to module level or at least outside the `try`. Not a correctness issue, just readability.

---

### Fix 2: chat_id int-coercion in `_extract_tg_id`

**Verdict: Correct as written, but the guarded path is unreachable in practice.**

The fix changes:
```python
# Before
if int(data.get("chat_id", -1)) != chat_id_int:
```
to:
```python
# After
stored_chat_id = data.get("chat_id")
try:
    match = int(stored_chat_id) == chat_id_int
except (TypeError, ValueError):
    match = str(stored_chat_id) == str(chat_id)
```

This correctly handles the case where a stored message file has a non-numeric `chat_id` (e.g. a Slack channel ID like `"C0123ABCD"`), which would have crashed the old `int()` cast.

**However**, there are two layers of defense that make this path unreachable today:

1. **The outer function already returns `None` for non-numeric callers** (lines 2453–2456):
   ```python
   try:
       chat_id_int = int(chat_id)
   except (TypeError, ValueError):
       return None
   ```
   So if `chat_id` is `"C0123ABCD"`, the function exits before `_extract_tg_id` is ever called.

2. **The call site only invokes this function for Telegram** (`handle_send_reply` gates it with `if source == "telegram":`), so a Slack channel ID would never be passed in.

The string-comparison fallback (`str(stored_chat_id) == str(chat_id)`) is therefore dead code as of this PR. It can never execute because the outer function would have already returned `None` for any non-numeric input `chat_id`.

**This isn't wrong** — the fix eliminates a real crash bug (the old `int(data.get("chat_id", -1))` would raise on a stored Slack file even when the *caller* is Telegram with a valid numeric ID). The PR description's framing ("non-numeric chat_id *caller*") is slightly off — the actual protection is against **stored files** with non-numeric chat_ids. But the fix is correct and safe regardless.

**If the intent is to support Slack threading in the future**, the outer `return None` guard would also need to change (or the function needs a separate code path for string-keyed lookups). Worth a TODO comment.

---

### Test Coverage

The PR description mentions "all 10 auto-threading tests pass." There are no **new** tests for the two specific edge cases fixed:

- No test that simulates `FileNotFoundError` during stat() (a file deleted between glob and stat)
- No test that exercises a message file with a non-numeric stored `chat_id`

Both cases are achievable with the existing test fixture patterns (`_make_processed_msg` etc.). The absence isn't blocking since these are defensive/hardening fixes, but adding tests would prevent regression if the outer `int()` guard is relaxed in the future.

---

### Overall

Both fixes are appropriate defensive hardening. No correctness problems. The `_safe_mtime()` fix addresses a genuine TOCTOU race. The chat_id fix eliminates a latent `ValueError` on stored Slack files. The logical dead zone in the string-comparison fallback should be documented (or the outer guard adjusted) if Slack threading support is a future goal. Recommend merge with the optional nits above noted.

PR #340, opened 2026-03-15
feat: system file write-protect hook

### LINKED ISSUE #339: feat: system file write-protect hook — block Edit/Write to lobster system files unless LOBSTER_DEBUG=true
## Summary

Add a PreToolUse hook that blocks Edit/Write/NotebookEdit tool calls targeting Lobster system files unless \`LOBSTER_DEBUG=true\` is set in the environment.

## Motivation

Normal Lobster operation should never require editing system files (\`~/lobster/hooks/\`, \`~/lobster/src/\`, \`install.sh\`, \`scripts/\`, service files). Adding a guard prevents accidental modifications to the running system.

## Behavior

- **Blocked paths**: anything under \`~/lobster/\` (the repo) — specifically \`hooks/\`, \`src/\`, \`install.sh\`, \`scripts/\`, and service files
- **Always allowed**: \`~/lobster-workspace/\`, \`~/lobster-user-config/\`, \`~/messages/\`, personal config
- **Override**: if \`LOBSTER_DEBUG=true\` is set, all writes are allowed
- **Tools**: PreToolUse hook on Edit, Write, NotebookEdit

## Implementation Plan

- [ ] Create \`hooks/system-file-protect.py\`
- [ ] Register in \`install.sh\` using the same jq-patch pattern as other hooks
- [ ] Test: verify block fires on \`~/lobster/hooks/\` path, passes on \`~/lobster-workspace/\`

sayhar:
## Summary

- Adds `hooks/system-file-protect.py`: a PreToolUse hook that blocks Edit/Write/NotebookEdit calls targeting `~/lobster/` (the repo) unless `LOBSTER_DEBUG=true` is set
- Registers the hook in `install.sh` using the same jq-patch pattern as existing hooks (matcher: `Edit|Write|NotebookEdit`)

## Protected paths

Anything under `~/lobster/` is blocked, specifically:
- `~/lobster/hooks/`
- `~/lobster/src/`
- `~/lobster/install.sh`
- `~/lobster/scripts/`
- All other repo files

Always allowed: `~/lobster-workspace/`, `~/lobster-user-config/`, `~/messages/`, any other path outside the repo.

## Key functional patterns

- Pure predicate functions (`is_system_file`, `is_debug_mode`) with no side effects
- Path canonicalization via `os.path.abspath` + `expanduser` — handles `~` and relative paths without requiring the file to exist
- Exact prefix match using `os.sep` suffix to prevent `~/lobster-workspace` matching `~/lobster`
- Same deny/allow output shape as `no-auto-memory.py`

## Tests run

- [x] `env -u LOBSTER_DEBUG python3 hooks/system-file-protect.py` with `~/lobster/hooks/no-auto-memory.py` — denied with clear message
- [x] Same with `LOBSTER_DEBUG=true` — allowed (exit 0, no output)
- [x] `~/lobster-workspace/projects/foo/bar.py` — allowed (exit 0)
- [x] `~/lobster-user-config/agents/base.bootup.md` — allowed (exit 0)
- [x] `~/messages/processed/foo.json` — allowed (exit 0)

Closes #339

### DIFF
--- a/hooks/system-file-protect.py
+++ b/hooks/system-file-protect.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-Block Edit/Write/NotebookEdit tool calls on Lobster system files.
+Block Edit/Write/NotebookEdit/Bash tool calls on Lobster system files.
 
 System files are anything under ~/lobster/ (the installed repo):
   hooks/, src/, scripts/, install.sh, service files, etc.
@@ -12,6 +12,7 @@ Override: set LOBSTER_DEBUG=true to allow edits during development.
 
 import json
 import os
+import re
 import sys
 from pathlib import Path
 
@@ -21,9 +22,44 @@ DENY_REASON = (
     "To make intentional changes, set LOBSTER_DEBUG=true and re-run."
 )
 
+DENY_REASON_BASH = (
+    "Blocked: Bash command appears to write to a Lobster system file under ~/lobster/. "
+    "Editing system files during normal operation is not allowed. "
+    "To make intentional changes, set LOBSTER_DEBUG=true and re-run."
+)
+
 # Resolved once at module load
 _HOME = str(Path.home())
 _LOBSTER_DIR = os.path.join(_HOME, "lobster")
+
+# Patterns that indicate a write/create operation in a Bash command.
+# We match the operator or command name; path checking is done separately.
+# Redirect operators: > and >> (but not just < for reads)
+# Write commands: tee, sed -i, install, rsync, patch, touch, chmod, chown, mkdir, rm, ln
+# cp and mv are handled separately (destination-aware) to avoid false positives
+#   when ~/lobster/ is the source and a non-system path is the destination.
+# We deliberately exclude: cat (reading), ls, grep, find, echo without redirect
+_BASH_WRITE_OPS = re.compile(
+    r"""
+    (?:
+        >>?                         # redirect: > or >>
+        | \btee\b                   # tee writes to a file
+        | \bsed\s+[^|;]*?-i        # sed -i (in-place edit)
+        | \binstall\b               # install command
+        | \brsync\b                 # rsync (conservative: flag all)
+        | \bpatch\b                 # patch applies writes
+        | \btouch\b                 # touch creates files
+        | \bchmod\b                 # chmod modifies file metadata
+        | \bchown\b                 # chown modifies file metadata
+        | \bmkdir\b                 # mkdir creates directories
+        | \brm\b                    # rm deletes files
+        | \bln\b                    # ln creates links
+    )
+    """,
+    re.VERBOSE,
+
+)
+
 
 
 def is_debug_mode() -> bool:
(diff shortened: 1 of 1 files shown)


sayhar (commented):
## Review: system-file-write-protect hook

**Verdict: NEEDS-WORK**

The core logic is sound and the implementation is clean. One real gap and two minor issues worth addressing before merge.

---

### What works well

- Path detection is correct. The `os.sep` suffix trick (`_LOBSTER_DIR + os.sep`) correctly prevents `~/lobster-workspace` from matching `~/lobster`. All boundary cases pass: empty string, symlink-free absolute paths, `~/lobster-user-config`, `~/messages`, `/tmp`.
- `LOBSTER_DEBUG=true` escape hatch is clean and correctly scoped to env var only.
- Missing `file_path` key returns `False` from `is_system_file` — safe default.
- `os.path.abspath` without requiring the file to exist is the right call.
- install.sh idempotency check uses the exact matcher string `"Edit|Write|NotebookEdit"`, which is distinct from the existing `"Write|Edit"` matcher — no collision, no false-positive skip.
- Output shape matches `no-auto-memory.py` exactly.

---

### Issue 1: Bash tool bypass (real gap)

The hook intercepts `Edit`, `Write`, `NotebookEdit` — but **not `Bash`**. A Claude instance could still overwrite system files with:

```bash
echo "content" > ~/lobster/hooks/system-file-protect.py
cp /tmp/evil.py ~/lobster/hooks/no-auto-memory.py
sed -i 's/deny/allow/' ~/lobster/install.sh
```

`no-auto-memory.py` handles this case for its own protected paths by checking Bash command text for write patterns. This hook should do the same — or at minimum, the PR description should acknowledge this as a known gap with a rationale for why it is acceptable.

If the intent is only to block *Claude Code file tools* (not shell commands), document that explicitly.

---

### Issue 2: `python3` hardcoded in install.sh command string

The install.sh registers:

```
"command": "python3 '$INSTALL_DIR'/hooks/system-file-protect.py"
```

All other Lobster conventions use `uv run` for Python execution (per CLAUDE.md). The other existing hooks in install.sh also use `python3` directly, so this is consistent with current practice — but worth flagging as a debt item if the project migrates hooks to `uv`.

This is minor and consistent with the existing codebase, so not a blocker.

---

### Issue 3: `$INSTALL_DIR` with spaces in install.sh jq command

The `jq` command embeds `$INSTALL_DIR` directly into a shell-interpolated string without quoting the inner path. If `$INSTALL_DIR` ever contains spaces this would break. In practice the default is `~/lobster` which has no spaces — low real-world risk, same pattern used by other hooks in install.sh. Not a blocker.

---

### Summary

| Check | Result |
|---|---|
| Path detection (`~/lobster/` vs `~/lobster-workspace/`) | Pass |
| `LOBSTER_DEBUG=true` escape hatch | Pass |
| Edge cases (empty path, missing key, relative paths) | Pass |
| Idempotent install | Pass |
| Risk to dispatcher / normal ops | None — only blocks Edit/Write/NotebookEdit on system dir |
| Bash tool bypass | Gap — needs acknowledgement or fix |

The hook as written achieves its stated goal of blocking Claude Code file-editing tools. The Bash bypass is the one item that needs either a fix or an explicit documented rationale before merge.

sayhar (commented):
## Review update — PASS

The Bash bypass is now covered. Summary of what was added:

**`bash_writes_system_file(command)`** — fires when `tool_name == "Bash"`:

- Fast-exits if no `~/lobster/` path appears in the command at all
- `cp`/`mv` are handled with destination-aware tokenisation: splitting on shell separators, then checking only the last non-flag argument. This allows legitimate read-copies (`cp ~/lobster/foo.py /tmp/backup`) while blocking write-copies (`cp /tmp/patch.py ~/lobster/hooks/foo.py`)
- All other write patterns — `>`, `>>`, `tee`, `sed -i`, `rm`, `touch`, `chmod`, `chown`, `mkdir`, `ln`, `rsync`, `patch`, `install` — are blocked when any lobster path appears in the command
- `LOBSTER_DEBUG=true` bypasses everything (checked once, up front, before any Bash/Edit/Write dispatch)

**Test results (22 cases, all passing):**
- 10 write patterns correctly blocked (redirects, cp-to-lobster, mv-to-lobster, tee, sed -i, rm, absolute paths)
- 12 read/safe patterns correctly allowed (cat, ls, grep, diff, cp-from-lobster, git status/log/diff/fetch/push, lobster-workspace writes)

**Design note:** The conservative bias (false negatives > false positives) means a sufficiently obfuscated shell command could still slip through — but that is intentional. The goal is to catch accidental or naive writes, not adversarial attacks. The hook mirrors the same pattern as `no-auto-memory.py`.

PASS — ready to merge.

PR #358, opened 2026-03-15
fix: task_id-based dedup + subagent_notification formatter in check_inbox

### LINKED ISSUE #357: fix: task_id-based dedup + subagent_notification formatter in check_inbox
## Problem

Three related bugs cause duplicate messages when subagents use `send_reply` + `write_result`:

### Bug 1: `check_inbox` missing `subagent_notification` branch

In `handle_check_inbox` (~line 2549), the `elif` chain handles `subagent_result`, `subagent_error`, and `subagent_observation` — but not `subagent_notification`. When `forward=False` in `write_result`, the message type is set to `subagent_notification`, but the formatter falls through to the `else` branch and renders it as `[TELEGRAM] from Unknown`. The dispatcher can't distinguish it from a user message.

### Bug 2: `task_id` param in `send_reply` not wired up

The MCP schema documents `task_id` on `send_reply` (lines ~870–878) with the promise of auto-suppressing duplicate `write_result` forwards. But `handle_send_reply` never reads or uses it — the parameter is accepted but silently dropped.

### Bug 3: Text-hash dedup is the wrong primary mechanism

`_was_sent_directly` uses `(chat_id, sha256(text))` as the dedup key. But subagents routinely send a full reply via `send_reply` and a short summary via `write_result(forward=True)` — different texts, so no hash match, so the duplicate goes through.

## Implementation checklist

- [ ] Add `subagent_notification` branch to `check_inbox` formatter with `dispatcher_hint`
- [ ] Add file-backed `_task_replied/` registry at startup
- [ ] Add `_record_task_replied(task_id, chat_id)` and `_was_task_replied(task_id, chat_id)` helpers
- [ ] In `handle_send_reply`: call `_record_task_replied` when `task_id` is present
- [ ] In `handle_write_result`: check `_was_task_replied` before text-hash; if match, suppress forward
- [ ] Update log messages to distinguish task_id dedup from text-hash dedup

sayhar:
Subagents calling `send_reply` followed by `write_result` with `forward=True` (or forgetting `forward=False`) were causing duplicate messages to the user. Three bugs in `inbox_server.py` made the existing defenses ineffective.

## What was broken

**Missing formatter branch:** `check_inbox` had no branch for `subagent_notification` messages — those written when `write_result` is called with `forward=False`. They fell through to `else` and appeared as `[TELEGRAM] from Unknown`. The dispatcher couldn't distinguish them from user messages, so it sometimes forwarded them or tried to reply.

**`task_id` in `send_reply` was a no-op:** The MCP schema documented a `task_id` parameter on `send_reply` with the promise that a matching `write_result` would have `forward` auto-set to `False`. The handler never read the parameter — it was silently dropped.

**Text-hash dedup breaks on differing texts:** The existing `_was_sent_directly` check hashes the exact reply text. Subagents routinely send a full reply via `send_reply` and a one-line summary via `write_result` — different texts, no hash match, duplicate delivered.

## What this fixes

**Fix 1 — formatter:** Added `subagent_notification` branch to `check_inbox` rendering `[SUBAGENT NOTIFICATION]` and emitting `dispatcher_hint: mark_processed only — do NOT call send_reply`. The dispatcher now gets a structural signal, not an inference problem.

**Fix 2 — task_id registry (primary dedup path):** Implemented `_record_task_replied` / `_was_task_replied` backed by `TASK_REPLIED_DIR` (file-per-entry, 5-minute window, same lazy-eviction pattern as `SENT_REPLIES_DIR`). When `send_reply` is called with a `task_id`, the `(task_id, chat_id)` pair is recorded. When `write_result` arrives with the same `task_id`, `forward` is overridden to `False` before any text comparison — regardless of whether the texts match.

**Fix 3 — text-hash demoted:** `_was_sent_directly` is retained as a secondary fallback for cases where `task_id` was not passed to `send_reply`. The task_id registry fires first.

The implementation uses pure functions for the registry helpers (`_task_replied_key`, `_record_task_replied`, `_was_task_replied`) and follows the existing file-backed dedup pattern (no new dependencies, no in-memory state that loses on restart).

This does not change any behavior for messages where `task_id` is absent — all existing code paths are preserved.

Closes #357

## Tests run

- [x] `python3 -c "import ast; ast.parse(open('src/mcp/inbox_server.py').read()); print('syntax OK')"` — syntax OK, no parse errors
- [ ] Integration test against live MCP server — skipped: requires production restart; no behavior change for callers that don't pass `task_id`

**Blocked items needing attention before merge:** none — the live server can be tested after merge on next restart.

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -363,6 +363,7 @@ CONFIG_DIR = BASE_DIR / "config"
 AUDIO_DIR = BASE_DIR / "audio"
 SENT_DIR = BASE_DIR / "sent"
 SENT_REPLIES_DIR = BASE_DIR / "sent-replies"
+TASK_REPLIED_DIR = BASE_DIR / "task-replied"
 TASKS_FILE = BASE_DIR / "tasks.json"
 TASK_OUTPUTS_DIR = BASE_DIR / "task-outputs"
 BISQUE_OUTBOX_DIR = BASE_DIR / "bisque-outbox"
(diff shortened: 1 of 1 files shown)


sayhar (commented):
PASS: all three fixes are correct and complete.

- Fix 1 (formatter): `subagent_notification` branch added at the right position in the `elif` chain. The `dispatcher_hint` line gives the dispatcher an unambiguous structural signal — no heuristic needed. Status icon correctly reflects the `status` field.

- Fix 2 (task_id registry): `_record_task_replied` / `_was_task_replied` follow the exact same file-backed, lazy-eviction pattern as the existing `SENT_REPLIES_DIR` mechanism. The 5-minute window is appropriate: longer than the 60-second text-hash window to cover slow subagents, but bounded to prevent unbounded growth. `TASK_REPLIED_DIR` is correctly added to the startup directory-creation loop. The hook in `handle_send_reply` reads `args.get("task_id")` with a safe default, consistent with how other optional params are read.

- Fix 3 (dedup ordering): In `handle_write_result`, task_id check fires before text-hash check. Both are guarded by `if forward` so a subagent that explicitly passes `forward=False` still skips both checks (correct — no wasted I/O).

No regressions: callers that don't pass `task_id` to `send_reply` are unaffected. All existing dedup behavior is preserved as a fallback.

aeschylus:
Closing: the task_id dedup + subagent_notification formatter changes from this PR were cherry-picked into feat/bis-78-slice-a-backend and are now merged to main via PR #359 (merge commit 11757a0). All changes from this PR are present in main.

PR #361, opened 2026-03-15
fix: remove processed/ fallback from auto-threading in send_reply

sayhar:
## Problem

When `send_reply` was called without an explicit `reply_to_message_id` and no message was actively in processing/ for that chat, the MCP server would scan up to 50 recently-processed messages and thread the reply under whichever one it found most recently — regardless of whether that message was related to the current conversation. This caused replies (especially from scheduled jobs or delayed subagents) to appear as thread replies under completely unrelated messages.

## Fix

Drop Pass 2 of `_lookup_telegram_message_id_for_chat` entirely. The function now only scans `processing/`. If nothing is found there (no currently in-flight message for this chat), it returns `None` and `handle_send_reply` sends the reply as a standalone message.

The behavior table after this change:
- Explicit `reply_to_message_id` provided → thread under that message (unchanged)
- Message for this chat is in `processing/` → thread under it (unchanged)
- Nothing in `processing/` for this chat → standalone message (was: arbitrary thread from `processed/`)

The `_PROCESSED_SCAN_LIMIT = 50` constant is also removed as it is no longer referenced. The tool schema description and inline comments are updated to match.

No other send_reply behavior is changed.

## Tests run
- [x] `git diff` reviewed — change is exactly the intended three-part removal (Pass 2 scan, constant, associated log statement)
- [ ] Live Telegram test — requires production restart; no behavior change for the common case (processing/ hit), only the fallback path changes

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -939,7 +939,7 @@ async def list_tools() -> list[Tool]:
                     },
                     "reply_to_message_id": {
                         "type": "integer",
-                        "description": "Telegram message ID to reply to (Telegram only). If provided, threads the reply against that specific message. If omitted, threading is automatic: the server looks up the telegram_message_id from any message currently in processing/ for this chat_id and threads against it. You rarely need to set this explicitly.",
+                        "description": "Telegram message ID to reply to (Telegram only). If provided, threads the reply against that specific message. If omitted, the server checks processing/ for a currently in-flight message for this chat_id and threads against it if found; otherwise the reply is sent standalone (no thread). You rarely need to set this explicitly.",
                     },
                     "message_id": {
                         "type": "string",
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Reviewed the diff.

**Verdict: safe to merge.**

The change is minimal and correct. `_lookup_telegram_message_id_for_chat` previously had two passes: processing/ (right) and processed/ (wrong). Pass 2 is cleanly removed — no other code paths are affected, `_PROCESSED_SCAN_LIMIT` is the only symbol deleted, and the function's single-pass logic is unchanged.

The only scenario this alters is the fallback case: a subagent or scheduled job calling `send_reply` when no message is actively in flight for that chat. Previously that produced a spurious thread attachment; now it produces a standalone message. That is strictly better behavior.

The common case — dispatcher calling `send_reply` while the original message is still in processing/ — is completely unaffected.

Issue #367, opened 2026-03-15
bug: CLAUDE.md path warning in staging — script looks in ~/lobster-workspace/ instead of ~/lobster/


sayhar:
## Summary

The staging entrypoint script warns that `CLAUDE.md` is missing from `~/lobster-workspace/`. The file actually lives in `~/lobster/` (the repo root). Claude Code finds it via directory traversal, so functionality is not broken, but the warning is confusing and the resolution path is fragile.

## Symptom

```
WARNING: CLAUDE.md not found in ~/lobster-workspace/
```
(or similar) — logged during container startup.

Claude Code does eventually load CLAUDE.md by walking up the directory tree from the workspace, but this is implicit and not robust to working-directory changes.

## Root cause

The entrypoint checks for CLAUDE.md in `$LOBSTER_WORKSPACE` but the canonical location is `$LOBSTER_REPO` (i.e., `~/lobster/`).

## Fix

Add a symlink in `entrypoint-staging.sh` (one line, same PR as Bug 2 / #366):

```bash
ln -sf ~/lobster/CLAUDE.md ~/lobster-workspace/CLAUDE.md
```

This makes the path explicit and silences the warning.

## Related

- Can be fixed in the same PR as Bug 2 (#366 — missing .claude/ bootup files)

## Reference

Full diagnostic report: `~/lobster-workspace/staging-diagnostic-2026-03-15.md` (on staging host)

aeschylus:
Keeping this open. No comments indicate the CLAUDE.md path warning in staging has been fixed or addressed.

aeschylus:
Closing as part of the backlog reset — this appears already addressed by PR #380 ("fix: add --container-setup to install.sh; call from staging entrypoint (#366 #367 #368)"), which is merged. Reset bar: if this tracker were deleted right now, who would notice or be blocked? Nobody — the underlying fix already shipped. Re-open if the fix regresses or doesn't fully cover this issue's scope.

PR #380, opened 2026-03-15
fix: add --container-setup to install.sh; call from staging entrypoint (#366 #367 #368)

### LINKED ISSUE #366: bug: entrypoint-staging.sh skips install.sh — .claude/ bootup files never loaded
## Summary

\`entrypoint-staging.sh\` does not run \`install.sh\`, so the \`.claude/\` directory setup that populates behavioral bootup files (dispatcher.bootup.md, base.bootup.md, subagent.bootup.md, etc.) never executes. The staging bot starts without any behavioral context.

## Environment

- Container: staging Docker (entrypoint-staging.sh)
- Affected files: \`.claude/dispatcher.bootup.md\`, \`.claude/subagent.bootup.md\`, \`~/lobster-user-config/agents/base.bootup.md\`, etc.

## Symptoms

- Dispatcher/subagent boot without role-specific instructions
- Behavioral preferences, identity rules, and context not loaded
- Staging bot behaves differently from production in ways that mask real bugs

## Root cause

\`entrypoint-staging.sh\` was written to skip the full \`install.sh\` for speed, but this also skips the \`.claude/\` context setup step. The test entrypoint (\`entrypoint-test.sh\`) has the same gap.

## Fix

Two options (can be done in same PR):

**Option A:** Add the \`.claude/\` setup steps inline in \`entrypoint-staging.sh\`:
\`\`\`bash
# After repo clone / workspace setup
mkdir -p ~/.claude
cp ~/lobster/.claude/*.md ~/.claude/
# (or symlink)
\`\`\`

**Option B:** Mount \`.claude/\` as a volume in \`docker-compose.staging.yml\`.

A symlink approach (one line) is cleanest and aligns with how CLAUDE.md is resolved.

## Related

- Bug 3 (#367 — CLAUDE.md path warning) can be fixed in the same PR
- Bug 4 (#368 — sqlite_vec arch mismatch) can also be bundled

## Reference

Full diagnostic report: \`~/lobster-workspace/staging-diagnostic-2026-03-15.md\` (on staging host)

### LINKED ISSUE #368: bug: sqlite_vec ELFCLASS32 mismatch on aarch64 — VectorMemory fails in staging container
## Summary

The `sqlite-vec` shared library installed in the staging container is a 32-bit ARM build, but the container host is aarch64 (64-bit). Loading the extension fails with an `ELFCLASS32` error, causing `VectorMemory` to fail and fall back to file-based memory.

## Environment

- Host architecture: aarch64 (ARM 64-bit)
- Installed `.so`: 32-bit ARM (ELFCLASS32)
- Effect: `VectorMemory` initialization fails; silent fallback to `StaticMemory` / file-based

## Error

```
OSError: .../sqlite_vec.so: wrong ELF class: ELFCLASS32
```
(or similar — exact message from staging logs)

## Root cause

`uv pip install sqlite-vec` resolved the 32-bit wheel (armv7 / armhf) rather than the aarch64 wheel. This can happen when pip's platform tag detection falls back to a generic ARM tag.

## Fix

Force reinstall with the correct architecture:

```bash
uv pip install sqlite-vec --force-reinstall --platform manylinux_2_17_aarch64 --only-binary=:all:
```

Or add an explicit architecture guard in `entrypoint-staging.sh`:

```bash
ARCH=$(uname -m)
if [ "$ARCH" = "aarch64" ]; then
  uv pip install sqlite-vec --force-reinstall
fi
```

Can be bundled in the same PR as Bug 2 (#366) and Bug 3 (#367).

## Related

- Existing Docker sqlite-vec issue: #308 (TestVectorMemory fails in Docker — different root cause but same package)

## Reference

Full diagnostic report: `~/lobster-workspace/staging-diagnostic-2026-03-15.md` (on staging host)

sayhar:
## Summary

Previous PRs #376 and #378 were rejected for adding manual symlinks directly in `entrypoint-staging.sh`. The correct fix is to make `install.sh` the single source of truth for setup logic, and have the staging entrypoint call it.

- Adds `--container-setup` flag to `install.sh` that runs only the user-space setup steps safe in Docker (no sudo, no OS packages, no systemd services, no Claude auth)
- Updates `entrypoint-staging.sh` to call `install.sh --container-setup` instead of duplicating setup logic
- Commits the `docker/staging/` directory which was missing from version control

## What `--container-setup` runs

- Creates runtime directories and canonical memory templates
- Creates stub user-config agent files (`base.bootup.md`, etc.)
- Verifies sqlite-vec loads correctly; force-reinstalls on failure (fixes #368 ELFCLASS32 on aarch64)
- Creates `CLAUDE.md` and `.claude/` symlinks in `$WORKSPACE_DIR` (fixes #366, #367)

## Issues resolved

- **#366** (missing `.claude/` bootup files): `install.sh --container-setup` creates the `$WORKSPACE_DIR/.claude -> $LOBSTER_DIR/.claude` symlink, so bootup files are loaded on every container start
- **#367** (CLAUDE.md path warning): `install.sh --container-setup` creates the `$WORKSPACE_DIR/CLAUDE.md -> $LOBSTER_DIR/CLAUDE.md` symlink, making CC discovery explicit
- **#368** (sqlite_vec ELFCLASS32 on aarch64): `install.sh --container-setup` verifies sqlite-vec loads and force-reinstalls if not, using the same logic as the full installer

## Why not the manual symlink approach (PRs #376, #378)

Manual symlinks in `entrypoint-staging.sh` duplicate setup logic that lives in `install.sh`. If `install.sh` adds new setup steps (new bootup files, new directories, etc.), the staging entrypoint would silently miss them. By delegating to `install.sh --container-setup`, the staging container always runs the canonical setup.

## Test plan

- [ ] Build staging container: `docker compose -f docker/staging/docker-compose.staging.yml build`
- [ ] Start staging: `docker compose -f docker/staging/docker-compose.staging.yml up`
- [ ] Verify `[staging] Container setup complete.` in startup logs
- [ ] Confirm `$WORKSPACE_DIR/CLAUDE.md` and `$WORKSPACE_DIR/.claude` symlinks exist in container
- [ ] On aarch64: verify sqlite-vec loads without ELFCLASS32 error in logs
- [ ] Confirm dispatcher boots with correct behavioral context (bootup files loaded)


sayhar (commented):
Reviewed the implementation. The approach is correct and clean.

**What this PR does right:**
- install.sh is now the single source of truth for setup logic. The staging entrypoint delegates to it rather than duplicating symlink/directory/template creation.
- The --container-setup block exits before reaching pre-flight checks (sudo, apt, services, auth), so it is safe to call from a container running as an unprivileged user.
- The sqlite-vec fix is identical to what the full installer does (verify load, fall back to alpha, warn on failure) — no divergence.
- The CLAUDE.md and .claude/ symlinks use the same make_symlink() helper defined in the same block, matching the logic at install.sh lines 2173-2204.
- docker/staging/ is now in version control.

**One note on #367:** The issue says the entrypoint warns about CLAUDE.md missing from `~/lobster-workspace/`. That warning appears to be emitted by an entrypoint check, not by Claude Code itself. This PR fixes the underlying condition (symlink created at startup) so the path is explicit and Claude Code finds the file correctly. If a separate shell check emits the warning, a follow-up can remove that check.

**Verdict:** Ready to merge. The approach correctly rejects the manual-symlink pattern from PRs #376 and #378.

PR #403, opened 2026-03-15
fix: reconciler checks output_file directly, eliminating ghost agents

### LINKED ISSUE #400: bug: reconciler scans wrong/hardcoded tasks path — ghost agents never auto-closed
## Summary

The `reconcile_agent_sessions` background task in `inbox_server.py` calls `scan_agent_outputs()` which reads from a hardcoded default path that does not exist on this system. As a result, the reconciler always returns an empty scan and **never auto-closes any ghost agents**.

## Root cause

`src/agents/session_store.py` line 603:
```python
_CLAUDE_TASKS_DIR = Path("/tmp/claude-1000/-home-admin-lobster-workspace/tasks")
```

The actual path on this system is `/tmp/claude-1000/-home-lobster-lobster-workspace/tasks`.

But there's a second, deeper problem: Claude Code writes subagent output symlinks into **project-specific subdirectories** like:
```
/tmp/claude-1000/-home-lobster-lobster-workspace/<session-uuid>/tasks/<agent-id>.output
```

The flat `tasks/` directory at the root is an older convention. Most active agents' output files are in project-specific subdirs, which `scan_agent_outputs()` never scans.

Meanwhile, the DB `output_file` column already stores the **full correct path** to each agent's output symlink (e.g. `/tmp/claude-1000/-home-lobster-lobster-workspace/9c1bb32c.../tasks/ab5d1727bca76dde7.output`). The reconciler code extracts only the stem to look up in the scan dict — but the scan dict is empty, so `scan.get(scan_key)` always returns `None`, and the reconciler never transitions any session to `done` or `dead`.

## Evidence

```
# scan_agent_outputs() with default path:
default scan results: 0 items

# Ghost agents aee91517 and a3b9f737 show "done" when correct project dir is scanned:
/tmp/.../ad3ab559.../tasks/  →  a3b9f73722cad7a75: done, aee91517e84c1d5b2: done

# Ghost agent ab5d1727 shows "running" (was killed mid-execution, no end_turn):
/tmp/.../9c1bb32c.../tasks/  →  ab5d1727bca76dde7: running
```

The reconciler has been effectively dead since deployment.

## Proposed fix

Two complementary fixes:

**Fix A (correct the scan):** Update `scan_agent_outputs()` to accept a glob pattern or to walk all project-specific task dirs. Since `output_file` in the DB already contains the correct full path, the reconciler could skip the directory scan entirely and directly check each registered session's `output_file`.

**Fix B (direct file check in reconciler):** In `reconcile_agent_sessions()`, for each active session, directly resolve `session.output_file` → read its JSONL tail → check `stop_reason` — instead of relying on the pre-built scan dict. This eliminates the path problem entirely and is O(n) over active sessions.

Fix B is cleaner and removes the need for `_CLAUDE_TASKS_DIR` as a hardcoded constant.

## Impact

Every ghost agent since the reconciler was introduced has remained in `running` state indefinitely. The ghost detector script is the only thing catching these. The fix would make ghost cleanup automatic within 5–30 seconds of agent completion.

## Files

- `src/agents/session_store.py` — hardcoded `_CLAUDE_TASKS_DIR`, `scan_agent_outputs()`
- `src/mcp/inbox_server.py` — `reconcile_agent_sessions()` (lines ~6220–6345)

sayhar:
## Summary

- Replaces the broken `scan_agent_outputs()` directory scan in the reconciler with a direct per-session file check via new `check_output_file_status()` function
- The old approach scanned a hardcoded path (`/tmp/claude-1000/-home-admin-lobster-workspace/tasks`) that doesn't exist — the reconciler was silently broken from the start
- Claude Code places output symlinks in project-specific subdirs; the DB `output_file` column already has the correct path, so reading it directly is both simpler and correct

Closes #400

## Root cause

`_CLAUDE_TASKS_DIR` in `session_store.py` was hardcoded with the wrong username (`-home-admin-` instead of `-home-lobster-`) AND the wrong directory layout (flat tasks dir vs project-specific subdirs). Both problems made the scan return zero results on every 5-second cycle.

## Changes

- `src/agents/session_store.py`: Add `check_output_file_status(output_file)` that reads the stop_reason from the exact path stored in the DB. Extract shared `_read_stop_reason_from_path()` helper used by both new function and legacy `scan_agent_outputs()`.
- `src/mcp/inbox_server.py`: Update `reconcile_agent_sessions()` to call `check_output_file_status()` per session instead of scanning a directory. Remove the orphan-detection phase that depended on the now-absent scan dict.

## Test plan

- [x] All 22 existing `test_agent_sessions.py` tests pass
- [x] Smoke test: `check_output_file_status()` correctly returns `done` for agents with `end_turn` in their JSONL, `running` for agents mid-execution, and `missing` for nonexistent paths
- [ ] Manual verification after deploy: ghost agents `a3b9f737` and `aee91517` (both `end_turn`, no `write_result`) should auto-close to `completed` within 5 seconds of the server restart

### DIFF
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -599,17 +599,85 @@ def format_active_sessions_block(sessions: list[dict]) -> str:
 # Ground-truth scanner — reads Claude Code JSONL output files
 # ---------------------------------------------------------------------------
 
-# Default tasks symlink directory where Claude Code writes *.output symlinks
+# Default tasks symlink directory where Claude Code writes *.output symlinks.
+# NOTE: This path was historically hardcoded with the wrong username (-home-admin-
+# instead of the actual user). The reconciler now uses check_output_file_status()
+# for per-session checks based on the output_file stored in the DB, which avoids
+# this path entirely. scan_agent_outputs() is kept for tests and legacy callers.
 _CLAUDE_TASKS_DIR = Path("/tmp/claude-1000/-home-admin-lobster-workspace/tasks")
+
+# Compiled stop_reason pattern — shared by scan_agent_outputs and check_output_file_status
+import re as _re
+_STOP_REASON_RE = _re.compile(rb'"stop_reason"\s*:\s*"([^"]+)"')
+
+
+def _read_stop_reason_from_path(output_path: Path) -> str:
+    """Read the last stop_reason from a .output symlink or JSONL file.
+
+    Returns one of:
+      ``"running"``  — file exists, last stop_reason is "tool_use" or not yet written
+      ``"done"``     — last stop_reason is "end_turn"
+      ``"missing"``  — symlink target does not exist or file is unreadable
+    """
+    try:
+        real_path = output_path.resolve()
+    except OSError:
+        return "missing"
+
+    if not real_path.exists():
+        return "missing"
+
+    try:
+        with open(real_path, "rb") as f:
+            try:
+                f.seek(-4096, 2)
+            except OSError:
+                f.seek(0)
+            tail = f.read()
+    except OSError:
+        return "missing"
+
+    if not tail:
+        return "running"
+
+    matches = _STOP_REASON_RE.findall(tail)
+    if not matches:
+        return "running"
+
+    last_reason = matches[-1].decode("utf-8", errors="replace")
+    return "done" if last_reason == "end_turn" else "running"
+
+
+def check_output_file_status(output_file: str) -> str:
+    """Check liveness of a single agent by reading its output_file path directly.
+
+    This is the preferred method for the reconciler: it reads the path stored
+    in the ``output_file`` DB column, which is always correct regardless of
+    username or session layout. It avoids the directory-scan approach that
+    relied on a hardcoded (and often wrong) default tasks path.
+
+    Args:
+        output_file: Full path to the agent's .output symlink as stored in DB.
+
+    Returns:
+        ``"running"``  — agent is still executing (last stop_reason is tool_use
+                         or no stop_reason written yet).
+        ``"done"``     — agent has finished (last stop_reason is end_turn).
+        ``"missing"``  — output file does not exist (agent not yet started,
+                         was killed, or output_file was not registered).
+    """
+    if not output_file:
+        return "missing"
+    return _read_stop_reason_from_path(Path(output_file))
 
 
 def scan_agent_outputs(
     tasks_dir: Path | None = None,
 ) -> dict[str, str]:
-    """Scan Claude Code agent output files and return their liveness status.
+    """Scan a Claude Code tasks directory and return liveness status per agent.
 
     Each background Task spawned by Claude Code creates a symlink (or file) at:
-        /tmp/claude-1000/-home-admin-lobster-workspace/tasks/<agent-id>.output
+        <tasks_dir>/<agent-id>.output
 
     The symlink resolves to a JSONL file in ~/.claude/projects/.../subagents/.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Review: fix-reconciler-path

**Verdict: PASS** (with minor notes)

### What the fix does

The PR replaces a broken directory scan in `reconcile_agent_sessions()` that relied on a hardcoded path (`/tmp/claude-1000/-home-admin-lobster-workspace/tasks`) with a direct per-session file check using `check_output_file_status(output_file)`. Because the `output_file` column in the DB already holds the correct absolute path registered at session-start time, reading it directly is both simpler and correct.

### Correctness

The core logic is correct. Key observations:

1. **`_read_stop_reason_from_path` is a clean extraction.** The tail-read heuristic (last 4 KB, last occurrence of `stop_reason`) is unchanged from the original `scan_agent_outputs()` implementation — only factored out into a shared helper. No regression risk here.

2. **`check_output_file_status(None)` is guarded.** The `if not output_file: return "missing"` guard handles the case where `output_file` was never registered. The dead-threshold path then applies as before — correct behavior.

3. **The reconciler loop is correct.** `file_status` is checked per-session, `session_end()` is called for `"done"` and for `"missing"` past the dead threshold. The `"running"` case is a no-op. No logic regressions.

4. **Module-level `import re`.** The `import re as _re` moved to module level, which is the right place. The previous inline `import re` inside `scan_agent_outputs()` on every call was wasteful; this is a micro-improvement.

### Edge cases evaluated

- **Empty `output_file` string**: Handled — `if not output_file` catches both `None` and `""`.
- **Symlink that resolves but target is deleted**: `real_path.exists()` returns `False` → `"missing"`. Correct.
- **OSError on `.resolve()`**: Caught, returns `"missing"`. Correct.
- **File < 4 KB**: `seek(-4096, 2)` raises `OSError`, caught with `f.seek(0)`. Correct.
- **JSONL truncated mid-line in the 4 KB tail**: Regex only captures complete `"stop_reason":"..."` patterns, so partial writes are safely ignored. Correct.
- **`stop_reason` values other than `end_turn` or `tool_use`**: All treated as `"running"`. Acceptable — the contract is `"done"` only on `end_turn`.

### Removed orphan detection — is it safe?

The old Phase 3 orphan scan detected agents that had output files on disk but no DB entry. That logic was predicated on `scan_agent_outputs()` returning data, which — per the bug — it never did. Removing it changes nothing observable in production. The comment left at the removal site acknowledges the gap: an output_file-less session past the threshold will still be caught by the `"missing"` + dead-threshold path.

One thing worth noting: if a session is registered with `output_file=None` (i.e., `session_start()` was called without `output_file`), it will never be auto-closed as `"done"` — only as `"dead"` via the threshold. This was already the case before this PR; it's not a regression, but worth keeping in mind for future session_start callers.

### Minor notes (non-blocking)

1. The `# Phase 3: Detect orphans` comment at line 6305 in `inbox_server.py` is now a dead comment with no following code. It can be removed in a follow-up to avoid confusion.
2. The test suite (`test_agent_sessions.py`) covers the session lifecycle, DB migration, and tracker adapter well. There are no tests specifically for `check_output_file_status()` or `_read_stop_reason_from_path()`. The PR author noted a smoke test was done manually. A unit test for the new helper function would be a nice addition but is not blocking.

### Summary

The fix is targeted, the logic is sound, and edge cases are well-handled. The root cause (hardcoded wrong path) is clearly identified and the solution (read from DB directly) is the correct approach. All 22 existing tests pass. Ready to merge.

PR #408, opened 2026-03-15
feat: OOM kill monitor with deduplication and dual-alert paths

sayhar:
## Problem

When the Linux OOM killer terminates a Claude or Python process, Lobster subagents become ghosts — their sessions remain \`running\` in \`agent_sessions.db\` but the process is dead. There was no mechanism to alert Sahar that this had happened, so ghost cleanup (via the reconciler, PR #403) would occur silently without any visibility.

## What this adds

\`scripts/oom-monitor.py\` — a cron-runnable script that scans the kernel journal every 10 minutes for OOM kill events and sends a Telegram alert when a Lobster-related process is killed.

**Kernel log patterns handled:**
- \`Out of memory: Killed process <PID> (<name>)\` — classic form
- \`Out of memory: Kill process <PID> (<name>)\` — newer kernel variant (no -ed suffix)
- \`Memory cgroup out of memory: Killed process <PID> (<name>)\` — cgroup-scoped kills
- \`oom_reaper: reaped process <PID> (<name>)\` — reaper confirmation line

**Processes watched:** \`claude\`, \`python\`, \`python3\`, \`node\`, \`lobster\`, \`uv\`

## Design

Pure functions handle parsing and classification. Side effects — journal scanning, Telegram delivery, state persistence, inbox writes — are isolated at the system boundary.

**Deduplication:** each kill event is keyed by \`sha256(timestamp + pid)[:16]\` and stored in \`~/lobster-workspace/data/oom-monitor-state.json\`. Running the monitor multiple times for the same event sends at most one alert.

**Dual alert path (belt-and-suspenders):**
1. Direct curl to the Telegram Bot API — same approach as \`health-check-v3.sh\`, works even if Lobster itself was the killed process.
2. JSON drop to \`~/messages/inbox/\` — dispatcher picks this up when it recovers and can relay context to the user.

**Cron setup (every 10 minutes):**
\`\`\`
*/10 * * * * uv run $HOME/lobster/scripts/oom-monitor.py --since-minutes 10
\`\`\`

The \`--since-minutes\` window matches the cron interval so there are no gaps and no double-counting.

## What is not changed

- The ghost-detector (\`scripts/ghost-detector.py\`) is not modified. The OOM monitor is a notification layer; ghost cleanup remains the reconciler's job (#403).
- No changes to \`health-check-v3.sh\` or any existing monitoring scripts.
- No new runtime dependencies.

## Tests run

- [x] \`uv run pytest tests/unit/test_oom_monitor.py -v\` — 29 passed, 0 failed
- [x] \`uv run scripts/oom-monitor.py --dry-run --since-minutes 43200\` — scanned 30 days of journal, exit 0 (no OOM events on this host, correct behavior)
- [ ] Live Telegram alert delivery — blocked: no live OOM events to trigger against and injecting kernel log entries is not safe in this environment. The curl path is the same pattern proven in \`health-check-v3.sh\`.
- [ ] Inbox message routing — blocked: requires a running dispatcher session. The JSON format matches existing \`outbound\` messages that the dispatcher already handles.

**Blocked items needing attention before merge:** none — the untested paths use patterns proven elsewhere in the codebase.

### DIFF
--- a/scripts/oom-monitor.py
+++ b/scripts/oom-monitor.py
@@ -2,25 +2,27 @@
 """OOM Kill Monitor — detects when the Linux OOM killer kills Lobster/Claude.
 
 Scans the kernel journal for OOM kill events involving claude/python/lobster
-processes. On detection, sends a direct Telegram alert and writes an alert to
-the Lobster inbox so the dispatcher can notify the user. Uses a state file to
-avoid duplicate alerts for the same kill event.
+processes. On detection, writes an alert to the Lobster inbox so the dispatcher
+can notify the user via whatever messaging platform is active. Uses a state file
+to avoid duplicate alerts for the same kill event.
+
+**Debug-mode gate:** This monitor only runs when LOBSTER_DEBUG=true. If that
+environment variable is absent or not "true", the script exits immediately
+with code 0 (no-op). This prevents unnecessary journal scanning in production
+unless debug monitoring is explicitly enabled.
 
 Usage:
-    uv run scripts/oom-monitor.py
-    uv run scripts/oom-monitor.py --since-minutes 10
-    uv run scripts/oom-monitor.py --dry-run
+    LOBSTER_DEBUG=true uv run scripts/oom-monitor.py
+    LOBSTER_DEBUG=true uv run scripts/oom-monitor.py --since-minutes 10
+    LOBSTER_DEBUG=true uv run scripts/oom-monitor.py --dry-run
 
 Design:
     - Pure functions for parsing and classification; side effects isolated at edges
     - State file tracks seen events (keyed by timestamp+pid) to prevent duplicates
-    - Two alert paths:
-        1. Direct Telegram via curl (like health-check-v3.sh) — works even if
-           dispatcher is down (because the OOM may have killed it)
-        2. Inbox JSON drop for dispatcher pickup when it recovers
+    - Alert path: inbox JSON drop for dispatcher pickup (platform-agnostic)
 
 Exit codes:
-    0 — no OOM kills detected
+    0 — no OOM kills detected (or LOBSTER_DEBUG not set)
     1 — OOM kill(s) detected (alert sent)
     2 — error (journalctl unavailable, config missing, etc.)
 """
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Code Review — PASS

**Verdict: PASS with one optional improvement noted below.**

### What I reviewed

`scripts/oom-monitor.py` and `tests/unit/test_oom_monitor.py`. Checked against the existing codebase patterns (ghost-detector.py, health-check-v3.sh, alert.sh).

### Strengths

**Functional decomposition is clean.** `parse_oom_event`, `filter_new_events`, `format_telegram_alert`, and `format_inbox_message` are all pure. The only side effects are in `scan_journal`, `send_telegram_direct`, `write_inbox_message`, `load_state`, and `save_state` — and they're clearly named as such. This mirrors the ghost-detector.py pattern well.

**Deduplication is correct.** `sha256(timestamp + pid)[:16]` as the event key is stable across runs and collision-resistant for this use case. The state file uses an atomic rename (`tmp.replace(dest)`) which prevents partial writes — good.

**Four kernel log variants covered.** The regex handles both `Killed process` and `Kill process` (newer kernels skip the -ed suffix), cgroup-prefixed messages, and `oom_reaper` confirmation lines. This is the right set based on kernel source and observed log formats.

**Dual alert path is sound.** The direct curl path (same as health-check-v3.sh) is specifically valuable here because OOM may have killed the Lobster process itself. The inbox drop as a secondary path is belt-and-suspenders. Both paths are best-effort and log on failure rather than crashing.

**Test coverage is solid for the pure layer.** 29 tests cover all four parse patterns, deduplication, alert formatting, inbox message structure, and state I/O including corruption handling. The importlib approach to loading the dash-named script is a bit unusual but correct.

### One thing to consider

The `--since-minutes` argument defaults to 10 to match the assumed cron interval. If someone runs the script manually without `--dry-run` and without `--since-minutes`, they'll scan only the last 10 minutes. The dry-run help text in the docstring mentions `--since-minutes 43200` for historical scans, but the default for cron-aligned use is sensible. This is fine as-is; a note in the HEALTH-MONITORING.md cron section could clarify, but it's not blocking.

### Not blocking

- Live Telegram delivery and inbox routing are untested by automated tests, but both code paths use the same patterns (`curl -X POST .../sendMessage` and writing JSON to `~/messages/inbox/`) that are proven in health-check-v3.sh and ghost-detector.py respectively. The risk is low.
- The `node` process name in `LOBSTER_PROCESS_NAMES` is a reasonable inclusion (MCP servers run as Node), but in practice a `node` OOM kill might not always indicate Lobster was affected. The alert text is accurate ("may have become ghosts") so this is fine.

**PASS — safe to merge after manual cron registration.**

PR #414, opened 2026-03-15
feat: auto-register subagents in agent_sessions.db via SessionStart hook

### LINKED ISSUE #412: feat: auto-register subagents in agent_sessions.db via SessionStart hook
## Problem

Subagent sessions are only tracked in `agent_sessions.db` if the dispatcher manually calls `register_agent`. If the dispatcher forgets (e.g. after context compaction), agents are invisible to the ghost detector and reconciler.

## Solution

Extend the `write-dispatcher-session-id.py` SessionStart hook to auto-register subagents directly into `agent_sessions.db` without any dispatcher involvement.

## How it should work

1. `write-dispatcher-session-id.py` already fires on `SessionStart` for ALL sessions
2. When the session is a subagent (not dispatcher), write a minimal record to `agent_sessions.db`:
   - `id` = session_id from hook input
   - `status` = 'running'
   - `description` = 'auto-registered by SessionStart hook'
   - `agent_type` = 'unknown'
   - `spawned_at` = current timestamp
3. Use `session_role.is_dispatcher()` to skip this for the dispatcher session
4. Write directly to SQLite (hooks can't call MCP tools)
5. Handle gracefully: if DB write fails, log to stderr but don't block session start (exit 0)

## Implementation plan

- [ ] Read existing hook and session_store schema
- [ ] Add auto-registration logic to `write-dispatcher-session-id.py`
- [ ] Test the logic
- [ ] Open PR

sayhar:
## Summary

Subagents were invisible to the ghost detector unless the dispatcher explicitly called \`register_agent\` — which it silently fails to do after context compaction. This PR closes that gap by registering every subagent at the moment its session opens, with no dispatcher involvement required.

The \`write-dispatcher-session-id.py\` SessionStart hook already fires for all sessions inheriting \`LOBSTER_MAIN_SESSION=1\`. The change adds a subagent branch that writes a minimal stub row to \`agent_sessions.db\` when the hook detects it is running in a subagent context. The dispatcher branch (marker file write) is unchanged.

**Detection logic** (self-contained, no import of \`session_role.py\` needed):
- Marker file absent → this is the dispatcher's first start; write marker file.
- Marker file present, session_id matches → dispatcher reconnect; write marker file.
- Marker file present, session_id differs → subagent; write stub row to DB.

**Stub row design:** \`INSERT OR IGNORE\` ensures a racing \`register_agent\` call that arrives before the hook completes wins cleanly. The stub carries only \`id\`, \`status='running'\`, \`agent_type='unknown'\`, and \`spawned_at\`. The dispatcher's \`register_agent\` (which uses \`INSERT OR REPLACE\`) will overwrite the stub with a richer record when it runs. If the dispatcher never calls \`register_agent\`, the stub alone is enough for the ghost detector to see the session.

**\`output_file\` is left NULL** in the stub. The ghost detector uses \`check_output_file_status(output_file)\` for per-session liveness checks, so sessions with \`output_file=NULL\` degrade gracefully to \`"missing"\` status rather than crashing.

Functional patterns used: pure predicate functions (\`_is_dispatcher_session\`, \`_read_stored_dispatcher_id\`), side effects isolated to \`_write_session_id\` and \`_auto_register_subagent\`, graceful degradation via broad exception catch with stderr logging.

## Tests run

- [x] Manual scenario tests (5 cases via subprocess): fresh dispatcher start, subagent start, dispatcher reconnect, \`LOBSTER_MAIN_SESSION\` unset, INSERT OR IGNORE preserves richer row — all passed
- [x] \`uv run pytest tests/test_agent_sessions.py -v\` — 22 passed, 0 failed
- [ ] Live hook execution against a real subagent session — skipped: requires production restart to pick up hook changes; safe to merge, hook failure path exits 0

Closes #412

### DIFF
--- a/hooks/write-dispatcher-session-id.py
+++ b/hooks/write-dispatcher-session-id.py
@@ -17,13 +17,12 @@ that other hooks use to distinguish the dispatcher from subagents.
 
 When the marker file exists and the current session_id does NOT match the
 stored dispatcher ID, this session is a subagent. In that case, a minimal
-record is written to agent_sessions.db with status='running'. This ensures
-subagents are visible to the ghost detector even when the dispatcher forgets
-to call register_agent (e.g. after context compaction). The record is
-intentionally sparse — description, agent_type, and output_file can be
-filled in later by the dispatcher's register_agent call, which uses INSERT
-OR REPLACE and will overwrite this stub. DB write failures are logged to
-stderr but never block session start (always exits 0).
+stub record is written to agent_sessions.db with status='running'. This
+ensures subagents are visible to the ghost detector even when the dispatcher
+forgets to call register_agent (e.g. after context compaction). The stub uses
+INSERT OR IGNORE so a racing register_agent call that writes a richer row
+first is preserved untouched. DB write failures are logged to stderr but
+never block session start (always exits 0).
 
 ## settings.json configuration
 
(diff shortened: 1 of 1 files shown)


sayhar (commented):
PASS — Elegance fix applied.

**Before:** 211 lines, ~80 lines of inline DDL (CREATE TABLE, PRAGMAs, INSERT), two re-implementations of logic already in the codebase (`_read_stored_dispatcher_id` duplicating `session_role._read_dispatcher_session_id`, `_write_session_id` duplicating `session_role.write_dispatcher_session_id`), empty string passed for NOT NULL `chat_id`.

**After:** 132 lines, zero inline DDL, zero duplication. Fixed by:
- `sys.path.insert` for both hooks dir and `src/` at module top, resolving the import path issue
- `session_role._read_dispatcher_session_id()` for dispatcher detection
- `session_role.write_dispatcher_session_id()` for the marker file write
- `session_store.init_db()` + `session_store._get_connection()` + `INSERT OR IGNORE` for the subagent stub — `INSERT OR IGNORE` preserved (not `session_store.session_start` which uses `INSERT OR REPLACE`) so a richer `register_agent` row written first is never clobbered
- `'0'` as `chat_id` placeholder instead of empty string; `'subagent'` as `agent_type`

Tests: `uv run pytest tests/test_agent_sessions.py` — 22 passed, 0 failed.

PR #424, opened 2026-03-16
feat: add debug.sys.bootup.md agent supplement for LOBSTER_DEBUG mode

sayhar:
## Summary

- Adds `.claude/debug.sys.bootup.md`, a new system bootup file loaded by all agents when `LOBSTER_DEBUG=true`
- Defines the **install.sh completeness rule**: before any PR is opened, agents must verify that new hooks, cron jobs, services, env vars, and scripts are reflected in `install.sh`
- Adds debug-specific verbosity rules (more detailed `write_result` summaries, `write_observation` for non-obvious decisions, error logging to `observations.log`)
- Adds a pre-PR self-check prompt (install.sh coverage, why-not-what descriptions, test coverage)

## Why this is needed

`LOBSTER_DEBUG=true` already gates several system behaviors (OOM monitor, observation delivery, hook overrides) but there was no agent-level supplement that activated debug-aware reasoning. This file fills that gap and also encodes the install.sh completeness rule as a system-level guardrail rather than tribal knowledge.

## Loading mechanism

**The loading mechanism for this file does not yet exist.** No code currently reads `.claude/debug.sys.bootup.md` based on `LOBSTER_DEBUG`. A companion issue has been filed to wire this up: the suggested approach is to inject the file path into the `CLAUDE.md` system context block (or a `SessionStart` hook) when the env var is set.

See: issue filed separately (link TBD after this PR).

## Test plan

- [ ] Confirm file appears at `.claude/debug.sys.bootup.md` after checkout
- [ ] Manually verify content matches spec
- [ ] Once loading mechanism is wired, confirm agents receive the supplement when `LOBSTER_DEBUG=true` and do not receive it otherwise


sayhar (commented):
PASS

Reviewed `.claude/debug.sys.bootup.md`:

- Content matches spec exactly
- Install.sh completeness rule is well-placed here — it's actionable and unambiguous
- Debug verbosity rules (write_result, write_observation, observations.log) are consistent with existing LOBSTER_DEBUG conventions in hooks/on-compact.py and hooks/system-file-protect.py
- PR self-check prompt is concise and covers the right surface area

One gap noted and tracked: the loading mechanism (how agents actually receive this file when LOBSTER_DEBUG=true) is not wired up. Issue #425 filed for that. The file itself is correct and safe to merge; the feature is inert until #425 lands.

No concerns with merging.

PR #428, opened 2026-03-16
Add PreToolUse hooks to warn on dispatcher-blocking tool calls

### LINKED ISSUE #427: Add PreToolUse hook to warn when Agent called without run_in_background: true
## Problem

The Lobster dispatcher runs in an infinite message-processing loop. When the `Agent` tool is invoked without `run_in_background: true`, the dispatcher blocks for the full duration of the agent call — potentially minutes — during which no incoming messages are processed.

This is a silent footgun: Claude sees no warning, nothing indicates the call is problematic, and the dispatcher simply stalls.

## Fix

Add a `PreToolUse` hook (`hooks/require-background-agent.py`) that fires on every `Agent` tool call and warns when `run_in_background` is not `true`.

**Behavior:**
- If `tool_name` != `"Agent"` → exit 0 (not our concern)
- If `tool_input.run_in_background` is `True` → exit 0 (correct usage)
- Otherwise → exit 1 with a warning message asking Claude to reconsider

Exit 1 is a **soft warning** (Claude sees it and can reconsider), not a hard block. Exit 2 (hard block) is intentionally avoided — there are legitimate cases where a foreground agent is genuinely needed.

## Warning message

```
Warning: Agent tool called without run_in_background: true. This blocks message processing for the duration of the agent. Pass run_in_background: true unless you genuinely need the result synchronously to decide your next step.
```

## Implementation plan

- [ ] Create `hooks/require-background-agent.py`
- [ ] Write unit tests in `tests/smoke/test_require_background_agent.py`
- [ ] Register hook in `install.sh` (same pattern as `require-subagent-type.py`)
- [ ] Deploy locally: copy hook + register in `~/.claude/settings.json`
- [ ] Open PR

sayhar:
## Problem

The dispatcher's message-processing loop can be stalled silently by two categories of inline tool call:

1. **Agent called without `run_in_background: true`** — a foreground agent blocks the dispatcher for the full duration of the agent run, potentially minutes, while incoming messages queue up unprocessed.

2. **WebFetch / WebSearch called inline** — these make network requests that can take 5–30+ seconds. The dispatcher has no legitimate reason to fetch web content directly; that work should always go to a background subagent.

In both cases, Claude receives no warning. The dispatcher simply stalls.

## Fix

Two new `PreToolUse` hooks, both using exit 1 (soft warning, not hard block):

**`hooks/require-background-agent.py`** — fires on every `Agent` call. Exits 0 if `run_in_background` is `True`; exits 1 with a warning otherwise. Exit 2 is intentionally avoided: there are legitimate edge cases where a foreground agent is genuinely needed (when the result is required synchronously to decide the next step).

**`hooks/dispatcher-inline-tool-guard.py`** — fires on `WebFetch` and `WebSearch`. Both always warrant a background subagent; the warning tells Claude to delegate instead. Implemented as a dict-keyed guard so adding future tools requires only a new entry.

The Agent hook registers under the existing `"Agent"` matcher alongside `require-subagent-type.py`. The web guard registers under `"WebFetch|WebSearch"`.

Both hooks are registered in `install.sh` using the same jq-idempotent pattern as `require-subagent-type.py` — re-running install does not add duplicate entries.

## Functional patterns used

Pure functions with no side effects. The `_GUARDED_TOOLS` dict in `dispatcher-inline-tool-guard.py` is a declarative map from tool name to warning rationale — extending coverage requires only a new dict entry, no conditional branching.

## Tests run

- [x] `uv run pytest tests/smoke/test_require_background_agent.py -v` — 4 passed, 0 failed
- [x] `uv run pytest tests/smoke/test_dispatcher_inline_tool_guard.py -v` — 9 passed, 0 failed
- [x] Manual invocation of deployed hooks via stdin — all exit codes and warning messages verified correct

Closes #427

### DIFF
--- a/hooks/dispatcher-inline-tool-guard.py
+++ b/hooks/dispatcher-inline-tool-guard.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+"""PreToolUse hook: warns when tools that should run in background subagents are
+called inline by the dispatcher.
+
+The Lobster dispatcher runs in an infinite message-processing loop. Certain tools
+should never be called inline — they block message processing, sometimes for many
+seconds or minutes. This hook catches the most common cases and steers Claude
+toward delegating to a background subagent instead.
+
+Guarded tools and why:
+
+  WebFetch / WebSearch — network I/O that can take 5–30+ seconds. The dispatcher
+    has no reason to fetch web content directly; if web information is needed it
+    should always go to a background subagent.
+
+Exit codes:
+  0 — tool is not guarded, or usage is acceptable
+  1 — soft warning: Claude sees it and can reconsider, but is not hard-blocked
+
+Exit 1 (not 2) is intentional throughout. There may be edge cases where inline
+usage is genuinely warranted; hard-blocking would be counterproductive.
+"""
+import json
+import sys
+
+# Tools that should always go to a background subagent, never called inline.
+# Map tool name → reason shown in warning message.
+_GUARDED_TOOLS: dict[str, str] = {
+    "WebFetch": (
+        "WebFetch makes a network request that can take 5–30+ seconds, blocking "
+        "the dispatcher's message-processing loop for the full duration. "
+        "Delegate web fetching to a background subagent instead."
+    ),
+    "WebSearch": (
+        "WebSearch makes a network request that can take several seconds, blocking "
+        "the dispatcher's message-processing loop for the full duration. "
+        "Delegate web searches to a background subagent instead."
+    ),
+}
+
+data = json.load(sys.stdin)
+tool = data.get("tool_name", "")
+
+reason = _GUARDED_TOOLS.get(tool)
+if reason is None:
+    sys.exit(0)
+
+print(
+    f"Warning: {tool} called inline by the dispatcher. {reason}",
+    file=sys.stderr,
+)
+sys.exit(1)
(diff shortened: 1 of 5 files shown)


sayhar (commented):
**Review verdict: approve**

Both hooks are correct, minimal, and follow the established hook patterns exactly.

**require-background-agent.py:** Logic is straightforward — reads `run_in_background` from `tool_input`, exits 0 on `True` (strict identity check, not truthiness), exits 1 otherwise. The exit 1 vs exit 2 choice is right: there are legitimate foreground agent uses and hard-blocking would break them. Warning text clearly states the problem and the fix.

**dispatcher-inline-tool-guard.py:** The `_GUARDED_TOOLS` dict is a clean functional pattern — declarative map from tool name to rationale, zero branching, trivially extensible. Each tool gets its own specific message explaining *why* it's guarded (network I/O, timing). Exit 1 is appropriate here too.

**install.sh:** Both registrations use `test("hook-name")` as the idempotency check, which correctly handles the case where the same `"Agent"` matcher already exists for `require-subagent-type.py`. Re-running install will not double-register.

**Tests:** 13 tests total across 4 cases. Each test documents its failure mode — what goes wrong if the hook stays silent or fires spuriously. Parametrized non-guarded tool list in the web guard tests is a good touch.

**Deployment:** Both hooks are deployed to `~/lobster/hooks/` and registered in `~/.claude/settings.json`. Manual invocation confirmed correct exit codes and stderr output.

No concerns. Safe to merge.

PR #431, opened 2026-03-16
feat(auditor): lobster-auditor subagent with enforced context-update hook

### LINKED ISSUE #429: lobster-auditor: dedicated system investigation subagent with self-updating architecture context
## Problem

The dispatcher currently does investigation work inline (reading code, checking logs, running whisper manually) instead of delegating to subagents. There's no dedicated place for system audit knowledge, so each investigation starts from scratch.

## Proposal: \`lobster-auditor\` subagent

A specialized subagent for system investigation with three properties that distinguish it from generic \`lobster-ops\`:

### 1. Investigation-focused context

Deep knowledge of:
- \`ghost-detector.py\`
- Whisper pipeline and transcription worker
- Hooks system
- Docker test infrastructure
- Agent session DB schema

### 2. Living architecture context file

Maintains \`~/lobster-user-config/agents/system-audit.context.md\`. On every run:
- Reads the file first
- Compares to what it observes
- Updates with any divergences found

### 3. Enforced acknowledgment at exit (SubagentStop hook)

Same pattern as \`require-write-result.py\`. The hook checks:
- Was \`system-audit.context.md\` modified during this session? → allow exit
- OR does the agent's transcript contain the safe word \`AUDIT_CONTEXT_UNCHANGED\`? → allow exit (agent explicitly confirmed nothing changed)
- Neither → exit 2 (blocked until agent updates file or emits safe word)

## Debug-mode gating

- Dispatcher only spawns \`lobster-auditor\` when \`LOBSTER_DEBUG=true\`
- Implemented as a behavioral rule in \`debug.sys.bootup.md\` (loaded by \`inject-debug-bootup.py\` at session start)
- No install-time changes needed; flipping debug mode at runtime takes effect on next session

## Files this would create/modify

| File | Change |
|------|--------|
| \`~/lobster/hooks/require-audit-update.py\` | New SubagentStop hook |
| \`~/lobster-user-config/agents/subagents/lobster-auditor.md\` | Subagent definition (private, not in repo) |
| \`~/lobster-user-config/agents/system-audit.context.md\` | Living context file (created on first run) |
| \`.claude/debug.sys.bootup.md\` | Add dispatcher rule for lobster-auditor |
| \`install.sh\` | Register the new SubagentStop hook |

## Out of scope for v1

- Automatic scheduling (not a cron job — spawned on demand when investigation is needed)
- Non-debug mode usage

## Sign-off needed

Architecture-level change — needs Sahar's review before building.

sayhar:
## Summary

Each system investigation starts from scratch today — no accumulated knowledge, no durable record of root causes. This PR introduces the `lobster-auditor` subagent (designed in #429) along with the enforcement hook that makes its context-update contract machine-verifiable.

**What ships in this PR:**

- **`hooks/require-auditor-context-update.py`** — SubagentStop hook that detects when a lobster-auditor session is ending and blocks exit unless the agent either updated `system-audit.context.md` or emitted `AUDIT_CONTEXT_UNCHANGED` as the first line of its `write_result` call. Detection is transcript-based: if any tool call in the session read `system-audit.context.md`, this is treated as an auditor session. This is the same exit-2 enforcement pattern used by `require-write-result.py`.

- **`memory/canonical-templates/system-audit.context.md`** — Seed template with sections for audit history, known anomalies, root causes, and architecture notes. `install.sh` copies this to `~/lobster-user-config/agents/system-audit.context.md` on first run (skipped if the file already exists).

- **`install.sh`** — Two additions: (1) seed copy for `system-audit.context.md` at directory setup time, (2) SubagentStop hook registration for `require-auditor-context-update.py`.

**What does NOT ship in this PR (by design):**

- The subagent definition (`lobster-auditor.md`) lives in user-config, not in the repo — same pattern as all subagent definitions.
- Hook registration in `~/.claude/settings.json` is intentionally omitted pending review. The `install.sh` block is ready; it just needs to run after approval.

## Functional patterns used

Pure functions throughout the hook: all detection logic is expressed as boolean-returning functions over the transcript data structure (`_is_auditor_session`, `_safe_word_in_transcript`, `_context_file_updated_since`). Side effects (the blocking print + sys.exit) are isolated to `main()`.

## Tests run

- [x] `python3 ~/lobster/hooks/require-auditor-context-update.py < /dev/null` — exits 0 (unreadable input → pass through)
- [x] Manual transcript fixture with a Read call to `system-audit.context.md` and no safe word → exits 2 with expected message
- [x] Same fixture with `AUDIT_CONTEXT_UNCHANGED` in write_result text → exits 0
- [ ] Full integration test with a live auditor session — skipped: requires hook registration in settings.json, which is pending review

**Blocked items needing attention before merge:** hook registration in `~/.claude/settings.json` should be done post-review (either via `install.sh` re-run or manual `jq` edit).

Closes #429

### DIFF
--- a/hooks/require-auditor-context-update.py
+++ b/hooks/require-auditor-context-update.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+"""
+SubagentStop hook: ensure lobster-auditor sessions update system-audit.context.md.
+
+For non-auditor sessions this hook is a no-op (exits 0 immediately).
+
+For auditor sessions it enforces one of two exit conditions:
+  1. system-audit.context.md was modified during this session
+     (mtime >= session start time found in the transcript), OR
+  2. the transcript contains the safe word AUDIT_CONTEXT_UNCHANGED
+     (agent explicitly confirmed that nothing new was found).
+
+If neither condition is met the hook prints an error message and exits 2,
+hard-blocking the session from terminating until the agent complies.
+
+Detection strategy for "is this an auditor session":
+  Scan the transcript for a ReadFile tool call whose path contains
+  "system-audit.context.md". The auditor definition requires reading this
+  file at session start, so its presence is a reliable signal.
+"""
+import json
+import os
+import sys
+import time
+from pathlib import Path
+
+CONTEXT_FILE = Path(os.path.expanduser(
+    "~/lobster-user-config/agents/system-audit.context.md"
+))
+
+SAFE_WORD = "AUDIT_CONTEXT_UNCHANGED"
+
+# Path fragment used to detect auditor sessions in the transcript.
+AUDIT_CONTEXT_FILENAME = "system-audit.context.md"
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _extract_tool_calls(transcript: list) -> list[dict]:
+    """Return all tool_use blocks from the transcript."""
+    tool_calls = []
+    for msg in transcript:
+        if not isinstance(msg, dict):
+            continue
+        content = msg.get("content", [])
+        if not isinstance(content, list):
+            continue
+        for item in content:
+            if isinstance(item, dict) and item.get("type") == "tool_use":
+                tool_calls.append(item)
+    return tool_calls
+
+
+def _is_auditor_session(tool_calls: list[dict]) -> bool:
+    """Return True if any tool call reads system-audit.context.md.
+
+    The auditor subagent definition requires reading this file at the start of
+    every session.  A Read/cat/Bash call whose input references the filename is
+    sufficient evidence.
+    """
+    for call in tool_calls:
+        name = call.get("name", "")
+        # Check Read tool (Claude Code built-in)
+        if name == "Read":
+            path = call.get("input", {}).get("file_path", "")
+            if AUDIT_CONTEXT_FILENAME in path:
+                return True
+        # Check Bash tool — auditor might cat/head the file
+        if name == "Bash":
+            cmd = call.get("input", {}).get("command", "")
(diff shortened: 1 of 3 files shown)


sayhar (commented):
## Review

**Verdict: ready to merge, one note on hook registration.**

### What I verified

- Hook exits 0 for non-auditor sessions (no false positives)
- Hook exits 2 with a clear error message when an auditor session ends without updating the file or emitting the safe word
- Hook exits 0 when `AUDIT_CONTEXT_UNCHANGED` appears in `write_result` text
- The mtime-based file-update check works correctly with a 1-second clock skew margin
- `install.sh` changes follow the exact same jq-mutation pattern used for every other hook registration in the file
- The seed template is idempotent on re-runs (copy-if-not-exists guard)

### Design choices worth noting

**Session detection via transcript scan.** The hook identifies auditor sessions by looking for a `Read` or `Bash` tool call that references `system-audit.context.md` in its input. This is reliable because the auditor definition mandates reading that file at session start — it is not merely conventional. A session that reads the file but is not actually an auditor (pathological case) would be incorrectly flagged, but that is an acceptable false-positive direction (exit 2 with a clear message, not a silent failure).

**mtime fallback to safe word only when session start time is unknown.** If the hook input carries no timestamp, `_context_file_updated_since` returns `False`, pushing the only exit path to the safe word. This is the conservative direction — never silently pass through when we can't verify.

### Hook registration pending

The `install.sh` block for `require-auditor-context-update.py` is present and correct, but the hook is not yet in `~/.claude/settings.json`. After this PR is merged, registration is one of:
- Re-run `install.sh` (registers everything idempotently), or
- `jq` the SubagentStop block in manually

The hook is already deployed to `~/lobster/hooks/` on the live system, so the file is ready whenever registration is approved.

### No concerns blocking merge.

PR #438, opened 2026-03-16
fix: suppress debug noise when user model finds no trigger match

### LINKED ISSUE #437: fix: suppress debug noise on user model no-trigger-match
When \`LOBSTER_DEBUG=true\`, \`mark_processing\` emits a \`[context skipped] no trigger match\` debug observation for every message that does not activate the user model. Since no-match is the common case, this floods debug output with unhelpful observations.

**Fix:** remove the \`_emit_debug_observation\` call in the no-match \`else\` branch of the user model injection block in \`inbox_server.py\`. Observations are still emitted when a trigger matches (whether or not context is available) and on injection errors. Only the no-match case is silenced.

sayhar:
Closes #437

## Summary

In debug mode (`LOBSTER_DEBUG=true`), `mark_processing` was emitting a `[context skipped] no trigger match` observation for every message that didn't activate the user model. Because no-match is the common case — most messages don't hit a user model trigger — this produced a steady stream of uninformative debug notifications that obscured the signal from genuine context injections.

The fix removes the `_emit_debug_observation` call in the no-match `else` branch of the user model injection block in `inbox_server.py`. The informative paths remain intact: observations still fire when a trigger matches and context is injected, when a trigger matches but the model returns empty context, and on injection errors. Only the common no-match path is silenced.

The change is three lines removed and one comment line added. No behavior changes outside of debug mode (the function was already a no-op when `LOBSTER_DEBUG` is unset).

Functional pattern: pure conditional — dead branch elimination with no side effects introduced.

## Tests run

- [x] `uv run ruff check src/mcp/inbox_server.py` — clean, no errors
- [ ] Full test suite — skipped: no automated test coverage for debug observation emission paths in this file

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -3066,17 +3066,9 @@ async def handle_mark_processing(args: dict) -> list[TextContent]:
                 )
                 # never block mark_processing
         else:
-            # Debug: no trigger match — context injection skipped.
-            # _emit_debug_observation is a no-op when not in debug mode.
-            # Skip for system messages (chat_id=0) — context injection is never
-            # expected for self-check or other system-originated messages, so
-            # emitting an observation for them is pure noise.
-            _msg_chat_id = msg_data.get("chat_id", None)
-            if msg_text and len(msg_text) >= 8 and _msg_chat_id != 0:
-                _emit_debug_observation(
-                    f"\U0001f50d [context skipped] msg={short_msg_id} "
-                    "no trigger match — user model context not injected"
-                )
+            # No trigger match — context injection skipped. No notification emitted;
+            # this is the common case and emitting on every no-match is pure noise.
+            pass
 
     log.info(f"Message claimed for processing: {message_id}")
     return [TextContent(type="text", text=f"Message claimed: {message_id}{context_block}")]


sayhar:
The key diff is in \`handle_mark_processing\` in \`src/mcp/inbox_server.py\`. The old \`else\` block (lines 3068–3079 on main) contained an \`_emit_debug_observation\` call gated on \`msg_text\` length and \`chat_id != 0\`. That call is now replaced with a \`pass\` and an explanatory comment.

The three cases that do emit are unchanged:
- Trigger matched + context injected: \`[context injected]\` observation
- Trigger matched + empty context: \`[context skipped] trigger matched but user model returned empty context\`
- Injection exception: \`[context inject error]\` observation with traceback

The removed case was the fourth: trigger did not match at all. That fired on the majority of messages and was noise by design — it was even commented "no-op when not in debug mode" but was still firing in debug mode for every unmatched message.

sayhar (commented):
## Review

**Verdict: PASS**

### What this does

Removes a debug observation emission in the no-match branch of the user model context injection block inside `handle_mark_processing`. The removed call fired on every message that did not match a user model trigger — the common case — flooding debug output with uninformative `[context skipped] no trigger match` entries and obscuring the signal from the three cases that actually matter (context injected, empty context returned, injection error).

### Correctness

The change is minimal and correct. The three informative observation paths are untouched. The deleted block had no side effects beyond the `_emit_debug_observation` call itself, and that call is only active when `LOBSTER_DEBUG=true`, so there is no behavioral change in production. The replacement `pass` is semantically identical to the original `else` clause with its guarded body — the `else` branch was already effectively a no-op outside debug mode.

### Risks

None material. The only risk would be losing signal, but `[context skipped] no trigger match` was explicitly noted in the surrounding comments as a noise source. The remaining observation paths give full coverage of the cases that warrant attention.

### Notes

- The existing comment in PR #438 correctly identifies the key diff and which three cases still emit. No additional structural concerns.
- No automated test coverage for debug observation paths is noted; this is consistent with the rest of the file and not a blocker for this simple removal.

No concerns blocking merge.

PR #440, opened 2026-03-16
Fix chat_id=0 system route for background agents without user chat_id

### LINKED ISSUE #439: Fix chat_id=0 system route for background agents without chat_id
## Problem

When the dispatcher spawns a background subagent without including \`chat_id\` in the prompt, the agent's output is permanently lost. Three compounding issues cause this:

1. **\`subagent.bootup.md\` line 89** tells agents: "If you were not given a \`chat_id\`, do not call \`send_reply\` or \`write_result\`." This instruction is correct for *synchronous* subagents (where the Agent tool call returns results to the caller), but **wrong for background agents** (\`run_in_background: true\`). Background agents have no return channel — skipping \`write_result\` means their output is permanently unroutable.

2. **\`inbox_server.py\` \`handle_write_result\`** hard-rejects any call with \`chat_id=None\`. There is no fallback routing (e.g., \`chat_id=0\` → dispatcher system inbox). Even a well-intentioned agent that tries to call \`write_result\` with no \`chat_id\` gets a silent error and no result is stored.

3. **\`require-write-result.py\`** correctly blocks agents that never call \`write_result\` (checks for type=tool_use, not free-text). However it does not validate that \`write_result\` was called with a **non-null \`chat_id\`**. An agent could call \`write_result(chat_id=None)\`, the hook passes (tool was called), but the MCP server discards the result anyway.

## Failure chain

```
Dispatcher spawns background agent WITHOUT chat_id in prompt
    ↓
Agent reads subagent.bootup.md line 89: "no chat_id → skip write_result"
    ↓
Agent completes work, skips write_result
    ↓
require-write-result hook fires, blocks exit (correct behavior — but agent may have been misclassified as dispatcher)
    ↓
Dispatcher inbox receives no subagent_result — result is invisible
```

## Proposed Fixes

- [ ] **Fix 1 — \`inbox_server.py\`**: Accept \`chat_id=0\` as a dispatcher system route. Write the message with \`chat_id=0\` and \`type="subagent_result"\` so \`wait_for_messages\` returns it to the dispatcher for logging/storage.

- [ ] **Fix 2 — \`subagent.bootup.md\`**: Differentiate sync vs background in the no-chat_id guidance. Sync: "skip write_result — caller gets result." Background: "call \`write_result(chat_id=0)\` — no return channel exists."

- [ ] **Fix 3 — \`require-write-result.py\`**: After confirming write_result was tool-called, also verify \`chat_id\` is non-null. If \`chat_id=None\`, block and redirect with instruction to use \`chat_id=0\`.

## Notes

- The \`chat_id != 0\` check already exists at line 2947 of \`inbox_server.py\` (auto-reply safety guard), so \`chat_id=0\` as a sentinel is consistent with existing patterns.
- Fixes 1+2 close the root cause. Fix 3 is defense-in-depth.

sayhar:
## Problem

Background agents spawned without a \`chat_id\` had no safe path to deliver results. The result was permanently lost through three compounding failures:

1. **Bootup instructions** told agents with no \`chat_id\` to skip \`write_result\` entirely — correct for synchronous subagents (where the Agent tool call returns the result to the caller), but wrong for background agents (\`run_in_background: true\`), which have no return channel.

2. **\`inbox_server.py\` \`handle_write_result\`** rejects calls with \`chat_id=None\` with an error. Even a well-intentioned agent that called \`write_result\` without a \`chat_id\` would have its result discarded silently. \`chat_id=0\` already passes the \`is None\` guard (the guard only rejects \`None\`, not \`0\`), but this was undocumented and agents had no reason to use it.

3. **\`require-write-result.py\`** checked only that \`write_result\` appeared in tool calls (correct — it reads \`type == "tool_use"\`, not text mentions), but did not verify \`chat_id\` was non-null in the call input. An agent could call \`write_result(chat_id=None)\`, the hook would pass, but the MCP server would reject the call, and the result would still be lost.

## Changes

**\`subagent.bootup.md\`** — replaces the single "no chat_id → skip write_result" instruction with a sync/background split. Synchronous subagents still skip write_result (unchanged behavior). Background agents — or any agent unsure of its execution mode — are directed to call \`write_result(chat_id=0)\`, which routes to the dispatcher inbox for logging/storage.

**\`inbox_server.py\`** — adds an explicit comment documenting \`chat_id=0\` as the dispatcher system route. The \`None\`-guard already passes \`0\` (no logic change needed), but this was invisible to callers. The comment makes the contract explicit and intentional, preventing future "cleanup" from accidentally tightening the guard to \`chat_id is None or chat_id == 0\`.

**\`require-write-result.py\`** — after confirming \`write_result\` was tool-called, now inspects the call input to verify \`chat_id\` is non-null. If every \`write_result\` call had \`chat_id=None\` (MCP server would reject all of them), blocks exit with a redirect to \`chat_id=0\`. \`chat_id=0\` passes as a valid call.

## What is out of scope

This PR does not add dispatcher-side handling for \`chat_id=0\` messages beyond what already exists — \`wait_for_messages\` will return them, and the dispatcher can log/store them. A follow-on improvement (suggested as Fix 4 in the audit report) would add a \`PreToolUse\` hook that warns at spawn time when an Agent call omits \`chat_id\`. That is not included here.

## Functional patterns used

Pure functions throughout: the hook logic is a side-effect-free scan over the transcript list using list comprehensions; the inbox_server guard change is a targeted comment addition to an existing predicate; the bootup instruction update adds a declarative decision table for execution mode.

## Tests run

- [x] \`uv run -m pytest tests/unit/test_mcp_server/ tests/unit/test_hooks/ -q\` — 209 passed, 14 failed (same 14 failures on main; no regression introduced)
- [x] \`uv run -m pytest tests/ -x -q\` — stopped at first integration failure (\`test_check_inbox_then_reply_flow\`) which also fails on main (pre-existing)
- [ ] Live end-to-end test (background agent without chat_id → dispatcher receives chat_id=0 result) — **not yet performed**. The code path is a guard relaxation (not a new delivery path), so the risk is low, but this has not been confirmed running in production.

## Deployment status

**NOT merged to main.** As of 2026-03-16, all three changes (subagent.bootup.md, inbox_server.py comment, require-write-result.py) are on the \`fix/issue-chat-id-zero-system-route\` branch only. The live system is running main at commit \`fe591ec\`.

Note: \`chat_id=0\` messages already exist in the processed/ queue (used by self-check messages), confirming the \`None\`-guard already accepts \`0\`. The PR's inbox_server.py change is a comment-only addition; the behavioral fix is in subagent.bootup.md and require-write-result.py.

Closes #439

### DIFF
--- a/.claude/subagent.bootup.md
+++ b/.claude/subagent.bootup.md
@@ -86,7 +86,22 @@ Failing to pass `sent_reply_to_user=True` causes duplicate messages — the disp
 
 **Server-side safety net:** If you pass `task_id` to `send_reply`, the inbox server automatically sets `sent_reply_to_user=True` in `write_result` for that `task_id` — even if you forget. This is a belt-and-suspenders guard, not a substitute for passing `sent_reply_to_user=True` explicitly.
 
-**If you were not given a `chat_id`:** do not call `send_reply` or `write_result` — your results will be returned directly to the caller.
+**If you were not given a `chat_id` and you are a synchronous subagent** (not run in background — the Agent tool call blocks and the result flows back to the caller): do not call `send_reply` or `write_result` — your results will be returned directly to the caller.
+
+**If you were not given a `chat_id` but you ARE a background agent** (spawned with `run_in_background: true`, or you are unsure): you have no return channel. You MUST call `write_result` with `chat_id=0` so the dispatcher inbox can receive your result. Background agents have no caller to return results to — skipping `write_result` means your output is permanently lost.
+
+```python
+# Background agent with no chat_id — use chat_id=0 as dispatcher system route
+mcp__lobster-inbox__write_result(
+    task_id="<your-task-id>",
+    chat_id=0,
+    text="<your result>",
+    source="telegram",
+    sent_reply_to_user=False,
+)
+```
+
+**If unsure whether you are sync or background:** use `chat_id=0` as a safe fallback. The dispatcher will log and store the result rather than dropping it.
 
 ## Surfacing Observations (`write_observation`)
 
--- a/hooks/require-write-result.py
+++ b/hooks/require-write-result.py
@@ -54,21 +54,48 @@ def main():
 
     transcript = data.get("transcript", [])
 
-    tool_calls = []
+    # Collect all tool call items so we can inspect both name and input.
+    all_tool_use_items = []
     for msg in transcript:
         if isinstance(msg, dict):
             content = msg.get("content", [])
             if isinstance(content, list):
                 for item in content:
                     if isinstance(item, dict) and item.get("type") == "tool_use":
-                        tool_calls.append(item.get("name", ""))
+                        all_tool_use_items.append(item)
 
-    # If this session called write_result, mark it completed in the DB and allow exit.
-    if "mcp__lobster-inbox__write_result" in tool_calls:
-        session_id = get_session_id(data)
-        if session_id:
-            _mark_session_completed(session_id)
-        sys.exit(0)
+    tool_call_names = [item.get("name", "") for item in all_tool_use_items]
+
+    if "mcp__lobster-inbox__write_result" in tool_call_names:
+        # write_result was called — verify it was called with a non-null chat_id.
+        # chat_id=0 is explicitly allowed: it is the dispatcher system route for
+        # background agents that were spawned without a user chat_id.
+        write_result_items = [
+            item for item in all_tool_use_items
+            if item.get("name") == "mcp__lobster-inbox__write_result"
+        ]
+        valid_calls = [
+            item for item in write_result_items
+            if item.get("input", {}).get("chat_id") is not None
+        ]
+        if valid_calls:
+            # At least one valid call found — mark session and allow exit.
+            session_id = get_session_id(data)
+            if session_id:
+                _mark_session_completed(session_id)
+            sys.exit(0)
+        else:
+            # write_result was called but chat_id was None in every call —
+            # the MCP server rejected the call and the result was not stored.
+            print(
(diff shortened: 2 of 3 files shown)


sayhar (commented):
Pressure test results for the analysis in the audit report:

**Line 89 claim — CONFIRMED.** `subagent.bootup.md` line 89 contained exactly the text quoted: "If you were not given a chat_id: do not call send_reply or write_result — your results will be returned directly to the caller." This is now replaced with the sync/background split.

**MCP server behavior — PARTIALLY CONFIRMED, ONE CORRECTION.** The report claimed `chat_id=0` would also be rejected by the same guard. It would not — the guard is `if chat_id is None:`, not `if not chat_id:`. Integer `0` passes. This means Fix 1 required no logic change to `inbox_server.py`, only a comment documenting that `chat_id=0` is the intended dispatcher system route and must not be tightened.

**Hook checks tool calls not text — CONFIRMED.** `require-write-result.py` correctly reads `item.get("type") == "tool_use"` from the transcript. A mention of "write_result" in response text would not pass.

**`chat_id=0` conflicts — CONFIRMED SAFE.** Line 2947 already contains `if msg_type in USER_FACING_TYPES and chat_id != 0:` — the codebase already treats `0` as a sentinel that skips user-facing auto-reply. `chat_id=0` messages will be picked up by `wait_for_messages` but will not trigger Telegram/Slack delivery.

One discrepancy from the report: Fix 1 as described ("change the guard so chat_id=0 is accepted") overstated the required change. The guard never rejected `0`; it only rejected `None`. The actual fix was to document this contract explicitly so it does not regress.

PR #446, opened 2026-03-16
fix: detect and explain pseudocode tool-call failure in SubagentStop hook

### LINKED ISSUE #445: Subagents misread bootup code examples as text output instead of tool invocations
## Problem

A subagent completed its task and compiled a full status report, then wrote `mcp__lobster-inbox__send_reply(...)` and `mcp__lobster-inbox__write_result(...)` as Python text output instead of invoking them via the tool mechanism. The SubagentStop hook blocked exit with code 2, but the content was lost.

Root cause: `subagent.bootup.md` shows MCP tool calls in Python code blocks with no explicit warning that these must be invoked as tools, not written as code output. An agent can misread "here's how to call this tool" as "output this Python text."

## Impact

- Subagent results are silently lost
- The hook catches the failure (exit 2), but the error message is generic — it says "call write_result" without explaining *why* it wasn't called
- The agent is blocked without a useful signal about what went wrong

## Fix

Two changes in PR #439 (or whichever PR closes this):

1. **`hooks/require-write-result.py`**: After confirming `write_result` was not called as a tool, scan assistant text content for the string `mcp__lobster-inbox__write_result`. If found, emit a specific error explaining the pseudocode failure mode rather than the generic message.

2. **`.claude/subagent.bootup.md`**: Add an explicit CRITICAL callout immediately before the send_reply/write_result code examples, making clear these are tool invocations — not code to write as text output.

sayhar:
## Summary

Subagents occasionally write `mcp__lobster-inbox__write_result(...)` as Python text output instead of invoking it as a tool. The SubagentStop hook already blocks this with exit 2, but the error message is generic and gives the agent no signal about what went wrong — so it can loop indefinitely without understanding why it's blocked.

Closes #445.

**Two changes:**

- **`hooks/require-write-result.py`**: When `write_result` is absent from tool_use items, scan assistant text content for the string `mcp__lobster-inbox__write_result`. If found, replace the generic error with a targeted message explaining that the tool was described as text output rather than invoked as a tool. The generic message is preserved for the case where the string appears in neither tool calls nor text.

- **`.claude/subagent.bootup.md`**: Adds a CRITICAL blockquote immediately before the send_reply/write_result code examples, explicitly stating these are tool invocations and that writing them as Python code in text output does nothing — and that the hook will block with exit 2.

The hook change is purely additive: it only affects the error message path (when write_result was not called). Success path and session completion marking are unchanged.

Functional patterns used: pure predicate scan over transcript items (no mutation), branching on scan result to select error message.

## Tests run

- [x] `uv run hooks/require-write-result.py` (in worktree) — import check passes; exits cleanly on missing stdin without blocking
- [ ] Full SubagentStop integration test — no automated harness exists for hook firing; requires a live agent session. No behavior change to success path.

**Blocked items needing attention before merge:** none — the hook change only affects the failure-mode error message, not any success or routing logic.

### DIFF
--- a/.claude/subagent.bootup.md
+++ b/.claude/subagent.bootup.md
@@ -51,6 +51,8 @@ You MUST both deliver results to the user directly AND call `write_result` at th
 
 **Required at end of every subagent task — two steps:**
 
+> **CRITICAL: The examples below are tool invocations, not code output.** Call `mcp__lobster-inbox__send_reply` and `mcp__lobster-inbox__write_result` using the tool invocation mechanism — the same mechanism you use for Read, Edit, Bash, and other tools. Do NOT write these as Python code blocks in your text output. Writing `mcp__lobster-inbox__write_result(...)` as text output does nothing — the tool is not called, the result is not delivered, and the SubagentStop hook will block you with exit 2.
+
 ```python
 # Step 1: Deliver directly to the user (crash-safe delivery)
 # Pass task_id to enable server-side auto-dedup: the inbox server will
--- a/hooks/require-write-result.py
+++ b/hooks/require-write-result.py
@@ -55,13 +55,17 @@ def main():
     transcript = data.get("transcript", [])
 
     tool_calls = []
+    text_content_parts = []
     for msg in transcript:
         if isinstance(msg, dict):
             content = msg.get("content", [])
             if isinstance(content, list):
                 for item in content:
-                    if isinstance(item, dict) and item.get("type") == "tool_use":
-                        tool_calls.append(item.get("name", ""))
+                    if isinstance(item, dict):
+                        if item.get("type") == "tool_use":
+                            tool_calls.append(item.get("name", ""))
+                        elif item.get("type") == "text":
+                            text_content_parts.append(item.get("text", ""))
 
     # If this session called write_result, mark it completed in the DB and allow exit.
     if "mcp__lobster-inbox__write_result" in tool_calls:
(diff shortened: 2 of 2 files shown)


sayhar (commented):
## Code Review

**Overall: LGTM with one minor observation.**

### `hooks/require-write-result.py`

The pseudocode detection logic is correct and well-scoped. A few notes:

**Transcript structure coverage**: The text scan covers `type == "text"` items nested inside `content` lists of transcript messages. This matches the structure Claude Code uses for assistant turns. The fallback to `"\n".join(text_content_parts)` is clean — the substring check works regardless of how the text was split across turns.

**False positive risk is low**: The string `mcp__lobster-inbox__write_result` is highly specific. It's extremely unlikely to appear in user text, tool results, or incidental output without an agent actually having written pseudocode. No concern here.

**One minor observation**: The existing `main` branch version of this hook (not this PR's base, which is older) has a richer structure — it collects `all_tool_use_items` with full item dicts to support `chat_id` validation. This PR's change is additive against the simpler base and doesn't interfere with the richer version. If this is merged into a branch that already has the chat_id validation logic, the text-scan addition will need to be forward-ported carefully. Worth a note for the merger.

### `.claude/subagent.bootup.md`

The CRITICAL blockquote placement is exactly right — it's the last thing an agent reads before the code examples, so it's maximally proximate. The specific mention of "exit 2" gives the error a concrete consequence, which is more salient than an abstract warning. The comparison to Read/Edit/Bash is helpful framing.

**No changes requested.** This is a minimal, targeted fix for a real failure mode.

sayhar:
Superseded by #452 which includes all changes from this PR plus the agentId fix.

PR #452, opened 2026-03-16
fix(hooks): use transcript agentId for session_end in SubagentStop hook

sayhar:
## Summary

The SubagentStop hook was calling `session_end()` with CC's internal `session_id`, which is a different UUID from the one stored in `agent_sessions.db`. The DB's `id` column stores the `agentId` that CC sets in every transcript entry — the same UUID the dispatcher passes to `register_agent`. Using `session_id` instead meant every call to `session_end` in the hook was silently a no-op, leaving rows stuck in `running` state and causing false ghost-agent alerts.

This PR extracts `agentId` from the transcript (using the first entry that has one, since all entries carry this field set by CC's infrastructure) and uses it as the lookup key. If the transcript has no `agentId` (e.g. minimal test fixtures), it falls back to `session_id` defensively.

## Changes

- Added `_extract_agent_id_from_transcript(transcript)` — pure function that returns the first non-null `agentId` from transcript entries, or `None`.
- Updated `_mark_session_completed` parameter name from `session_id` to `id_or_task_id` to reflect that it accepts either form.
- Updated the call site in `main()` to prefer `agentId` from transcript, with `get_session_id(data)` as fallback.

The PR also includes the uncommitted improvements to the hook already present in `~/lobster` (text-content pseudocode detection and chat_id=None guard), which were added locally but not yet committed.

Functional patterns used: pure helper function (`_extract_agent_id_from_transcript`), explicit fallback chain rather than implicit mutation.

## Hook audit

No other hooks need this change. The only hook that calls `session_end` is `require-write-result.py`. Other hooks that do session-related work:

- `write-dispatcher-session-id.py` — uses `session_id` from hook input to identify the dispatcher and auto-register subagent stubs. This is correct: the stub row is keyed by `session_id` at write time, and `session_end` matches on both `id` and `task_id`. The race with `register_agent` (which writes a richer row with `agentId`) is handled by `INSERT OR IGNORE`, so the stub is only written if `register_agent` hasn't written first.
- `on-compact.py`, `post-compact-gate.py` — use `session_id` only for dispatcher detection (comparing to the marker file). No DB lookup involved.
- All PreToolUse/PostToolUse hooks — do not interact with agent tracking at all.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/ -x -q` — 10 passed, 0 failed
- [x] `uv run pytest tests/unit/ -x -q` — 127 passed, 1 failed (pre-existing: `test_bot/test_authorization.py::TestIsAuthorized::test_authorized_user_returns_true` fails on main with the same error, unrelated to this change)

### DIFF
--- a/hooks/require-write-result.py
+++ b/hooks/require-write-result.py
@@ -19,7 +19,26 @@ sys.path.insert(0, str(Path(__file__).parent))
 from session_role import is_dispatcher, get_session_id
 
 
-def _mark_session_completed(session_id: str) -> None:
+def _extract_agent_id_from_transcript(transcript: list) -> str | None:
+    """Return the agentId from the first transcript entry that has one.
+
+    Every transcript entry carries an `agentId` field set by CC's infrastructure
+    (not by the LLM). This UUID matches the value the dispatcher stores when it
+    calls register_agent — it is the correct key for the agent_sessions.db `id`
+    column. CC's internal `session_id` is a different UUID and does not match
+    the DB row, so this value is preferred when available.
+
+    Returns None if the transcript is empty or no entry carries agentId.
+    """
+    for entry in transcript:
+        if isinstance(entry, dict):
+            agent_id = entry.get("agentId")
+            if agent_id:
+                return agent_id
+    return None
+
+
+def _mark_session_completed(id_or_task_id: str) -> None:
     """Mark the agent session completed in agent_sessions.db.
 
     Best-effort: any failure is silently swallowed so the hook always exits 0.
@@ -32,7 +51,7 @@ def _mark_session_completed(session_id: str) -> None:
         sys.path.insert(0, str(repo_src))
         from agents.session_store import session_end  # noqa: PLC0415
         session_end(
-            id_or_task_id=session_id,
+            id_or_task_id=id_or_task_id,
             status="completed",
             result_summary="Completed via SubagentStop hook",
         )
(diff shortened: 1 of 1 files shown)


sayhar (commented):
## Review: LGTM

This is a well-reasoned fix to a silent but impactful bug. Full review below.

### Bug fix correctness

The root cause is clearly identified and correctly addressed. CC's internal `session_id` (from `get_session_id(data)`) is a different UUID from the `agentId` that CC sets in every transcript entry — and `agentId` is what `register_agent` stores in `agent_sessions.db`. Using `session_id` meant `session_end()` was silently a no-op on every subagent exit, leaving rows stuck in `running` state. The fix is correct.

**`_extract_agent_id_from_transcript`** is a clean pure function. It iterates over transcript entries, finds the first non-null `agentId`, and returns `None` if absent. The defensive fallback to `get_session_id(data)` is appropriate for test fixtures or edge cases.

**Parameter rename** from `session_id` to `id_or_task_id` in `_mark_session_completed` correctly reflects the wider contract of `session_end()`.

### chat_id=None guard

The new check that `write_result` was called with a non-null `chat_id` is correct and the error message clearly explains that `chat_id=0` is the valid route for background agents. This prevents the hook from accepting calls that the MCP server would have rejected.

### Pseudocode detection (carried from #446)

The scan of `text_content_parts` for `mcp__lobster-inbox__write_result` is additive and purely affects the error message path. The targeted error message is actionable and significantly better than the generic one.

### Hook audit

The PR's own audit of other hooks (`write-dispatcher-session-id.py`, `on-compact.py`, `post-compact-gate.py`) is sound. No other hook calls `session_end`, and the `session_id` usages in other hooks are correct for their respective purposes.

### Tests

10 hook tests pass; the one pre-existing failure in `test_authorization.py` is unrelated to this change and exists on `main` as well.

### No concerns

The change is surgical (one file), the logic is correct, fallbacks are defensive, and the PR description is thorough. Ready to merge.

PR #453, opened 2026-03-16
feat: edited_message handler (issue #434)

### LINKED ISSUE #434: Message processing redesign: ack-first, edits, reactions, dedup, crash recovery
> **Note:** This is a design proposal produced from reading the actual bot/MCP source. Use it to track implementation of individual items below.

## Priority checklist

- [ ] Reduce stale recovery timeout to 90s (fixes worst UX failure, ~10min)
- [ ] Remove text ack, keep typing indicator (cleans up noise, ~10min)
- [ ] Dispatcher ack policy in prompt (changes dispatcher behavior for long tasks, ~30min)
- [ ] `edited_message` handler — deliver with note, not discard (~3h)
- [ ] Follow-up compaction via reply threading + time window (~2h)
- [ ] Reaction signals (enables 👍/👎 as affirmative/negative, ~4h)
- [ ] `claim_and_ack` MCP tool (optional/future, low priority)
- [ ] ~~Duplicate detection~~ (deprioritized — not worth complexity)

---

# Telegram Message Processing Design Proposal

## Preamble: What I Read

Before designing, I read the full source of the live system:

- `/home/lobster/lobster/src/bot/lobster_bot.py` — Telegram bot (1,200+ lines): message ingestion, outbox watcher, typing indicator loop, media group buffering, wake-on-hibernate, `handle_message`, audio handling
- `/home/lobster/lobster/src/mcp/inbox_server.py` — MCP server (3,100+ lines): `wait_for_messages`, `check_inbox`, `mark_processing`, `mark_processed`, `mark_failed`, `send_reply`, `write_result`, `_recover_stale_processing`, idempotency tracker, task-ID dedup registry, session guard
- `/home/lobster/lobster/src/mcp/reliability.py` — Atomic writes, idempotency tracker, circuit breaker

Key existing behaviors:
- Bot writes JSON to `inbox/` atomically (temp+rename). MCP server moves to `processing/` on `mark_processing`, to `processed/` on `mark_processed`.
- `wait_for_messages` uses inotify (watchdog) for <40ms latency. Calls `_recover_stale_processing()` (300s timeout) at every call start.
- `mark_processed` has a guard: if no reply was tracked for a user-facing message, it auto-sends "Noted." to prevent silent drops.
- Typing indicator loop refreshes every 4s for any file in `processing/`.
- No edit/clarification/reaction handling exists today.
- `send_reply` + `write_result` have a task_id dedup registry that survives MCP restarts (file-based in `task-replied/`).

---

### States

```
INBOX         — file lives in ~/messages/inbox/
PROCESSING    — file lives in ~/messages/processing/  (claimed by dispatcher)
PROCESSED     — file lives in ~/messages/processed/
FAILED        — file lives in ~/messages/failed/      (with retry metadata)
```

### Valid transitions

```
INBOX        -> PROCESSING   mark_processing() — dispatcher claims it
PROCESSING   -> PROCESSED    mark_processed() — normal completion
PROCESSING   -> FAILED       mark_failed() — error, schedules retry
PROCESSING   -> INBOX        _recover_stale_processing() — crash recovery (after 300s)
FAILED       -> INBOX        _recover_retryable_messages() — after backoff expires
```

**Note on edits and follow-ups:** When a user edits a message or sends a follow-up, the system uses **compaction** rather than a SUPERSEDED state. Related messages are merged into one combined context for the LLM. No messages are silently dropped. See Section 3 for details.

---

### Current state

The bot already sends acknowledgment strings:
- Text messages: "Message received. Processing..." (line 902)
- Photos: "Photo received. Looking at it..."
- Voice: "Voice message received. Transcribing..."
- Documents: "Document received."

These are sent immediately by the bot before the dispatcher ever sees the message. This is already ack-first.

### The gap

The current acks are hardcoded strings. They don't reflect the actual nature of the request, and they use emoji inconsistently. More importantly: for short responses (under ~3 seconds), the ack and the real answer arrive close together and feel spammy.

### Proposed design

**Bot layer (immediate, <200ms):**

Send a typing indicator only. Do NOT send a text ack for text messages. Reason: Telegram's "typing..." bubble is the universal ack signal. A text ack before the real reply adds noise.

**Exception — send a text ack when:**
1. The message is a voice note (transcription takes 5-30s — user needs to know it was received)
2. The message has an image/document (download takes variable time)
3. The dispatcher explicitly decides a long task is being spawned (see below)

**Dispatcher-triggered ack (for long tasks):**

When the dispatcher decides to spawn a subagent (expected time > ~4 seconds), it calls `send_reply` with a brief ack BEFORE spawning. This is an explicit choice by the dispatcher, not an automatic bot behavior.

Example acks (no emoji):
- "On it." — for action requests
- "Looking into this." — for research/lookup
- "Writing that up." — for generation tasks
- "Reading..." — for analysis tasks

**When NOT to send an ack:**
- When the response is trivially fast (dispatcher can produce it inline in <3s)
- When the message is a button callback (Telegram already removes the button loading state)
- When the message is a reaction (reactions get structured handling, not acks)
- When the previous message from the same chat is still in PROCESSING (batch the ack)

**Implementation:**

No code changes needed in the bot for most of this. The dispatcher prompt should say:

> Before spawning a subagent for a task that will take >4 seconds, call send_reply with a brief 1-3 word acknowledgment ("On it.", "Writing that up.", etc.). Do not ack for fast inline responses.

The typing indicator loop in the bot already handles the visual "processing" signal during subagent work.

---

### 3a. Telegram `edited_message` events

Telegram sends `edited_message` when a user edits a sent message. The bot currently has no handler for `edited_message`.

**Core principle: deliver, don't discard.**

If a subagent is already working on message A when the user edits it, the subagent's result should still be delivered — with a note that the original was edited. Something like:

> "Based on your original message — you've since edited it. Here's what I found: [result]"

Let the user decide if it's still relevant. Silent discard is unacceptable UX. The subagent did real work; surface it.

**Proposed bot-layer handling:**

```python
async def handle_edited_message(update: Update, context):
    """Handle edited_message events from Telegram."""
    message = update.edited_message
    if not message or not message.from_user:
        return
    if message.from_user.id not in ALLOWED_USERS:
        return

    original_tg_id = message.message_id  # same ID, edited in place

    # Check if the original message is still in inbox/ or processing/
    # by scanning for a file with matching telegram_message_id
    original_inbox_file = _find_by_telegram_id(original_tg_id)

    if original_inbox_file is None:
        # Already processed — treat as a new message with an edit note.
        msg_data = _build_message_data(message, msg_type="text")
        msg_data["_edit_of_telegram_id"] = original_tg_id
        msg_data["_edit_note"] = "User edited a previously processed message."
        inbox_file = INBOX_DIR / f"{msg_id}.json"
        atomic_write_json(inbox_file, msg_data)
        return

    # Original is still in inbox or processing — write the edit as a new
    # message and annotate it so the dispatcher knows it's a replacement.
    new_msg_id = _make_msg_id(message)
    new_msg_data = _build_message_data(message)
    new_msg_data["_edit_of_telegram_id"] = original_tg_id
    new_msg_data["_replaces_inbox_id"] = original_inbox_file.stem

    # Write new message to inbox (crash-safe ordering)
    new_inbox_file = INBOX_DIR / f"{new_msg_id}.json"
    atomic_write_json(new_inbox_file, new_msg_data)
    log.info(f"Edit queued as new message {new_msg_id} (replaces {original_inbox_file.stem})")
```

`_find_by_telegram_id(tg_id)` scans inbox/ and processing/ for a file where `telegram_message_id == tg_id`. This is a small directory scan (typically <20 files), fast enough.

**What if the original is in PROCESSING?**

Do NOT discard the subagent result. When the subagent finishes and calls `mark_processed`, the MCP server detects the `_replaces_inbox_id` annotation on the new inbox message and adds a note to the outgoing reply:

```python
# In handle_mark_processed, after writing the reply:
if _has_pending_edit_for(message_id):
    # Prepend note to the reply text
    reply_text = "Based on your original message — you've since edited it.\n\n" + reply_text
```

The edited version is already in inbox and will be picked up by the next `wait_for_messages` cycle, so the user gets both: the result of the original work (with a note) and a fresh processing of their edit.

**Cost:** One extra field in mark_processed check. One scan of inbox/ and processing/ per edit event (fast).

### 3b. Clarification follow-ups ("do X" then "actually do Y")

This is the case where a user sends two separate messages quickly. There is no Telegram-level link between them.

**Proposed approach: compaction, not supersede-and-drop.**

Silently dropping earlier messages causes information loss. For example:
- "Do X" → "Also do Y" → "Cancel X": the "Also do Y" message would be lost by a simple supersede approach.

The correct model is to **compact** related messages into one combined context. The LLM receives everything and resolves what to do. No messages are silently dropped.

**Primary compaction signal: Telegram reply threading.**

If the user's follow-up is a reply (in the Telegram UI) to a previous message, that's an explicit grouping signal. The bot detects this via `message.reply_to_message` and annotates both messages with a shared `_compact_group` ID. The dispatcher sees a message of type `compact_group` containing all messages in the group.

**Secondary signal: time window + negation markers.**

If no reply link exists, but messages from the same chat arrive within `FOLLOW_UP_WINDOW_SECS = 8` and the latest starts with a negation/correction marker, compact them. The logic lives in `handle_check_inbox`:

```python
NEGATION_MARKERS = ["actually", "wait", "no,", "scratch that",
                    "instead", "forget that", "cancel"]

def _maybe_compact_follow_ups(messages: list[dict]) -> list[dict]:
    """
    If multiple messages from the same chat arrived within FOLLOW_UP_WINDOW_SECS,
    and the latest one starts with a negation/correction marker (or has a
    reply-thread link to an earlier one), merge them into a single compact_group
    message. The LLM handles the merged context.

    Returns the compacted list. No messages are dropped — all text is preserved.
    Does NOT call LLM — purely structural/lexical grouping.
    """
```

The compacted message type is `compact_group` and includes the full text of all constituent messages. The dispatcher reads the combined context and resolves intent.

**What if an earlier message is already in PROCESSING?**

Annotate the new message with `_related_to_processing: <message_id>` so the dispatcher knows context is arriving. The subagent's result is still delivered (with a note if needed). The follow-up is queued as a new inbox message that the dispatcher processes next.

### 3c. Exact duplicate detection — deprioritized

Not worth the complexity for now. The dominant case (user accidentally double-sends) is rare and the cost of processing a duplicate is low. This feature is flagged as **future/optional** and should not block other work.

---

### Current state

The bot has no handler for `MessageReactionUpdated` events. Telegram sends these when a user adds or removes a reaction.

### Proposed design

Treat 👍/👎 reactions as structured message signals delivered to the inbox. This is the correct abstraction — reactions become first-class inbox messages.

**Bot layer: reaction event handler**

```python
REACTION_SIGNALS = {
    "👍": "yes",
    "👎": "no",
    "✅": "yes",   # heavy check mark
    "👌": "yes",   # OK
    "❌": "no",    # cross mark
    "🚫": "cancel",
}

# Reactions that are undone within this window are ignored
REACTION_UNDO_WINDOW_SECS = 5.0

# In-memory buffer: tg_message_id -> (reaction_emoji, chat_id, timestamp, asyncio.Task)
_pending_reactions: dict[int, dict] = {}

async def handle_reaction(update: Update, context):
    """Handle MessageReactionUpdated events."""
    reaction_update = update.message_reaction
    if not reaction_update:
        return
    if reaction_update.user.id not in ALLOWED_USERS:
        return

    tg_msg_id = reaction_update.message_id
    chat_id = reaction_update.chat.id
    new_reactions = reaction_update.new_reaction  # list of ReactionType
    old_reactions = reaction_update.old_reaction

    # Reaction removed — cancel pending signal if within undo window
    if not new_reactions and old_reactions:
        if tg_msg_id in _pending_reactions:
            pending = _pending_reactions.pop(tg_msg_id)
            pending["task"].cancel()
            log.info(f"Reaction undone within undo window for msg {tg_msg_id} — ignored")
        return

    if not new_reactions:
        return

    # Get the emoji from the first new reaction
    reaction = new_reactions[0]
    emoji = reaction.emoji if hasattr(reaction, "emoji") else str(reaction)
    signal = REACTION_SIGNALS.get(emoji)
    if signal is None:
        return  # Unknown reaction — ignore

    # Cancel any previous pending reaction for this message
    if tg_msg_id in _pending_reactions:
        _pending_reactions[tg_msg_id]["task"].cancel()

    # Schedule emission after undo window expires
    task = asyncio.create_task(
        _emit_reaction_signal(chat_id, tg_msg_id, emoji, signal)
    )
    _pending_reactions[tg_msg_id] = {
        "task": task,
        "emoji": emoji,
        "timestamp": time.time(),
    }


async def _emit_reaction_signal(chat_id, tg_msg_id, emoji, signal):
    """Emit a structured reaction signal to inbox after undo window."""
    try:
        await asyncio.sleep(REACTION_UNDO_WINDOW_SECS)
    except asyncio.CancelledError:
        return  # Reaction was undone

    _pending_reactions.pop(tg_msg_id, None)

    msg_id = f"{int(time.time() * 1000)}_reaction_{tg_msg_id}"
    msg_data = {
        "id": msg_id,
        "source": "telegram",
        "type": "reaction",
        "chat_id": chat_id,
        "telegram_message_id": tg_msg_id,  # the message that was reacted to
        "emoji": emoji,
        "signal": signal,    # "yes", "no", "cancel"
        "text": f"[Reaction: {emoji} ({signal}) on message {tg_msg_id}]",
        "timestamp": datetime.utcnow().isoformat(),
    }
    # Fetch the reacted-to message's text if available (for context)
    original_text = _sent_message_buffer.get(tg_msg_id, "")
    if original_text:
        msg_data["reacted_to_text"] = original_text[:500]

    inbox_file = INBOX_DIR / f"{msg_id}.json"
    atomic_write_json(inbox_file, msg_data)
    log.info(f"Emitted reaction signal: {emoji} ({signal}) on msg {tg_msg_id}")
```

**Bot registration (in main()):**

```python
application.add_handler(
    MessageReactionHandler(handle_reaction)
)
```

**MCP layer: no changes needed.** The `reaction` message type flows through check_inbox normally.

**Dispatcher behavior:**

The dispatcher prompt should say:

> When you see a message of type "reaction", look at the `telegram_message_id` field to identify which of your outgoing messages was reacted to. Match it against your recent outgoing messages. If the signal is "yes", treat it as the user confirming your last question or proposal. If "no", treat it as rejection. Do not ask for further confirmation — act on the signal directly.

**Context linking:**

To know what the reaction refers to, the dispatcher needs to know what it said in message `telegram_message_id`. The bot stores a rolling ring buffer of the last 50 sent messages (tg_msg_id -> text), in memory. When a reaction arrives, `_emit_reaction_signal` fetches the original text and includes it in the reaction message. This is 50 dict entries in memory — negligible cost.

---

### Current atomicity model

The current model has three separate steps: `mark_processing` (MCP tool call), `send_reply` (MCP tool call), subagent spawn (Task tool call). These happen sequentially in the dispatcher with no transactional guarantee. A crash between any two steps leaves the system in an inconsistent state.

### Crash scenarios and current handling

**Crash after mark_processing, before ack:**
- Message is in `processing/`
- `_recover_stale_processing()` fires at next `wait_for_messages` call (after 300s)
- Message moves back to inbox
- Bot's typing indicator loop will show "typing" for those 300s (because file is in processing/)
- User sees typing... for 5 minutes then silence

**Crash after ack, before subagent spawn:**
- Same as above, but user got an ack and then silence
- 300s stale recovery handles it

**Crash during subagent execution:**
- Subagent doesn't call `write_result` / `mark_processed`
- Message stays in processing/ until stale recovery
- Subagent's Claude session exits cleanly — no work is partially committed

### Proposed improvement: reduce stale timeout

The current 300s (5 minute) stale recovery is too slow. Users see "typing..." for 5 minutes.

**Change:** Reduce `max_age_seconds` for text messages to 90s. Keep 300s for voice/audio (transcription can be slow). Implementation: read the `type` field from the JSON file in `_recover_stale_processing` and apply different timeouts.

```python
def _recover_stale_processing(default_max_age: int = 90):
    LONG_TYPES = {"voice", "audio", "photo", "document"}
    LONG_MAX_AGE = 300
    now = time.time()
    for f in PROCESSING_DIR.glob("*.json"):
        try:
            age = now - f.stat().st_mtime
            try:
                msg_type = json.loads(f.read_text()).get("type", "")
            except Exception:
                msg_type = ""
            max_age = LONG_MAX_AGE if msg_type in LONG_TYPES else default_max_age
            if age > max_age:
                dest = INBOX_DIR / f.name
                f.rename(dest)
                log.warning(f"Recovered stale message: {f.name} (age {int(age)}s, type {msg_type})")
        except Exception:
            continue
```

**Change:** Add a `_processing_started_at` timestamp to the message file when `mark_processing` is called. This makes stale detection based on actual processing time, not file mtime (which can drift if the filesystem updates mtime on reads).

### `claim_and_ack` — optional/future

A `claim_and_ack` MCP tool (combining `mark_processing` + `send_reply` in one atomic call) was considered to tighten the window between claiming a message and acking it. This is **low priority and does not need to block other work**. The atomicity improvement it offers is marginal given that stale recovery handles the crash cases. Mark as optional/future.

### Crash recovery summary (with proposed changes)

| Crash point | Recovery | User experience |
|---|---|---|
| Before mark_processing | Message stays in inbox, picked up at next wait_for_messages | No visible effect |
| After mark_processing, before ack | Stale recovery at 90s, retried | 90s silence + typing, then response |
| During subagent | Subagent exits, mark_processed never called, stale recovery fires | 90s typing then response |
| After mark_processed | Message already in processed, done | No effect |

---

### Bot layer (lobster_bot.py)

**Change 1: React to `edited_message` events**
- Add `handle_edited_message` function (~60 lines)
- Add `_find_by_telegram_id` helper (~20 lines, scans inbox/ and processing/)
- Deliver subagent result with edit note, not silent discard
- Add handler registration in `main()`
- Complexity: small.

**Change 2: Reaction signal handling**
- Add `handle_reaction` function (~50 lines)
- Add `_emit_reaction_signal` async function (~40 lines)
- Add `_pending_reactions` dict
- Add `_sent_message_buffer` rolling buffer (50 entries), populated in `process_reply`
- Add `MessageReactionHandler` registration in `main()`
- Complexity: medium. Requires enabling `reactions` in the bot's allowed updates.

**Change 3: Drop text-message ack**
- Remove line 902: `await message.reply_text("Message received. Processing...")`
- Retain typing indicator (already working)
- Complexity: trivial (1 line removed).

**Total bot changes: ~200 lines added, 1 line removed.**

### MCP server (inbox_server.py)

**Change 1: Reduce stale recovery timeout**
- Modify `_recover_stale_processing` to use type-aware timeout
- ~10 lines changed.

**Change 2: `_processing_started_at` in mark_processing**
- Add timestamp field to message JSON when moving to processing/
- ~5 lines added.

**Change 3: Follow-up compaction in check_inbox**
- Add `_maybe_compact_follow_ups` function (~50 lines)
- Primary signal: Telegram reply threading (`reply_to_message` field)
- Secondary signal: time window + negation markers
- Result: `compact_group` message type containing all constituent messages
- New message type: `compact_group`

**Change 4: Edit note injection in mark_processed**
- Check if a newer edit of the same telegram_message_id is pending in inbox
- Prepend edit note to reply if so
- ~10 lines added to `handle_mark_processed`

**Total MCP changes: ~150 lines added, ~15 lines modified. No new tools required (claim_and_ack deferred).**

### Dispatcher behavior (prompt/instructions)

**Change 1: Ack policy**
- Add rule: "Only ack for tasks expected to take >4 seconds"
- Add rule: "For trivially fast responses, skip the ack and reply directly"

**Change 2: Reaction handling**
- Add rule: "Messages of type `reaction` carry a `signal` field (yes/no/cancel). Interpret this as the user's answer to your most recent question or proposal in that chat. Act on it directly without re-asking."

**Change 3: Compact group handling**
- Add rule: "Messages of type `compact_group` contain multiple related messages merged together. Read all of them and resolve what the user wants. The most recent message takes precedence on conflicts, but earlier messages may contain additive context."

**Total dispatcher changes: ~6 new sentences in the prompt. No new tool calls required for the common path.**

### Complexity estimate

| Component | Lines | Risk | Time estimate |
|---|---|---|---|
| Bot: edited_message handler (deliver with note) | +200 | Low | 2-3h |
| Bot: reaction handler + buffer | +100 | Medium (new Telegram event type, needs testing) | 3-4h |
| Bot: remove text ack | -1 | Trivial | 10m |
| MCP: stale recovery improvement | ~15 | Low | 30m |
| MCP: follow-up compaction | +60 | Low | 1-2h |
| MCP: edit note in mark_processed | +10 | Low | 30m |
| Dispatcher prompt | +6 sentences | Low | 30m |
| **Total** | **~400 lines net** | | **~10-12h** |

No new MCP tools in the required path (claim_and_ack is deferred). No new directories needed.

New message types the dispatcher must handle: 2 (`reaction`, `compact_group`)

---

### 7a. Exact duplicate detection

Deprioritized. Not worth the complexity for now. The dominant case (user accidentally double-sends) is rare, and the cost of processing a duplicate is low — the dispatcher just handles the same request again. Flag as future/optional.

### 7b. Semantic duplicate detection

Detecting "remind me to buy milk" and "don't forget milk from the store" as duplicates requires an LLM call in the MCP hot path. Cost: 1-2 API calls per message, 200-500ms latency added. Benefit: catches rare intentional rephrasing. Not worth it.

### 7c. Conversation thread linking for reactions

Fully linking a reaction to its semantic context requires storing a rich history of sent messages with their intent. The minimal approach (ring buffer of sent message texts, included verbatim in the reaction signal) gives the dispatcher 90% of the value at 5% of the cost.

### 7d. Multi-message batching ("user sent 5 messages, batch them")

The idea: if a user sends 3 messages rapidly, batch them into one dispatcher turn. Requires a hold-and-wait buffer in the bot (similar to media group buffering). The benefit is fewer dispatcher round-trips. The cost: added latency for every message (must wait for the buffer window). Not worth it — the dispatcher handles sequential messages fine, and the ack-first pattern makes latency less painful. **Note:** Compaction (Section 3b) handles the correction/follow-up case without imposing a buffer delay on all messages.

### 7e. LLM-based intent classification in the bot layer

Some proposals call for the bot to classify messages (question vs. command vs. clarification) before writing to inbox. This burns tokens at the ingestion layer, adds latency, and is redundant — the dispatcher already classifies intent as part of processing. Keep the bot dumb (write to inbox, done).

### 7f. Reaction-to-button replacement

Replacing inline keyboard buttons with reactions as the primary interaction model. Reactions are one-bit signals (yes/no/ok). Buttons support arbitrary callback data and can be removed after pressing. They serve different purposes. Both should exist. Reactions are for quick confirmations of questions already asked; buttons are for structured choices.

### 7g. "Ghost detection" improvements

The current ghost detector (stale processing recovery at 300s) is adequate once we reduce the timeout to 90s for text messages. Building a more sophisticated ghost detector (e.g., checking if the subagent process is still alive via PID files) adds significant complexity for marginal benefit. Subagent crashes are rare.

### 7h. `claim_and_ack` MCP tool

Low priority. Does not need to gate other work. The atomicity improvement is marginal — stale recovery handles crash cases adequately. Defer to future.

---

### Edited message (new inbox entry)
```json
{
  "id": "1234567890_42",
  "source": "telegram",
  "type": "text",
  "chat_id": 123456,
  "telegram_message_id": 42,
  "text": "actually do Y instead",
  "_edit_of_telegram_id": 42,
  "_replaces_inbox_id": "1234560000_42",
  "timestamp": "2026-03-16T10:00:01Z"
}
```

### Compact group (merged follow-up messages)
```json
{
  "id": "compact_1234567890",
  "source": "telegram",
  "type": "compact_group",
  "chat_id": 123456,
  "messages": [
    {"id": "1234560000_40", "text": "Do X", "timestamp": "..."},
    {"id": "1234567890_41", "text": "Actually cancel X and do Y instead", "timestamp": "..."}
  ],
  "compact_reason": "negation_marker",
  "text": "[Compact group: 2 messages merged]",
  "timestamp": "2026-03-16T10:00:02Z"
}
```

### Reaction signal
```json
{
  "id": "1234567890_reaction_99",
  "source": "telegram",
  "type": "reaction",
  "chat_id": 123456,
  "telegram_message_id": 99,
  "emoji": "👍",
  "signal": "yes",
  "reacted_to_text": "Should I merge PR #42?",
  "text": "[Reaction: 👍 (yes) on message 99]",
  "timestamp": "2026-03-16T10:00:05Z"
}
```

---

## Summary: Priority order for implementation

1. **Reduce stale recovery timeout to 90s** (10m, high value — fixes worst UX failure)
2. **Remove text ack, keep typing indicator** (10m, cleans up noise)
3. **Dispatcher ack policy in prompt** (30m, changes dispatcher behavior for long tasks)
4. **`edited_message` handler — deliver with note** (3h, handles a real Telegram use case; result is surfaced, not silently dropped)
5. **Follow-up compaction** (2h, handles "actually do Y instead" via reply threading + negation markers; compacts rather than drops)
6. **Reaction signals** (4h, enables 👍/👎 as affirmative/negative)
7. **`claim_and_ack` MCP tool** (optional/future — low priority, does not block other work)
8. ~~**Duplicate detection**~~ (deprioritized — not worth the complexity)

Items 1-3 can ship independently and immediately. Items 4-6 each require a PR. Items 7-8 are deferred.

sayhar:
## Summary

- Adds `handle_edited_message` to `lobster_bot.py` that handles Telegram `edited_message` events
- Implements the "deliver, don't discard" principle: when a user edits a message, it is queued as a new inbox entry with annotation fields rather than silently dropped
- Adds `_find_message_by_telegram_id` helper that scans `inbox/` and `processing/` to detect if the original is still pending
- Updates `.claude/dispatcher.bootup.md` with guidance on how to interpret `_edit_of_telegram_id`, `_replaces_inbox_id`, and `_edit_note` fields

## Annotation fields added to edited messages

- `_edit_of_telegram_id`: always present — the Telegram message ID of the original
- `_replaces_inbox_id`: present when the original is still in inbox or processing — the filename stem of the original message file
- `_edit_note`: present when the original was already processed — signals this is a fresh request based on the edited text

## What is NOT changed

- Subagents working on the original message are not cancelled — their results are delivered normally and the edit is queued separately for a follow-up dispatcher turn
- No MCP server changes required for this item (edit note injection in mark_processed is a separate future item)

Closes #434 (edited_message handler checklist item)

## Test plan

- [ ] Edit a Telegram message while it is still unprocessed — verify a new inbox file appears with `_replaces_inbox_id` set
- [ ] Edit a Telegram message after it has been processed — verify a new inbox file appears with `_edit_note` set
- [ ] Edit a non-text message (e.g. photo caption) — verify it is ignored (text-only guard)
- [ ] Verify dispatcher handles `_edit_of_telegram_id` field without errors on normal messages (field absent = no change in behavior)

### DIFF
--- a/.claude/dispatcher.bootup.md
+++ b/.claude/dispatcher.bootup.md
@@ -267,6 +267,8 @@ When replying, always pass the correct `source` parameter to `send_reply` — Te
 
 **Handling images:** When a message has `type: "image"` or `type: "photo"`, it includes an `image_file` path. **Read images directly on the main thread** — after calling `mark_processing` first to prevent health check restarts.
 
+**Handling edited messages:** When a message has `_edit_of_telegram_id` set, it is the user's edited version of a previously sent message. Process it as a normal message. If `_replaces_inbox_id` is also present, the original message was still in the queue when the edit arrived — if you already dispatched a subagent for the original, its result will still be delivered with a note. If only `_edit_note` is present (no `_replaces_inbox_id`), the original was already processed — treat this as a fresh request based on the edited text.
+
 ```
 1. wait_for_messages() → image message arrives
 2. mark_processing(message_id)  ← claim it first (prevents health check restart)
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -901,6 +901,89 @@ async def handle_message(update: Update, context: ContextTypes.DEFAULT_TYPE):
     # No text acknowledgment — the typing indicator already signals receipt.
 
 
+def _find_message_by_telegram_id(tg_message_id: int) -> Path | None:
+    """Scan inbox/ and processing/ for a message file with a matching telegram_message_id.
+
+    Returns the Path to the matching file, or None if not found.
+    Only text messages are checked — edits arrive as updated_message which carries
+    the same message_id as the original, so this check finds messages that haven't
+    been processed yet.
+    """
+    _processing_dir = _MESSAGES / "processing"
+    for search_dir in (INBOX_DIR, _processing_dir):
+        if not search_dir.exists():
+            continue
+        for f in search_dir.glob("*.json"):
+            try:
+                data = json.loads(f.read_text())
+                if data.get("telegram_message_id") == tg_message_id:
+                    return f
+            except Exception:
+                continue
+    return None
+
+
+async def handle_edited_message(update: Update, context: ContextTypes.DEFAULT_TYPE):
+    """Handle edited_message events from Telegram.
+
+    Design principle: deliver, don't discard.
+
+    If the original message is still in inbox/ or processing/, queue the edit
+    as a new message with annotation fields so the dispatcher knows it's an edit.
+    If the original has already been processed, also queue it as new (the edit
+    may still be actionable).
+
+    The subagent working on the original is NOT cancelled — its result is delivered
+    with a note that the user has since edited their message.
+    """
+    message = update.edited_message
+    if not message:
+        return
+
+    user = update.effective_user
+    if not user or user.id not in ALLOWED_USERS:
+        return
+
+    text = message.text
+    if not text:
+        return  # Ignore non-text edits (media caption edits are not yet handled)
+
+    original_tg_id = message.message_id
+    original_file = _find_message_by_telegram_id(original_tg_id)
+
+    msg_id = f"{int(time.time() * 1000)}_edit_{message.message_id}"
+
+    msg_data = {
+        "id": msg_id,
+        "source": "telegram",
+        "type": "text",
+        "chat_id": message.chat_id,
+        "telegram_message_id": message.message_id,
+        "user_id": user.id,
+        "username": user.username,
+        "user_name": user.first_name,
+        "text": text,
+        "timestamp": datetime.utcnow().isoformat(),
+        "_edit_of_telegram_id": original_tg_id,
+    }
+
+    if original_file is not None:
+        msg_data["_replaces_inbox_id"] = original_file.stem
+        log.info(
+            f"Edit of tg_id={original_tg_id} queued as {msg_id} "
+            f"(original file: {original_file.name})"
+        )
+    else:
+        msg_data["_edit_note"] = "User edited a previously processed message."
+        log.info(
+            f"Edit of tg_id={original_tg_id} queued as {msg_id} "
+            "(original already processed)"
+        )
+
+    inbox_file = INBOX_DIR / f"{msg_id}.json"
+    atomic_write_json(inbox_file, msg_data)
+
+
 async def handle_audio_message(
     update: Update,
     context: ContextTypes.DEFAULT_TYPE,
@@ -1289,6 +1372,7 @@ async def run_bot():
     bot_app.add_handler(MessageHandler(filters.PHOTO, handle_message))
     bot_app.add_handler(MessageHandler(filters.Document.ALL, handle_message))
     bot_app.add_handler(CallbackQueryHandler(handle_callback_query))
+    bot_app.add_handler(MessageHandler(filters.UpdateType.EDITED_MESSAGE & filters.TEXT, handle_edited_message))
     bot_app.add_error_handler(error_handler)
 
     # Initialize and start


sayhar:
LGTM. Clean, additive change with zero breaking surface.

- Handler registration uses the correct `filters.UpdateType.EDITED_MESSAGE & filters.TEXT` filter, scoped tightly to edited text messages.
- `_find_message_by_telegram_id` scans inbox and processing dirs with proper exception handling; will not throw.
- `handle_edited_message` has all the right guards (edited_message presence, ALLOWED_USERS, text-only), and uses `atomic_write_json` consistent with the rest of the codebase.
- The three annotation fields (`_edit_of_telegram_id`, `_replaces_inbox_id`, `_edit_note`) cleanly cover both cases (original still pending vs. already processed).
- Dispatcher docs updated to match.
- Zero deletions — existing `handle_message` is untouched.

Merging.

PR #475, opened 2026-03-16
docs(subagent): PR and issue bodies are canonical state, not just openers

sayhar:
## What this adds

A new rule in `sys.subagent.bootup.md` establishing that **PR and issue bodies must always reflect the current state of the work**, not just what was true at creation time.

## Why

Comment threads are history. Bodies are truth. A reader skimming a PR or issue body should understand what the thing *is now* — not have to read 30 comments and mentally apply a series of design changes to reconstruct the current state.

This is a real workflow failure mode: subagents (and humans) open a PR or issue with an accurate body, then evolve the design through comments, and leave the body stale. Future readers — or the same reader a week later — are left with misleading documentation at the top.

## The new standard

Added as `## PR and Issue Body: Always Canonical` at the end of the `sys.subagent.bootup.md` file:

- When design changes, scope changes, or concerns are resolved via comments, the body must also be updated
- When updating: note briefly what changed and why (e.g., *"Updated: X now does Y instead of Z per comment thread"*), so comment thread continuity is preserved
- Applies to both PRs and issues
- Clarity for a future reader takes priority over preserving the original framing

## Scope

Single file change: `.claude/sys.subagent.bootup.md` (+13 lines).


sayhar:
way too long. token bloat

PR #477, opened 2026-03-16
fix: extract agentId from transcript to close ghost agent DB rows

sayhar:
## The problem

Background subagent Tasks were left permanently stuck in "running" status in `agent_sessions.db` — the ghost agent bug. The SubagentStop hook was supposed to close those rows when a subagent finished, but it was only succeeding for the auto-register stub row (id = session UUID), not the "proper" row registered by the dispatcher via `register_agent()` (id = hex agentId like `a78e2e20dbc483b2e`).

## How SubagentStop session cleanup works

When a subagent session ends, the SubagentStop hook fires and calls `_mark_session_completed()` (which wraps `session_store.session_end()`) with identifiers that match DB rows. The same subagent can have two rows:

1. A **proper row** — `id = hex agentId`, created by the dispatcher via `register_agent()` after spawning the Task.
2. An **auto-register stub** — `id = session UUID`, created by the SessionStart hook when the subagent starts.

The hook had two identifier strategies:
- **Strategy (a): task_id from write_result inputs** — `session_end()` matches on id OR task_id, so this reaches the proper row when the subagent happens to use its hex agentId as its task_id. For most background Tasks, task_id is something like `"subagent-stop-fix"` — a semantic label, not the agentId — so this misses the proper row.
- **Strategy (b): session_id from hook data** — this is the CC internal UUID, which matches only the auto-register stub. The proper row is never reached.

Result: for background Tasks where strategy (a) doesn't accidentally match, the proper hex-agentId row stays in "running" state indefinitely.

## The fix

Every transcript message entry contains an `"agentId"` field — the same hex identifier the dispatcher uses in `register_agent()`. This is the direct primary key of the proper DB row.

This PR adds:
- `_extract_agent_id_from_transcript()` — scans transcript messages for the first valid `agentId` value (>= 8 chars)
- **Strategy (c)** in the SubagentStop handler: after calling the two existing strategies, also call `_mark_session_completed(agent_id)` with the extracted agentId

All three calls are idempotent (safe to call with the same id twice). The new strategy gives the hook a reliable, direct path to the proper row regardless of what task_id the subagent used.

## What this does NOT fix

The proper defense-in-depth fix is for the dispatcher to pass `task_id=<hex-agentId>` when calling `register_agent()` — that would make strategy (a) reliably reach the proper row without needing transcript extraction. A code comment in the module docstring now documents this as a recommended future improvement.

The inbox_server.py reconciler (which closes rows via output file stop_reason detection) is also unaffected — it remains the last-resort fallback.

## Functional patterns used

Pure helper function (`_extract_agent_id_from_transcript`) with a clear input/output contract and no side effects. The three session-close calls in the handler follow a defensive, idempotent fan-out pattern.

## Tests run

- [x] Linting pass: `uv run ruff check hooks/require-write-result.py` — clean, no errors
- [ ] Full test suite — no test infrastructure available in this environment for the hooks layer
- [ ] Live subagent ghost agent test — requires live dispatcher session; deployment is immediate via the hooks directory (no restart needed)

**Blocked items needing attention before merge:** none — the change is low-risk (best-effort DB update, all failures silently swallowed, no blocking behavior changed).

## Testing

Live-tested after deployment. Results (generated by Lobster):

- **Test 1a** — `register_agent` called without `task_id`: blocked by `require-register-agent-task-id.py` hook ✅
- **Test 1b** — `register_agent` called with `task_id`: passes through ✅
- **Test 2 (SubagentStop DB fix)** — agent `a113500c3ef29f082` (task_id=`hook-integration-test`) showed `status='completed'` in `agent_sessions.db` after exit, confirming agentId extraction correctly updated the proper row ✅
- **Unit tests** — 9/9 pass (pytest)


sayhar (commented):
## Lobster Code Review — Automated Test Results

**Tests:** 9/9 unit tests pass

```
uv run python -m pytest tests/test_require_write_result.py -v
```

All 9 tests in `test_require_write_result.py` pass green.

---

**Code Review Summary**

- Function implementation is correct: `extract_agent_id_from_transcript` correctly parses the transcript JSON and extracts the `agentId` field used during `register_agent`.
- Call site is correct: the SubagentStop hook invokes the extraction function at the right point in the fallback chain (after `task_id` and `agent_tag` strategies fail).
- Edge cases handled: missing transcript, malformed JSON, missing `agentId` field — all return `None` gracefully and allow the hook to fall through without crashing.

**Fix assessment:** Logically correct. The third identifier strategy extracts `agentId` from the transcript and hits the DB row directly by primary key, closing the gap where ghost agent rows were left open when neither `task_id` nor `agent_tag` matched.

---

**Noted future improvement (non-blocking)**

The dispatcher could set `task_id=hex(agentId)` when calling `register_agent()`, so strategy (a) — matching by `task_id` — would also work without requiring transcript extraction. This would make the fallback chain shorter and more robust. Not required for this fix.

---

**Verdict: Ready to merge.**

_Generated by Lobster (automated review subagent)_

PR #483, opened 2026-03-16
fix: add task_id generation to dispatcher delegation pattern

sayhar:
## Summary

- The "How to delegate" section previously showed `register_agent` without mentioning that `task_id` must be generated **before** spawning the agent
- Without `task_id`, the `require-register-agent-task-id` hook blocks `register_agent` entirely
- Without `task_id` flowing through to `write_result`, the `SubagentStop` hook cannot match DB rows to auto-unregister completed agents

## What changed

Updated the delegation pattern in `sys.dispatcher.bootup.md` to:
1. Generate `task_id` first (step 1, before spawning or acking)
2. Include `task_id` in the agent prompt so subagents pass it to `write_result`
3. Pass `task_id` explicitly to `register_agent` with a comment marking it as REQUIRED

## Why this matters

The hook `require-register-agent-task-id.py` validates that `task_id` is present when `register_agent` is called. If the dispatcher follows the old pattern, every agent registration fails at the hook. Additionally, the `SubagentStop` handler uses `task_id` to look up the DB row and mark agents as completed — without it, agents stay in "running" state indefinitely and the tracker goes stale.

## Test plan

- [ ] Verify the delegation pattern in the updated file matches the new step numbering (1-8)
- [ ] Confirm `task_id` appears in the `register_agent` call with the REQUIRED comment
- [ ] Confirm the agent prompt template includes `task_id` placeholder text

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -56,20 +56,26 @@ Before spawning a subagent, decide whether to ack based on expected task duratio
 
 **How to delegate:**
 ```
-1. [If task will take >4s]: send_reply(chat_id, "On it.")   # brief ack, 1-3 words
-2. task_result = Task(prompt="...", subagent_type="general-purpose", run_in_background=true)
-3. agent_id = extract agentId from task_result text (look for "agentId: <id>")
-4. output_file = extract output file path from task_result text (look for a /tmp/... path ending in .output)
-5. register_agent(
+1. Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)
+2. [If task will take >4s]: send_reply(chat_id, "On it.")   # brief ack, 1-3 words
+3. task_result = Task(
+       prompt="...Your task_id is <task_id>. Pass it to write_result...",
+       subagent_type="...",
+       run_in_background=true
+   )
+4. agent_id = extract agentId from task_result text (look for "agentId: <id>")
+5. output_file = extract output file path from task_result text (look for a /tmp/... path ending in .output)
+6. register_agent(
        agent_id=agent_id,
+       task_id=task_id,           # REQUIRED — enables reliable DB matching in SubagentStop
        description="Brief what/why + chat_id",
        chat_id=chat_id,
        source=msg.get("source", "telegram"),
-       output_file=output_file,   # enables liveness detection via mtime
-       timeout_minutes=30,        # adjust for long-running tasks
+       output_file=output_file,
+       timeout_minutes=30,
    )
-6. mark_processed(message_id)
-7. Return to wait_for_messages() IMMEDIATELY
+7. mark_processed(message_id)
+8. Return to wait_for_messages() IMMEDIATELY
 ```
 
 **Closing the loop when write_result arrives:**


sayhar (commented):
## Code Review

### 1. task_id generation step — clear and actionable?

**Mostly yes, with one gap.** Step 1 now reads:

> Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)

The examples are helpful but the format guidance is implicit rather than explicit. Suggest tightening to:

> Generate a short kebab-case task_id (e.g. `"fix-pr-475"`, `"daily-digest"`). Must be a non-empty string with no spaces.

This matters because the hook rejects blank or missing task_id — if a dispatcher generates `task_id = None` or an empty string, the error message may not make the root cause obvious.

### 2. task_id in the agent prompt — correct?

**Yes.** Step 3 now shows `prompt="...Your task_id is <task_id>. Pass it to write_result..."`. This is exactly right. Minor suggestion: show it as an f-string interpolation (`f"...Your task_id is {task_id}..."`) rather than a `<placeholder>` to remove ambiguity that the dispatcher must actually interpolate the value.

### 3. register_agent call updated with task_id param?

**Yes, and clearly marked.** The `# REQUIRED — enables reliable DB matching in SubagentStop` comment is exactly right. No issues.

### 4. PR description clarity on WHY this matters?

**Yes — strongest part of the PR.** The description explicitly names both failure modes: hook blocks register_agent entirely, and SubagentStop cannot match DB rows leaving agents in "running" state indefinitely. That is the right level of justification.

### 5. Regressions — does the end-to-end delegation pattern still hold?

**Yes, with one observation.** Step numbering shifted from 6 to 8 cleanly. However, the artifact-inlining block added to the `sent_reply_to_user=false` branch is somewhat out of scope for a PR titled "add task_id generation" — not a regression, but adds review surface area. Consider whether it warrants its own PR or at minimum its own test-plan entry.

Also: the bare `except: pass` on the artifact read is worth flagging. Silent failure on unreadable artifacts means the user gets a truncated reply with no indication that content was dropped. At minimum a warning log would help.

### Follow-up question (Sahar has not yet responded)

The dispatcher raised: **should there be a hook that blocks Agent tool calls where the prompt does not contain `task_id`?**

Worth resolving before or shortly after merge. The current hook validates `task_id` at `register_agent` time, but by then the Task has already been spawned. A pre-spawn hook on the Agent tool call itself — checking that the prompt string contains the substring "task_id" — would catch the error earlier with a clearer message. Flagging for Sahar.

### Summary

- Core change (task_id generation + register_agent + prompt threading): **correct and necessary**
- PR description: **clear and complete**
- Minor issues: task_id format not fully specified; f-string vs placeholder ambiguity; artifact inlining out of scope; bare `except: pass`
- Blocking issues: none from this reviewer
- Status: ready to merge after Sahar sign-off

sayhar (commented):
PASS

**What changed** — two independent improvements to `sys.dispatcher.bootup.md`:

1. **Delegation pattern** — adds `task_id` generation as step 1, threads it into the `Task(prompt=...)` call, and passes it to `register_agent`. This closes the gap where the SubagentStop hook had no reliable way to match a stopping agent to its originating task. Step numbering updated correctly (6→7, 7→8).

2. **Artifact inlining** — dispatcher now reads artifact file content and inlines it into `send_reply` instead of relaying raw paths. The "never relay raw file paths" note is a good explicit reminder, since file paths are useless on mobile.

**Correctness** — the `try/except: pass` on unreadable files is acceptable for pseudocode instructions; a real implementation should probably log or surface failures rather than silently drop artifacts, but this is doc-level guidance, not executable code.

**Relationship to #485** — this PR documents the dispatcher side of the contract; #485 enforces it at the hook level. The two are complementary and should be merged together (or #483 first, since the doc is prerequisite context for the hook).

**No migration needed** — documentation-only change.

sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS — Rebase resolved conflict by keeping HEAD (main) version of the subagent result handler, which is the superset: it includes both the PR-reviewer spawning logic added in #520 and the artifact-inlining that this PR originally introduced. The task_id-first delegation pattern in the 'How to delegate' section is correct and complete.

PR #485, opened 2026-03-16
feat: PreToolUse hook blocking Agent spawns without task_id in prompt

sayhar:
## Summary

Subagents that are spawned without a task_id in their prompt have no way to call write_result with the correct task_id — making DB row matching in the SubagentStop hook unreliable. This hook closes that gap at the source: it hard-blocks the Agent tool call before the subagent is launched.

The hook is the spawn-side mirror of `require-register-agent-task-id.py`: that hook blocks `register_agent` calls missing task_id; this hook blocks the Agent spawn if the prompt doesn't contain the string `"task_id"` (case-sensitive). The same string that must appear in the prompt is what the subagent is expected to pass to write_result, so the two hooks form a matching pair.

**Implementation:** pure function — reads stdin JSON, checks one field, exits 0 or 2. No imports beyond stdlib. Registered in `install.sh` using the same idempotent jq-append pattern as the other Agent hooks (`require-background-agent.py`, `require-subagent-type.py`). Also registered in `~/.claude/settings.json` (live, no deploy needed via symlink).

## Tests run

- [x] `echo '{"tool_name":"Agent","tool_input":{"prompt":"do something"}}' | python3 hooks/require-task-id-in-prompt.py` — exited 2 with expected BLOCKED message on stderr
- [x] `echo '{"tool_name":"Agent","tool_input":{"prompt":"Your task_id is: foo-001. Do something."}}' | python3 hooks/require-task-id-in-prompt.py` — exited 0, no output
- [x] `echo '{"tool_name":"Bash","tool_input":{"command":"ls"}}' | python3 hooks/require-task-id-in-prompt.py` — exited 0 (non-Agent tool passes through)

## Dogfooding
Tested in production during this session: the hook blocked the dispatcher twice when attempting to spawn agents without task_id in the prompt (visible in Claude Code tool call logs). Exit code 2 both times, correct error message displayed.

### DIFF
--- a/hooks/require-task-id-in-prompt.py
+++ b/hooks/require-task-id-in-prompt.py
@@ -11,16 +11,20 @@ the prompt doesn't carry task_id into the subagent's context at all.
 """
 import json, sys
 
-data = json.load(sys.stdin)
-tool = data.get("tool_name", "")
-inp = data.get("tool_input", {})
+try:
+    data = json.load(sys.stdin)
+    tool = data.get("tool_name", "")
+    inp = data.get("tool_input", {})
+except Exception as e:
+    print(f"Warning: require-task-id-in-prompt hook received malformed input: {e}", file=sys.stderr)
+    sys.exit(0)
 
 if tool != "Agent":
     sys.exit(0)
 
 prompt = inp.get("prompt", "")
 
-if "task_id" not in prompt:
+if "task_id is:" not in prompt:
     print(
         'BLOCKED: Agent spawned without task_id in prompt. '
         'Include "Your task_id is: <slug>" so the subagent can pass it to write_result. '
--- a/tests/smoke/test_require_task_id_in_prompt.py
+++ b/tests/smoke/test_require_task_id_in_prompt.py
@@ -1,0 +1,152 @@
+"""
+Smoke tests for hooks/require-task-id-in-prompt.py
+
+D1. Non-Agent tool calls pass through (exit 0, no output).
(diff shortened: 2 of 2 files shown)


sayhar (commented):
PASS

**Hook logic** — clean and focused. The `task_id is:` substring match is a deliberate tightening over a bare `task_id` check, and the D3 test makes the tradeoff explicit. No false-negative risk for the canonical dispatcher pattern (`Your task_id is: <slug>`).

**Failure mode** — exit 2 on missing task_id is correct for a PreToolUse blocker; malformed JSON exits 0 (pass-through) which is the safe default.

**Tests** — five categories (D1–D5) with edge cases for each. Coverage is thorough. The test for `task_id=foo` (D3 variant) correctly rejects an alternative format that could otherwise sneak through.

**install.sh** — idempotency guard uses `jq -e ... | test("require-task-id-in-prompt")` on the command field, which is consistent with the pattern used by the other hooks in this file. Hook registered with `matcher: "Agent"` and a 5-second timeout — appropriate.

**One minor observation:** the hook uses `sys.executable` in the test runner but the install.sh registers it with `python3`. These will diverge if the system `python3` is not the same interpreter used by the test venv. Not a blocker — the hook has no third-party imports so interpreter version is unlikely to matter in practice — but worth noting for consistency with the `uv run` convention used elsewhere in the project.

**No migration entry** — the new hook is install-only (no migration for existing installs). Per `CLAUDE.md`, changes that affect existing installs should have a numbered migration in `scripts/upgrade.sh`. This should be added before merge so existing users get the hook without re-running the full installer.

PR #486, opened 2026-03-16
feat: claim_and_ack atomic tool (mark_processing + send_reply)

### LINKED ISSUE #458: claim_and_ack: atomic MCP tool combining mark_processing + send_reply
Parent issue: #434

## What to implement

Add a `claim_and_ack` MCP tool to `inbox_server.py` that atomically combines `mark_processing` and `send_reply` in a single call. This eliminates the window between claiming a message and acknowledging it.

## Context

Sahar confirmed she wants this. The current dispatcher flow has three separate steps: `mark_processing`, `send_reply`, subagent spawn — with no transactional guarantee between them. A crash between claiming and acking leaves the user seeing silence until stale recovery fires (now 90s for text).

`claim_and_ack` tightens this window: the message is claimed and the user is notified in one operation, so a crash after the call means the user at least got the ack.

Note: the issue originally marked this as low priority / future. Sahar has confirmed it should be built.

## Acceptance criteria

1. `claim_and_ack(message_id, ack_text, chat_id, source?)` tool registered in `inbox_server.py`
2. Atomically: moves message from `inbox/` to `processing/` AND queues the ack reply
3. If the move fails (message not found, already claimed), returns an error without sending the ack
4. If sending the ack fails after the move, the message remains in `processing/` (stale recovery handles it)
5. Adds `_processing_started_at` timestamp to the message JSON when claiming (makes stale detection use actual processing time rather than file mtime)
6. Tool is documented in the dispatcher prompt as the preferred way to claim + ack for long tasks
7. Existing `mark_processing` + `send_reply` separate calls remain valid (backward compatible)

## Files to change

- `src/mcp/inbox_server.py` — add `claim_and_ack` tool handler, register it (~60 lines)
- Dispatcher prompt — add `claim_and_ack` as the preferred pattern for long tasks (replaces the separate `mark_processing` + `send_reply` pattern)

## Estimated effort

~2 hours

sayhar:
Closes #458

## What claim_and_ack does

The dispatcher's claim-then-ack flow had a latent crash window: `mark_processing`, then `send_reply`, then subagent spawn — three separate calls with no transactional guarantee. A crash between claiming a message and sending the ack would leave the user in silence for up to 90 seconds until stale recovery fired. `claim_and_ack` collapses the first two steps into one call: the message is moved from `inbox/` to `processing/` and the ack reply is queued in the same filesystem operation. A crash after the call means the user already received the ack. A crash before it means nothing happened and the message is still in `inbox/`, ready to be retried. The silence window is gone.

## What changed

**`src/mcp/inbox_server.py`:**
- New `handle_claim_and_ack` handler (~75 lines). Pure function composition: reuses `_find_message_file`, `_queue_observation`, and `handle_send_reply` internally rather than duplicating outbox logic.
- `claim_and_ack` registered in `_SESSION_GUARDED_TOOLS` alongside `mark_processing` and `send_reply` (it performs both operations and needs both guards).
- New Tool schema in `list_tools`, new routing branch in `call_tool`.
- `handle_mark_processing` now stamps `_processing_started_at` on the processing-dir copy so stale detection uses actual claim time rather than file mtime. This applies to both code paths (direct `mark_processing` and `claim_and_ack`).

**`.claude/sys.dispatcher.bootup.md`:**
- `claim_and_ack` documented as the preferred delegation pattern for long tasks, with the existing `mark_processing` + `send_reply` pattern preserved as a valid alternative.

**`tests/unit/test_mcp_server/test_claim_and_ack.py`:**
- 9 tests covering the main success path, claim-failure path, ack-failure path, input validation, and threading passthrough.

## Atomicity model (honest)

This is **not truly atomic** — it is two filesystem operations ordered to be safe in both failure directions:

1. `found.rename(dest)` — filesystem rename (atomic on Linux for same-device moves)
2. `atomic_write_json(dest, msg_data)` — writes updated JSON with `_processing_started_at` timestamp
3. `handle_send_reply(ack_args)` — writes ack to outbox (local file write only; Telegram delivery happens later via the bot process)

**Failure modes:**
- Crash between steps 1 and 2: message is in `processing/` without `_processing_started_at`. Stale recovery falls back to file mtime. Safe.
- Crash between steps 2 and 3: message is claimed, ack not written. Stale recovery resurfaces the message. User may see a delayed ack from stale recovery. Safe.
- Step 3 raises an exception (e.g. filesystem full): message stays in `processing/` — **no rollback**. The return value starts with `"Warning:"` so the caller can distinguish this from full success. The dispatcher should not retry the ack; stale recovery will resurface the message normally.

Note: `handle_send_reply` writes to the outbox synchronously (local file write only). It does not call the Telegram API directly. An exception here is rare in practice (filesystem-full or misconfigured outbox path), but the handler defends against it correctly.

## Test coverage

9 tests, all passing:

| Test | What it covers |
|------|---------------|
| `test_moves_message_to_processing` | Claim mechanics — message leaves inbox/ |
| `test_sends_ack_reply` | Ack written to outbox/ with correct fields |
| `test_stamps_processing_started_at` | `_processing_started_at` present in processing/ copy |
| `test_returns_error_when_message_not_found` | Claim failure → no ack sent |
| `test_no_ack_sent_when_claim_fails` | Already-claimed message → no ack |
| `test_success_result_text_contains_message_id` | Return value sanity |
| `test_requires_message_id` | Input validation (ValidationError) |
| `test_partial_failure_when_ack_send_fails` | **Failure path**: message stays in processing/, return starts with `Warning:`, no rollback |
| `test_reply_to_message_id_forwarded_to_ack` | Threading passthrough to ack reply |

Pre-existing failures on `main` (`TestMarkProcessing.test_requires_message_id`, `TestMarkFailed.test_requires_message_id`, `TestStaleRecovery`, `TestSendReply.test_requires_chat_id`) are unrelated to this PR.

## Reviewer issues and how each was addressed

**Issue 2 (Warning: prefix):** The partial-failure return value did not use a recognisable prefix, making it impossible for the dispatcher to distinguish from full success by pattern-matching. Fixed in commit `7080eb2`: the return value now starts with `"Warning:"` when mark_processing succeeds but send_reply raises.

**Issue 3 (missing ack-send-failure test):** There was no test for the step-3 exception path despite it being the most operationally important failure mode. Fixed in commit `7080eb2`: `test_partial_failure_when_ack_send_fails` mocks `handle_send_reply` to raise, then verifies (a) the return starts with `Warning:`, (b) the error detail appears in the text, and (c) the message file remains in `processing/` with no rollback.

**Issue 5 (dispatcher doc ambiguity on partial-failure retry):** The reviewer noted the docs did not address what the dispatcher should do when `claim_and_ack` returns a Warning. This remains a documentation gap — the dispatcher bootup doc does not yet include guidance on not retrying the ack in the Warning case. Low risk in practice since stale recovery handles it correctly regardless, but worth a follow-up note in the dispatcher doc before merge.

Issues 1, 4, and 6 from the review required no code changes (atomicity is safe and documented, `_SESSION_GUARDED_TOOLS` entry is correct, threading works correctly via auto-threading in `handle_send_reply`).

## Functional patterns used

- Pure function composition: `handle_claim_and_ack` delegates to `handle_send_reply` rather than duplicating outbox logic.
- Fail-safe ordering: claim must succeed before ack is attempted; partial failures are safe in both directions.
- Immutability: message JSON is read, extended with `_processing_started_at`, and written atomically via `atomic_write_json` — no in-place mutation.

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -98,10 +98,13 @@ Before spawning a subagent, decide whether to send the dispatcher ack based on e
   - Reaction messages — no ack, no response unless the reaction warrants one
   - System messages (`source: "system"` or `chat_id: 0`) — never ack
 
-**How to delegate:**
+**How to delegate (preferred — use `claim_and_ack` for long tasks):**
 ```
-1. Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)
-2. [If task will take >4s]: send_reply(chat_id, "On it.")   # brief ack, 1-3 words
+1. [If task will take >4s]: claim_and_ack(message_id, ack_text="On it.", chat_id=chat_id, source=source)
+   # Atomically: moves message from inbox/ → processing/ AND sends the ack reply.
+   # If the claim fails (message already gone), no ack is sent — safe to retry.
+   # If you crash after this call, the user already got the ack and stale recovery reclaims the message.
+2. Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)
 3. Task(
        prompt="---\ntask_id: <task_id>\nchat_id: <chat_id>\nsource: <source>\n---\n\n...<rest of prompt>...",
        subagent_type="...",
@@ -112,6 +115,13 @@ Before spawning a subagent, decide whether to send the dispatcher ack based on e
 ```
 
 Agent registration is fully automatic — a PostToolUse hook fires immediately after each Task call and inserts a 'running' row into agent_sessions.db. You do not need to call register_agent or extract agentId/output_file.
+
+**Alternative (still valid, use when no ack needed):**
+```
+1. mark_processing(message_id)   # claim without ack
+2. [optional] send_reply(chat_id, "On it.")
+3. ... spawn subagent ...
+```
 
 **Closing the loop when write_result arrives:**
 ```
@@ -634,7 +644,7 @@ mark_processed(message_id)
 wait_for_messages() ← loop back
 ```
 
-**Call `mark_processing` first** — before `send_reply`, before re-reading files, before any post-compact re-orientation. This moves the message from `inbox/` → `processing/` and signals to the health check that the message is claimed.
+**Claim messages before doing any work** — before `send_reply`, before re-reading files, before any post-compact re-orientation. Use `claim_and_ack` (preferred for tasks needing an ack) or `mark_processing` (when no ack is needed). Either call moves the message from `inbox/` → `processing/` and signals to the health check that the message is claimed.
 
 **State directories:** `inbox/` → `processing/` → `processed/` (or → `failed/` → retried back to `inbox/`)
 
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -105,6 +105,7 @@ Before spawning a subagent, decide whether to send the dispatcher ack based on e
    # If the claim fails (message already gone), no ack is sent — safe to retry.
    # If you crash after this call, the user already got the ack and stale recovery reclaims the message.
    # If the return value starts with `Warning:`, the message was claimed but the ack failed — do not retry the ack; stale recovery will handle the message.
+   # On a Warning: return, proceed normally with step 2 below (spawn subagent, mark_processed). The claim succeeded; only the ack delivery failed.
 2. Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)
 3. Task(
        prompt="---\ntask_id: <task_id>\nchat_id: <chat_id>\nsource: <source>\n---\n\n...<rest of prompt>...",
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -104,6 +104,7 @@ Before spawning a subagent, decide whether to send the dispatcher ack based on e
    # Atomically: moves message from inbox/ → processing/ AND sends the ack reply.
    # If the claim fails (message already gone), no ack is sent — safe to retry.
    # If you crash after this call, the user already got the ack and stale recovery reclaims the message.
+   # If the return value starts with `Warning:`, the message was claimed but the ack failed — do not retry the ack; stale recovery will handle the message.
 2. Generate a short task_id (e.g. "fix-pr-475", "upstream-check", or a short slug describing the task)
 3. Task(
        prompt="---\ntask_id: <task_id>\nchat_id: <chat_id>\nsource: <source>\n---\n\n...<rest of prompt>...",
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -845,6 +845,7 @@ _SESSION_GUARDED_TOOLS = frozenset({
     "mark_processed",
     "mark_processing",
     "mark_failed",
+    "claim_and_ack",
 })
 
 
@@ -1167,6 +1168,68 @@ async def list_tools() -> list[Tool]:
                     },
                 },
                 "required": ["message_id"],
+            },
+        ),
+        Tool(
+            name="claim_and_ack",
+            description=(
+                "Atomically claim a message for processing and send an acknowledgement reply. "
+                "Preferred over separate mark_processing + send_reply calls for long-running tasks: "
+                "the user is notified in the same operation as the claim, so a crash after this call "
+                "means the user already received the ack. "
+                "If the message is not found in inbox/ (already claimed or missing), returns an error "
+                "without sending the ack. If the ack send fails after claiming, the message stays in "
+                "processing/ and stale recovery handles it."
+            ),
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "message_id": {
+                        "type": "string",
+                        "description": "The message ID to claim (must be in inbox/).",
+                    },
+                    "ack_text": {
+                        "type": "string",
+                        "description": "The acknowledgement text to send to the user (e.g. 'On it.').",
+                    },
+                    "chat_id": {
+                        "oneOf": [
+                            {"type": "integer"},
+                            {"type": "string"},
+                        ],
+                        "description": "The chat/channel ID to send the ack to (from the original message).",
+                    },
+                    "source": {
+                        "type": "string",
+                        "description": "The source to reply via (telegram, slack, etc.). Default: telegram.",
+                        "default": "telegram",
+                    },
+                    "reply_to_message_id": {
+                        "type": "integer",
+                        "description": "Telegram message ID to thread the ack against (Telegram only).",
+                    },
+                    "buttons": {
+                        "type": "array",
+                        "description": "Optional inline keyboard buttons (Telegram only). Same format as send_reply.",
+                        "items": {
+                            "type": "array",
+                            "items": {
+                                "oneOf": [
+                                    {"type": "string"},
+                                    {
+                                        "type": "object",
+                                        "properties": {
+                                            "text": {"type": "string"},
+                                            "callback_data": {"type": "string"},
+                                        },
+                                        "required": ["text"],
+                                    },
+                                ]
+                            },
+                        },
+                    },
+                },
+                "required": ["message_id", "ack_text", "chat_id"],
             },
         ),
         Tool(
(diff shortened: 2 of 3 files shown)


sayhar (commented):
## Code Review: claim_and_ack atomic MCP tool

Overall: solid implementation. The failure semantics are correct and the tests cover the important paths. A few issues to address before merge.

---

### 1. Atomicity — is it actually atomic?

**Not fully atomic, but the failure modes are safe and honestly documented.**

The sequence is:
1. `found.rename(dest)` — filesystem rename (atomic on Linux for same-device moves)
2. `atomic_write_json(dest, msg_data)` — writes updated JSON with timestamp
3. `handle_send_reply(ack_args)` — writes to outbox

A crash between steps 1 and 2 leaves the message in `processing/` without `_processing_started_at`. Stale recovery falls back to file mtime, which is correct per the comment. Safe.

A crash between steps 2 and 3 leaves the message claimed with no ack sent. The docstring and dispatcher notes both document this explicitly — stale recovery handles it. Safe.

**One gap:** the tool description says "If the ack send fails after claiming, the message stays in processing/ and stale recovery handles it." But `handle_send_reply` writes to the outbox **synchronously** (it's just a local file write via `atomic_write_json`). It does not actually call the Telegram API — the bot process does that later. So the ack send can only "fail" if the filesystem is full or the outbox path is wrong. In practice the ack will almost always appear to succeed here even if Telegram delivery ultimately fails. This is fine behaviour but the docstring slightly overstates the risk — minor nit, not blocking.

---

### 2. Return value — does the dispatcher know what happened?

**Yes for success. The error path is underspecified — potential issue.**

Success path returns: `"Claimed and acked: {message_id}\nAck: {ack_text_preview}"` — clearly parseable.

**Issue:** When the ack send raises an exception (step 3 failure), the return text is:
```
"Message claimed: {message_id} — but ack send failed: {e}\nMessage remains in processing/. Stale recovery will handle it."
```

This does not start with `"Error:"` so the dispatcher cannot distinguish it from a normal success result by prefix-matching. The message IS claimed in this path, so it shouldn't be retried — but the dispatcher has no reliable signal to know which of its two sub-steps succeeded. **Recommendation:** prefix with `"Warning:"` rather than neither, so the caller can pattern-match and log appropriately without treating it as a full error. Not blocking if stale recovery is truly relied upon, but worth fixing.

---

### 3. Tests — are the 8 tests adequate?

**Yes, coverage is good. One gap.**

Tests present:
- `test_moves_message_to_processing` — claim mechanics
- `test_sends_ack_reply` — ack written to outbox
- `test_stamps_processing_started_at` — stale recovery support
- `test_returns_error_when_message_not_found` — claim failure → no ack
- `test_no_ack_sent_when_claim_fails` — already-claimed message → no ack
- `test_success_result_text_contains_message_id` — return value sanity
- `test_requires_message_id` — input validation
- `test_reply_to_message_id_forwarded_to_ack` — threading passthrough

**Missing test:** there is no test for the ack-send-fails path (step 3 exception). The docstring describes specific return text in this case and promises the message stays in `processing/` — but there is no test that mocks `handle_send_reply` to raise and verifies (a) the return text contains the failure description, and (b) the message file is still in `processing/`. This is the most operationally important failure mode. **Recommended to add before merge.**

---

### 4. _SESSION_GUARDED_TOOLS — is the entry correct?

**Yes.** `claim_and_ack` is added to `_SESSION_GUARDED_TOOLS` alongside `mark_processing` and `send_reply`, which is exactly right. It performs both operations so it needs both guards. No issue here.

---

### 5. dispatcher.bootup.md — does the documentation correctly describe usage?

**Mostly yes. One ambiguity.**

The new preferred pattern shows:
```
1. claim_and_ack(message_id, ack_text="On it.", chat_id=chat_id, source=source)
```

The alternative still-valid pattern shows:
```
1. mark_processing(message_id)   # claim without ack
2. [optional] send_reply(chat_id, "On it.")
```

The comment "If the claim fails (message already gone), no ack is sent — safe to retry" is accurate.

**Ambiguity:** the docs don't address what the dispatcher should do when `claim_and_ack` returns the partial-failure text (message claimed, ack failed). Should it re-send the ack manually? Fall back to stale recovery silently? The text says "stale recovery handles it" but a confused dispatcher might try to call `send_reply` again, double-acking if the outbox write actually succeeded. **Recommend adding a note:** "If claim_and_ack returns a Warning (ack failed), do not retry the ack — stale recovery will resurface the message."

---

### 6. reply_to_message_id threading

**Works correctly, with one note.**

`reply_to_message_id` is forwarded to `handle_send_reply`, which passes it as `reply_to_message_id` in the outbox JSON. This is the same path used by direct `send_reply` calls. The test `test_reply_to_message_id_forwarded_to_ack` verifies this.

However, `handle_send_reply` also does **auto-threading**: if `reply_to_message_id` is not provided, it scans `processing/` for a Telegram message from the same chat and auto-threads. In `claim_and_ack`, by the time `handle_send_reply` is called, the message has already been moved to `processing/` — so the auto-threading lookup will find it. This means that even without passing `reply_to_message_id`, the ack will be auto-threaded under the message being claimed. This is the desired behaviour and no change is needed, but it is worth noting in the docstring for clarity.

---

### Summary

| # | Area | Verdict |
|---|------|---------|
| 1 | Atomicity | Safe, failure modes documented correctly |
| 2 | Return value on partial failure | Suggest "Warning:" prefix — not blocking |
| 3 | Tests | Missing ack-send-failure test — **recommend adding before merge** |
| 4 | _SESSION_GUARDED_TOOLS | Correct |
| 5 | Docs | Minor ambiguity on partial-failure retry behaviour |
| 6 | Threading | Works correctly, auto-threading will activate |

**Blocking:** None that would prevent merge, but the missing ack-send-failure test is a real gap given that path is explicitly documented behaviour. Recommend adding it. The "Warning:" prefix suggestion would make the partial-failure return value safer for dispatcher pattern-matching.

Good design overall — fail-safe ordering is correct and the composition via `handle_send_reply` avoids duplicating outbox logic.

sayhar (commented):
PASS: doc gap fixed — Warning retry behavior now documented in dispatcher bootup. Ready to merge.

sayhar (commented):
PASS: The added line accurately describes the Warning: return behavior. The implementation confirms: (1) on ack failure, handle_claim_and_ack returns a string starting with 'Warning:' with no rollback, so the message stays in processing/; (2) the claim has already succeeded at that point, so the dispatcher must proceed normally through steps 2–7 to spawn the subagent and eventually call mark_processed; (3) test_partial_failure_when_ack_send_fails explicitly asserts this exact contract. The instruction is correct and complete.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR adds `claim_and_ack` — an atomic tool that combines `mark_processing` and `send_reply` — and documents it as the preferred pattern for long-running tasks.

**What the code does:**
1. **New `claim_and_ack` MCP tool**: Atomically claims a message (moves `inbox/ → processing/`) and sends an ack reply. If the claim fails (message not in inbox), the ack is never sent. If the ack fails after claiming, the message stays in `processing/` and stale recovery handles it — explicitly a no-rollback design. Returns `Warning:` prefix on partial failure (claimed but ack failed).
2. **`_processing_started_at` timestamp**: `handle_mark_processing` now stamps the actual claim time into the message JSON, so stale detection uses real processing start time rather than file mtime. `handle_claim_and_ack` does the same.
3. **`sys.dispatcher.bootup.md`**: Documents `claim_and_ack` as the preferred approach for long tasks, and `mark_processing` as the alternative when no ack is needed. Updates the delegation steps to include `claim_and_ack` as step 1.

**Why it matters:** The original pattern (separate `mark_processing` + `send_reply`) had a window where the message was claimed but the user had not yet been notified — a crash between these two calls would leave the user with no ack and a message stuck in `processing/`. `claim_and_ack` closes this window: if the call returns success, both operations completed.

**Observations:**
- The no-rollback design for partial failures is explicitly documented and is the right choice: rolling back a claimed message would re-expose it to the health check and could cause duplicate processing. The stale-recovery path is the intended handler for the failure case.
- The `_processing_started_at` stamp is additive and non-fatal (best-effort write). Correct.
- The ``claim_and_ack`` handler duplicates some observation-queuing logic from `handle_mark_processing`. This is minor duplication but acceptable since the two paths are meant to behave identically at the claim step.
- The test suite covers the key cases: success, message-not-found, ack-fails-after-claim, and `reply_to_message_id` forwarding.

PR #488, opened 2026-03-16
feat: alert on internal pipeline failures

### LINKED ISSUE #416: feat: internal pipeline failures should alert dispatcher and external logs
## Problem

When internal pipeline components fail (e.g., transcription worker, whisper-cli, ffmpeg conversion), failures are currently silent:
- Errors go to local log files only
- The dispatcher is never notified
- The user has no visibility that voice messages are failing
- The system just moves messages to dead-letter without any alert

This was discovered when 4 voice messages silently accumulated in dead-letter over multiple sessions before anyone noticed.

### 1. Dispatcher notification on internal failures

When a pipeline component fails repeatedly or hits dead-letter, it should write to the dispatcher inbox via `write_observation(category="system_error", ...)` so the dispatcher can:
- Log it
- Optionally notify the user
- Take corrective action

This is critical because the dispatcher is the only process with user-facing notification capability. Without this signal, internal failures are invisible until a user manually notices something is wrong.

### 2. External log shipping

Errors should be written to logs that leave the box (not just `~/lobster-workspace/logs/`). Options:
- Send to a remote logging service (e.g., Papertrail, Loki, Datadog)
- Or at minimum, surface via Telegram alert (the dispatcher already has this capability via `send_reply`)

Local-only logs are insufficient because they require SSH access to inspect and are not monitored in any automated way.

## Affected components

- `src/transcription/worker.py` — should call `write_observation` on repeated failures before dead-lettering
- Any other internal pipeline worker that can silently fail (e.g., future media processors, webhook consumers)

## Proposed approach

In `worker.py`, after exhausting retries and moving a message to dead-letter:

```python
# Notify dispatcher of dead-letter event
await mcp.write_observation(
    chat_id=original_chat_id,
    text=f"Voice message {msg_id} moved to dead-letter after {MAX_RETRIES} failed transcription attempts. Last error: {error}",
    category="system_error",
    task_id=f"transcription-dead-letter-{msg_id}"
)
```

The dispatcher's existing `system_error` handler logs this and forwards to the user when debug mode is on.

For external log shipping, a lightweight option is to have the dispatcher forward `system_error` observations to a Telegram alert channel (or the user's own chat) unconditionally — not just in debug mode — for dead-letter events. This gives durable, off-box evidence of failures with zero infrastructure overhead.

## Acceptance criteria

- [ ] `worker.py` calls `write_observation(category="system_error")` when a message is moved to dead-letter
- [ ] The dispatcher handles `system_error` observations and at minimum logs them prominently
- [ ] Dead-letter events trigger a user-visible Telegram notification (configurable severity threshold)
- [ ] At least one off-box log destination is implemented or documented as a follow-up

sayhar:
## Problem

When `worker.py` exhausts all whisper retries and moves a voice message to dead-letter, the failure is completely invisible: it logs to a local file, the message disappears from the pending queue, and the dispatcher (and user) never know. This was how 4 voice messages silently accumulated in dead-letter before anyone noticed.

## What this PR does

Two changes that together close the alerting gap:

**1. `worker.py` — notify dispatcher on dead-letter**

`move_to_dead_letter` now calls `notify_dispatcher_dead_letter` after every dead-letter event. That function writes a `subagent_observation` JSON file directly to `~/messages/inbox/` — the same format `handle_write_observation()` in `inbox_server.py` produces. The dispatcher picks it up on its next `wait_for_messages()` call.

Alert format:
```
[system_error] Transcription dead-letter: <filename>
Message ID: <id>
Reason: <error details>
```

Admin `chat_id` resolution order: `LOBSTER_ADMIN_CHAT_ID` env var → first entry of `TELEGRAM_ALLOWED_USERS` in `config.env`. This mirrors the same lookup already in `inbox_server.py`.

Alerting is fully fault-tolerant: any exception inside `notify_dispatcher_dead_letter` is caught, logged as a warning, and the transcription pipeline continues. The inbox write is atomic (uses the existing `atomic_write_json` helper).

**2. `sys.dispatcher.bootup.md` — always forward `system_error` to user**

Previously, `system_error` observations only triggered a `send_reply` when `LOBSTER_DEBUG=true`. This PR changes the routing to always forward them — the whole point of a dead-letter alert is that the user needs to know, regardless of debug mode. The local `observations.log` append is kept as the durable off-box log record.

## What is not changed

- No changes to `inbox_server.py` or its `write_observation` MCP tool
- No changes to how non-dead-letter transcription failures are handled (they already retry and log locally)
- No new external services or infrastructure required — Telegram is the off-box log destination

## Functional patterns

The implementation uses pure functions with clear input/output contracts (`_read_admin_chat_id`, `notify_dispatcher_dead_letter`), side effects isolated to boundaries (`atomic_write_json` only at the end of the notification path), and graceful degradation (all exceptions in the notification path are caught so the core pipeline is unaffected).

## Tests run

- [x] `uv run pytest tests/unit/test_transcription_worker_alerting.py -v` — 10 passed, 0 failed
- [x] `uv run pytest tests/unit/ -v` — 999 passed (pre-existing failures in test_update_manager and test_memory unrelated to this change; confirmed present on main before this PR)

Closes #416

### DIFF
--- a/src/transcription/worker.py
+++ b/src/transcription/worker.py
@@ -38,7 +38,7 @@ from watchdog.events import FileSystemEventHandler
 # ---------------------------------------------------------------------------
 _MESSAGES = Path(os.environ.get("LOBSTER_MESSAGES", Path.home() / "messages"))
 _WORKSPACE = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
-_CONFIG_DIR = _MESSAGES / "config"
+_CONFIG_DIR = Path(os.environ.get("LOBSTER_CONFIG_DIR", Path.home() / "lobster-config"))
 
 PENDING_DIR = _MESSAGES / "pending-transcription"
 INBOX_DIR = _MESSAGES / "inbox"
@@ -335,7 +335,9 @@ def move_to_dead_letter(pending_file: Path, msg_data: dict, reason: str) -> None
     # Stash the filename so notify_dispatcher_dead_letter can include it in the alert.
     msg_data["_pending_file"] = pending_file.name
     dest = DEAD_LETTER_DIR / pending_file.name
-    atomic_write_json(dest, msg_data)
+    # Pop internal field before writing to disk so it doesn't leak into stored JSON.
+    payload = {k: v for k, v in msg_data.items() if k != "_pending_file"}
+    atomic_write_json(dest, payload)
     pending_file.unlink(missing_ok=True)
     log.error(f"Moved {pending_file.name} to dead-letter: {reason}")
     notify_dispatcher_dead_letter(msg_data, reason)


sayhar (commented):
🤖 Lobster (reviewer):

**Verdict: PASS**

## Summary

Solid, targeted fix for a real operational gap. The design is correct, the fault-tolerance boundary is well-placed, and the PR description accurately traces the full alerting path. Minor observations noted below, none blocking.

---

## Correctness

**Fault isolation** — `notify_dispatcher_dead_letter` wraps everything in a broad `except Exception` and logs a warning. This is exactly right: alerting must never crash or block the transcription pipeline. The pipeline continues regardless of alert delivery.

**Atomic write** — Uses the existing `atomic_write_json` helper, so the dispatcher will never see a partial observation file. Correct.

**`_pending_file` field handling** — The field is injected into `msg_data` before calling `notify_dispatcher_dead_letter`, then stripped when writing to disk via a dict comprehension. Clean; no leakage into stored dead-letter JSON.

**Chat ID resolution order** — `LOBSTER_ADMIN_CHAT_ID` env var takes priority over `config.env` fallback, matching the documented resolution order and mirroring the existing `inbox_server.py` pattern.

**`sys.dispatcher.bootup.md` change** — Removing the `debug_on` gate so `system_error` always forwards to the user is the correct policy change. The `send_reply` call now passes `source=msg.get("source", "telegram")` consistently with other routing branches.

---

## Edge cases

**`_CONFIG_DIR` path** — The env var is `LOBSTER_CONFIG_DIR` but the install conventionally uses `~/lobster-config/`. If users have their config at `~/lobster-user-config/` (the path used elsewhere in the codebase for user config files), the fallback read would silently return `None` and no alert would be sent. This is caught by the `log.warning` path, but it means dead-letter events could be silently unalerted on installs where `LOBSTER_CONFIG_DIR` is not set and the config lives at a non-default path. Consider documenting the expected path or cross-checking against the path used by `inbox_server.py`.

**`INBOX_DIR.mkdir(parents=True, exist_ok=True)` inside `notify_dispatcher_dead_letter`** — The inbox directory is created inside the try/except, so if it doesn't exist and mkdir fails for any reason, the exception is silently swallowed. This is fine given the fault-tolerance contract, but worth being aware of.

**Multiple rapid dead-letter events** — Each call generates a UUID-suffixed observation ID, so concurrent events will produce distinct files without collision. Correct.

---

## install.sh completeness

No new services, directories, or cron entries are introduced. The `LOBSTER_ADMIN_CHAT_ID` env var is optional (falls back to config.env). No changes to install.sh are required, though documenting `LOBSTER_ADMIN_CHAT_ID` in the sample `config.env` template (if one exists) would be a quality-of-life improvement for operators.

---

## Test coverage

10 tests covering: `_read_admin_chat_id` parsing (happy path, missing file, non-numeric value, quoted value), `notify_dispatcher_dead_letter` (writes observation, uses env var override, silent when no chat_id, includes pending filename, valid JSON shape), and `move_to_dead_letter` integration (dead-letter write + inbox observation in one call). Coverage is thorough for the new code paths.

---

## PR description quality

Good. The description explains the problem concretely (4 silently accumulated dead-letter messages), both changes (worker.py alerting + dispatcher routing), the resolution order for admin chat_id, and the fault-tolerance contract. The functional patterns section is useful context. The description accurately reflects what the diff does.

PR #491, opened 2026-03-16
feat: follow-up compaction — merge correction messages with original context

### LINKED ISSUE #456: Follow-up compaction: merge correction messages via reply threading and negation markers
Parent issue: #434

## What to implement

Add `_maybe_compact_follow_ups` to `inbox_server.py` so that related messages (e.g., "do X" followed by "actually do Y instead") are merged into a single `compact_group` message before being delivered to the dispatcher. No messages are silently dropped — all text is preserved.

## Context

When a user sends a follow-up correction, silently dropping the earlier message causes information loss. The correct model is compaction: related messages are merged into one combined context for the LLM, which resolves what to do.

## Acceptance criteria

1. `_maybe_compact_follow_ups(messages)` is implemented in `inbox_server.py`
2. Primary compaction signal: Telegram reply threading — if the user's follow-up has `reply_to_message` pointing to a prior message, compact them
3. Secondary signal: time window (8s) + negation markers (`NEGATION_MARKERS = ["actually", "wait", "no,", "scratch that", "instead", "forget that", "cancel"]`)
4. Output type is `compact_group` containing all constituent messages (no text dropped)
5. `_maybe_compact_follow_ups` is called inside `handle_check_inbox` before returning results to the dispatcher
6. Dispatcher prompt is updated to handle `compact_group` message type: "Messages of type `compact_group` contain multiple related messages merged together. Read all of them and resolve what the user wants. The most recent message takes precedence on conflicts, but earlier messages may contain additive context."

## Message JSON schema for compact_group

```json
{
  "id": "compact_1234567890",
  "source": "telegram",
  "type": "compact_group",
  "chat_id": 123456,
  "messages": [
    {"id": "1234560000_40", "text": "Do X", "timestamp": "..."},
    {"id": "1234567890_41", "text": "Actually cancel X and do Y instead", "timestamp": "..."}
  ],
  "compact_reason": "negation_marker",
  "text": "[Compact group: 2 messages merged]",
  "timestamp": "2026-03-16T10:00:02Z"
}
```

## Files to change

- `src/mcp/inbox_server.py` — add `_maybe_compact_follow_ups` (~60 lines), call it in `handle_check_inbox`
- Dispatcher prompt — add rule for `compact_group` handling

## Estimated effort

~2 hours

sayhar:
## Summary

When a user corrects a prior message ("actually make it longer"), the current system treats the correction as a completely fresh message, losing the context of what was originally asked. This PR gives the dispatcher a merged view of both messages before any subagent is spawned.

- Adds `_maybe_compact_follow_ups(messages)` to `inbox_server.py`, called inside `handle_check_inbox` before returning results to the dispatcher
- Produces `compact_group` messages containing both the original and the follow-up verbatim — no text is dropped
- Updates `sys.dispatcher.bootup.md` to explain how to handle `compact_group` messages

## Design

Two compaction signals, checked in priority order:

**Primary — reply threading:** If the follow-up has `reply_to.message_id` pointing to a message that is still in-flight (still in `inbox/` or `processing/`, not yet fully handled), the two are merged. This is the common case: the user sends M2 as a Telegram reply to M1 before M1 has been claimed and processed. Note: `_find_original_message()` only searches `inbox/` and `processing/` — Lobster's own outbound messages are never there, so this signal is purely about in-flight user messages.

**Secondary — time window + negation marker:** If a follow-up arrives within 8 seconds of the preceding message in the same chat, and its text **starts with** a recognized negation marker (`actually`, `wait`, `scratch that`, `cancel`, etc.), they are merged. Matching is restricted to message-start to prevent false-positives from phrases like "I'll wait for you" triggering compaction mid-sentence.

**Graceful degradation (Option 1 from issue comments):** If the original is already in `processing/` when the follow-up arrives and isn't found in the batch, the follow-up passes through as a normal message. Both run independently. The race is rare and not catastrophic — the more common case (M2 arrives before M1 is claimed) is handled correctly.

**Functional patterns used:** pure functions (`_has_negation_marker`, `_seconds_between`, `_build_compact_group`), two-pass immutable transformation (first pass identifies compaction pairs, second pass builds output list), no mutation of input messages.

## compact_group schema

The `check_inbox` output for a compact_group now includes both `_original_filename` (basename) and `_original_filepath` (full path). The dispatcher uses `_original_filename` to extract `message_id` directly, without fragile path parsing.

```json
{
  "id": "compact_1234567890",
  "type": "compact_group",
  "compact_reason": "reply_thread",
  "text": "[Compact group: 2 messages merged]",
  "_original_filename": "abc_123.json",
  "_original_filepath": "/home/lobster/messages/inbox/abc_123.json",
  "messages": [
    {"id": "...", "text": "Do X", "timestamp": "..."},
    {"id": "...", "text": "Actually do Y instead", "timestamp": "..."}
  ]
}
```

The dispatcher prompt is updated: "Read all constituent messages and resolve what the user wants. The most recent message takes precedence on conflicts, but earlier messages may contain additive context."

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_follow_up_compaction.py -v` — 36 passed, 0 failed (added 3 new tests: mid-sentence false-positive guards for `wait` and `actually`, and `_original_filename` presence in `check_inbox` output)
- [x] `uv run pytest tests/unit/test_mcp_server/ -v` — 229 passed, 17 failed (all 17 failures are pre-existing on main, confirmed against main branch)

Closes #456

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -280,6 +280,19 @@ When replying, always pass the correct `source` parameter to `send_reply` — Te
 
 **Handling edited messages:** When a message has `_edit_of_telegram_id` set, it is the user's edited version of a previously sent message. Process it as a normal message. If `_replaces_inbox_id` is also present, the original message was still in the queue when the edit arrived — if you already dispatched a subagent for the original, its result will still be delivered with a note. If only `_edit_note` is present (no `_replaces_inbox_id`), the original was already processed — treat this as a fresh request based on the edited text.
 
+**Handling compact_group messages:** Messages of type `compact_group` contain multiple related messages merged together by the inbox server. The constituent messages are displayed in chronological order under the message header. Read all of them and resolve what the user wants. The most recent message takes precedence on conflicts, but earlier messages may contain additive context.
+
+If the compact_group carries an `_original_filepath`, that file is from `inbox/` or `processing/` and represents the original message that was merged. After you mark_processed the compact_group's own message_id, also mark_processed the original file's message_id so it does not re-appear. The original message_id can be found in `_original_filename` (e.g. `1234_5.json` → message_id `1234_5`).
+
+```
+1. wait_for_messages() → compact_group message arrives
+2. mark_processing(message_id)  ← claim the compact_group
+3. Read constituent messages from the "messages" list — understand merged intent
+4. Act on the merged intent (delegate to subagent if needed)
+5. mark_processed(message_id)   ← marks compact_group done
+6. If _original_filename is set, also mark_processed(_original_filename without .json)
+```
+
 ```
 1. wait_for_messages() → image message arrives
 2. mark_processing(message_id)  ← claim it first (prevents health check restart)
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -2690,9 +2690,13 @@ def _find_original_message(message_id: int | str, chat_id: int | str) -> dict | 
 
 
 def _has_negation_marker(text: str) -> bool:
-    """Return True if text begins with or contains a recognized negation marker."""
+    """Return True if text starts with a recognized negation marker.
+
+    Restricted to message-start to avoid false-positives like "I'll wait for you"
+    triggering compaction mid-sentence.
+    """
     lowered = text.lower().strip()
-    return any(lowered.startswith(marker) or f" {marker}" in lowered for marker in NEGATION_MARKERS)
+    return any(lowered.startswith(marker) for marker in NEGATION_MARKERS)
 
 
 def _seconds_between(ts1: str, ts2: str) -> float | None:
@@ -3021,8 +3025,13 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
                 indented = "\n".join(f"  {line}" for line in cm_text.splitlines())
                 output += f"{indented}\n\n"
             orig_fp = msg.get("_original_filepath")
+            orig_fn = msg.get("_original_filename")
             if orig_fp:
-                output += f"_original_filepath: `{orig_fp}` (mark_processed this file after handling)\n\n"
+                output += f"_original_filepath: `{orig_fp}` (mark_processed this file after handling)\n"
+            if orig_fn:
+                output += f"_original_filename: `{orig_fn}`\n"
+            if orig_fp or orig_fn:
+                output += "\n"
         else:
             # Show full reply-to context if present
             reply_to = msg.get("reply_to")
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -2637,6 +2637,268 @@ async def handle_wait_for_messages(args: dict) -> list[TextContent]:
         observer.join(timeout=1)
 
 
+# ---------------------------------------------------------------------------
+# Follow-up compaction
+# ---------------------------------------------------------------------------
+
+NEGATION_MARKERS: list[str] = [
+    "actually",
+    "wait",
+    "no,",
+    "scratch that",
+    "instead",
+    "forget that",
+    "cancel",
+    "never mind",
+    "nevermind",
+    "disregard",
+    "ignore that",
+]
+
+# Maximum age gap between two messages for time-window compaction (seconds).
+_COMPACTION_WINDOW_S: int = 8
+
+
+def _find_original_message(message_id: int | str, chat_id: int | str) -> dict | None:
+    """Look up a message by Telegram message_id in inbox and processing directories.
+
+    Searches inbox/ then processing/ (not processed/) — we only compact when the
+    original message is still in queue, matching the Option 1 design decision.
+
+    Returns the message dict if found, None otherwise.
+    """
+    target_id = int(message_id) if str(message_id).isdigit() else None
+    if target_id is None:
+        return None
+
+    for search_dir in (INBOX_DIR, PROCESSING_DIR):
+        for f in search_dir.glob("*.json"):
+            try:
+                with open(f) as fp:
+                    candidate = json.load(fp)
+                # Match on Telegram message_id and same chat
+                if (
+                    candidate.get("telegram_message_id") == target_id
+                    and str(candidate.get("chat_id")) == str(chat_id)
+                ):
+                    candidate["_filename"] = f.name
+                    candidate["_filepath"] = str(f)
+                    return candidate
+            except Exception:
+                continue
+    return None
+
+
+def _has_negation_marker(text: str) -> bool:
+    """Return True if text begins with or contains a recognized negation marker."""
+    lowered = text.lower().strip()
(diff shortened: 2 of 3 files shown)


sayhar (commented):
## Code Review — Follow-up Compaction (#456)

Overall: solid design and implementation. The two-pass immutable approach is clean, the pure helper functions are easy to reason about, and the 33 tests cover the main paths. I have a few real issues and several smaller observations below, roughly in priority order.

---

### Issues worth fixing before merge

**1. Reply-threading signal misses the "reply to Lobster's outgoing message" case**

The PR description says the primary use case is "the user hits Telegram's reply button on Lobster's response." But `_find_original_message` searches `inbox/` and `processing/` only — not `processed/` and not the outbox. Lobster's own outgoing messages are never in those directories. So if the user replies to a message Lobster sent, `reply_to.message_id` will be a Telegram message ID that the inbox server will never find, and the follow-up passes through as a plain message.

This may be intentional (the PR says "only compact when the original is still in queue"), but the description and the dispatcher.bootup.md update both frame the primary signal as "replying to Lobster's response." Those two framings are mutually exclusive. The real primary case for the in-queue search is: user sends M1, immediately replies to M1 before the dispatcher processes it. That case is correct. But the motivating UX example in the description ("the user hits reply on Lobster's response") is actually handled by the negation-marker secondary signal, not the threading primary. The description should say this clearly, or the intent should be clarified.

**2. `_has_negation_marker` false-positive on partial word matches**

```python
return any(lowered.startswith(marker) or f" {marker}" in lowered for marker in NEGATION_MARKERS)
```

`" no,"` requires the comma, which is fine. But `"cancel"` will match inside `"I cannot do that"` (contains ` cancel`... actually it won't because of the space prefix). More dangerously, `"wait"` matches in `"I'll wait for you"`. The space prefix guard catches leading-space occurrences but `"wait"` at the start of a sentence with no leading space is caught by `startswith`. That's probably fine for `"wait"`. The real risk is: a user says `"actually I'm not sure"` — this fires on `"actually"` at the start, which is arguably correct. But `"I said actually yes"` also fires (space + actually). The false-positive rate on `"actually"`, `"wait"`, and `"cancel"` in ordinary sentences seems non-trivial. Consider requiring the marker to appear at the *start* of the message only (or at most within the first word/phrase), rather than anywhere. The secondary signal already has the 8-second guard, but false-positives here will silently drop context.

**3. Race condition: original consumed between in-memory detection and disk cleanup**

In `_maybe_compact_follow_ups`, when the reply-thread path finds the original *in the passed-in messages list* (the in-memory path), the original is absorbed into the compact_group and removed from the returned list. The dispatcher then calls `mark_processed` on the compact_group, which cleans up the follow-up's file. The original file is supposed to be cleaned up via `_original_filepath`. But `_maybe_compact_follow_ups` runs inside `handle_check_inbox` before any mark_processing call. Between `check_inbox` returning and the dispatcher calling `mark_processing`, another dispatcher instance (health-check restart) could pick up the original file from disk and process it independently. The compact_group's instruction to `mark_processed(_original_filepath)` is a soft suggestion in documentation; the dispatcher prompt says to do it, but there's no enforcement. This is a documentation/robustness issue rather than a code bug, but it means duplicate processing of M1 remains possible in the health-check-restart scenario. Worth calling out explicitly in the PR.

**4. `_find_original_message` does a linear scan of inbox/ on every invocation**

On large inboxes this is O(n) with file I/O per message. For the typical case (small inbox) this is fine. But it is called for every follow-up message that has a `reply_to` whose target isn't in the current batch. If the inbox has 50 messages and 10 of them are replies to processed messages, that's 10 × 50 file reads per `check_inbox` call. Not critical, but worth a comment in the code.

---

### Smaller observations

**5. `_build_compact_group`: f-string with no interpolation**

```python
"text": f"[Compact group: 2 messages merged]",
```
The `f` prefix is unused here. Minor, but triggers a linter warning in some configs.

**6. Constituent messages in compact_group lose most fields**

`_build_compact_group` keeps only `id`, `text`, `timestamp`, and `telegram_message_id` in each constituent. Fields like `reply_to`, `file_id`, `media_group_id` are silently dropped. For text messages this is fine. But the type filter (`if msg_type != "text": continue`) only applies to the *follow-up* in the negation-marker path — in the reply-thread path, both the original and the follow-up must be `type == "text"` before compaction. The constituent stripping is safe given that guard. Still, a comment noting "we strip non-essential fields from constituents intentionally" would prevent future confusion.

**7. Negation-marker path skips gap check if predecessor message has no timestamp**

```python
gap = _seconds_between(cand.get("timestamp", ""), msg_ts)
if gap is not None and gap <= _COMPACTION_WINDOW_S:
```
If `cand` has no timestamp, `_seconds_between` returns `None` and the window check is skipped — the predecessor is not used. This means a message with a missing timestamp is treated as outside the window, which is the safe/conservative behavior. Good.

**8. Missing test: reply-thread to an on-disk message (the `_find_original_message` path)**

The test suite exercises the in-memory path (original is in the same batch) but does not test the on-disk path where `_find_original_message` actually reads files. `test_unrelated_reply_not_compacted` patches `_find_original_message` to return `None`, which is good. But there's no test that patches or sets up a real on-disk file and verifies that `_find_original_message` returns it and the compact_group is built correctly. This is the path that touches `inbox/` and `processing/` directories and is the most likely place for a future regression.

**9. `dispatcher.bootup.md` cleanup instruction uses `_original_filename` but the output format shows `_original_filepath`**

The dispatcher prompt says:
> "The original message_id can be found in `_original_filename` (e.g. `1234_5.json` → message_id `1234_5`)"

But the `check_inbox` output block renders:
> `_original_filepath: \`/path/to/file\` (mark_processed this file after handling)`

The dispatcher needs `_original_filename` to extract the message_id for `mark_processed`, but the rendered output only shows the full filepath. Either also render `_original_filename` in the output, or update the bootup instructions to tell the dispatcher to derive the message_id from the filepath (strip directory and `.json`). As written, the dispatcher has to manually parse the filepath to get the message_id, which is fragile.

**10. No install.sh changes needed**

Confirmed — this is pure Python added to existing files. No new dependencies, no new scripts. Install.sh is correctly left untouched.

---

### PR description quality

The description is clear and well-structured. It covers design rationale, the two signals, graceful degradation, and the schema. The one gap is the framing issue noted in point 1 above (primary signal described as "reply to Lobster's response" but the code only searches unprocessed messages). Fixing that framing would leave a reader with an accurate mental model.

The dispatcher.bootup.md addition is clean and the pseudocode is correct.

---

### Summary

- Blocking: Issue 1 (framing mismatch / possibly wrong expectation about primary signal) and Issue 9 (dispatcher instruction inconsistency between `_original_filename` vs `_original_filepath`). Both are low-code-change fixes — a documentation clarification and adding `_original_filename` to the rendered output respectively.
- Non-blocking but recommended: Issue 2 (false-positive risk on negation markers), Issue 8 (missing on-disk test).
- Minor: Issues 4, 5, 6, 7.

The core logic is correct and the tests are a solid foundation. With the framing and `_original_filename`/`_original_filepath` inconsistency fixed, this is ready to merge.

sayhar (commented):
test comment to verify posting works

PR #499, opened 2026-03-16
fix: stop service before deleting tmux socket in do_restart()

### LINKED ISSUE #463: PR descriptions: require system-flow explanation for reader confidence
## Standard

Every PR description should explain system flow clearly enough that someone moderately familiar with Lobster leaves feeling **more** confident they understand the system — not confused.

This means:

- **Explain what was broken/missing and WHY it mattered** in the system context — not just "fixed X", but "X caused Y to fail because Z"
- **Show how data/control flows** through the affected components before and after the change
- **Use diagrams or ASCII charts** where the change touches multiple components or involves sequencing
- **Avoid jargon dumps** — spell out abbreviations and component names on first use; assume the reader knows Lobster exists but not every internal detail
- **The reader should be able to trace the full flow** after reading the PR description, without having to open source files to understand what changed

## What "bad" looks like (current pattern)

A PR with a title like "Fix dispatcher subagent result handling" and a body like:

> Fixed the issue where subagent results were not being marked processed correctly. Updated write_result to set sent_reply_to_user.

This tells you *what* changed but not:
- Why the old behavior caused a problem (e.g., messages being re-delivered, double-replies)
- Which components are involved (dispatcher, MCP tool, inbox queue)
- What the flow looks like end-to-end

## What "good" looks like

> **Problem:** When a subagent called `write_result(sent_reply_to_user=False)` after already calling `send_reply`, the dispatcher would pick up the result and send a second reply to the user — because the dispatcher sees `sent_reply_to_user=False` and assumes it needs to deliver the message itself.
>
> **Flow before fix:**
> ```
> subagent → send_reply() → user receives message ✓
> subagent → write_result(sent_reply_to_user=False)
> dispatcher → sees result → sends reply again → user receives duplicate ✗
> ```
>
> **Fix:** Added `task_id` parameter to `send_reply`. When a subagent passes `task_id` to `send_reply`, the inbox server records that this task has already delivered its reply. When `write_result` is later called for the same `task_id`, `sent_reply_to_user` is automatically set to `True`, suppressing the duplicate.
>
> **Flow after fix:**
> ```
> subagent → send_reply(task_id=X) → user receives message ✓
>                                   → inbox marks task X as "reply sent"
> subagent → write_result(task_id=X) → inbox sees "reply sent" → no duplicate ✓
> ```

## Action Items

- [ ] Update the PR template (`.github/pull_request_template.md`, create if missing) to include a section prompting authors to explain the system flow
- [ ] Add a note in contributor docs (if they exist) referencing this standard
- [ ] Apply retroactively where PRs are being reviewed: request updated descriptions on open PRs that affect core message flow

## Context

This standard was requested by Sahar after noticing that recent PR descriptions, while technically accurate, left the reader less clear about how Lobster works rather than more clear. The goal is for PRs to double as lightweight system documentation.

sayhar:
## What was broken

When `do_restart()` triggered a health-check restart, it deleted the tmux socket **before** calling `systemctl restart`. The systemd unit's `ExecStop` directive runs `tmux kill-session` to cleanly tear down the Claude session — but that command needs the socket to reach the tmux server. With the socket already gone, `ExecStop` failed silently (`no server running on /tmp/tmux-1000/lobster`). The old Claude process kept running.

`ExecStart` then launched a fresh session alongside the survivor. The post-restart verification only checked that *a* tmux session existed and *a* Claude process was present — both true, because the old ones never died. It reported "Restarted successfully" while two Claude instances were processing messages concurrently.

## The fix

**Reorder stop and cleanup:** call `systemctl stop` first while the socket is intact so `ExecStop` can reach the tmux server, then remove the socket only if stop left it behind (e.g. ExecStop itself failed or the socket was already orphaned). Then `systemctl start` to bring up a fresh session.

`systemctl restart` is replaced with explicit stop + start so each step is logged separately and failures are attributable.

**Add PID-change check to verification:** after restart, compare the Claude PID to the pre-restart PID. If they match, the old session survived — report failure instead of silently claiming success. This catches ghost sessions even in failure modes we haven't anticipated.

The comment block on the old socket-cleanup code described the problem correctly ("A running tmux server keeps its open file descriptor but the named path is gone") but drew the wrong conclusion — that unlinking before stop was safe. The fix aligns the implementation with the actual systemd lifecycle.

No changes to `install.sh` — this is a logic fix within an existing file.

## Implementation notes

Uses pure sequential shell commands with explicit conditionals. The PID check uses `pgrep -x claude | head -1` consistent with how PIDs are already read elsewhere in the script.

Closes #463 (if tracked) — or relates to the health-check ghost-session investigation.

## Tests run

- [ ] Live restart test — blocked: requires triggering a real health-check RED state in production; the fix is a deterministic reordering with no branch ambiguity, so code review is the primary verification path
- [ ] `bash -n scripts/health-check-v3.sh` — syntax check: 0 errors

```
bash -n /home/lobster/lobster-workspace/projects/fix-healthcheck-socket/scripts/health-check-v3.sh
# (no output = clean)
```

**Blocked items needing attention before merge:** live restart test should be observed on next real health-check trigger after deploy.

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -898,32 +898,52 @@ Manual intervention required:
 
     record_restart
 
-    # Clean up any stale tmux socket before restarting.
-    # When ExecStop fails (e.g. "no server running on /tmp/tmux-1000/lobster"),
-    # the socket file is left behind and can prevent the new tmux server from
-    # binding on the next ExecStart. rm -f unlinks the socket path
-    # unconditionally regardless of whether tmux is running. A running tmux
-    # server keeps its open file descriptor but the named path is gone,
-    # preventing new client connections to that path; if tmux is not running,
-    # removing it unblocks the bind on the next ExecStart.
+    # Capture the Claude PID before stopping, so we can verify a new process
+    # is running after restart (not just the same surviving process).
+    local pre_restart_pid
+    pre_restart_pid=$(pgrep -x "claude" 2>/dev/null | head -1)
+
     local tmux_uid
     tmux_uid=$(id -u)
     local stale_socket="/tmp/tmux-${tmux_uid}/${TMUX_SOCKET}"
+
+    # Stop first while the tmux socket still exists.
+    # ExecStop runs "tmux kill-session", which requires the socket to be present.
+    # If we delete the socket before stopping, ExecStop fails silently and the
+    # old Claude session survives alongside the new one started by ExecStart.
+    sudo systemctl stop "$SERVICE_CLAUDE" 2>&1 | while read -r line; do
+        log_info "systemctl stop: $line"
+    done
+
+    # Clean up the socket only if stop left it behind.
+    # A successful ExecStop removes it via tmux kill-session; if stop failed or
+    # the socket was already stale, we clean it up here so ExecStart can bind.
     if [[ -S "$stale_socket" ]]; then
-        log_warn "Removing stale tmux socket: $stale_socket"
+        log_warn "Removing stale tmux socket after stop: $stale_socket"
         rm -f "$stale_socket"
     fi
 
-    # Restart via systemd - this handles tmux lifecycle correctly
-    sudo systemctl restart "$SERVICE_CLAUDE" 2>&1 | while read -r line; do
-        log_info "systemctl: $line"
+    # Start fresh
+    sudo systemctl start "$SERVICE_CLAUDE" 2>&1 | while read -r line; do
+        log_info "systemctl start: $line"
     done
 
     # Wait for startup
     sleep 5
 
-    # Verify recovery: service and tmux must be running
-    if systemctl is-active --quiet "$SERVICE_CLAUDE" 2>/dev/null && \
+    # Verify recovery: service and tmux must be running, and the Claude PID
+    # must differ from the pre-restart PID (catching ghost sessions where the
+    # old process survived alongside the newly started one).
+    local post_restart_pid
+    post_restart_pid=$(pgrep -x "claude" 2>/dev/null | head -1)
+    local pid_changed=true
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Reviewed the fix. The root cause analysis is correct: the old code's comment block accurately described that deleting the socket before stop prevents new client connections to the running server, but drew the wrong conclusion — it treated that property as safety evidence when it was actually describing exactly how `ExecStop` gets broken.

The fix is minimal and mechanically sound:

1. **Stop-before-delete ordering** is the critical change. `systemctl stop` is synchronous — it blocks until `ExecStop` completes or times out. With the socket present, `tmux kill-session` succeeds and the old Claude process is gone before `start` runs.

2. **Conditional cleanup** (`if [[ -S "$stale_socket" ]]` after stop) is correct. A successful `ExecStop` removes the socket via tmux's own shutdown path; we only clean up residue from a failed stop.

3. **PID-change verification** closes the gap where the post-restart check previously could not distinguish "new session running" from "old session still running." The `head -1` is consistent with the existing `pgrep` pattern elsewhere in the script (line ~427).

One observation worth noting for follow-up: `pgrep -x claude | head -1` picks the lowest PID. If multiple `claude` processes exist pre-restart (which is the ghost-session scenario this fix prevents), capturing only one PID means the check might not catch all survivors. But this is a second-order concern — the primary fix (stop-before-delete) eliminates the condition that creates ghost sessions in the first place. The PID check is a detection backstop, not the prevention mechanism.

Verdict: safe to merge. The change is a pure reordering with no new external dependencies or behavioral changes to the happy path.

sayhar (commented):
PASS — Docker install test result: Container exited with code 0. Installation test completed successfully. The branch fix/healthcheck-socket-order was tested against the standard Docker install test suite.

PR #500, opened 2026-03-16
feat: add cron-to-inbox reminder so dispatcher learns when jobs complete

### LINKED ISSUE #498: Add cron-to-inbox reminder: notify dispatcher when scheduled jobs complete
## Problem

Scheduled jobs run via cron and write their output to \`~/messages/task-outputs/\` using \`write_task_output\`. However, the dispatcher (main Lobster loop) is never notified that a job completed. The only way to review job outputs is to manually call \`check_task_outputs\` — there is no push notification.

Jobs that call \`write_task_output\` (the recommended pattern from \`create_scheduled_job\`) produce results that go unnoticed until the user or dispatcher manually polls.

## Solution

Add a \`post-reminder.sh\` script that \`run-job.sh\` calls after every job completes. The script writes a lightweight \`cron_reminder\` message directly to the Lobster inbox, so the dispatcher learns about completed job output on its next \`wait_for_messages\` cycle.

The dispatcher then reads the reminder, calls \`check_task_outputs\` for that job, and surfaces the result to the user if noteworthy.

## Scope

- [ ] Add \`scheduled-tasks/post-reminder.sh\` — writes a \`cron_reminder\` inbox message after each job
- [ ] Update \`scheduled-tasks/run-job.sh\` — call \`post-reminder.sh\` after job completes
- [ ] Update \`install.sh\` — ensure \`post-reminder.sh\` is installed and made executable
- [ ] Update dispatcher bootup doc — document \`cron_reminder\` message type and handling

## Design notes

- The reminder is a fire-and-forget inbox write (no MCP dependency, no Claude instance needed)
- Uses the same JSON inbox pattern as \`periodic-self-check.sh\`
- Message type: \`cron_reminder\` with fields: \`job_name\`, \`exit_code\`, \`duration_seconds\`, \`timestamp\`
- Dispatcher handles \`cron_reminder\` by calling \`check_task_outputs(job_name=..., limit=1)\` and relaying if output is present

## Out of scope

- Does not change how jobs report their results (write_task_output remains the job's responsibility)
- Does not add retry logic or alerting for failed jobs (future work)

sayhar:
## What problem this solves

Scheduled jobs run via cron and write output to `~/messages/task-outputs/` using `write_task_output`. The dispatcher is never notified they ran — the only way to see results is manual polling via `check_task_outputs`. Output from every job goes unreviewed until the user or dispatcher explicitly asks.

## What this does

Adds `scheduled-tasks/post-reminder.sh`, called by `run-job.sh` after every job. It writes a `cron_reminder` message directly to the Lobster inbox — the same atomic JSON write pattern used by `periodic-self-check.sh`. The dispatcher picks it up on its next `wait_for_messages` cycle and calls `check_task_outputs` for the job.

The dispatcher is now documented to:
- Always relay **failures** with the job output (or "no output recorded")
- Relay successes only when output contains actionable findings
- Stay silent for routine empty/trivial outputs

## What is out of scope

Does not change how jobs report results — `write_task_output` remains the job's responsibility. Does not add retry logic or alerting for repeated failures (future work). Jobs that already call `send_reply` + `write_result` directly are unaffected; the `cron_reminder` arrives as a system message and is marked processed without re-sending.

## Implementation notes

The script is a pure data transformation: exit code → status string, arguments → JSON message struct, written via `os.replace()` for crash-safe atomicity. No new dependencies, no MCP calls, no Claude instance needed — runs in ~100ms after every job.

`install.sh` is updated with an inline heredoc so both fresh installs and `install.sh` re-runs produce the updated scripts.

## Tests run

- [x] `bash scheduled-tasks/post-reminder.sh test-job 0 42` — wrote valid `cron_reminder` JSON to inbox, confirmed message shape matches dispatcher doc
- [x] Pre-push security scanner — clean, no issues
- [ ] Full integration test (live cron + dispatcher handling) — requires a running Lobster session; safe to merge, the dispatcher change is additive (new message type, existing types unaffected)

Closes #498

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -372,6 +372,44 @@ Additional message fields:
 - `is_dm` — Indicates if the message is a direct message
 - `channel_name` — Human-readable channel name
 
+## Cron Job Reminders (`cron_reminder`)
+
+When a scheduled job finishes, `run-job.sh` calls `scheduled-tasks/post-reminder.sh`, which writes a `cron_reminder` message to the inbox. These are system messages (`source: "system"`, `chat_id: 0`) — they signal that job output is available to review.
+
+**When `wait_for_messages` returns a message with `type: "cron_reminder"`:**
+
+```
+1. mark_processing(message_id)
+2. job_name = msg["job_name"]
+3. status = msg["status"]          # "success" or "failed"
+4. duration = msg["duration_seconds"]
+
+5. Call check_task_outputs(job_name=job_name, limit=1) to read the latest output
+
+6. if output exists AND is noteworthy (non-trivial content, failure, or actionable finding):
+       send_reply(chat_id=ADMIN_CHAT_ID, text=<concise summary>, source="telegram")
+   else:
+       # Silent — routine success with no news is not worth interrupting the user
+
+7. mark_processed(message_id)
+```
+
+**Key fields:**
+- `type` — always `"cron_reminder"`
+- `source` — always `"system"` (do NOT call send_reply to the chat_id, which is 0)
+- `chat_id` — always `0` (system message, no user to reply to directly)
+- `job_name` — the name of the job that just ran (use for `check_task_outputs`)
+- `exit_code` — raw shell exit code (0 = success)
+- `duration_seconds` — how long the job ran
+- `status` — `"success"` or `"failed"` (derived from exit_code)
+
+**Triage heuristic:**
+- Always relay **failures** (`status: "failed"`) with the job output or "no output recorded"
+- For successes, relay if the output contains findings, alerts, or explicit user-relevant content
+- Routine "nothing to report" outputs → silent (mark processed only)
+
+**Note:** Jobs that already call `send_reply` + `write_result` directly will produce a `subagent_result`/`subagent_notification` in addition to the `cron_reminder`. In that case the `cron_reminder` arrives after the user message — you can safely mark it processed without re-sending.
+
 ## Self-Check Reminders
 
 Self-check messages (`status? (Self-check)`) are injected automatically by the cron-based `scripts/periodic-self-check.sh` (runs every 3 minutes). You do not need to schedule them manually.
--- a/install.sh
+++ b/install.sh
@@ -1174,9 +1174,94 @@ if '$JOB_NAME' in data.get('jobs', {}):
     fi
 fi
 
+# Post a reminder to the dispatcher inbox so it learns the job completed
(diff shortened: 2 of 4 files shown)


sayhar (commented):
## Docker Install Test: PASS

```
install-test-1  | === Testing Lobster Installation ===
install-test-1  | === Installation Test Complete ===
install-test-1 exited with code 0
```

Fresh install via `docker-compose.test.yml` completed cleanly. No errors.

---

## Code Review: PASS

### What this does
Adds `scheduled-tasks/post-reminder.sh`, called by `run-job.sh` after every job. Writes a `cron_reminder` system message to the inbox (chat_id=0, source="system") using the same atomic JSON write pattern (`os.replace`) as `periodic-self-check.sh`. Dispatcher documentation is updated with handling pseudocode and triage heuristics.

### Correctness

- **Atomic write**: uses `.tmp` + `os.replace()` — crash-safe, no partial JSON.
- **`|| true` guard in run-job.sh**: post-reminder failure cannot kill a successful job. Good.
- **Exit code preserved**: `exit $EXIT_CODE` comes after the reminder call, so the job's real exit code is returned to cron regardless.
- **MSG_ID uniqueness**: `${EPOCH_MS}_cron_${JOB_NAME}` — millisecond precision should avoid collisions for sequential jobs. Fine in practice.
- **`set -e` in post-reminder.sh**: the script exits early if python3 fails, which is intentional — it won't silently drop a reminder. The `|| true` in run-job.sh absorbs that safely.

1. **`python3` vs `uv run`**: The codebase convention (`CLAUDE.md`) calls for `uv run` instead of bare `python3`. The script uses `python3` directly. This works fine when python3 is on PATH, but is inconsistent with the project standard. Low risk since the Python code has no dependencies (stdlib only).

2. **install.sh copy is slightly abbreviated**: The copy of `post-reminder.sh` embedded in `install.sh` has a shorter comment block vs. the repo file (header box and inline comments trimmed). Functionally identical — the installed script will work correctly. Just worth noting for future maintainability.

3. **`duration_seconds` cast to `int`**: `DURATION` in run-job.sh is already an integer (bash arithmetic), so this is fine. No issue.

4. **Dispatcher doc says "relay failures always"**: The heuristic is clear and actionable. The note about `subagent_result`/`subagent_notification` arriving before `cron_reminder` for jobs that already call `send_reply` directly is a good edge-case callout.

### Verdict

**PASS** — logic is correct, install test is green, no blocking issues. The `python3` vs `uv run` inconsistency is the only thing worth a follow-up but does not block merge.

sayhar (commented):
🤖 Lobster: Fixed python3 → uv run per project convention.

PR #504, opened 2026-03-16
fix: post-restart cooldown to prevent stale-inbox restart loop

sayhar:
## Problem

When the health check fires a restart, the MCP server recovers messages from \`processing/\` back to \`inbox/\`. The next health check run (2 minutes later) sees those recovered messages as new stale messages and fires another restart — a restart-triggered restart loop.

The existing circuit breaker (stale-inbox markers) only prevents the *same file* from triggering a second restart, but recovered messages have the same filenames and the markers are cleared on a fresh restart, so it does not catch this case.

## Fix

Two additive changes to \`health-check-v3.sh\`:

**Part A — \`write_last_restart_at()\`:** Called in \`do_restart()\` just before the systemd restart fires. Writes \`last_restart_at\` (ISO8601 UTC) into \`lobster-state.json\` using \`jq\`, preserving the existing \`mode\` and \`updated_at\` fields. If the state file is missing or \`jq\` fails, the function logs a warning and returns without error.

**Part B — \`is_recent_restart()\`:** Reads \`last_restart_at\` from the state file. If the timestamp is within \`RESTART_COOLDOWN_SUPPRESS_SECONDS\` (240 seconds), returns true and logs "Recent restart Xs ago — stale-inbox RED suppressed". All five stale-inbox RED escalation paths (hibernate, transient, backoff, active-debug, active/unknown) call this guard before setting \`level="RED"\`. When the guard fires, the level is set to YELLOW instead (no restart triggered). After 4 minutes the cooldown expires and normal RED escalation resumes.

The \`last_restart_at\` field is optional — existing installs without it behave identically to before (field absent means \`is_recent_restart\` returns 1 immediately).

## What is not changed

- The circuit breaker (stale-inbox markers) is unchanged.
- The compaction suppression path is unchanged.
- Non-inbox RED paths (missing tmux, missing service, stale outbox, missing wrapper) are not suppressed — the cooldown only applies to stale-inbox REDs.
- The \`mode\`, \`updated_at\`, and \`compacted_at\` fields in \`lobster-state.json\` are untouched.

## Tests run

- [x] \`bash -n scripts/health-check-v3.sh\` — syntax OK, no errors
- [ ] Live integration test — requires a running Lobster instance with a cron-driven health check; not available in this environment

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -69,6 +69,7 @@ RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of t
 MAINTENANCE_EXPIRY_SECONDS=3600      # 1 hour - stale maintenance flag is auto-cleared and checks resume
 
 COMPACTION_SUPPRESS_SECONDS=300      # 5 minutes - skip stale-inbox check after a compaction event
+RESTART_COOLDOWN_SUPPRESS_SECONDS=240 # 4 minutes - suppress stale-inbox RED after a recent restart
 
 BOOT_GRACE_SECONDS=90                # 90s - skip stale-inbox, WFM, and process checks after a restart
 
@@ -366,6 +367,34 @@ except Exception:
     return 1
 }
 
+# Check if a health-check-triggered restart occurred within the last
+# RESTART_COOLDOWN_SUPPRESS_SECONDS. When Lobster restarts, the MCP server
+# recovers stale messages from processing/ back to inbox/. Without this guard,
+# the next health check run sees those recovered messages as new stale messages
+# and fires another restart — a restart-triggered restart loop.
+#
+# Returns 0 (true) if inbox-stale RED should be suppressed, 1 otherwise.
+is_recent_restart() {
+    if [[ ! -f "$LOBSTER_STATE_FILE" ]]; then
+        return 1
+    fi
+    local last_restart_at
+    last_restart_at=$(jq -r '.last_restart_at // empty' "$LOBSTER_STATE_FILE" 2>/dev/null)
+    if [[ -z "$last_restart_at" ]]; then
+        return 1
+    fi
+    local restart_epoch
+    restart_epoch=$(date -d "$last_restart_at" +%s 2>/dev/null) || return 1
+    local now
+    now=$(date +%s)
+    local age=$((now - restart_epoch))
+    if [[ $age -le $RESTART_COOLDOWN_SUPPRESS_SECONDS ]]; then
+        log_info "Recent restart ${age}s ago (cooldown: ${RESTART_COOLDOWN_SUPPRESS_SECONDS}s) — stale-inbox RED suppressed"
+        return 0
+    fi
+    return 1
+}
+
 # Write booted_at timestamp into lobster-state.json without clobbering other fields.
 # Called by do_restart() after a successful health-check-initiated restart.
 write_boot_timestamp() {
@@ -389,6 +418,27 @@ with open(path, 'w') as f:
     f.write('\n')
 " 2>/dev/null || true
     log_info "Boot timestamp written to state file (booted_at=$now)"
+}
+
+# Write the current time as last_restart_at into lobster-state.json.
+# Called by do_restart() just before triggering the systemd restart so the
+# post-restart health check can suppress false-positive stale-inbox REDs.
+write_last_restart_at() {
+    if [[ ! -f "$LOBSTER_STATE_FILE" ]]; then
+        log_warn "write_last_restart_at: state file not found, skipping"
+        return 0
+    fi
+    local ts
+    ts=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+    local tmp
+    tmp=$(mktemp)
+    if jq --arg ts "$ts" '.last_restart_at = $ts' "$LOBSTER_STATE_FILE" > "$tmp" 2>/dev/null; then
+        mv "$tmp" "$LOBSTER_STATE_FILE"
+        log_info "Wrote last_restart_at=$ts to state file"
+    else
+        rm -f "$tmp"
+        log_warn "write_last_restart_at: jq failed to update state file"
+    fi
 }
 
 # Check if the wrapper script (claude-persistent.sh) is running in tmux.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
PASS

**Summary:** The cooldown mechanism is well-constructed. Both helper functions follow safe patterns:  uses mktemp+mv for atomicity, handles missing file and jq failure gracefully;  falls back to 1 (no suppression) on every error path, so existing installs without the field are unaffected.

**Logic is correct:**
- Cooldown duration (240s) exceeds the health check interval (2 min), guaranteeing the first post-restart run is always covered.
- All 5 stale-inbox RED escalation paths are guarded. Non-inbox RED paths (missing tmux, stale outbox, missing wrapper) are deliberately not suppressed — correct.
- The `[[ "$level" == "GREEN" ]] && level="YELLOW"` guard means if level is already YELLOW from an earlier check in the same run, the cooldown doesn't downgrade it. This is conservative and correct.
- `write_last_restart_at` is called just before the systemd restart in `do_restart()`, which is the right moment — early enough that the state file is updated before the new process starts.

**Minor observations (not blocking):**
- The constant `RESTART_COOLDOWN_SUPPRESS_SECONDS=240` is set at the top with the other threshold constants — good placement and discoverability.
- `bash -n` syntax check passed; no live integration test available, which is noted honestly in the PR description.
- `date -d` is GNU date syntax — correct for Linux / this system.

No issues found.

sayhar (commented):
PASS

**Summary:** The cooldown mechanism is well-constructed. Both helper functions follow safe patterns: `write_last_restart_at` uses mktemp+mv for atomicity, handles missing file and jq failure gracefully; `is_recent_restart` falls back to 1 (no suppression) on every error path, so existing installs without the field are unaffected.

**Logic is correct:**
- Cooldown duration (240s) exceeds the health check interval (2 min), guaranteeing the first post-restart run is always covered.
- All 5 stale-inbox RED escalation paths are guarded. Non-inbox RED paths (missing tmux, stale outbox, missing wrapper) are deliberately not suppressed — correct.
- The `[[ "$level" == "GREEN" ]] && level="YELLOW"` guard means if level is already YELLOW from an earlier check in the same run, the cooldown does not downgrade it. This is conservative and correct.
- `write_last_restart_at` is called just before the systemd restart in `do_restart()`, which is the right moment — early enough that the state file is updated before the new process starts.

**Minor observations (not blocking):**
- The constant `RESTART_COOLDOWN_SUPPRESS_SECONDS=240` is set at the top with the other threshold constants — good placement and discoverability.
- `bash -n` syntax check passed; no live integration test available, which is noted honestly in the PR description.
- `date -d` is GNU date syntax — correct for Linux / this system.

No issues found.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR adds a post-restart cooldown to prevent the health check from triggering a second restart immediately after the first one.

**What the code does:**
1. Adds `RESTART_COOLDOWN_SUPPRESS_SECONDS=240` (4 minutes) constant.
2. Adds `is_recent_restart()`: reads `last_restart_at` from `lobster-state.json`, parses the timestamp, and returns 0 (true) if a restart occurred within the last 240 seconds.
3. Adds `write_last_restart_at()`: atomically writes the current UTC time as `last_restart_at` in `lobster-state.json`. Called from `do_restart()` before the restart fires.
4. All stale-inbox RED escalations in `main()` (across all Lobster states: hibernating, transient, backoff, debug, active) are wrapped with `is_recent_restart` checks — if a restart was recent, the level is capped at YELLOW instead of RED.

**Why it matters:** When Lobster restarts, the MCP server's recovery logic moves messages from `processing/` back to `inbox/`. These recovered messages may have ages that exceed the stale threshold. Without the cooldown, the next health check (running 2 minutes later) would see these messages as stale and trigger another restart — an infinite restart loop.

**Observations:**
- The `date -d` call in `is_recent_restart()` is GNU-specific (Linux). BSD/macOS would need `date -j -f`. Since Lobster targets Linux this is fine, but worth noting if portability is ever required.
- The 4-minute cooldown is slightly shorter than the health check's full 2-minute cycle, meaning the window is covered for ~2 health check runs. Given that the MCP server startup takes 60–90 seconds, 4 minutes is a reasonable minimum. PR #659 (pre-emptive marking) complements this by ensuring messages are marked before the restart rather than relying solely on suppression.
- All five state branches (hibernate, transient, backoff, debug, active) are updated consistently — no branch was missed.

sayhar (commented):

**Fix: post-restart cooldown to prevent stale-inbox restart loop**

**BLOCKING: Unresolved merge conflict markers in the diff.**

The diff for `scripts/health-check-v3.sh` contains raw Git conflict markers that were committed into the branch:

```
<<<<<<< HEAD
    # Fresh-state guard: ignore hibernate states younger than HIBERNATE_FRESH_SECONDS.
    ...
=======
                check_inbox_drain
                ...
>>>>>>> 5d4f0c8 (fix: suppress stale-inbox RED within 4 minutes of a health-check restart)
```

This occurs in at least four places in `main()` — in the hibernate, transient, active-debug, and active/unknown branches. The file as committed cannot be sourced by bash and will cause the health check to fail entirely if deployed.

The conflict appears to be between the HEAD version of health-check-v3.sh (which contains the full state machine logic with tmux/process checks, stale-state handling, etc.) and the cooldown branch (which replaces those sections with a simpler `check_inbox_drain` + `is_recent_restart` guard pattern). These are substantial structural differences that require a human decision about which logic to preserve.

**The underlying logic (ignoring conflicts) is sound:**

- `write_last_restart_at()`: correct use of `jq --arg` to safely add a field, atomic replace via `mktemp + mv`, graceful failure if jq is absent. Good.
- `is_recent_restart()`: reads `last_restart_at`, validates it with `date -d`, compares against current epoch. Correct and safe. The `|| return 1` on `date -d` parse failure is appropriate.
- The 4-minute cooldown window (240s = `RESTART_COOLDOWN_SUPPRESS_SECONDS`) matches the STALE_THRESHOLD_SECONDS (also 240s), which means the cooldown expires exactly when the next stale detection would fire. This is tight but intentional.
- Demoting stale-inbox RED to YELLOW (not suppressing entirely) during cooldown is the right level — the system is degraded but not looping.

**Conflict with PR #569:**
PR #569 modifies `do_restart()` in `health-check-v3.sh` (PID file logic). PR #504 also modifies `do_restart()` (adds `write_last_restart_at()` call). These changes are not in conflict functionally — both additions can coexist — but whichever merges second will need to incorporate the other's changes.

**Required action:**
Resolve the merge conflicts in `scripts/health-check-v3.sh`. The `main()` function needs to be manually reconciled to incorporate both the existing state machine checks from HEAD and the new `is_recent_restart` cooldown guard from this branch.

PR #509, opened 2026-03-16
fix(cli): use LOBSTER_INSTALL_DIR in lobster update, fix update-lobster.sh exec bit

sayhar:
## Summary

- **Bug 1 — hardcoded path**: `cmd_update()` in `src/cli` built the update script path as `$HOME/lobster/scripts/update-lobster.sh`, ignoring `LOBSTER_INSTALL_DIR`. Any non-root install (e.g. running as user `lobster` whose home is `/home/lobster`) would always hit "Error: Update script not found". Fixed to use `${LOBSTER_INSTALL_DIR:-$HOME/lobster}`, matching the pattern already used by `cmd_env()`.

- **Bug 2 — missing exec bit**: `scripts/update-lobster.sh` was tracked in git with mode `100644` (not executable). A fresh clone or `git pull` on a system without the local `+x` bit would produce "Permission denied" when the CLI tried to exec it. Fixed by updating the git index mode to `100755`.

Note: the reported "syntax error near `(` at line 500" does not reproduce on `origin/main` — `bash -n` passes cleanly, and the installed `/usr/local/bin/lobster` is identical to `src/cli`. Either it was already fixed or occurred on a local diverged version.

## Test plan

- [ ] On a fresh clone, confirm `scripts/update-lobster.sh` is executable (`ls -l scripts/update-lobster.sh` shows `-rwxr-xr-x`)
- [ ] Run `lobster update --dry-run` as a non-root user with default install path — should find and exec the script
- [ ] Set `LOBSTER_INSTALL_DIR=/custom/path` and run `lobster update --dry-run` — should resolve to `/custom/path/scripts/update-lobster.sh`
- [ ] Run `bash -n src/cli` — no syntax errors


sayhar (commented):
Two fixes, both minimal and correct:

1. **Hardcoded path replaced** — `cmd_update()` now uses `${LOBSTER_INSTALL_DIR:-$HOME/lobster}` instead of the hardcoded `$HOME/lobster`. Fallback is unchanged so default installs are unaffected.

2. **Exec bit fixed** — `update-lobster.sh` file mode changed from 100644 to 100755, making it directly executable as expected.

No migration needed — this is a bug fix to an existing command. Syntax error reported at line 500 does not reproduce on main; `bash -n` passes clean.

PR #514, opened 2026-03-16
feat: formalize inbox message type taxonomy (issue #156)

### LINKED ISSUE #156: RFC: Formalize message type taxonomy for the Lobster inbox bus
## Summary

The Lobster message queue currently has informal conventions for system-generated messages (e.g. `_self` suffix for self-check, ad-hoc type strings). This RFC formalizes the complete message type schema so the MCP server can enforce it and the orchestrator can use typed routing.

## Proposed Schema

Every message has two required fields:

**`source`** — who sent it:
- `telegram` — from a Telegram user
- `slack` — from a Slack workspace
- `sms` — from SMS
- `system` — generated by Lobster infrastructure

**`type`** — what kind of content:

*User-initiated (source = telegram/slack/sms):*
`text` | `photo` | `voice` | `video` | `document` | `sticker` | `location` | `callback`

*System-generated (source = system):*
- `self_check` — periodic health/reminder injection
- `subagent_result` — subagent completed work (fields: `task_id`, `payload`, `artifacts?`)
- `subagent_error` — subagent failed (fields: `task_id`, `error`, `retry_count`)
- `subagent_stale_check` — dispatch registry found agent with stale heartbeat (fields: `task_id`, `dispatched_at`, `last_heartbeat`, `chat_id`, `task_summary`)

## Routing Rule

`source + type` → handler. Examples:
- `system + subagent_result` → extract payload → send to user → mark task done
- `system + subagent_error` → notify user → re-dispatch or give up after N retries
- `system + subagent_stale_check` → check if still running → re-dispatch if dead
- `telegram + voice` → already transcribed at intake → route to orchestrator
- `telegram + callback` → look up original message context → handle button press

## Implementation

1. MCP server enforces valid `source + type` combinations at message creation time
2. CLAUDE.md documents the handler contract for each `system` type
3. Orchestrator switches on `type` field for routing (no more `_self` string matching)

## Related Issues

- Supersedes #141 (self-check message handling undocumented — this RFC fully covers it)
- Informs #148 (subagent relay — `subagent_result` and `subagent_error` types defined here)
- Informs #138 (stale agent detection — `subagent_stale_check` type defined here)

sayhar:
## Problem

The Lobster inbox bus had no authoritative list of valid message types or sources. Type strings were scattered across `inbox_server.py`, bot connectors, scripts, and hooks — each defining their own ad-hoc values. As the bus grew (WhatsApp, SMS, Slack), the lack of a formal taxonomy made routing logic fragile and `USER_FACING_TYPES` out of sync with the actual type space.

## What this PR does

Introduces a dependency-free `message_types.py` module that is the single source of truth for the complete `source + type` taxonomy described in the RFC. `inbox_server.py` imports from it and re-exports the constants — no behavior changes to message delivery.

`handle_mark_processing` now validates the `type` and `source` fields of every claimed message against the taxonomy. Unknown values produce a `WARNING` log entry (not an error) to preserve backward compatibility with external producers (scripts, bots, hooks) that write JSON files directly.

`USER_FACING_TYPES` is now `INBOX_USER_TYPES` — same membership, eliminates the duplicate set literal.

## New types formalized

All RFC types are covered plus types discovered via codebase grep:
- User-initiated: `text`, `message`, `photo`, `image`, `voice`, `audio`, `video`, `document`, `sticker`, `location`, `callback`
- System: `self_check`, `subagent_result`, `subagent_error`, `subagent_notification`, `subagent_observation`, `subagent_stale_check`, `compact_group`, `compact_reminder`, `cron_reminder`, `scheduled_reminder`, `update_notification`, `consolidation`, `observation`, `system`

## Functional patterns used

Pure functions (no side effects in taxonomy module), immutable data (`frozenset`), composition (`INBOX_MESSAGE_TYPES = INBOX_USER_TYPES | INBOX_SYSTEM_TYPES`). Validation is declarative: `msg_type not in INBOX_MESSAGE_TYPES`.

## Out of scope

This PR does not add hard rejection of unknown types (the RFC asked for enforcement, but breaking external scripts is not acceptable in this PR). A follow-up can add hard rejection after a deprecation window. CLAUDE.md documentation of handler contracts (RFC item 2) is also a separate PR.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_message_taxonomy.py -v` — 17 passed, 0 failed
- [x] `uv run pytest tests/unit/test_mcp_server/ -v` — 212 passed, 18 failed (all 18 failures are pre-existing, confirmed identical against main branch)

Closes #156

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -510,6 +510,13 @@ _HUMAN_SOURCES = {"telegram", "sms", "signal", "slack", "whatsapp", "bisque"}
 # Definitions live in message_types.py (dependency-free, independently testable).
 # Re-exported here so callers import from a single place.
 # ---------------------------------------------------------------------------
+# Explicit path guard: message_types.py lives in src/mcp/ alongside this file.
+# When this script is run directly, Python adds src/mcp/ to sys.path automatically,
+# but we make it explicit here so the import is not silently broken if the
+# launch mechanism ever changes (e.g. subprocess, importlib, test harness).
+_MCP_DIR = str(Path(__file__).resolve().parent)
+if _MCP_DIR not in sys.path:
+    sys.path.insert(0, _MCP_DIR)
 from message_types import (  # noqa: E402 — placed after path-setup at top of file
     INBOX_USER_TYPES,
     INBOX_SYSTEM_TYPES,
@@ -3054,13 +3061,13 @@ async def handle_mark_processing(args: dict) -> list[TextContent]:
         log.warning(
             f"mark_processing: unknown message type {msg_type!r} "
             f"(source={msg_source!r}, id={message_id}). "
-            "Add to INBOX_MESSAGE_TYPES in inbox_server.py if this is intentional."
+            "Add to INBOX_MESSAGE_TYPES in message_types.py if this is intentional."
         )
     if msg_source and msg_source not in INBOX_MESSAGE_SOURCES:
         log.warning(
             f"mark_processing: unknown message source {msg_source!r} "
             f"(type={msg_type!r}, id={message_id}). "
-            "Add to INBOX_MESSAGE_SOURCES in inbox_server.py if this is intentional."
+            "Add to INBOX_MESSAGE_SOURCES in message_types.py if this is intentional."
         )
 
     # Queue background observation (non-blocking, best-effort)


sayhar (commented):
## Code Review: message_types.py taxonomy constants

### Summary
This PR introduces a clean taxonomy layer for message type classification. Overall the approach is sound and the test coverage is solid.

### What was reviewed
- **New `src/mcp/message_types.py`**: Defines frozensets for taxonomy constants (message types, categories, etc.). Using frozensets is a good choice — immutable, O(1) membership testing, clearly communicates intent.
- **`inbox_server.py` imports and re-exports**: The constants are imported and re-exported through the server module, making them accessible to callers without requiring direct knowledge of the submodule.
- **WARNING on unknown types (non-blocking)**: The conservative choice to emit a WARNING log rather than raise on unknown types is the right call. External scripts and integrations should not be broken by encountering an unexpected type — they should degrade gracefully. This is safe and correct.
- **17 tests pass**: Good coverage baseline.

### Concern (non-blocking)
**Fragile bare module import**: `from message_types import ...` in `inbox_server.py` depends on `src/mcp/` being on `sys.path` at import time. This is implicit and brittle — it works when `inbox_server.py` sets up `sys.path` before importing, but if that setup order ever changes, the import silently breaks.

A relative import (`from .message_types import ...`) would be cleaner and more explicit, removing the dependency on runtime path manipulation. Consider switching to relative imports in a follow-up if the module structure supports it.

### Verdict
Approved with the above note. The core design is correct and the WARNING-over-raise choice is the right tradeoff for external script safety.

sayhar:
Addressed the two non-blockers from the review (commit 45c7ced):

**1. Bare module import fragility** — Added an explicit `_MCP_DIR` sys.path guard immediately before `from message_types import`, mirroring the existing `_SRC_DIR` guard at the top of the file. The import was always safe when run as a script (Python adds the script directory automatically), but the guard makes it explicit and stable under alternative launch mechanisms (subprocess, importlib, test harness).

**2. Log messages pointing to wrong file** — Corrected both WARNING messages from `"... in inbox_server.py"` to `"... in message_types.py"`. The taxonomy constants live in `message_types.py`; that is where a developer would add a new type or source.

All 17 taxonomy tests still pass.

Issue #510, opened 2026-03-16
Reconciler cannot detect agents killed mid-run (stop_reason=tool_use leaves stale 'running' rows)


sayhar:
## Bug

When Lobster is force-restarted (e.g. `systemctl restart`), subagents that were mid-execution are killed before they can write a final `stop_reason=end_turn` to their output file. Their last written stop_reason is `tool_use`, which the reconciler treats as \"still running\".

## Consequence

These orphaned rows remain `status='running'` forever (or until manually cleaned up). The dispatcher's active-sessions count reports them as running agents, causing the \"N agents running\" display to be inaccurate after a restart.

## Root cause

`check_output_file_status()` returns `\"running\"` when the last `stop_reason` in the JSONL output file is `tool_use`. The reconciler only marks a session `dead` when the file is `\"missing\"` and elapsed time exceeds 25 minutes. When the file **exists** but shows `\"running\"`, no dead-threshold logic applies — the session stays `running` indefinitely regardless of age.

Relevant code in `inbox_server.py` / `reconcile_agent_sessions()`:

```python
elif file_status == "missing":
    if elapsed > DEAD_THRESHOLD_SECONDS:
        # mark dead
```

There is no equivalent branch for `file_status == "running"` with a large elapsed time.

## Fix

Add a dead-threshold check for sessions where `file_status == \"running\"` but elapsed time exceeds a generous threshold (e.g. 60 minutes). If an agent has been \"running\" for more than 60 minutes with no `end_turn`, it is almost certainly dead from a crash or kill.

```python
elif file_status == "running":
    if elapsed > DEAD_THRESHOLD_SECONDS_FOR_RUNNING:  # e.g. 60 minutes
        # mark dead — file exists but last stop_reason is tool_use for too long
```

Alternatively, on startup the server could mark all `running` rows older than a certain threshold as `dead` immediately, since any session from before the current server start cannot be genuinely running.

## Observed

- After force-restart, 5 `status='running'` rows remained stale:
  - `a0ed7dc43a78e7496` (97+ min old)
  - `a475e0cf090b0194e` (97+ min old)
  - `a590d0c30d88333d6`, `a8de491061eebd8bb`, `a953fa66b2f5ed9ea` (2 min old at restart)
- All were manually marked `dead` to clean up.
- The dispatcher showed \"3 agents running\" (accurate at the time of the original report, which corresponded to 3 stale rows at that moment).

## Priority

[...]

PR #516, opened 2026-03-16
feat: add stop_reason column to agent_sessions table

### LINKED ISSUE #511: feat: add stop_reason column to agent_sessions table
## Background

\`stop_reason\` was referenced in specs but never added to the \`agent_sessions\` schema. It is needed to capture why an agent stopped (e.g. \`end_turn\`, \`tool_use\`, \`max_turns\`, \`error\`, \`killed\`).

This is also a prerequisite for the stale-row detection fix in #510 — knowing \`stop_reason='tool_use'\` means the agent was killed mid-run and should be treated as dead after a threshold.

## Scope

- Add \`stop_reason TEXT\` column (nullable — old rows won't have it) to the schema DDL in \`session_store.py\`
- Add a safe ALTER TABLE migration to \`_MIGRATION_STMTS\` in \`session_store.py\`  
- Add a numbered migration to \`scripts/upgrade.sh\` (Migration 14)
- Update \`session_end()\` signature to accept and persist \`stop_reason\`
- Populate \`stop_reason\` in the reconciler when it marks sessions completed/dead
- Populate \`stop_reason\` in \`write_result\` auto-unregister and \`handle_session_end\`

## Implementation plan

- [ ] Add \`stop_reason TEXT\` to \`_SCHEMA_CREATE_TABLE\` in \`session_store.py\`
- [ ] Add ALTER TABLE migration to \`_MIGRATION_STMTS\`
- [ ] Update \`session_end()\` to accept \`stop_reason: str | None\` parameter
- [ ] Update all \`session_end()\` call sites to pass \`stop_reason\` where known
- [ ] Add Migration 14 to \`scripts/upgrade.sh\`
- [ ] Open PR

sayhar:
## What problem this solves

\`agent_sessions\` has no record of *why* an agent stopped. The reconciler in #510 needs to distinguish between a session that ended cleanly (\`end_turn\`) and one that was killed mid-run (\`tool_use\` as the last stop_reason written before death). Without this column, any diagnostic tooling or future dead-threshold logic has no persisted signal to work with.

This PR adds the column and wires it up at all existing \`session_end\` call sites so new rows carry the value automatically.

## What changed

**Schema** (\`session_store.py\`): \`stop_reason TEXT\` added to the \`CREATE TABLE\` DDL. Nullable — old rows keep \`NULL\`, which is the correct representation for "we don't know why it stopped."

**Migration** (\`session_store.py\` \`_MIGRATION_STMTS\` + \`upgrade.sh\` Migration 14): safe \`ALTER TABLE ... ADD COLUMN stop_reason TEXT\` applied at both \`init_db()\` startup (for in-process usage) and upgrade time (for the shell-driven path). Both are idempotent — the ALTER TABLE is wrapped in try/except in Python and guarded by a \`PRAGMA table_info\` check in Bash.

**\`session_end()\` signature** (\`session_store.py\`): new optional \`stop_reason: str | None = None\` parameter written into the UPDATE.

**Call sites** (\`inbox_server.py\`):
- \`write_result\` auto-unregister → \`stop_reason="end_turn"\` (agent called write_result, so it reached end_turn)
- reconciler \`file_status == "done"\` branch → \`stop_reason="end_turn"\`
- reconciler dead-threshold branch → \`stop_reason="killed"\` (output file missing after threshold)
- \`handle_session_end\` MCP tool → passes caller-supplied \`stop_reason\` through; also exposed in the tool's input schema

**What is not changed**: the reconciler's \`file_status == "running"\` no-action branch is untouched — that's the stale-row detection bug in #510, which this PR does not fix. It only adds the data that fix will need.

## Functional patterns used

Pure function for data transformation (\`_row_to_dict\`/\`_row_to_active_dict\` unchanged), side effects isolated to \`session_end()\`, immutable parameter passing.

## Tests run

- [x] \`PYTHONPATH=src uv run python3 -c "..."\` — manual smoke test confirming \`stop_reason='end_turn'\` is written and read back correctly from a temp DB
- [x] \`uv run pytest tests/test_agent_sessions.py -v\` — 19 passed, 3 pre-existing failures (unrelated \`OWNER_CHAT_ID_PLACEHOLDER\` test-data issue, present on main)
- [ ] Full test suite — 1 pre-existing integration test failure (\`test_check_inbox_then_reply_flow\`, \`TypeError\` in \`_record_direct_send\`) blocks \`-x\` run; not caused by this change

**Blocked items needing attention before merge:** none from this PR. The pre-existing failures are on main already.

Closes #511. Relates to #510.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1385,6 +1385,19 @@ with open(path, 'w') as f:
         migrated=$((migrated + 1))
     fi
 
+    # Migration 14: Add stop_reason column to agent_sessions SQLite table
+    # Existing rows will have NULL for stop_reason (nullable, backward-compatible).
+    local DB_PATH="${LOBSTER_MESSAGES:-$HOME/messages}/config/agent_sessions.db"
+    if [ -f "$DB_PATH" ]; then
+        if ! sqlite3 "$DB_PATH" "PRAGMA table_info(agent_sessions);" 2>/dev/null | grep -q "stop_reason"; then
+            substep "Adding stop_reason column to agent_sessions table..."
+            sqlite3 "$DB_PATH" "ALTER TABLE agent_sessions ADD COLUMN stop_reason TEXT;" 2>/dev/null && \
+                success "stop_reason column added to agent_sessions" || \
+                warn "Failed to add stop_reason column (may already exist)"
+            migrated=$((migrated + 1))
+        fi
+    fi
+
     if [ "$migrated" -eq 0 ]; then
         success "No migrations needed"
     else
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -68,7 +68,8 @@ CREATE TABLE IF NOT EXISTS agent_sessions (
     notified_at         TEXT,
     trigger_message_id  TEXT,
     trigger_snippet     TEXT,
-    reply_message_ids   TEXT
+    reply_message_ids   TEXT,
+    stop_reason         TEXT
 );
 """
 
@@ -111,6 +112,7 @@ _MIGRATION_STMTS = [
     "ALTER TABLE agent_sessions ADD COLUMN trigger_message_id TEXT",
     "ALTER TABLE agent_sessions ADD COLUMN trigger_snippet TEXT",
     "ALTER TABLE agent_sessions ADD COLUMN reply_message_ids TEXT",
+    "ALTER TABLE agent_sessions ADD COLUMN stop_reason TEXT",
 ]
 
 # Additive migrations for the reports table (BIS-85 multi-instance prep).
@@ -296,6 +298,7 @@ def session_end(
     id_or_task_id: str,
     status: str,
     result_summary: str | None = None,
+    stop_reason: str | None = None,
     path: Path | None = None,
 ) -> None:
     """Mark an agent session as completed or failed.
@@ -310,6 +313,9 @@ def session_end(
         id_or_task_id:  The id or task_id of the session to end.
         status:         Final status: 'completed' | 'failed' | 'dead'.
         result_summary: Optional short summary of the outcome.
+        stop_reason:    Why the agent stopped: 'end_turn', 'tool_use',
+                        'max_turns', 'error', 'killed', etc. NULL for
+                        sessions that ended before this column existed.
         path:           DB path override (for tests).
     """
     resolved = path if path is not None else _DEFAULT_DB_PATH
@@ -324,10 +330,10 @@ def session_end(
     conn.execute(
         """
         UPDATE agent_sessions
-        SET status = ?, completed_at = ?, result_summary = ?
+        SET status = ?, completed_at = ?, result_summary = ?, stop_reason = ?
         WHERE (id = ? OR task_id = ?) AND status IN ('running', 'starting')
         """,
-        (status, now, result_summary, id_or_task_id, id_or_task_id),
+        (status, now, result_summary, stop_reason, id_or_task_id, id_or_task_id),
     )
     conn.commit()
 
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -631,18 +631,16 @@ def get_active_sessions(path: Path | None = None) -> list[dict]:
     """Return all currently running sessions with elapsed time.
 
     Returns:
-        List of dicts with keys: id, task_id, agent_type, description,
-        chat_id, source, status, output_file, timeout_minutes, parent_id,
-        spawned_at, elapsed_seconds.
+        List of dicts with all session columns plus elapsed_seconds.
+        Uses SELECT * so new columns (e.g. stop_reason) are included
+        automatically without requiring updates to this function.
     """
     resolved = path if path is not None else _DEFAULT_DB_PATH
     conn = _get_connection(resolved)
 
     cursor = conn.execute(
         """
-        SELECT id, task_id, agent_type, description, chat_id, source, status,
-               output_file, timeout_minutes, parent_id, spawned_at
-        FROM agent_sessions
+        SELECT * FROM agent_sessions
         WHERE status IN ('running', 'starting')
         ORDER BY spawned_at ASC
         """
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1845,6 +1845,13 @@ async def list_tools() -> list[Tool]:
                         "type": "string",
                         "description": "Optional short summary of the outcome.",
                     },
+                    "stop_reason": {
+                        "type": "string",
+                        "description": (
+                            "Why the agent stopped. Known values: 'end_turn', 'tool_use', "
+                            "'max_turns', 'error', 'killed'. NULL for legacy rows."
+                        ),
+                    },
                 },
                 "required": ["agent_id", "status"],
             },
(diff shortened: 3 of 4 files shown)


sayhar (commented):
## Review: stop_reason column addition

Overall: looks solid and safe. A few observations below.

### upgrade.sh migration (Migration 14)

**Safe.** The PRAGMA table_info guard correctly prevents double-application of the ALTER TABLE. The fallback `warn` on failure is appropriate since SQLite will error if the column already exists (and the guard should prevent that, but the fallback handles any edge case).

One minor issue: the `migrated=$((migrated + 1))` increment is **inside the outer `if [ -f "$DB_PATH" ]` block but outside the inner guard**. This means if the DB file doesn't exist, no increment happens (correct), but if it does exist and the column is already present (guard fires), the migration count is also not incremented (correct). The logic is fine.

The DB path derivation `${LOBSTER_MESSAGES:-$HOME/messages}/config/agent_sessions.db` matches the pattern used in session_store.py's `_DEFAULT_DB_PATH`, so no path mismatch risk.

**One gap:** if `$DB_PATH` doesn't exist yet (fresh install before first run), the migration is silently skipped. That's fine because `session_store.py`'s `_MIGRATION_STMTS` list handles the in-process path at `init_db()` time, so new DBs get the column from CREATE TABLE DDL anyway. But this means the upgrade.sh migration only matters for existing DBs — which is exactly the correct semantic.

### session_store.py

- Column added to `CREATE TABLE` DDL (nullable TEXT, no default) — correct. Nullable is the right choice; existing rows should be NULL, not a fake value.
- `_MIGRATION_STMTS` append is correct and matches the same pattern as the prior three columns.
- `session_end()` signature: `stop_reason: str | None = None` as a new optional parameter is backward-compatible. No existing callers are broken.
- UPDATE query correctly includes `stop_reason = ?` in the SET clause with the value passed positionally. Binding order looks right: `(status, now, result_summary, stop_reason, id_or_task_id, id_or_task_id)`.

### inbox_server.py call sites

All four session_end call sites are accounted for and wired correctly:

| Site | stop_reason value | Correct? |
|---|---|---|
| `handle_write_result` auto-unregister | `"end_turn"` | Yes — subagent reached completion |
| reconciler `file_status == "done"` | `"end_turn"` | Yes — same semantic |
| reconciler dead-threshold | `"killed"` | Yes — agent died without clean exit |
| `handle_session_end` MCP tool | caller-supplied via args | Yes — passthrough, nullable |

The tool schema update in `list_tools()` correctly documents the known enum values without making them required (good — future values are possible).

### Risk to existing rows

None. The column is nullable with no default, so existing rows get NULL automatically. The UPDATE in `session_end()` only touches rows with `status = 'running'` (the existing guard), which is unchanged.

### Minor suggestions (non-blocking)

1. The `stop_reason` parameter is inserted between `result_summary` and `path` in the function signature. Alphabetical or semantic grouping would be fine either way, but worth noting for consistency with any callers that use positional args (none appear to, so this is safe).
2. The `"killed"` value used by the dead-threshold reconciler branch is reasonable but undocumented in the schema description — the tool description lists `'end_turn', 'tool_use', 'max_turns', 'error', 'killed'` so `"killed"` is covered. Good.
3. No `_row_to_dict` or `_row_to_active_dict` changes — if these functions return `stop_reason` from the SELECT, it would be available to callers; if not, diagnostics querying the store won't see it. Worth confirming these return all columns (or add `stop_reason` explicitly). Not a blocker for merge.

### Verdict

Migration is idempotent and safe. Column type and constraints are appropriate. All session_end call sites updated. No risk of breaking existing rows. Ready to merge.

sayhar (commented):
## Test Fix — PASS

Three tests were failing with `NameError: name 'OWNER_CHAT_ID_PLACEHOLDER' is not defined`. The same bug also blocked all 36 tests in `test_ghost_detector_filesystem.py` from collecting (import-time error in `ghost-detector.py`).

**Root cause:** `OWNER_CHAT_ID_PLACEHOLDER` was used as a bare Python identifier in two places where it should be a string literal:

1. `scripts/ghost-detector.py` line 591: `RELAUNCH_CHAT_ID = OWNER_CHAT_ID_PLACEHOLDER` — module-level assignment where no such variable is defined
2. `tests/test_agent_sessions.py` lines 232, 362, 396: three tests passing the bare name as an argument

**Fix:**
- `ghost-detector.py`: changed to `os.environ.get("LOBSTER_ADMIN_CHAT_ID", "OWNER_CHAT_ID_PLACEHOLDER")` — reads the real chat_id from env (consistent with `alert.sh` and `run-job.sh`) and falls back to the sentinel string
- `tests/test_agent_sessions.py`: quoted the three bare identifiers as `"OWNER_CHAT_ID_PLACEHOLDER"`

**Tests run:**
- `uv run pytest tests/test_agent_sessions.py tests/unit/test_ghost_detector_filesystem.py -v` — 58 passed, 0 failed
- `uv run pytest tests/test_agent_sessions.py::test_optional_fields tests/test_agent_sessions.py::test_tracker_add_with_all_params tests/test_agent_sessions.py::test_json_migration -v` — 3/3 passed (confirmed fail before fix, pass after)

The broader test suite has 69 pre-existing failures unrelated to this PR (daemon module not found, auth message text changes, memory DB errors) — those exist on `main` as well and are out of scope.

**Verdict: PASS** — the 3 targeted failures are fixed and no new failures introduced.

sayhar (commented):
PASS: All 3 OWNER_CHAT_ID_PLACEHOLDER test failures are fixed (bare name → string literal in test_agent_sessions.py + os.environ.get fallback in ghost-detector.py). 22 agent_sessions tests pass, 36 ghost-detector tests pass. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR adds a `stop_reason` column to `agent_sessions` to track why an agent stopped, with full migration support.

**What the code does:**
1. **`session_store.py`**: Adds `stop_reason TEXT` column to the schema and `_MIGRATIONS` list. Updates `session_end()` to accept and persist `stop_reason`. Changes `get_active_sessions()` from explicit column list to `SELECT *` so future column additions don't require updates here.
2. **`inbox_server.py`**: Exposes `stop_reason` in the `session_end` MCP tool schema. In `handle_write_result()`, auto-sets `stop_reason="end_turn"` when a subagent calls `write_result` (since calling write_result means the agent completed its turn). In `handle_session_end()`, passes `stop_reason` through from the caller.
3. **`scripts/upgrade.sh` (Migration 14)**: Uses `ALTER TABLE ... ADD COLUMN` to add `stop_reason` to existing installs. Guards with a `PRAGMA table_info` check to detect if the column already exists (idempotent).
4. **`scripts/ghost-detector.py`**: Fixes `RELAUNCH_CHAT_ID` to read from `LOBSTER_ADMIN_CHAT_ID` env var instead of a hardcoded placeholder string.

**Why it matters:** `stop_reason` is critical for the startup cleanup logic (PR #647 depends on reading it from output files). Having it in the DB as well enables reporting and future queries ('which agents crashed vs. completed?').

**Observations:**
- The `SELECT *` change in `get_active_sessions()` is a reasonable tradeoff: less brittle to schema changes, but callers now depend on undefined column order/presence. Since the function returns dicts (`row_factory = sqlite3.Row`), this is safe.
- `stop_reason="end_turn"` auto-set in `handle_write_result` is correct semantics: an agent that completes its task and calls write_result has reached an end_turn state.
- The ghost-detector env var fix is a nice cleanup — a hardcoded placeholder string in production code would cause a runtime error if `send_alert` was ever called with it.

PR #517, opened 2026-03-16
fix: compact-reminder sorts first in inbox after compaction

### LINKED ISSUE #513: fix: compact-reminder sorts after queued user messages, delaying re-orientation
## Bug

When context compaction fires, \`on-compact.py\` writes the compact-reminder with \`ts_ms = int(time.time() * 1000)\`. Because inbox files are sorted lexicographically by filename (which starts with the timestamp), the compact-reminder filename sorts **after** any user messages that were already queued at compaction time.

The dispatcher processes user messages before seeing the re-orientation signal, which means it may act on incoming requests without first restoring its identity context from \`CLAUDE.md\` and \`handoff.md\`.

## Root cause

\`write_reminder()\` in \`hooks/on-compact.py\`:

\`\`\`python
ts_ms = int(time.time() * 1000)
message_id = f"{ts_ms}_compact"
\`\`\`

\`handle_check_inbox()\` in \`src/mcp/inbox_server.py\` sorts with:

\`\`\`python
for f in sorted(INBOX_DIR.glob("*.json")):
\`\`\`

Lexicographic sort on filenames means \`1741234567890_compact.json\` sorts after any earlier-timestamped user messages.

## Fix

Write the compact-reminder with \`ts_ms = 0\`, producing filename \`0_compact.json\`, which sorts before any real message filename.

## Implementation plan

- [x] Change \`ts_ms = int(time.time() * 1000)\` to \`ts_ms = 0\` in \`write_reminder()\` in \`hooks/on-compact.py\`
- [x] Verify \`handle_check_inbox\` sorting confirms \`0_compact.json\` sorts first
- [x] Update \`already_pending()\` check — no change needed (it scans by \`subtype\`, not filename)
- [x] Verify \`message_id\` field in the JSON payload (\`0_compact\`) is still unique/valid

sayhar:
## Problem

After context compaction, the dispatcher was processing queued user messages before reading the compact-reminder re-orientation signal. This meant the dispatcher could act on incoming requests without first restoring its identity from \`CLAUDE.md\` and \`handoff.md\`.

Root cause: \`write_reminder()\` in \`hooks/on-compact.py\` timestamped the reminder with \`int(time.time() * 1000)\`, producing a filename like \`1741234567890_compact.json\`. The \`handle_check_inbox\` path in \`src/mcp/inbox_server.py\` iterates \`sorted(INBOX_DIR.glob("*.json"))\` — pure lexicographic order on filenames. Any user messages already in the inbox with an earlier timestamp sorted before the compact-reminder, so they were processed first.

## Fix

Use \`ts_ms = 0\` in \`write_reminder()\`, producing the filename \`0_compact.json\`. The digit \`0\` sorts before any real millisecond timestamp (which starts at \`1...\`), so the compact-reminder is always first in the sorted inbox.

The \`already_pending()\` deduplication guard is unaffected — it matches messages by \`subtype\` field in the JSON body, not by filename.

The \`message_id\` and \`timestamp\` fields in the JSON payload remain independent: \`message_id\` becomes \`"0_compact"\` (still unique, since only one compact-reminder is ever written at a time) and \`timestamp\` continues to record the actual wall-clock time.

## Functional pattern

Pure function: \`write_reminder()\` has no shared mutable state. The change is a constant substitution with no side-effect widening.

## Tests run

- [x] \`uv run pytest tests/unit/test_hooks/ tests/unit/test_mcp_server/ -q --tb=short\` — 206 passed, 17 pre-existing failures in unrelated \`test_mcp_server\` tests (confirmed not introduced by this change)
- [ ] Live compaction test — blocked: requires triggering a real compaction cycle in a running dispatcher session; safe to merge, behavior is a pure sort-order shift with no other observable change

Closes #513

### DIFF
--- a/hooks/on-compact.py
+++ b/hooks/on-compact.py
@@ -82,8 +82,12 @@ def write_reminder() -> None:
     """Write a compact-reminder system message to the inbox."""
     INBOX_DIR.mkdir(parents=True, exist_ok=True)
 
-    # Timestamp in milliseconds, matching the pattern used by other inbox files.
-    ts_ms = int(time.time() * 1000)
+    # Use ts_ms=0 so the filename ("0_compact.json") sorts lexicographically
+    # before any real user-message filename (which starts with the current epoch
+    # in milliseconds, e.g. "1741234567890_...").  This guarantees the
+    # compact-reminder is the first message the dispatcher sees after
+    # compaction, regardless of how many user messages were queued beforehand.
+    ts_ms = 0
     message_id = f"{ts_ms}_compact"
     timestamp = time.strftime("%Y-%m-%dT%H:%M:%S", time.gmtime()) + ".000000"
 
--- a/tests/unit/test_hooks/test_compact_reminder.py
+++ b/tests/unit/test_hooks/test_compact_reminder.py
@@ -1,0 +1,47 @@
+"""
+Unit tests for the compact-reminder sort-order fix in hooks/on-compact.py.
+
+The fix uses ts_ms=0 so the filename "0_compact.json" sorts before any real
+user-message filename (which begins with the current epoch in milliseconds,
+e.g. "1773695000000_msg.json").  These tests assert that sort invariant
+directly, without importing the hook.
+"""
+
+
+def test_compact_reminder_sorts_before_real_messages():
+    """0_compact.json must be first after sorted(), regardless of how many
+    real epoch-ms filenames are present."""
+    filenames = [
+        "1773695000000_msg.json",
+        "1773695001234_msg.json",
+        "0_compact.json",
+        "1741234567890_compact.json",  # old-style filename (pre-fix)
+        "1773695999999_msg.json",
+    ]
+    result = sorted(filenames)
+    assert result[0] == "0_compact.json", (
+        f"Expected '0_compact.json' to be first but got '{result[0]}'; "
+        f"full order: {result}"
+    )
+
+
+def test_compact_reminder_sorts_before_single_real_message():
+    """Minimal case: one compact-reminder and one user message."""
+    filenames = ["1773695000000_msg.json", "0_compact.json"]
+    result = sorted(filenames)
+    assert result[0] == "0_compact.json"
+
+
+def test_real_messages_sort_among_themselves_unchanged():
+    """Sorting real epoch-ms filenames still produces ascending timestamp order."""
+    filenames = [
+        "1773695003000_c.json",
+        "1773695001000_a.json",
+        "1773695002000_b.json",
+    ]
+    result = sorted(filenames)
+    assert result == [
+        "1773695001000_a.json",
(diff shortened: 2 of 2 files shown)


sayhar (commented):
## Review: fix/issue-513-compact-reminder-sort

**Summary:** The fix is correct and safe to merge. All four review questions check out. One minor gap worth noting.

---

### 1. Does ts_ms=0 actually sort before all real message filenames?

Yes. `handle_check_inbox` (inbox_server.py:2646) sorts with `sorted(INBOX_DIR.glob("*.json"))` — pure Python lexicographic order on the full filename string, not numeric order. The resulting filename is `0_compact.json`. Real messages use `int(time.time() * 1000)` (inbox_server.py:4396, 4491, 6098), which today produces ~17-digit strings starting with `1`. Lexicographically, `"0..."` < `"1..."` always. This holds permanently — the `1...` prefix is structural to Unix epoch milliseconds and won't change in our lifetimes.

### 2. Any edge case where ts_ms=0 could conflict with another system message?

No conflict found. Surveying all inbox writers:
- `handle_write_observation` (line 4491): uses real epoch ms, suffix `_observation_<uuid>`
- `_write_inbox_message` for cron results (line 4396): uses real epoch ms, suffix `_<task_id>`
- `_write_reconciler_message` (line 6098): uses real epoch ms, suffix `_reconciler_<id>`
- `whatsapp_bridge_adapter.py` (line 312): uses real epoch ms, writes to WA_COMMANDS_DIR, not INBOX_DIR

The only file that could collide with `0_compact.json` is another call to `write_reminder()` — and the `already_pending()` deduplication guard (which checks `subtype == "compact-reminder"` in the JSON body) prevents that. The idempotency test A4 covers this regression explicitly.

### 3. Is write_reminder() the right function to change?

Yes. `write_reminder()` is the sole point of origin for compact-reminder messages. The change is minimal, isolated, and the comment explaining the rationale is clear. No callers pass a timestamp in; this is a pure internal constant substitution.

### 4. Are there tests for the sort order?

**This is the one gap.** The smoke tests (A1–A4 in `tests/smoke/test_on_compact.py`) verify that the file is written with correct fields and the hook is idempotent. They do **not** assert that the resulting filename actually sorts before a typical real-message filename. There is no test like:

```python
def test_compact_reminder_sorts_before_user_messages():
    reminder = "0_compact.json"
    user_msg = f"{int(time.time() * 1000)}_telegram_12345.json"
    assert sorted([user_msg, reminder])[0] == reminder
```

This is a low-risk gap (the sort property is structurally guaranteed by lexicographic ordering of `"0"` vs `"1..."`), but a targeted regression test would lock in the invariant explicitly and catch any future refactor that changes the sort key or filename scheme.

---

**Verdict:** Correct fix, clean implementation, safe to merge. Suggest adding a sort-order assertion to the test suite as a follow-up (can be a separate issue/PR rather than blocking this one).

PR #518, opened 2026-03-16
fix(tests): rewrite test_require_write_result for current hook implementation

sayhar:
## Summary

PR #462 removed `_extract_agent_id_from_transcript()` from the hook and replaced it with `_extract_write_result_task_ids()`. The test file was never updated and continued testing the removed function, causing all 9 tests to fail with `AttributeError` on every run.

The tests now cover what the hook actually does:

- **`_extract_write_result_task_ids()`** (8 tests): extraction of task_ids from write_result tool_use blocks, deduplication, and rejection of empty/non-string/missing values.
- **`main()` integration** (10 tests): all exit paths — exit 0 for valid write_result with chat_id, exit 2 when write_result is absent or every call has a null chat_id, pseudocode detection path (write_result mentioned in text but not called as a tool), dispatcher session exemption, chat_id=0 background-agent route, invalid stdin JSON, and missing transcript key.

One implementation detail worth noting: the hook uses `from session_role import is_dispatcher, get_session_id`, which binds both names directly on the `require_write_result` module object. The previous test style (`patch("session_role.is_dispatcher")`) patched the source module but not the binding used by `main()`. The new tests use `patch.object(mod, "is_dispatcher")` so the dispatcher-exemption path is correctly controlled.

No changes to the hook itself or any other code.

## Tests run

- [x] `cd ~/lobster && uv run python -m pytest tests/test_require_write_result.py -v` — 18 passed, 0 failed (replaced the 9 that were all failing)

### DIFF
--- a/hooks/require-write-result.py
+++ b/hooks/require-write-result.py
@@ -32,6 +32,32 @@ The proper hex-ID row is the primary target but is only reachable when
 task_id matches. When it is not reachable via this hook, the reconciler in
 inbox_server.py will close it when it detects stop_reason=end_turn in the
 output file.
+
+## SubagentStop vs Stop transcript handling
+
+Neither SubagentStop nor Stop hooks receive an inline `transcript` field in
+CC 2.1.76+. Both pass a file path instead:
+- Stop: `transcript_path` (JSONL file of the current session's conversation)
+- SubagentStop: `agent_transcript_path` (JSONL file of the subagent's conversation)
+
+This hook reads the appropriate file path for each event type. An inline
+`transcript` field is supported as a legacy fallback for older CC versions.
+
+## JSONL message format
+
+Each line of the JSONL transcript file has the structure:
+    {"type": "assistant", "message": {"role": "assistant", "content": [...]}, ...}
+
+Tool use items are nested under entry["message"]["content"], NOT entry["content"].
+`_collect_tool_use_and_text` handles both the JSONL format and the legacy inline
+format where content is directly on the message dict.
+
+## Suppressing feedback injection on success
+
+Claude Code injects a "Stop hook feedback: ... No stderr output" system message
+into the agent even when the hook exits 0. To prevent this feedback from
+triggering a new agent turn, the hook outputs JSON with `{"suppressOutput": true}`
+on all success paths. Per the CC hook spec, this suppresses feedback injection.
 """
 import json
 import sys
@@ -40,6 +66,17 @@ from pathlib import Path
 # Import shared session role utility.
 sys.path.insert(0, str(Path(__file__).parent))
 from session_role import is_dispatcher, get_session_id
+
+# JSON to emit on every successful (allow) exit — suppresses the
+# "Stop hook feedback: No stderr output" injection that CC 2.1.76+ produces
+# even when the hook exits 0 with no output.
+_SILENT_OK = json.dumps({"suppressOutput": True})
+
+
+def _exit_ok() -> None:
+    """Exit 0 with JSON that suppresses CC feedback injection."""
+    print(_SILENT_OK)
+    sys.exit(0)
 
 
 def _extract_write_result_task_ids(all_tool_use_items: list) -> list[str]:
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Review of test_require_write_result.py (18 tests)

Good foundational coverage. The core scenarios are all exercised and the patch.object fix for the module-level name binding is correct and important. A few issues worth addressing before merge:

---

### Bug: pseudocode test asserts on stdout, hook writes to stderr

`test_exits_2_with_pseudocode_message_when_write_result_in_text` asserts `captured.out` but the hook calls `print(..., file=sys.stderr)` on all error paths. The test comment acknowledges uncertainty ("The hook prints to stdout in this version") — this needs to be verified. As written the assertion will silently pass only if capsys happens to capture stderr in `.out`, which it won't. Should be `captured.err`.

Similarly, the two `sys.exit(2)` error messages in `main()` (lines 301–309 and 320–333) both write to `sys.stderr`. The tests that check the exit code alone will still pass, but any future test asserting on the error text needs to use `captured.err`.

---

### Missing: test_exits_0_when_missing_transcript_key has a wrong docstring

The docstring says "exits 2" but the function name says "exits_0". The assertion is `assert code == 2`. The name is the bug — should be `test_exits_2_when_missing_transcript_key`. Minor, but misleading.

---

### Missing: JSONL transcript path / SubagentStop event

The hook has two distinct transcript-loading paths:
- `SubagentStop`: reads from `agent_transcript_path` (JSONL file)  
- `Stop`: reads from `transcript_path` (JSONL file) or falls back to inline `transcript`

All 10 `main()` tests exercise only the inline legacy fallback path via `data["transcript"]`. The JSONL path (`_load_transcript_from_jsonl`) and the SubagentStop branch are completely untested. At minimum, one test for the "no `agent_transcript_path` provided → allow exit" branch (line 225–227 of the hook) would be valuable since that's a silent `_exit_ok()` with no verification.

---

### Missing: _mark_session_notified is never tested

`_mark_session_completed` has one test (`test_marks_session_completed_with_task_id_on_success`). `_mark_session_notified` — which is called immediately after on the same ids — has zero tests. The PR summary doesn't mention this function at all. It's best-effort so failures are swallowed, but the calling sequence (completed → notified, for both task_ids and session_id) has no coverage.

---

### Minor: session_id used in _mark_session_completed is untested

`test_marks_session_completed_with_task_id_on_success` only checks that `_mark_session_completed` is called with the write_result `task_id`. The hook also calls `_mark_session_completed(session_id)` when `get_session_id` returns a truthy value. There's no test confirming that the session_id path is covered.

---

### What's good

- The `patch.object(mod, "is_dispatcher")` fix is correct and well-explained in the PR description. The old `patch("session_role.is_dispatcher")` approach genuinely would not have intercepted the bound name.
- `_extract_write_result_task_ids` has thorough, clean coverage: empty input, no matches, missing task_id, empty string, dedup, ordering, non-string type. These 8 tests are solid.
- `_run_main` helper with `_mark_session_completed` patched is the right design.
- `test_exits_0_when_chat_id_is_zero` explicitly tests the chat_id=0 background-agent route — a real edge case.
- `test_exits_0_when_stdin_invalid_json` confirms the hook never blocks on bad input.

---

### Summary

The test rewrite is a clear improvement over 9 failing tests. The `captured.out` vs `captured.err` bug in the pseudocode test and the misleading function name for the missing-transcript test should be fixed. JSONL path coverage and `_mark_session_notified` are gaps but can be addressed in a follow-up if scope is a concern.

sayhar:
All four review issues addressed in the latest commit (00b1f75):

**1. stderr vs stdout** — `test_exits_2_with_pseudocode_message_when_write_result_in_text` now asserts `captured.err`. Removed the stale comment claiming the hook writes to stdout. The hook on this branch was at an older revision that used plain `print()`; updating the hook to the current version (which uses `file=sys.stderr` on all error paths) made the correct assertion clear.

**2. Misleading function name** — `test_exits_0_when_missing_transcript_key` renamed to `test_exits_2_when_missing_transcript_key`.

**3. JSONL/SubagentStop path** — Two new test classes added:
- `TestLoadTranscriptFromJsonl` (4 tests): valid JSONL, blank lines, nonexistent file, invalid JSON lines
- `TestCollectToolUseAndText` (4 tests): legacy inline format, JSONL nested format, text extraction, empty input

Plus two new `main()` tests: SubagentStop with no `agent_transcript_path` exits 0 (can't verify, allow), and SubagentStop reading a real JSONL temp file exits 0 when write_result is present.

**4. `_mark_session_notified`** — New `TestMarkSessionNotified` class (2 tests): best-effort error swallow when `session_store` import fails, and successful `set_notified()` call when available. The existing success test also expanded to verify `_mark_session_notified` is called with the same task_id.

The hook itself was also updated to the current version (adds SubagentStop JSONL path, `_mark_session_notified`, `_collect_tool_use_and_text`, `_exit_ok`, and `suppressOutput` on success exits — it was at an older snapshot). `_run_main` now patches both `_mark_session_completed` and `_mark_session_notified`.

**30 tests pass, 0 fail.**

PR #519, opened 2026-03-16
fix: enforce review-before-notify rule in functional-engineer agent

sayhar:
## Problem

The rule "open PR → post review → surface both together, never notify without review" existed in `user.base.bootup.md` but was **absent from `functional-engineer.md`**. The old reporting template explicitly modeled:

```
open PR → send_reply
```

No review step. This is why functional-engineer subagents kept surfacing PRs without reviews — the rule was nowhere in their operative context.

## What this changes

Three additions to `functional-engineer.md`:

1. **HARD RULE blockquote in section 6** (Pull Request Creation) — immediately after the PR-open bullet, before the PR description guidance. Includes the exact `gh pr review` command. States "no exceptions — even for trivial changes."

2. **Updated reporting template** — the success code example now shows the required three-step sequence: post review → send_reply → write_result. Comment at the top says "DO NOT call send_reply before the review is posted."

3. **Updated section header** — the "Reporting Results Back to the User" intro now leads with the hard rule in bold, with the full ordered sequence spelled out.

## What this does not change

No behavior changes to the dispatcher, MCP servers, or any other agent definition. This is purely a documentation fix to a single agent file.

## Tests run

- [x] Read the modified file and verified both edit points look correct
- [ ] Live agent invocation — not applicable; this is a doc/instruction fix

### DIFF
--- a/.claude/agents/functional-engineer.md
+++ b/.claude/agents/functional-engineer.md
@@ -90,6 +90,22 @@ git branch -d feature/issue-42-my-feature
 - Reference the issue in the PR description using keywords (Closes #XX, Fixes #XX, or Relates to #XX)
 - **Set "Main Board" project status to "In Review"** after PR is opened
 
+> **HARD RULE — REVIEW BEFORE NOTIFY**
+>
+> **YOU MUST POST A REVIEW TO THE PR BEFORE YOU CALL `send_reply` TO NOTIFY THE USER.**
+>
+> The rule: **open PR → post review → surface both together, never notify without review.**
+>
+> After opening the PR, run `gh pr review` to post an inline review comment:
+>
+> ```bash
+> gh pr review <PR_NUMBER> --repo <owner/repo> \
+>   --comment \
+>   --body "Automated self-review: [summary of approach, tradeoffs, anything the human reviewer should focus on]"
+> ```
+>
+> Only then call `send_reply`. If you skip the review and call `send_reply` first, you are violating this rule. There are no exceptions — even for trivial changes.
+
 **Writing the PR description:**
 
 A PR description is a communication artifact, not a changelog. Its audience is a maintainer reviewing on mobile who needs to answer one question: "Is this safe to merge?" Write for that person.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Self-review: This fix closes a concrete gap in the agent's operative instructions. The change is minimal and surgical — two edit points, no logic changes, no new dependencies.

The HARD RULE blockquote in section 6 is placed at exactly the right moment in the workflow: after the agent knows it needs to open a PR, before it has any guidance on what to do next. The rule is hard to miss in this position.

The reporting template update is the more important of the two changes. Agents model their behavior heavily on code examples. The old example showed 'open PR, then send_reply' — two steps, in that order, with no review step anywhere. The new example shows the correct three-step sequence with a comment at the top that explicitly says 'DO NOT call send_reply before the review is posted.' This is the change most likely to actually alter agent behavior.

One thing worth considering for a follow-up: the rule in user.base.bootup.md and this file now say the same thing, but independently. If the rule ever changes, both will need updating. A future refactor could consider making the canonical statement live in one place only. That's out of scope here.

Safe to merge as-is.

sayhar (commented):
PASS: The change is correct and well-placed. The hard rule is inserted at both the right moment in section 6 (immediately after the PR-open bullet, before the agent can proceed) and at the top of section 8 (the reporting template), giving redundant enforcement. The `gh pr review --comment` flag is correct. The code template accurately models the four-step sequence with clear DO-NOT comments.

Minor cosmetic note: the PR description says the sequence is steps 1-2-3-4, but the code block labels the `gh pr review` bash step inline in comments rather than as "Step 1" — the `send_reply` call is labeled "Step 2" and `write_result` is "Step 3". The actual order is right; the numbering just differs from the description. Not a blocker.

Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR enforces the review-before-notify rule in the functional-engineer agent definition, making it a hard rule with a clear sequence and blocking language.

**What the code does:** Updates `.claude/agents/functional-engineer.md` in two places:
1. Adds a prominent HARD RULE block immediately after the PR workflow section: 'open PR → post review → surface both together, never notify without review.' Includes the exact `gh pr review` command to use.
2. Updates the 'Reporting Results Back to the User' section with a HARD RULE header and the required 4-step sequence: (1) open PR → (2) post review → (3) send_reply → (4) write_result. Clarifies the message to include 'and has a self-review posted'.

**Why it matters:** The previous version had this as documentation but without the blocking emphasis. The new version uses repeated HARD RULE / NEVER language to make the constraint unmissable. Engineers reviewing their own work without posting a review creates an information asymmetry where the user gets 'PR opened' but no visibility into what was done and why.

**Observations:**
- The review step here is a self-review (the engineer reviewing their own PR), which is different from the independent reviewer spawned by the dispatcher (as documented in PR #583). These are two different things: the self-review is a structured summary of approach/tradeoffs; the independent review is a quality gate. Both are required in the full flow.
- The 4-step sequence in the code block matches the described flow. The PR is internally consistent.
- Note: PR #583 changed this file to have the engineer call `write_result` without `send_reply` (letting the dispatcher spawn a reviewer first). This PR (519) has the engineer call `send_reply` after posting the self-review. These two PRs are in conflict. They should not both be merged without reconciliation — whichever represents the current intended flow should be applied and the other updated.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: enforce review-before-notify rule in functional-engineer agent**

The change is targeted and correct. This is a documentation-only PR with no behavior changes to code.

**What changed:**
Three edits to `.claude/agents/functional-engineer.md`:
1. HARD RULE blockquote added immediately after the PR-open bullet in section 6, with the exact `gh pr review` command.
2. Reporting template updated to show the required 3-step sequence with explicit "DO NOT call send_reply before the review is posted" comment.
3. Section header updated to lead with the hard rule and the full ordered sequence.

**Why this is the right fix:**
The root cause was that the rule existed in `user.base.bootup.md` but not in the operative context for functional-engineer subagents. Subagents read their agent file; the base bootup may not always be loaded or prioritized. Putting the rule directly in the agent file, in the exact section where the agent is performing the action (PR creation), is the correct placement.

**The blockquote format:**
The `> **HARD RULE**` blockquote stands out visually and the explicit "no exceptions" language matches the intent. The exact `gh pr review` command is included inline, removing any ambiguity about how to comply.

**Potential gap:**
The rule is now in `functional-engineer.md` and `user.base.bootup.md`. It is not in `sys.subagent.bootup.md` or other agent files. If future agents are created for PR work, they would need the same rule. This is out of scope for this PR but worth tracking.

No blocking concerns.

PR #523, opened 2026-03-16
fix(health-check): retry PID verification after restart and alert on failure

### LINKED ISSUE #522: health-check: PID verification race + silent failure on restart
## Bug 1: PID check timing race condition

After `systemctl start lobster-claude`, the script waits 5 seconds then reads the new Claude PID to verify the restart succeeded. In practice 5 seconds is sometimes not enough — the new process hasn't written its PID yet. The comparison sees the old PID still in place and logs:

```
Restart verification failed: Claude PID XXXX unchanged — old session may have survived
```

...even though the restart actually succeeded.

**Fix:** Replace the single `sleep 5` + immediate PID read with a retry loop: after the initial wait, retry the PID check up to 3 times with 3-second gaps before declaring verification failed.

## Bug 2: No alert on verification failure

When PID verification fails (line 973), `send_telegram_alert` is never called. The script logs the error and returns 1, but the user receives no Telegram notification. Restarts can silently fail with no user-visible signal.

**Fix:** Add a `send_telegram_alert` call in the verification-failure `else` branch, e.g.:
> "Restart attempted but PID unchanged — restart may have failed. Check `lobster-claude` service."

## Affected file

`scripts/health-check-v3.sh` — `do_restart()` function, lines ~933–975

sayhar:
## Problem

Two gaps in \`do_restart()\` in \`health-check-v3.sh\` caused silent or spurious failures:

**Race condition:** After \`systemctl start lobster-claude\`, the script waited 5 seconds then read the Claude PID once. On slower starts the new process hadn't written its PID yet, so the comparison saw the old PID still in place and logged "Restart verification failed: Claude PID XXXX unchanged" — even though the restart actually succeeded and the system was healthy.

**Silent failure:** When verification genuinely did fail (PID unchanged after waiting), \`send_telegram_alert\` was never called. The error was written to the log and the function returned 1, but the user received no Telegram notification. Failing restarts could go unnoticed indefinitely.

## Changes

Both fixes are contained in the \`do_restart()\` function in \`scripts/health-check-v3.sh\`.

**Bug 1 — retry loop replaces single read:** The immediate post-\`sleep 5\` PID read is replaced with a \`while\` loop that retries up to 3 times with 3-second gaps. The loop exits early as soon as the PID changes (happy path adds no delay). Only after all 3 attempts still show the same PID does the script conclude the restart failed. Total maximum extra wait: 9 seconds, acceptable given the system is already in RED state.

**Bug 2 — alert on verification failure:** A \`send_telegram_alert\` call is added in the \`else\` branch that executes when \`pid_changed=false\` or the service/tmux checks fail. The alert message includes the restart reason and a prompt to check the \`lobster-claude\` service.

Both fixes are pure behavior changes; no new dependencies, no changes to the overall restart flow or escalation ladder.

## Tests run

- [x] \`bash -n scripts/health-check-v3.sh\` — syntax check passed, no errors
- [ ] Live restart integration test — skipped: requires production environment; no safe way to trigger a real restart in CI

Closes #522

### DIFF
--- a/tests/test-health-check-pid-timing.sh
+++ b/tests/test-health-check-pid-timing.sh
@@ -1,0 +1,364 @@
+#!/bin/bash
+#===============================================================================
+# Test Suite: Health Check v3 - PID Timing / Restart Verification
+#
+# Tests the retry loop introduced in issue #522:
+#   - pgrep -x "claude" is retried up to 3 times with 3-second gaps
+#   - Success is declared as soon as the returned PID differs from the
+#     pre-restart PID (i.e. a new process appeared)
+#   - Failure is declared only after all 3 attempts still return the same PID
+#
+# The retry loop is extracted and exercised with a mock pgrep that returns
+# configurable values per call, so no real systemctl or sleep is needed.
+#
+# Usage: bash tests/test-health-check-pid-timing.sh
+#===============================================================================
+
+set -eE
+
+# Helper: capture exit code without triggering set -e
+run_and_capture_rc() {
+    "$@" && RC=$? || RC=$?
+}
+
+# Colors
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+BOLD='\033[1m'
+NC='\033[0m'
+
+PASS=0
+FAIL=0
+TOTAL=0
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)/scripts"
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Lobster (reviewer): PASS — retry loop correctly addresses the PID timing race by polling up to 3 times with 3-second gaps before declaring failure. Alert on verification failure closes the silent-failure gap. Both changes are minimal and scoped to do_restart(). Syntax check passed. Authorized for merge by Sahar.

PR #525, opened 2026-03-16
fix: on-compact Telegram notify silent after compaction (moves guards)

### LINKED ISSUE #524: on-compact.py: Telegram debug notify silently skipped after context compaction
## Bug

When LOBSTER_DEBUG=true, the debug Telegram notification in \`on-compact.py\` never fires despite direct curl delivery working and the hook running (compacted_at state file updates correctly).

## Root Cause

PR #397 added an \`is_dispatcher()\` guard to prevent subagent compactions from writing spurious compact-reminders. The guard exits early for non-dispatcher sessions — but during context compaction, Claude Code assigns a **new session_id** to the continuing dispatcher. \`is_dispatcher()\` compares this new session_id against the stored \`dispatcher-session-id\` marker file, which still holds the pre-compact session_id. The mismatch causes \`is_dispatcher()\` to return \`False\` (conservative default, since SessionStart hooks have no transcript to fall back on), and the hook exits before \`maybe_send_dev_telegram_notify()\` or \`write_compacted_at()\` run.

The Telegram send code itself is correct. The bug is solely in guard placement.

## Secondary Bug

The same PR #397 broke \`tests/smoke/test_on_compact.py\`. The \`json.load(sys.stdin)\` call in \`main()\` hits pytest's captured stdin and raises \`OSError: pytest: reading from stdin while output is captured\`. All 4 smoke tests fail.

## Fix

Move \`write_compacted_at()\` and \`maybe_send_dev_telegram_notify()\` before the \`is_dispatcher()\` guard. Inbox reminder and sentinel writes remain gated on \`is_dispatcher()\`. Also fix the 4 broken smoke tests.

## Affected files

\`hooks/on-compact.py\`, \`tests/smoke/test_on_compact.py\`

sayhar:
Fixes #524.

## Summary

- After context compaction, Claude Code assigns a new session_id to the continuing dispatcher. `is_dispatcher()` compares this against the stored `dispatcher-session-id` marker file (still holding the pre-compact session_id), sees a mismatch, returns `False`, and the hook silently exits before sending the Telegram notification or writing `compacted_at`.
- PR #397 placed `write_compacted_at()` and `maybe_send_dev_telegram_notify()` after the `is_dispatcher()` guard. Both should fire for any compaction. Only the inbox reminder and sentinel writes need the dispatcher guard.
- PR #397 also broke all 4 `test_on_compact.py` smoke tests: `json.load(sys.stdin)` hits pytest's captured stdin and raises `OSError`.

## Changes

- `hooks/on-compact.py`: move `write_compacted_at()` and `maybe_send_dev_telegram_notify()` before the `is_dispatcher()` guard
- `tests/smoke/test_on_compact.py`: add `_stdin_from_module()` context manager, stub `is_dispatcher` to True, wrap all `mod.main()` calls

## Test plan

- [ ] `uv run python -m pytest tests/smoke/test_on_compact.py -v` passes (all 4)
- [ ] With `LOBSTER_DEBUG=true`, trigger a compaction and confirm Telegram message is received
- [ ] Subagent compacts (rare) now also send Telegram notify in debug mode — acceptable, informational only

### DIFF
--- a/hooks/on-compact.py
+++ b/hooks/on-compact.py
@@ -241,30 +241,37 @@ def main() -> None:
     except (json.JSONDecodeError, ValueError):
         data = {}
 
-    # Only act for the dispatcher session. Subagent compactions must not
-    # inject compact-reminders into the shared inbox or write the sentinel,
-    # because those signals are only meaningful to the dispatcher.
+    # Always record compaction timestamp — runs for both dispatcher and subagent
+    # compactions.  The health check reads this to suppress false-positive
+    # "stale inbox" restarts during any compaction pause window.
+    write_compacted_at()
+
+    # Always send the debug Telegram notification when LOBSTER_DEBUG=true.
+    # This must fire even for the dispatcher compact, where is_dispatcher() would
+    # return False because context compaction assigns a new session_id that no
+    # longer matches the stored dispatcher-session-id marker file.
+    # NOTE: In debug mode this also fires for subagent compactions (rare), which
+    # is acceptable — the notification is informational.
+    maybe_send_dev_telegram_notify()
+
+    # Guard the inbox reminder and sentinel writes to the dispatcher only.
+    # Subagent compactions must not inject compact-reminders into the shared
+    # inbox or write the compact-pending sentinel, because those signals are
+    # only meaningful to the dispatcher.
     if not is_dispatcher(data):
         sys.exit(0)
-
-    # Always record compaction timestamp first — before any early returns.
-    # This ensures the health check can suppress false-positive restarts even
-    # if a compact-reminder is already pending.
-    write_compacted_at()
 
     if already_pending():
         # Sentinel still needs refreshing even if the inbox reminder is a dupe
         # (double compaction without intervening wait_for_messages). Touch resets
         # the TTL clock so the gate keeps blocking correctly.
         write_sentinel()
-        maybe_send_dev_telegram_notify()
         return
     write_sentinel()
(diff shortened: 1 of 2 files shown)


sayhar (commented):
PASS — fix is minimal and correct.

**Root cause confirmed:** `is_dispatcher()` in `session_role.py` uses session_id comparison against a marker file. During compaction, the new session_id never matches the pre-compact stored ID, and the SessionStart hook has no transcript to fall back on, so `is_dispatcher()` returns `False` (conservative default). The hook exits before `maybe_send_dev_telegram_notify()` runs.

**Fix is correct:** Moving `write_compacted_at()` and `maybe_send_dev_telegram_notify()` before the `is_dispatcher()` guard ensures they fire on all compaction events. The inbox reminder write and sentinel touch remain correctly gated.

**Side effects are acceptable:**
- `write_compacted_at()` running on subagent compacts: extends health check suppression window slightly — safe, conservative direction.
- `maybe_send_dev_telegram_notify()` on subagent compacts: sends informational debug notification in `LOBSTER_DEBUG=true` mode. Subagent compactions are rare and the notification is debug-only.

**Tests:** 4 previously failing smoke tests now pass. The test fixes are correct — `_stdin_from_module()` is clean, `is_dispatcher` stub is appropriate since tests are exercising compaction behavior not dispatcher detection.

**No migration needed** — hook file changes are live immediately (hooks/ is in the repo, not user-config).

PR #521, opened 2026-03-16
docs: clarify ghost-writing rule — authorship honesty, not labeling

sayhar:
## Summary

- Clarifies the ghost-writing rule in `ONBOARDING.md`: the rule is about not presenting generated content as if written by the user, not about adding labels to every output
- Updates the live rule in `~/lobster-user-config/agents/user.base.bootup.md` (private, not in repo) to match the clarified intent
- Documents the authorship honesty principle in the public onboarding doc for visibility

## What changed

**Old rule intent (too broad):** Default to "Lobster voice" or an explicit marker (🤖 Lobster:) on any non-user-audience output.

**New rule intent:** Don't lie about authorship. For substantial external-facing content (PR descriptions, emails, issue comments, documents), offer drafts rather than framing them as ready-to-send "from Sahar." Routine replies, API results, and short functional outputs are exempt from any labeling requirement. Explicit approval by Sahar for specific content is always permitted.

## Test plan
- [ ] Read the updated `ONBOARDING.md` section to confirm the new text is clear and accurate
- [ ] Verify the live `user.base.bootup.md` reflects the same intent (already updated)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #528, opened 2026-03-16
fix: derive agent tasks path dynamically from uid and workspace hash

sayhar:
## What was broken

`periodic-self-check.sh` was always emitting generic `status?` pings instead of structured `[Task Completed]` completion messages. The root cause was two bugs in the hardcoded path used to locate Claude Code task output files:

1. **Wrong username segment**: path contained `-home-admin-lobster-workspace` but the actual directory on this system is `-home-lobster-lobster-workspace`
2. **No session UUID awareness**: Claude Code stores task outputs in per-session UUID subdirectories (`<base>/<uuid>/tasks/*.output`), but the code looked directly in the parent directory with `find -maxdepth 1`

Both functions that scanned for task files (`scan_agent_status` and `scan_completed_tasks`) shared the same broken default.

## What this fixes

Adds `_default_tasks_glob()`, a shared helper that derives the correct path at runtime:
- Computes `claude_tmp_base` from the actual running user's UID: `/tmp/claude-$(id -u)`
- Computes `workspace_hash` by replacing `/` with `-` in the `LOBSTER_WORKSPACE` path (matching what Claude Code does)
- Returns a glob that covers all session UUID subdirs: `<base>/<hash>/*/tasks/*.output`

Both `scan_agent_status` and `scan_completed_tasks` now call `_default_tasks_glob()` and use `compgen -G` to expand it, replacing the `find -maxdepth 1` calls.

The `AGENT_TASKS_DIR` env override is preserved unchanged — it still short-circuits the entire derivation for test isolation.

Functional patterns: pure function (`_default_tasks_glob` has no side effects), composable helper used by both scanning functions.

## Tests run

- [x] `source scripts/agent-status.sh && scan_agent_status` — found 94+ agent entries across real session dirs (correct)
- [x] `source scripts/agent-status.sh && scan_completed_tasks` — returned completed task summaries from real session dirs (correct)
- [x] `AGENT_TASKS_DIR=/tmp/fake-dir scan_agent_status` — correctly read from override directory, found `fake-task-abc123` (PASS)
- [ ] Full `periodic-self-check.sh` with new script live — can only be verified after merge since the script sources from `$HOME/lobster/scripts/agent-status.sh` (the installed path, not the worktree)

Closes #521 (if tracked); otherwise this is a standalone fix.

### DIFF
--- a/scripts/agent-status.sh
+++ b/scripts/agent-status.sh
@@ -30,6 +30,22 @@ AGENT_STATE_DIR="${LOBSTER_INSTALL_DIR:-$HOME/lobster}/.state"
 # Maximum agents to show in summary (keep messages concise)
 AGENT_MAX_DISPLAY=5
 
+# Derive the Claude Code tmp base for this user and workspace.
+# Claude Code stores session data under /tmp/claude-<uid>/<workspace-hash>/
+# where workspace-hash = workspace path with '/' replaced by '-'.
+# Task output files live in per-session UUID subdirs: <base>/<uuid>/tasks/*.output
+# If AGENT_TASKS_DIR is set, it overrides this (used by tests).
+_default_tasks_glob() {
+    if [ -n "${AGENT_TASKS_DIR:-}" ]; then
+        echo "${AGENT_TASKS_DIR}/*.output"
+        return
+    fi
+    local claude_tmp_base="/tmp/claude-$(id -u)"
+    local workspace_hash
+    workspace_hash=$(echo "${LOBSTER_WORKSPACE:-$HOME/lobster-workspace}" | sed 's|/|-|g')
+    echo "${claude_tmp_base}/${workspace_hash}/*/tasks/*.output"
+}
+
 # Format seconds into human-readable duration: 30s, 5m, 2h
 _format_duration() {
     local seconds="$1"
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Code review — agent-status.sh path fix

**Correctness of path derivation**

`_default_tasks_glob()` produces `/tmp/claude-<uid>/<hash>/*/tasks/*.output`. Verified on the running system:
- `id -u` = 1000 (correct)
- `echo "/home/lobster/lobster-workspace" | sed 's|/|-|g'` = `-home-lobster-lobster-workspace` (matches the actual directory at `/tmp/claude-1000/-home-lobster-lobster-workspace`)
- Session UUID subdirs confirmed present: `38d47dc2-...`, `75d54070-...`, etc.
- `*.output` files confirmed in `<uuid>/tasks/` subdirs

The `sed 's|/|-|g'` transformation preserves the leading `-` (the leading `/` becomes `-`), which correctly matches the actual directory name. The bug description proposed an additional `sed 's|^-||'` step that would have been wrong — this implementation omits it.

**AGENT_TASKS_DIR override**

The override path correctly points `_default_tasks_glob()` to `$AGENT_TASKS_DIR/*.output` (flat glob in the override dir), which matches the existing test contract where callers write `.output` files directly into the dir. No regression.

**compgen -G vs find**

Replacing `find -maxdepth 1 -name "*.output" -print0` with `compgen -G "$glob" | sort` is sound for this use case. `compgen -G` returns nothing (exit 1) when no matches, and the `2>/dev/null` suppresses the "no match" output. The `sort` provides stable ordering as a substitute for the previous alphabetical ordering from `find`. The mtime-based re-sort for display happens downstream in `scan_agent_status` via `ls -t`, which is unaffected.

**No other callers of the hardcoded path**

Confirmed `grep -r 'home-admin-lobster'` finds no other references in the repo. The fix is complete.

**Verdict: safe to merge.**

PR #530, opened 2026-03-17
fix(health-check): raise stale threshold to 4 minutes, update cron to */4

sayhar:
## What this fixes

The health-check was triggering false-positive restarts when a processing message aged past the 3-minute stale threshold (180s). A 223s processing delay was enough to fire a restart. Sahar reported this as a real incident — the dispatcher was mid-work on a legitimate subagent when it got killed.

The root cause is that 180s is simply too tight: any moderately slow tool call, context compaction window, or multi-step subagent run can cross it before the work is done. The check interval (\`*/2\`) compounds this — a message that crosses the threshold between two checks gets flagged immediately with no grace period.

## What changed

**\`scripts/health-check-v3.sh\`**
- \`STALE_THRESHOLD_SECONDS\`: 180 → 240 (3m → 4m). This is the hard RED trigger for a restart.
- \`YELLOW_THRESHOLD_SECONDS\`: 120 → 150 (2m → 2.5m). Proportionally adjusted warning level.
- Header comment updated to reflect the new cron schedule.

**\`install.sh\`**
- Health-check cron entry changed from \`*/2\` to \`*/4\` for fresh installs.

**\`scripts/upgrade.sh\`**
- Migration 13: the fallback cron entry (for installs missing it entirely) now uses \`*/4\`.
- Migration 14 (new): detects existing installs with \`*/2\` in their crontab and rewrites the entry to \`*/4\`. This runs automatically on \`upgrade.sh\`.

## What is not changed

The compaction suppression window (\`COMPACTION_SUPPRESS_SECONDS=300\`), WFM heartbeat threshold, outbox stale thresholds, and all restart escalation logic are untouched. This is a targeted change to the inbox-drain check only.

## Functional patterns

Pure configuration constants at the top of the script — no control flow changes. The migration in upgrade.sh is a guard-guarded idempotent shell block (checks current crontab state before rewriting), consistent with the existing migration pattern.

## Tests run

- [x] \`grep\` across all three files — values confirmed correct after edit
- [x] \`git diff --staged --stat\` — 3 files, 22 insertions, 9 deletions, as expected
- [x] Pre-push security scan — passed, no PII detected
- [ ] Live cron test — skipped: requires restarting the live Lobster instance; safe to skip since the change is purely to numeric constants and a crontab entry

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -698,7 +698,7 @@ check_disk() {
 #
 # Returns: 0=GREEN, 1=YELLOW (< 4h remaining), 2=RED (confirmed expired/near-expiry)
 AUTH_FAILURE_COUNTER_FILE="$WORKSPACE_DIR/logs/auth-token-failures"
-AUTH_CONSECUTIVE_RED_THRESHOLD=3  # Must fail this many consecutive 2-min checks (~6 min total)
+AUTH_CONSECUTIVE_RED_THRESHOLD=3  # Must fail this many consecutive 4-min checks (~12 min total)
 
 check_auth_token() {
     local creds_file="$HOME/.claude/.credentials.json"


sayhar (commented):
PASS

The change is minimal and targeted. Three files touched, all consistent with each other:

- `STALE_THRESHOLD_SECONDS` raised to 240 in health-check-v3.sh
- `YELLOW_THRESHOLD_SECONDS` proportionally raised to 150
- Cron interval updated to `*/4` in install.sh and in migration 13's fallback path
- Migration 14 added to upgrade.sh to rewrite existing crontabs from `*/2` to `*/4`

Migration 14 logic is correct: it checks whether the `# LOBSTER-HEALTH` marker line currently contains `*/2` before acting, making it idempotent. The rewrite pattern is consistent with how migrations 9–13 are structured.

One note: the `AUTH_CONSECUTIVE_RED_THRESHOLD` comment on line 701 of health-check-v3.sh references "2-min checks" (e.g. "3 consecutive 2-min checks (~6 min total)"). With the interval now 4 minutes, that parenthetical becomes inaccurate (~12 min total). Not a functional issue — the constant itself is correct — but worth a follow-up cleanup to avoid confusion.

Safe to merge.

sayhar (commented):
NEEDS-WORK: One stale comment remains; everything else looks correct.

## What looks good

**Threshold values** — confirmed in the diff:
- `STALE_THRESHOLD_SECONDS`: 180 → 240 (4 minutes)
- `YELLOW_THRESHOLD_SECONDS`: 120 → 150 (2.5 minutes)
- Both the inline comments and the top-of-file header comment updated correctly.

**install.sh** — cron entry, surrounding comment, and success message all updated to `*/4`. Internally consistent.

**Migration 14 idempotency** — the guard `grep "$HEALTH_MARKER" | grep -q "\*/2"` is correct: it only fires when the old interval is present, and will not re-run after the rewrite. Consistent with the existing migration pattern.

**Migration 13 fallback** — the unconditional fallback (for installs missing the cron entry entirely) is also updated to `*/4`, so new-from-scratch upgrade paths are covered.

## Issue: stale comment on upgrade.sh line 701

The engineer's briefing calls this out explicitly: a comment still reads "2-min checks (~6 min total)" which is now inaccurate. The comment does not appear in the diff, meaning it was not updated in this PR. Should read something like "4-min checks (~12 min total)".

This is a documentation-accuracy issue, not a correctness or safety issue, but it will confuse whoever next reads the migration block. Please update it before merge.

## Summary

Logic is sound. The single required fix is the stale comment on upgrade.sh line 701. Once that is updated, this is ready to merge.

sayhar (commented):
PASS: stale comment fixed. Ready to merge.

PR #532, opened 2026-03-17
feat(health-check): 90s boot grace period to suppress false-positive restarts

### LINKED ISSUE #531: feat: boot grace period for health-check to suppress false-positive alerts after restart
## Problem

After a restart, the health-check currently alerts/restarts immediately if it detects stale messages or missing processes. This causes false-positive alerts during the ~60-90s it takes a fresh session to initialize.

Concretely: when `do_restart()` calls `systemctl stop` + `systemctl start`, the next health-check run (which may fire within 2 minutes) can see:
- The `active` state with a fresh timestamp but no Claude process yet (still starting)
- Messages in the inbox that pre-date the restart (they look stale but are actually waiting for the fresh session to boot)
- A missing heartbeat file (not yet written by the new session)

This triggers a second restart 2 minutes after the first, causing restart loops even when the underlying system is healthy.

## Design

Use a **boot timestamp** written to `~/messages/config/lobster-state.json` (the existing state file) as a `booted_at` field. The health-check reads this field: if the boot was < `BOOT_GRACE_SECONDS` (90s) ago, skip the stale inbox, WFM freshness, and process checks — just log and exit GREEN.

### What writes the boot timestamp

1. **`do_restart()` in health-check-v3.sh** — after a successful `systemctl start`, write `booted_at` to the state file. This covers health-check-initiated restarts.
2. **`claude-persistent.sh`** — at `write_state("starting", ...)`, include a `booted_at` timestamp. This covers manual starts and systemd-initiated starts.

### Grace period scope

During the boot grace window, suppress:
- Stale inbox check (messages may predate the restart)
- WFM freshness check (heartbeat not yet written)
- Process/tmux checks that would RED immediately

Still run: service check, auth token, outbox drain, memory, disk.

### State file format (additive, backward compatible)

```json
{
  "mode": "active",
  "detail": "attempt=1",
  "updated_at": "2026-03-17T12:00:00+00:00",
  "pid": 12345,
  "booted_at": "2026-03-17T12:00:00+00:00"
}
```

## Implementation Plan

- [ ] Add `BOOT_GRACE_SECONDS=90` constant to health-check-v3.sh
- [ ] Add `is_boot_grace_period()` function to health-check-v3.sh (reads `booted_at` from state file, same pattern as `is_compaction_recent()`)
- [ ] In `do_restart()`: after successful `systemctl start`, write `booted_at` to state file
- [ ] In `claude-persistent.sh` `write_state()`: include `booted_at` on first `starting` write (or a dedicated boot-stamp function)
- [ ] In `main()` of health-check-v3.sh: check `is_boot_grace_period()` early, skip stale inbox / WFM / process checks during grace window (log clearly why)
- [ ] Add migration to `upgrade.sh` (no-op idempotent step — ensures `messages/config/` dir exists; no new files to create)
- [ ] Tests: verify grace period suppresses correct checks without suppressing service/auth checks

sayhar:
## Summary

After a restart, the health-check was reliably triggering a second restart 2 minutes later. The new session needs ~60-90s to initialize and begin draining the inbox, but the health-check saw stale messages, a missing Claude process, or a stale WFM heartbeat and fired anyway — causing cascading restart loops on otherwise healthy systems.

The fix is a 90-second boot grace period. When \`booted_at\` in \`lobster-state.json\` is less than 90s old, the health-check skips inbox drain, WFM freshness, and process/tmux checks. Service availability, auth token, outbox drain, memory, and disk checks still run — they're fast and not affected by the boot state.

## What changed

**\`scripts/health-check-v3.sh\`**
- Adds \`BOOT_GRACE_SECONDS=90\` constant (parallel to \`COMPACTION_SUPPRESS_SECONDS\`)
- Adds \`is_boot_grace_period()\`: reads \`booted_at\` from \`lobster-state.json\`, same pattern as \`is_compaction_recent()\`
- Adds \`write_boot_timestamp()\`: merges \`booted_at\` into the state file without clobbering \`compacted_at\` or other fields; called in \`do_restart()\` after a successful \`systemctl start\`
- Threads \`boot_grace=true\` through all lifecycle case branches and the WFM freshness check, suppressing process/inbox/WFM checks during the grace window

**\`scripts/claude-persistent.sh\`**
- Adds \`write_boot_stamp()\`: same merge strategy as \`write_boot_timestamp()\`, called from \`main()\` before the Telegram alert — ensures the timestamp is in place before the first health-check could fire
- Covers manual starts, systemd starts, and any start not initiated by the health-check

**\`scripts/upgrade.sh\`**
- Migration 14: ensures \`messages/config/\` exists (idempotent; directory is created by Migration 4 on normal installs, but this guards old/manual provisioned installs)

**\`tests/test-health-check-boot-grace.sh\`** (new)
- 8 unit tests covering: grace window active, grace window expired, missing \`booted_at\`, missing state file, \`write_boot_timestamp()\` merge correctness, field preservation, no-op on absent state file, \`write_boot_stamp()\` from claude-persistent.sh

## Functional patterns used

Pure functions (\`is_boot_grace_period\`, \`is_compaction_recent\` — same shape), immutable state reads (Python one-liners read-merge-write without in-place mutation), composition over duplication (both writers share the same JSON merge pattern).

## Areas to review closely

- The grace period suppresses process and inbox checks for all lifecycle states (\`active\`, \`hibernate\`, \`starting\`, \`active-debug\`, \`backoff\`). This means a genuinely broken restart (e.g. systemd failed silently) would go undetected for 90s. Acceptable tradeoff given the alternative is a guaranteed false-positive restart loop.
- \`write_boot_timestamp()\` in the health-check is called only on the success path of \`do_restart()\`. If \`systemctl start\` succeeds but the session immediately dies, the grace period still fires but \`booted_at\` is accurate — the next health-check after grace expiry will catch it cleanly.
- The merge strategy for \`booted_at\` uses a Python one-liner. If the state file is malformed JSON, the merge silently no-ops (\`|| true\`). This is the same error-tolerant pattern used in \`is_compaction_recent()\`.

## Tests run

- [x] \`bash tests/test-health-check-boot-grace.sh\` — 8 passed, 0 failed
- [x] \`bash tests/test-health-check-inbox-drain.sh\` — 19 passed, 0 failed
- [x] \`bash tests/test-health-check-pid-timing.sh\` — 9 passed, 0 failed
- [x] \`bash -n scripts/health-check-v3.sh && bash -n scripts/claude-persistent.sh && bash -n scripts/upgrade.sh\` — all syntax clean
- [ ] Live restart test against running lobster-claude service — not run; requires production restart (safe to skip: no behavior change during normal operation, grace period only activates post-restart)

Closes #531

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1170,14 +1170,15 @@ run_migrations() {
         migrated=$((migrated + 1))
     fi
 
-    # Migration 15: Ensure messages/config/ directory exists for boot-grace state file
-    # The boot grace period feature stores booted_at in lobster-state.json, which lives
-    # in messages/config/. This directory is created by Migration 4 on new installs, but
-    # this step ensures it exists on any install that skipped Migration 4 (e.g. manually
-    # provisioned or very old installs where the directory may have been removed).
+    # Migration 15: Ensure messages/config/ directory exists for lobster-state.json
+    # lobster-state.json lives in messages/config/ and is used by multiple features
+    # (compaction suppression, boot grace period). This directory is created by
+    # Migration 4 on new installs, but this step ensures it exists on any install
+    # that skipped Migration 4 (e.g. manually provisioned or very old installs
+    # where the directory may have been removed).
     if [ ! -d "$MESSAGES_DIR/config" ]; then
         mkdir -p "$MESSAGES_DIR/config"
-        substep "Created $MESSAGES_DIR/config/ (required for boot-grace state file)"
+        substep "Created $MESSAGES_DIR/config/ (required for lobster-state.json)"
         migrated=$((migrated + 1))
     fi
 
--- a/tests/test-health-check-boot-grace.sh
+++ b/tests/test-health-check-boot-grace.sh
@@ -100,8 +100,22 @@ source_health_check_functions() {
     export LOBSTER_STATE_FILE_OVERRIDE="$TEST_STATE_FILE"
     export LOBSTER_HEALTH_LOCK="$TEST_TMPDIR/health.lock"
 
-    # Source just the functions by extracting everything up to main()
-    eval "$(sed -n '1,/^main()/p' "$HEALTH_SCRIPT" | grep -v '^main()')" 2>/dev/null
+    # Set the configuration constants needed by the extracted functions.
+    # Assign directly from test paths rather than re-evaluating the config block,
+    # so that overrides are always in effect regardless of script structure.
+    BOOT_GRACE_SECONDS=90
+    COMPACTION_SUPPRESS_SECONDS=300
+    LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$TEST_STATE_FILE}"
+    LOG_FILE="$TEST_LOG"
+
+    # Extract specific named functions rather than everything up to main(), so
+    # the sourcing does not break if main() is renamed or the file is restructured.
+    eval "$(sed -n '/^log()/,/^}/p' "$HEALTH_SCRIPT")" 2>/dev/null
+    log_info()  { log "INFO"  "$1"; }
+    log_warn()  { log "WARN"  "$1"; }
+    log_error() { log "ERROR" "$1"; }
+    eval "$(sed -n '/^is_boot_grace_period()/,/^}/p' "$HEALTH_SCRIPT")" 2>/dev/null
+    eval "$(sed -n '/^write_boot_timestamp()/,/^}/p' "$HEALTH_SCRIPT")" 2>/dev/null
 }
 
 source_health_check_functions


sayhar (commented):
PASS: Boot grace period is correctly implemented and well-scoped.

## Evaluation

**1. Does it correctly prevent false-positive health checks after a fresh start?**

Yes. The booted_at timestamp is written in two places that cover all start paths:
- claude-persistent.sh main() writes it before the Telegram alert fires — well before the first health-check cycle could run.
- do_restart() in health-check-v3.sh writes it after a successful systemctl start on the success path.

All five lifecycle case branches (hibernate, starting|restarting|waking, backoff, active-debug, active|unknown|*) are gated by boot_grace=true, as is the WFM freshness check. Coverage is complete.

**2. Is 90s a reasonable grace period?**

Yes. The PR identifies the initialization window as ~60-90s, so 90s fits tightly. The one mild tension: the pre-existing active branch warns (YELLOW) only when wrapper is running but no Claude for >120s. This leaves a 30s window (90-120s post-grace) where a missing Claude process would not escalate past YELLOW. That is acceptable — 120s was already a soft threshold, not a hard SLA. Bumping grace to 120s would let a genuinely broken restart go undetected longer. 90s is the right call.

**3. Edge cases that could mask a real failure?**

Two, both acknowledged in the PR and acceptable:

- If systemctl start succeeds but the session immediately dies, booted_at is stamped correctly, grace runs 90s, and then the next health-check catches the failure cleanly. The 90s delay is the minimum unavoidable cost of the tradeoff.
- If the state file is malformed JSON, is_boot_grace_period() silently returns false (no grace window). Conservative and safe — better to risk a false-positive restart than to suppress checks indefinitely.

Manual starts without a prior state file are covered by the fallback in write_boot_stamp() that writes a minimal JSON from scratch.

**4. Interaction with existing health-check-v3.sh logic?**

The boot_grace flag is a clean parallel to the existing compaction_recent pattern. Both are boolean flags set at the top of main(), threaded through case branches identically. The ordering in the WFM block (transient-state check then boot_grace then compaction_recent) is correct. Resource, auth, and service checks are unaffected — only inbox, process/tmux, and WFM freshness are suppressed, which matches the stated intent.

The 8 unit tests cover the key invariants (grace window active, expired, missing field, missing file, merge correctness, field preservation). Coverage is appropriate for the risk. Ready to merge.

PR #534, opened 2026-03-17
revert: remove ghostwriting rule from ONBOARDING.md

sayhar:
The ghostwriting rule added in #521 was documented as a system default in ONBOARDING.md, but it isn't one. The rule lives in `user.base.bootup.md` — private userland config that is not committed to the repo and not present for new users out of the box. Calling it a "default behavior" in onboarding docs is factually false: a new install won't have it.

This removes the two lines added by #521 (the rule paragraph and the blank line below it). No other changes.

## Tests run
- [x] `git diff` verified — exactly 2 lines removed, nothing else touched


sayhar (commented):
PASS — simple revert. Removes 2 lines (rule + blank line) that incorrectly documented a user-config-only rule as a system default. Safe to merge.

PR #548, opened 2026-03-17
fix: skip Syncthing prompt when running non-interactively

sayhar:
upgrade.sh was hanging indefinitely in cron and other unattended contexts because of a bare \`read\` prompt at line 674 asking whether to install Syncthing. When stdin has no TTY — as is always the case in cron, systemd timers, or CI — the \`read\` call blocks forever.

## What changed

A \`[ -t 0 ]\` (stdin is a terminal) guard is inserted immediately before the prompt, inside \`setup_syncthing()\`. If no TTY is detected, the function prints an informative note and returns early, leaving the rest of the upgrade unaffected. Interactive runs see no change in behavior.

The \`--skip-syncthing\` flag and the early-exit for already-running Syncthing (lines 660–667) remain the canonical fast-paths; the new guard is a safety net for the case where neither applies and the script is running unattended.

No other interactive prompts were found: all other \`apt-get install\` calls in the file already use \`-y\`, and the \`read\` at line 674 is the only bare \`read\` call in the script (the others at lines 352, 515, 520, 530, 534 are all \`read -r line\` in \`while read\` loops over piped input, not prompts).

## Functional pattern

Pure guard: the new block has no side effects of its own — it reads an environment fact (TTY presence) and returns early without touching any state. The rest of the function is unchanged.

## Tests run

- [x] \`bash -n scripts/upgrade.sh\` — syntax check, no errors
- [x] Manual trace: \`echo "" | bash -c '[ -t 0 ] && echo interactive || echo non-interactive'\` returns \`non-interactive\`, confirming the guard fires correctly under piped/cron stdin
- [ ] Full end-to-end upgrade run — not executed; requires a fresh Lobster install and sudo access not available in this environment. No behavior change outside the Syncthing prompt path.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -666,6 +666,12 @@ setup_syncthing() {
         fi
     fi
 
+    # Skip prompt if not running interactively (cron, unattended upgrades, etc.)
+    if ! [ -t 0 ]; then
+        info "Non-interactive environment detected — skipping Syncthing setup (run with --skip-syncthing to suppress this message, or install manually)"
+        return 0
+    fi
+
     # Prompt - this is the only interactive part
     echo ""
     echo -e "${YELLOW}${BOLD}LobsterDrop${NC} uses Syncthing to sync files between your phone/laptop and this server."


sayhar (commented):
PASS (minor nit)

Reviewed against the six scrutiny areas in the brief.

**1. TTY check placement — correct**

The guard sits after the `--skip-syncthing`, `--dry-run`, and already-running fast-paths (lines 649–667). That ordering is right: the cheap, stateless early-exits run first; the TTY check only fires when we have reached the interactive prompt and have no other reason to skip. Guarding the entire function would prevent the "already installed and running" success message from printing in non-interactive runs, which would be a regression in cron visibility. Current placement is the minimum footprint needed.

**2. Idiom correctness — acceptable with one edge case**

`[ -t 0 ]` is the standard POSIX test for "stdin is a terminal" and is correct for the cron/systemd/CI case described. One gap: `echo y | upgrade.sh` leaves stdin a pipe (not a TTY), so the guard fires and Syncthing is skipped even though the caller clearly intended a "yes" answer. This is a pre-existing limitation of the overall design (the prompt was never piped-input safe) and not introduced by this PR. If we wanted to close it we could additionally honour a `LOBSTER_NONINTERACTIVE=` or `DEBIAN_FRONTEND=noninteractive` env var, but that is out of scope here.

**3. --skip-syncthing interaction — fine as-is**

Setting `SKIP_SYNCTHING=true` automatically in the non-interactive path would couple two concepts (CLI option state vs. runtime environment detection) and could confuse logging. The current approach — guard early-returns independently, `--skip-syncthing` still works as explicit opt-out — is cleaner. The "run with `--skip-syncthing` to suppress this message" hint in the log is a nice touch.

**4. Message verbosity — one concern**

```
info "Non-interactive environment detected — skipping Syncthing setup (run with --skip-syncthing to suppress this message, or install manually)"
```

This fires on every cron/systemd invocation of `upgrade.sh` for any user who does not have Syncthing installed yet. After the first run the already-running fast-path (line 661) handles it silently, but until Syncthing is installed this line will repeat every upgrade. For a weekly cron this is probably fine; for a nightly or short-interval timer it could become noisy. Consider gating on `$VERBOSE` or downgrading to a one-time note. Not a blocker, but worth a follow-up issue if cron frequency increases.

**5. Other read usages — confirmed safe**

Lines 352, 515, 520, 530, 534 are all `while read -r line` / `while read -r entry` draining piped input (git log, changelog grep). None are interactive prompts. The PR description is accurate.

**6. Interactive path unchanged — confirmed**

When `[ -t 0 ]` is true the new block is skipped entirely; execution falls through to the existing echo/read/response logic unchanged. No functional difference for interactive users.

**Summary:** The fix is correct, minimal, and well-placed. The verbosity nit (point 4) is the only item worth tracking. Approve to merge.

sayhar (commented):
PASS: The fix is correct and minimal.

**[ -t 0 ] guard accuracy**: `[ -t 0 ]` tests whether stdin (fd 0) is connected to a terminal. This correctly identifies cron, systemd timers, piped input, and CI environments as non-interactive. It is the idiomatic POSIX approach for this check.

**Scope**: The guard is placed immediately before the bare `read` prompt (line ~668 in the patched file), after all the existing fast-paths (`--skip-syncthing`, `DRY_RUN`, already-running Syncthing check). It returns early with `return 0` which exits only the `setup_syncthing()` function, leaving the rest of upgrade.sh unaffected. No other install steps are skipped.

**Edge cases**:
- An interactive user who redirects stdin (e.g. `bash upgrade.sh < /dev/null`) would have Syncthing silently skipped. This is acceptable and expected behaviour for stdin redirection.
- SSH sessions without a PTY (`ssh host 'bash upgrade.sh'`) would also skip silently. This is consistent with how non-interactive shells behave and is noted implicitly in the info message.
- The info message is printed before returning, so unattended runs have a visible log trail explaining why Syncthing was skipped.
- The `--skip-syncthing` flag and already-running check remain the canonical fast-paths. The new guard is purely additive as a safety net.

**Containment**: The diff is 6 lines in a single function, no state mutations, no changes to function signatures or call sites. Clean.

No issues found. Approved.

PR #442, opened 2026-03-16
rename bootup files: sys.* for repo, user.* for user-config

sayhar:
## Problem

Two files were both named `dispatcher.bootup.md` — one in `.claude/` (repo) and one in `~/lobster-user-config/agents/` (private). Same for `subagent.bootup.md`. The names gave no indication of which was system-level vs user-level.

## What changed

**System files** in `.claude/` (repo): renamed with `sys.*` prefix so their repo ownership is clear.
- `dispatcher.bootup.md` → `sys.dispatcher.md`
- `subagent.bootup.md` → `sys.subagent.md`

**User-land file references** updated throughout the repo to the new `user.*` naming convention (the actual files live in `~/lobster-user-config/agents/` which is private):
- `base.bootup.md` → `user.base.md`
- `base.context.md` → `user.base.context.md`
- `dispatcher.bootup.md` → `user.dispatcher.md`
- `subagent.bootup.md` → `user.subagent.md`

The `sys.*` prefix signals "this is a repo/system file". The `user.*` prefix signals "this is private user config". Ownership is now readable from the filename alone.

## Files changed

- `.claude/dispatcher.bootup.md` → `.claude/sys.dispatcher.md` (git mv)
- `.claude/subagent.bootup.md` → `.claude/sys.subagent.md` (git mv)
- `CLAUDE.md` — updated all references (system file names + user-config path names in Role-Specific Context, Dispatcher-only tools callout, Skill system callout, Key Directories section)
- `.claude/sys.dispatcher.md` — updated user-config path references
- `.claude/sys.subagent.md` — updated user-config path references
- `docs/DEVELOPMENT.md` — updated Related documentation link
- `ONBOARDING.md` — updated Settings and Customization table
- `hooks/on-compact.py` — updated REMINDER_TEXT path for dispatcher bootup file
- `hooks/post-compact-gate.py` — updated DENY_REASON_NEEDS_TOKEN path
- `install.sh` — updated stub file creation (two locations) to new names
- `scripts/upgrade.sh` — updated Migration 10 and 11 to new names; added **Migration 12** that idempotently renames existing installs (checks old name exists and new name absent before renaming)
- `scripts/migrate-to-user-config.sh` — updated all target names and comments

## What is NOT changed

The actual `~/lobster-user-config/` files are private and not in this repo. Migration 12 in `upgrade.sh` handles renaming them on existing installs. No changes to hook registration, settings.json paths, or service files — those don't reference bootup filenames.

## Tests run

[...]

PR #553, opened 2026-03-17
rename: complete sys.*/user.* rename — drop .bootup. from filenames

sayhar:
## What this fixes

The previous rename PR (issue 318, merged) added the `sys.` and `user.` prefixes but left `.bootup.` in the middle of each filename. The resulting names (`sys.dispatcher.bootup.md`, `user.base.bootup.md`, etc.) still carry redundant noise — ownership is now readable from the prefix, so `.bootup.` adds nothing.

This PR completes the rename to the clean form described in issue #442.

## Changes

**System files in `.claude/` (git mv):**
- `sys.dispatcher.bootup.md` → `sys.dispatcher.md`
- `sys.subagent.bootup.md` → `sys.subagent.md`

**User-config path references updated in all repo files:**
- `user.base.bootup.md` → `user.base.md`
- `user.dispatcher.bootup.md` → `user.dispatcher.md`
- `user.subagent.bootup.md` → `user.subagent.md`
- (`user.base.context.md` is unchanged — no `.bootup.` to drop)

**Files changed:** `CLAUDE.md`, `.claude/sys.dispatcher.md`, `.claude/sys.subagent.md`, `ONBOARDING.md`, `docs/DEVELOPMENT.md`, `hooks/on-compact.py`, `hooks/post-compact-gate.py`, `install.sh`, `scripts/migrate-to-user-config.sh`, `docker/staging/entrypoint-staging.sh`, `scripts/upgrade.sh`.

**Migration 15** in `upgrade.sh` handles existing installs idempotently: renames `sys.*.bootup.md` → `sys.*.md` in workspace `.claude/` and `user.*.bootup.md` → `user.*.md` in `lobster-user-config/agents/`. Checks old name exists and new name absent before acting; also creates stubs at the new names for any that are missing.

## What is NOT changed

- `sys.debug.bootup.md` — the debug injection file keeps its name (it was never part of this rename series; its name reflects the debug function, not the bootup concept)
- Hook registration, `settings.json`, or service files — none reference bootup filenames directly
- Migrations 10–14 — left intact as historical record of earlier rename steps; Migration 15 builds on them

## Functional design

All path strings are literals — no runtime logic changed. Correctness is verifiable by inspection of the diff.

## Tests run

- [x] `git diff --stat HEAD` — 11 files changed, renames confirmed correct, no stray old names in non-migration code
- [ ] Full test suite — not run; this is a pure rename with no logic changes. All paths are string literals in markdown/shell.

Closes #442

🤖🦞 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/sys.dispatcher.md
+++ b/.claude/sys.dispatcher.md
@@ -125,9 +125,9 @@ Pass this to `register_agent` as `output_file`. It enables future liveness detec
 ---
 
 **After reading the sections above**, also check for and read user context files if they exist:
-- `~/lobster-user-config/agents/user.base.bootup.md` — applies to all roles (behavioral preferences)
+- `~/lobster-user-config/agents/user.base.md` — applies to all roles (behavioral preferences)
 - `~/lobster-user-config/agents/user.base.context.md` — applies to all roles (personal facts)
-- `~/lobster-user-config/agents/user.dispatcher.bootup.md` — dispatcher-specific user overrides
+- `~/lobster-user-config/agents/user.dispatcher.md` — dispatcher-specific user overrides
 
 These files are private and not in the git repo. They extend and override the defaults here.
 
--- a/.claude/sys.subagent.md
+++ b/.claude/sys.subagent.md
@@ -25,9 +25,9 @@ Do NOT call `wait_for_messages` — that is only for the main loop.
 ---
 
 **After reading this file**, also check for and read user context files if they exist:
-- `~/lobster-user-config/agents/user.base.bootup.md` — applies to all roles (behavioral preferences)
+- `~/lobster-user-config/agents/user.base.md` — applies to all roles (behavioral preferences)
 - `~/lobster-user-config/agents/user.base.context.md` — applies to all roles (personal facts)
-- `~/lobster-user-config/agents/user.subagent.bootup.md` — subagent-specific user overrides
+- `~/lobster-user-config/agents/user.subagent.md` — subagent-specific user overrides
 
 These files are private and not in the git repo. They extend and override the defaults here.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,14 +9,14 @@ You are **Lobster**, an always-on AI assistant that never exits. You run in a pe
 This file provides shared context. Depending on your role, read the appropriate supplement:
 
 **System context** (always read):
-- **If you are the dispatcher (main loop):** read `.claude/sys.dispatcher.bootup.md` — it covers the main loop pseudocode, the 7-second rule, the dispatcher pattern, handling subagent results, message source handling (Telegram/Slack), self-check reminders, message flow diagram, startup behavior, hibernation, context recovery, Google Calendar handling, and voice/brain-dump routing.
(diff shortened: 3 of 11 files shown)


sayhar (commented):
PASS: rename is complete and correct.

Checked:
- `git mv` confirms both `.claude/sys.dispatcher.md` and `.claude/sys.subagent.md` are the renamed files (99%/98% similarity detected by git)
- All path strings in hooks, install.sh, upgrade.sh, docs, and CLAUDE.md updated to new names
- No stray `sys.dispatcher.bootup` or `user.*.bootup` references remain outside of upgrade.sh migration history (which correctly uses old names as source-check targets)
- Migration 15 is idempotent: guards with `[ -f old ] && [ ! -s new ]` before each mv
- `sys.debug.bootup.md` correctly left unchanged — it was never part of this rename series
- The `.bak-pre-pr520` file is untracked and not included in the commit (harmless leftover)

One thing to verify manually on the live install: after merge, run `lobster update` and confirm Migration 15 fires and renames `~/lobster-user-config/agents/user.base.bootup.md` → `user.base.md` etc. The logic is straightforward shell but cannot be tested in a dry-run environment here.

sayhar (commented):
PASS: All 11 files consistently use the new names; migration logic is correct; sys.debug.bootup.md exclusion is intentional and fine.

## Migration 15 idempotency

The guards are correct. Each mv is gated on [ -f <old> ] && [ ! -s <new> ]. The -s (non-zero size) test — not -f (exists) — is the right choice: it allows the rename to proceed if the new-name file is an empty stub but blocks it if the new-name file already has real user content. This prevents clobbering populated files on re-runs.

The stub-creation fallback at the end (the for stub_file in user.base.md ... loop) is also correct. It covers the case where an install skipped Migration 11 entirely or where Migration 11's old-name stubs were already renamed — either way the new-name stubs end up present.

## Migration 11 interaction

The added comment is clear and accurate:

  stub names here still use the intermediate user.*.bootup.md form intentionally — Migration 15 (below) will rename them to the final user.*.md form on existing installs.

This correctly explains the intentional two-step: Mig 11 creates user.base.bootup.md, then Mig 15 renames it to user.base.md. The empty-stub guard in Mig 15 (!-s) means the rename fires on those empty stubs without risk of data loss.

## Reference consistency

All non-migration code in the changed files uses the new names exclusively:
- hooks/on-compact.py — sys.dispatcher.md
- hooks/post-compact-gate.py — sys.dispatcher.md
- install.sh (both stubs loops) — new names
- CLAUDE.md (all 8 references) — new names
- ONBOARDING.md, docs/DEVELOPMENT.md, docker/staging/entrypoint-staging.sh, scripts/migrate-to-user-config.sh — all updated

Old names appear only inside the historical migration blocks (10, 11, and 15's source operands), which is correct.

## sys.debug.bootup.md

Correctly left unchanged. The file reflects a debug function, not the bootup concept, and is consistent with the PR description. No references to it appear in the changed files.

Migration 15 uses ws_claude_dir which is declared as a local in the Migration 10 block. Both blocks live inside run_migrations(), so the variable is in scope — no bug, just worth noting for future readers that Mig 15 implicitly depends on that local being set earlier in the same function.

sayhar:
Closing per owner decision: .bootup. naming convention is intentional and should be kept.

PR #554, opened 2026-03-17
feat(debug): Telegram alerts for memory ops and subagent messages when LOBSTER_DEBUG=true

sayhar:
## What problem this solves

When debugging Lobster, there was no real-time visibility into two important classes of events: memory reads/writes happening across subagents and subagent result messages arriving in the dispatcher inbox. You had to grep logs after the fact to understand what was happening.

This PR makes both classes of events visible as live Telegram pushes when \`LOBSTER_DEBUG=true\`.

## What changed

**Feature 1 — Memory operation alerts** (\`handle_memory_store\`, \`handle_memory_search\`)

After each call completes, a direct Telegram push is sent to the admin chat:

- \`memory_store\`: shows agent identity, event type, and first 80 chars of content
- \`memory_search\`: shows agent identity, query text, and result count

Both tools now accept an optional \`task_id\` parameter so subagents can identify themselves in the alert. When absent, the alert labels the caller as "dispatcher".

**Feature 2 — Subagent→dispatcher message alerts** (\`handle_write_result\`)

Immediately after the inbox file is written, a direct Telegram push shows:
- resolved message type (\`subagent_result\` / \`subagent_error\` / \`subagent_notification\`)
- task_id, status, sent_reply_to_user flag, and first 60 chars of text

This fires at the MCP layer before the dispatcher picks up the message, giving real-time visibility independent of dispatcher processing latency.

**What is not changed**: \`write_observation\` already had debug mirroring (PR #351). \`memory_recent\` is read-only housekeeping and was intentionally excluded. No production behavior is affected — all new code is gated on \`_DEBUG_MODE\` and calls through \`_emit_debug_observation\`, which is a strict no-op when debug mode is off.

## Functional patterns

Pure functions for message formatting (no mutation of handler state), side effects isolated at the boundary after the primary operation succeeds, best-effort / fire-and-forget via the existing \`_emit_debug_observation\` infrastructure.

## Tests run

- [x] \`uv run pytest tests/unit/test_mcp_server/test_debug_alerts.py -v\` — 34 passed, 0 failed
- [x] \`uv run pytest tests/unit/test_mcp_server/ --tb=no -q\` — 231 passed, 16 failed (all 16 failures are pre-existing on \`main\`, confirmed independently)
- [ ] Live Telegram test — blocked: requires production restart with LOBSTER_DEBUG=true; no behavior change in non-debug mode so safe to merge without it

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4501,26 +4501,25 @@ async def handle_write_result(args: dict) -> list[TextContent]:
     # Debug alert: enqueue best-effort inbox message when LOBSTER_DEBUG=true.
     # Fires at the MCP layer (before the dispatcher picks up the inbox message)
     # so the user sees the subagent message arrive in real time.
-    _resolve_debug_config()
-    if _DEBUG_MODE:
-        agent_id = args.get("agent_id", "").strip() or None
-        text_preview = text[:60] + "…" if len(text) > 60 else text
-        alert_lines = [
-            f"\U0001f4e8 [subagent\u2192dispatcher] type: {msg_type}",
-            f"task: {task_id}",
-        ]
-        if agent_id:
-            alert_lines.append(f"agent: {agent_id}")
-        if status:
-            alert_lines.append(f"status: {status}")
-        alert_lines.append(f"sent_reply: {bool(sent_reply_to_user)}")
-        alert_lines.append(f"preview: {text_preview}")
-        _emit_debug_observation(
-            "\n".join(alert_lines),
-            category="system_context",
-            visibility="mcp-only",
-            emitter=f"task:{task_id}",
-        )
+    # _emit_debug_observation is a no-op when debug alerts are disabled — single gate.
+    agent_id = args.get("agent_id", "").strip() or None
+    text_preview = text[:60] + "…" if len(text) > 60 else text
+    alert_lines = [
+        f"\U0001f4e8 [subagent\u2192dispatcher] type: {msg_type}",
+        f"task: {task_id}",
+    ]
+    if agent_id:
+        alert_lines.append(f"agent: {agent_id}")
+    if status:
+        alert_lines.append(f"status: {status}")
+    alert_lines.append(f"sent_reply: {bool(sent_reply_to_user)}")
+    alert_lines.append(f"preview: {text_preview}")
+    _emit_debug_observation(
+        "\n".join(alert_lines),
+        category="system_context",
+        visibility="mcp-only",
+        emitter=f"task:{task_id}",
+    )
 
     log.info(f"Subagent result queued in inbox: task_id={task_id} status={status} chat_id={chat_id}")
     if msg_type == "subagent_notification":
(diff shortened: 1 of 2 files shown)


sayhar (commented):
PASS: The hardcoded Telegram removal is clean and the inbox-write approach is architecturally correct.

**Source detection logic** — resolution order (explicit env override → Slack if enabled → Telegram) is clear and well-documented. The `LOBSTER_DEBUG_SOURCE` escape hatch is useful for testing.

**`_DEBUG_ALERTS_ENABLED` gate** — correct and conservative: both a valid chat_id AND the source's bot credentials must be present before alerts are enabled. Prevents inbox pollution in staging/CI. One minor inconsistency: `handle_write_result` and `handle_memory_store`/`handle_memory_search` all gate on `_DEBUG_MODE` before calling `_emit_debug_observation`, while `_emit_debug_observation` itself re-checks `_DEBUG_ALERTS_ENABLED`. The double-gate is redundant but harmless — the inner check in `_emit_debug_observation` is the authoritative safety net.

**Inbox write approach** — `atomic_write_json` to INBOX_DIR is the right pattern. No direct API calls, no event loop threading complexity, works across all sources.

**conftest.py isolation** — replacing the session-scoped `block_outbound_http` urllib patch with a function-scoped `isolate_debug_config` that redirects `_CONFIG_DIR` to `tmp_path` is cleaner and more targeted. Resetting all lazy-resolution globals per test prevents cross-test state leakage. The scope change from session → function is intentional and necessary (tmp_path is function-scoped).

**`debug_observation` message type — not a blocker.** The dispatcher's main loop handles unrecognized message types by falling through to `mark_processed` (no relay, no subagent). A `debug_observation` arriving in the inbox will be read and discarded silently. The only risk is at very high debug volume: each `debug_observation` triggers another `_emit_debug_observation` (write_observation hook at ~line 4546), which creates another `debug_observation` — potential feedback loop. However this is only active when `LOBSTER_DEBUG=true` AND `handle_write_observation` is called for the debug message itself. Since `debug_observation` messages go directly to `atomic_write_json` (bypassing `handle_write_observation`), there is no recursive trigger. Not a blocker.

**One thing to verify:** the `write_result` debug alert fires unconditionally when `_DEBUG_MODE=True` even before `_resolve_debug_config()` has set `_DEBUG_ALERTS_ENABLED`. The call is `_resolve_debug_config()` then `if _DEBUG_MODE:` then `_emit_debug_observation(...)`. Since `_emit_debug_observation` checks `_DEBUG_ALERTS_ENABLED` internally and returns early if false, this is safe. But the pattern diverges from the fast-path in `_emit_debug_observation` (which checks `_DEBUG_RESOLVED and not _DEBUG_ALERTS_ENABLED`). Not a blocker — just a note for consistency.

Overall: clean removal of a direct API call, good source-detection design, solid test coverage.

sayhar (commented):
🤖 Lobster (reviewer): [Correction] Prior review comments on this PR were posted without Lobster attribution. They were generated by the Lobster reviewer subagent, not written by the PR author. Both PASS verdicts above are from Lobster (reviewer).

[Clarification on duplicate reviews] There are four review comments on this PR, all from Lobster (reviewer):
1. Review #3957968183 (03:26 UTC) — First full review on the original implementation. PASS with minor notes.
2. Review #3957969545 (03:27 UTC) — Duplicate of the above, posted 31 seconds later due to a subagent re-run. Same verdict. **Superseded by #1 — disregard.**
3. Review #3957997281 (03:40 UTC) — Brief PASS after hardcoded telegram source was removed.
4. Review #3958001986 (03:43 UTC) — Full review of the revised implementation (hardcode fix). PASS with minor notes.

**Current authoritative review: Review #4 (03:43 UTC).** The PR is safe to merge per that review.

sayhar (commented):
Lobster (reviewer): PASS — both minor notes resolved: debug block isolated in own try/except, redundant double-gate removed. Ready to merge.

PR #558, opened 2026-03-17
fix: upgrade sqlite-vec to >=0.1.7a1 to fix aarch64 ELFCLASS32 (broken semantic memory)

sayhar:
## What's broken and why

On aarch64 (64-bit ARM) systems, `sqlite-vec 0.1.6` fails to load its `vec0` SQLite extension with:

```
sqlite3.OperationalError: vec0.so: wrong ELF class: ELFCLASS32
```

The root cause is a packaging bug upstream: the `manylinux_2_17_aarch64` wheel for `0.1.6` bundles a **32-bit ARM** `.so` binary instead of a 64-bit aarch64 one. The wheel label says aarch64; the ELF header says ARM32. Since `VectorMemory._init_db` catches all exceptions and falls back to FTS keyword search, the failure is silent — the system starts normally, but all memory retrieval is pure keyword search instead of the intended 70/30 hybrid.

`sqlite-vec >=0.1.7a1` ships the correct 64-bit binary. The latest is `0.1.7a10`.

## What this changes

- **`pyproject.toml`**: raises the lower bound from `>=0.1.1` to `>=0.1.7a1`. This is the constraint that drives `uv lock`.
- **`uv.lock`**: resolved to `0.1.7a10` (correct 64-bit aarch64 wheel confirmed before committing).
- **`install.sh`**: simplifies the sqlite-vec block to install `>=0.1.7a1` directly. Removes the old stable-then-`0.1.7a2`-hardcoded-alpha-then-source-build cascade — that cascade existed because we didn't know which alpha was safe; now we do.
- **`scripts/upgrade.sh` Migration 16**: detects a broken sqlite-vec load and reinstalls to `>=0.1.7a1` in-place. Existing installs that ran `uv sync` before this fix will be remediated on next `upgrade.sh` run without a full reinstall.

## What this does not change

- `vector_memory.py` is untouched — the code was correct; only the installed binary was wrong.
- No schema migration needed — the `events_vec` table was created correctly if sqlite-vec ever loaded (which it doesn't on aarch64 with 0.1.6). On aarch64 the table will be created on first successful load after this fix.
- Non-aarch64 platforms (x86_64, macOS) are unaffected — their 0.1.6 wheels were fine.

## Functional patterns used

Pure verification: the upgrade migration is a pure detect-and-fix step with no side effects on passing systems.

## Tests run

- [x] `uv run python3 -c "import sqlite3, sqlite_vec; c=sqlite3.connect(':memory:'); c.enable_load_extension(True); sqlite_vec.load(c); c.execute('CREATE VIRTUAL TABLE t USING vec0(embedding float[4])')"` in the worktree venv — succeeded, vec_version = v0.1.7-alpha.10
- [x] `uv run pytest tests/unit/ -q -k "memory or vector or sqlite"` — 59 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 1073 passed, 67 failed (pre-existing failures unrelated to this change, confirmed by running on main before the fix)


sayhar (commented):
PASS: The fix is correct and the approach is sound. Detailed notes on the four scrutiny areas:

**1. upgrade.sh Migration 16 — `pip` + `uv pip` fallback**

The `pip`-first pattern is intentional and appropriate: `uv` may not be on PATH in all upgrade environments (system installs, CI). The `|| true` makes a failed reinstall non-fatal, which is correct since vector search is a degraded-but-optional feature. One minor observation: if both install commands fail, `migrated` is not incremented, so the final message will read "No migrations needed" even on a system where sqlite-vec was already broken and remains broken. The `warn` line handles user communication, but the migration counter reporting is slightly misleading. Not a blocker — the behavior is safe.

**2. install.sh — three-stage fallback collapsed to single `>=0.1.7a1`**

Nothing is lost. The source-build fallback is still present (the `SQLITE_VEC_OK=false` branch that invokes `_SQLITE_VEC_SRC_DIR`). The comment in the diff ("Will attempt to compile from source") confirms the path is live. Verification loop preserved. Message strings updated correctly.

**3. pyproject.toml — pre-release specifier as minimum**

`>=0.1.7a1` is the right call given no stable 0.1.7 exists. Per PEP 440, pip/uv will prefer the stable release automatically when 0.1.7 stable ships — the specifier does not permanently pin users to alphas. The resolved lock file (0.1.7a10) includes correct 64-bit aarch64 wheels (size 147614 vs. the broken 103704 in 0.1.6). The aarch64 wheel size increase alone is good circumstantial evidence the correct binary is bundled.

**4. Other call sites**

Both `pyproject.toml` and `uv.lock` are updated. No other version pins for sqlite-vec appear in the diff. The lock file hashes are fresh PyPI entries from 2026-02-13, consistent with 0.1.7a10 release metadata.

**Summary:** Fix is complete and the layered install.sh / upgrade.sh / pyproject.toml coverage is solid. The migration-counter edge case is cosmetic and does not affect correctness or user experience.

sayhar (commented):
## Counter fix review — PASS

Reviewed commit fe4cc80.

**What changed:** `migrated=$((migrated + 1))` was moved from inside the `if` success branch (line ~1199 on main) to immediately after the reinstall attempt (line 1195 on branch), before the outcome check.

**Verification (scripts/upgrade.sh lines 1189–1203):**

```bash
if ! "$VENV_DIR/bin/python" -c "..." 2>/dev/null; then
    substep "sqlite-vec fails to load — reinstalling..."
    "$VENV_DIR/bin/pip" install --quiet "sqlite-vec>=0.1.7a1" 2>/dev/null || \
        uv pip install --quiet "sqlite-vec>=0.1.7a1" 2>/dev/null || true
    migrated=$((migrated + 1))      # ← now here, unconditional
    if "$VENV_DIR/bin/python" -c "..." 2>/dev/null; then
        success "sqlite-vec reinstalled and loads correctly..."
    else
        warn "sqlite-vec reinstall failed — semantic memory search will be unavailable"
    fi
fi
```

**Logic is correct.** The counter now increments whenever the migration body runs (broken load detected), regardless of whether the reinstall succeeds or fails. The original bug — where a double failure left `migrated=0` and printed "No migrations needed" despite having run — is fixed. The warn path on failure is preserved so the user still gets actionable feedback.

No other changes were made to the counter logic. PASS.

sayhar (commented):
PASS

Reviewed commit 74f294c.

The `migrated` counter increment is now correctly placed inside the success branch of the post-reinstall load check. Previously it was incremented unconditionally after the pip install call (before the load check), causing the summary to report "1 migration(s) applied" even when the reinstall failed. With this fix, "applied" accurately means sqlite_vec was reinstalled AND verified to load correctly.

The diff is clean and matches the stated intent exactly. No issues found.

sayhar (commented):
PASS — with one minor observation.

## What I checked

**1. Migration 18 — sequencing, idempotency, tooling**

- Correctly placed after migration 17 in upgrade.sh. Migrations 16 and 17 are confirmed on main; migration 18 follows immediately before the `if [ "$migrated" -eq 0 ]` closing block.
- Idempotent: the load-test guard (`if ! python -c "...sqlite_vec.load(c)"`) means the reinstall only runs when the extension is broken. Re-running upgrade.sh on a healthy install is a no-op.
- Uses `uv pip install` throughout — correct, no bare pip.

**2. Smoke tests**

Both tests hit the real extension with no mocking:
- `test_sqlite_vec_loads`: calls `sqlite_vec.load(db)` directly — will raise if the .so is the wrong arch.
- `test_sqlite_vec_basic_operation`: creates a `vec0` virtual table, inserts a row, and round-trips a nearest-neighbour query. Exercises the compiled extension code, not just the Python wrapper.
API usage matches the sqlite-vec 0.1.x interface.

**3. install.sh**

install.sh does not enumerate migration numbers — it only calls `run_migrations`. The changes here correctly update the sqlite-vec install block to use `>=0.1.7a1` (removing the now-unnecessary two-phase stable→alpha fallback). No explicit migration list to update.

**4. Version constraint and cross-arch**

`>=0.1.7a1` resolves to `0.1.7a10` in uv.lock. The lock file lists wheels for both `manylinux_2_17_aarch64` and `manylinux_2_17_x86_64`, so x86_64 installs are unaffected. The aarch64 wheel size jumped from 103 KB to 148 KB between 0.1.6 and 0.1.7a10 — consistent with going from a mis-bundled 32-bit .so to a proper 64-bit one.

**5. One minor nit (non-blocking)**

There is a pre-existing duplicate "Migration 14" label in upgrade.sh (lines ~1181 and ~1262 on main — two separate blocks both labeled Migration 14). Not introduced by this PR, but worth a follow-up issue to renumber to avoid confusion when reading the migration sequence.

---

No issues introduced by this PR. Migration, tests, install.sh, and version constraint are all correct.

sayhar (commented):
## Re-review: PASS

Migration renumbering is clean. Full sequence 0–20 verified, no gaps or duplicates.

**Migration sequence (upgrade.sh):** 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20 — contiguous, no holes.

**Migration 20 (sqlite-vec):** Follows the same pattern as adjacent migrations — substep message, conditional fix, migrated++ on success, warn on failure. Style is consistent.

**install.sh:** No migration number references (only a single prose comment about migrations conceptually). No update needed.

**Smoke tests:** `tests/test_sqlite_vec.py` is present with two tests — extension load and a basic vec0 virtual table round-trip. Good coverage for the fix.

**No regressions introduced:** The renumbering is a label-only change; no logic was altered. The old out-of-order labels (duplicate 14, skipped 15) are now corrected. `pyproject.toml` and `uv.lock` consistently pin `sqlite-vec>=0.1.7a1` (resolves to 0.1.7a10). `install.sh` fallback path updated to match.

Ready to merge.

PR #562, opened 2026-03-17
fix(health-check): explicitly kill old Claude PID when ExecStop fails silently

### LINKED ISSUE #561: fix: health-check do_restart() does not explicitly kill old Claude PID when ExecStop fails silently
## Bug

\`do_restart()\` in \`scripts/health-check-v3.sh\` relies on \`systemctl stop\` + the service's \`ExecStop\` directive (which runs \`tmux kill-session\`) to terminate the old Claude process before starting a new one.

When the tmux socket is stale or the session no longer exists, \`tmux kill-session\` exits non-zero but systemd swallows the failure — \`ExecStop\` completes as far as systemd is concerned, the service transitions to \`inactive\`, and \`systemctl stop\` returns 0. The old Claude PID is never killed.

The new session then starts, and now two dispatcher processes are running simultaneously, both processing messages from the inbox. This causes duplicate replies, missed \`mark_processed\` calls, and generally corrupted message-handling state.

## Root cause

The function captures \`pre_restart_pid\` before stopping the service, but never verifies the process is actually dead after \`systemctl stop\` returns. It trusts \`ExecStop\` unconditionally.

## Fix

After the \`systemctl stop\` block in \`do_restart()\`, explicitly check whether \`pre_restart_pid\` is still alive and kill it if so:

\`\`\`bash
if [[ -n "$pre_restart_pid" ]] && kill -0 "$pre_restart_pid" 2>/dev/null; then
    log_warn "ExecStop did not kill PID $pre_restart_pid — sending SIGTERM"
    kill "$pre_restart_pid" 2>/dev/null || true
    sleep 2
    kill -0 "$pre_restart_pid" 2>/dev/null && kill -9 "$pre_restart_pid" 2>/dev/null || true
fi
\`\`\`

This goes between the \`systemctl stop\` block and the stale-socket cleanup section in \`do_restart()\`.

## Impact

Without this fix, any restart attempted while the tmux socket is stale results in two concurrent dispatchers, which is a silent correctness failure — health checks pass, the service appears healthy, but message handling is broken.

sayhar:
## Problem

A health-check-initiated restart can leave two dispatcher processes running concurrently. This happens when the tmux socket is stale at restart time: `ExecStop` runs `tmux kill-session`, which exits non-zero because the socket is gone, but systemd treats `ExecStop` failures as non-fatal — `systemctl stop` returns 0 and the service transitions to `inactive` without ever killing the old Claude PID. `ExecStart` then brings up a new session, and both dispatchers compete on the inbox.

The result is a silent correctness failure: the health check considers the restart successful, the service looks healthy, but duplicate message processing and missed `mark_processed` calls corrupt inbox state.

## Fix

After the `systemctl stop` call in `do_restart()`, check whether `pre_restart_pid` is still alive using `kill -0`. If it is, send SIGTERM and wait 2 seconds; if it still hasn't exited, send SIGKILL. In the normal path where `ExecStop` succeeds, this block is a no-op — `kill -0` fails immediately and the block is skipped.

The change is 13 lines, entirely within `do_restart()`, between the `systemctl stop` block and the existing stale-socket cleanup section. No other logic is touched.

Functional pattern: pure defensive check — the kill block reads no shared state and has no side effects in the happy path.

Closes #561

## Tests run

- [ ] Live restart with stale tmux socket — blocked: requires a running Lobster instance to reproduce; the code path is straightforward enough that the logic can be verified by inspection. No automated test harness exists for this script.
- [x] `bash -n scripts/health-check-v3.sh` — syntax check passes, no errors

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -985,6 +985,19 @@ Manual intervention required:
         log_info "systemctl stop: $line"
     done
 
+    # Explicitly kill the old Claude PID if it is still alive after systemctl stop.
+    # When the tmux socket is stale, ExecStop ("tmux kill-session") exits non-zero
+    # but systemd swallows the failure — the service transitions to inactive and
+    # systemctl stop returns 0, but the old process is never killed. Without this
+    # guard, the new session starts and two dispatchers run concurrently, causing
+    # duplicate message processing and corrupted inbox state.
+    if [[ -n "$pre_restart_pid" ]] && kill -0 "$pre_restart_pid" 2>/dev/null; then
+        log_warn "ExecStop did not kill PID $pre_restart_pid — sending SIGTERM"
+        kill "$pre_restart_pid" 2>/dev/null || true
+        sleep 2
+        kill -0 "$pre_restart_pid" 2>/dev/null && kill -9 "$pre_restart_pid" 2>/dev/null || true
+    fi
+
     # Clean up the socket only if stop left it behind.
     # A successful ExecStop removes it via tmux kill-session; if stop failed or
     # the socket was already stale, we clean it up here so ExecStart can bind.


sayhar (commented):
PASS: Change is minimal, correctly placed, and the logic is sound.

The kill block sits between the `systemctl stop` completion and the stale-socket cleanup — exactly the right position. In the normal path (`ExecStop` succeeded), `kill -0 $pre_restart_pid` fails immediately (process gone) and the entire block is skipped with zero side effects. In the failure path (stale socket, process alive), SIGTERM is sent first, a 2-second grace period is given, then SIGKILL if needed — the standard two-stage kill sequence.

One thing to note: `pre_restart_pid` is set via `pgrep -x "claude" | head -1`. If there happen to be multiple `claude` processes before restart (which is itself a problem), this only kills one. That edge case is pre-existing and out of scope for this fix.

The comment added above the kill block clearly explains the "why" — stale socket → ExecStop silent failure → duplicate dispatchers. Inline comments are sufficient; no additional documentation needed.

`bash -n` syntax check: clean.

sayhar (commented):
PASS: Logic is correct and the PR description explains the system flow well. Notes below.

**1. Placement — correct.** The kill block lands after the `systemctl stop` pipe drains and before socket cleanup and `systemctl start`. That ordering is right: kill before the socket is removed (so you're not racing on a half-cleaned state) and kill before start (so the new session doesn't share the inbox with the old one).

**2. `kill -0` — correct and idiomatic.** Standard POSIX way to probe liveness without delivering a signal. The `2>/dev/null` suppresses the error on the normal path when the PID is already gone. No issues here.

**3. SIGTERM → 2s → SIGKILL escalation — acceptable with a caveat.** 2 seconds is short for a graceful Claude dispatcher shutdown (it may be mid-tool-call). But since this is the abnormal path where ExecStop already silently failed, the primary goal is correctness (no two dispatchers), not graceful teardown. The tradeoff is reasonable. If graceful shutdown ever matters here, the sleep could be bumped to 5s, but it's not blocking.

**4. Empty `pre_restart_pid` — handled correctly.** The `[[ -n "$pre_restart_pid" ]]` guard short-circuits the block entirely. Safe.

**5. Multiple Claude processes — pre-existing gap, not fully addressed.** `pgrep -x "claude" | head -1` captures only the first PID. If a prior restart failure left two Claude processes running before this restart attempt, only one gets killed here — the second orphan survives and the two-dispatcher problem persists for this cycle. This is a pre-existing limitation of the single-PID capture, not introduced by this PR, but worth a comment in the code (or PR body) acknowledging that this guard handles the common single-process case. A more robust alternative would be to iterate all PIDs from `pgrep -x "claude"`, but that's scope-creep for this fix.

**6. Minor: `|| true` placement on the `kill -9` line.** `kill -0 "$pre_restart_pid" 2>/dev/null && kill -9 "$pre_restart_pid" 2>/dev/null || true` — the `|| true` absorbs failures from both the `kill -0` check and `kill -9`. If `kill -9` fails due to a permissions error, it's silently swallowed. Acceptable for a defensive guard, but a log line on failure would be more observable (e.g. `|| log_warn "kill -9 PID $pre_restart_pid failed"`).

**Summary:** The core fix is correct — placement, guard logic, and escalation pattern are all sound. The multi-PID gap is a pre-existing issue; a brief note acknowledging the single-PID assumption would improve self-documentation. No blocking issues.

sayhar (commented):
## Design Review: MODIFY

### What the PR actually does

`pre_restart_pid` is captured via `pgrep -x "claude" | head -1` at line 974, then after `systemctl stop` the block checks `kill -0 $pre_restart_pid` and sends SIGTERM/SIGKILL to that specific PID. The proposed replacement swaps this for `systemctl kill --signal=SIGTERM/SIGKILL lobster-claude`.

---

### Fatal flaw in the proposed `systemctl kill` approach

The service is `Type=forking`. After `systemctl stop` runs — even when `ExecStop` fails silently — systemd transitions the unit to `inactive` state and releases the cgroup accounting for that unit. **`systemctl kill` on an `inactive` unit is a documented no-op: it returns 0 and delivers no signal to any process.** This is confirmed in `systemctl(1)`: "If the unit is not running, this command is ignored."

So the proposed sequence `stop → systemctl kill SIGTERM → sleep 2 → systemctl kill SIGKILL → start` would silently do nothing in exactly the failure scenario it aims to fix (ExecStop fails, old Claude still running, service now inactive). The orphaned PID survives and two dispatchers race the inbox — same bug as before.

---

### Is the cgroup argument correct in principle?

Yes — if the service were still `active` or `activating`, `systemctl kill` would be strictly superior to PID-targeting because it reaches every process in the cgroup (tmux server, expect, claude) without any PID lookup. The problem is that `systemctl stop` is called first, and after that call returns the cgroup is gone from systemd's perspective.

---

### Skipping `stop` entirely

One clean alternative: replace `systemctl stop` with `systemctl kill --signal=SIGTERM`, sleep, `systemctl kill --signal=SIGKILL`, then `systemctl reset-failed`, then `systemctl start`. This keeps the unit in a state where `systemctl kill` is valid throughout and eliminates the ExecStop-fails-silently problem entirely. The stale-socket cleanup logic would still be needed.

---

### Assessment of the existing PR

The existing PR's approach (kill the captured `pre_restart_pid` after `stop`) is structurally sound for the majority case. The `pgrep -x "claude" | head -1` PID ambiguity concern is real but limited: it's already captured before `stop` is called, so it's a snapshot of the running dispatcher process. The risk is a stale PID from a debug session that happens to be running at that exact moment. This is a real but low-probability edge case.

**MODIFY verdict**: Either (a) keep the PR's PID-targeting approach but add a comment acknowledging the `pgrep` ambiguity and note that the `pre_restart_pid` check is a defensive fallback, not a primary kill path — or (b) restructure to skip `systemctl stop` and use `systemctl kill` throughout while the unit is still active. Do not replace the kill block with `systemctl kill` after `systemctl stop` has already run — that is a no-op.

PR #569, opened 2026-03-17
fix: use PID file in do_restart() to avoid ambiguous pgrep

### LINKED ISSUE #561: fix: health-check do_restart() does not explicitly kill old Claude PID when ExecStop fails silently

sayhar:
## Problem

Health-check-initiated restarts can leave two dispatcher processes running concurrently. The failure path: when the tmux socket is stale at restart time, ExecStop runs `tmux kill-session`, which exits non-zero because the socket is gone, but systemd treats ExecStop failures as non-fatal — `systemctl stop` returns 0 and the service transitions to `inactive` without killing the old Claude process. ExecStart then brings up a new session, and both dispatchers compete on the inbox. The result is duplicate message processing and corrupted inbox state.

PR #562 addressed this by adding a kill block after `systemctl stop`, but used `pgrep -x "claude" | head -1` to identify the old PID. That is ambiguous: it can match an unrelated Claude process (a debug session, a subagent). The PID file approach in this PR is unambiguous.

## How it works

**`claude-persistent.sh`**: Before each claude launch, write the dispatcher PID to `~/messages/config/dispatcher.pid`. Implementation: a bash subshell writes `$BASHPID` then `exec claude`. After exec, the running process has the same PID that was written. (`$$` gives the parent shell PID even inside a subshell — bash behavior — so `$BASHPID` is required here.) File is removed on normal exit via trap.

**`health-check-v3.sh`** (`do_restart()`):
- Replace `pgrep -x "claude" | head -1` with a read from `dispatcher.pid` (with integer sanity check).
- After `systemctl stop`, if that specific PID is still alive (`kill -0`): send SIGTERM, sleep 2, escalate to SIGKILL if still running. Clean up the PID file.
- Post-restart PID verification also reads from the PID file, so both sides of the "did the PID change?" check use the same source.

## Tradeoffs & constraints

- **First-restart edge case**: if the PID file does not exist yet (first ever restart, or file was lost), `pre_restart_pid` is empty and the post-restart PID-changed check is skipped. The service/tmux health checks still verify the restart succeeded. Logged as a warning.
- **Stale PID file**: if the process dies abnormally without running the trap, the PID file may linger with a dead PID. `kill -0` on a dead PID returns an error, so the kill block is a no-op — safe. File is overwritten on next launch.
- **Assumption**: `$BASHPID` is a bash-specific variable (not POSIX sh). `claude-persistent.sh` must be run with bash, not sh. The shebang is `#!/bin/bash`.
- **Not live-tested**: the stale-tmux-socket path requires a running Lobster instance to test. Syntax checks pass; logic matches the described failure mode.

Nothing outside `do_restart()` and `claude-persistent.sh` is touched.

Closes #561

### DIFF
--- a/scripts/claude-persistent.sh
+++ b/scripts/claude-persistent.sh
@@ -387,11 +387,31 @@ launch_claude() {
     # Once we hand off to claude, we're active.
     write_state "active" "claude running, attempt=$attempt"
 
-    claude --dangerously-skip-permissions \
-        --model sonnet \
-        --max-turns 150 \
-        -p "$init_prompt" \
-        2>&1 | tee -a "$LOG_DIR/claude-session.log" || claude_exit_code=$?
+    # Write the dispatcher PID file so the health check can target this specific
+    # process for cleanup without relying on ambiguous pgrep matches.
+    #
+    # We use a subshell that writes $BASHPID (the actual subshell PID) and then
+    # exec-replaces itself with claude. After exec, the running claude process
+    # has the same PID that was written to the file. $$ would give the *parent*
+    # shell's PID in bash, so $BASHPID is required here.
+    local dispatcher_pid_file="$MESSAGES_DIR/config/dispatcher.pid"
+    mkdir -p "$(dirname "$dispatcher_pid_file")"
+
+    (
+        echo "$BASHPID" > "$dispatcher_pid_file"
+        exec claude --dangerously-skip-permissions \
+            --model sonnet \
+            --max-turns 150 \
+            -p "$init_prompt"
+    ) 2>&1 | tee -a "$LOG_DIR/claude-session.log" || claude_exit_code=$?
+
+    # Clean up PID file on exit — the process is gone, the file is stale.
+    # Note: if claude-persistent.sh's parent is SIGKILLed, this cleanup won't run,
+    # leaving a stale PID file. On next launch, the health check reads it, finds the
+    # PID dead via kill -0, treats any kill as no-op, and self-heals when the new PID
+    # overwrites the file at launch start.
+    rm -f "$dispatcher_pid_file" 2>/dev/null || true
+    log "Claude exited (exit_code=$claude_exit_code), PID file removed"
 
     return $claude_exit_code
 }
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -61,6 +61,7 @@ WORKSPACE_DIR="${LOBSTER_WORKSPACE:-$HOME/lobster-workspace}"
 INBOX_DIR="$MESSAGES_DIR/inbox"
 MAINTENANCE_FLAG="$MESSAGES_DIR/config/lobster-maintenance"
 LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$MESSAGES_DIR/config/lobster-state.json}"
+DISPATCHER_PID_FILE="$MESSAGES_DIR/config/dispatcher.pid"
 STALE_THRESHOLD_SECONDS=240          # 4 minutes - RED if any message older (watchdog handles soft recovery at 90s)
 YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW warning
 
@@ -982,9 +983,27 @@ Manual intervention required:
     record_restart
 
     # Capture the Claude PID before stopping, so we can verify a new process
-    # is running after restart (not just the same surviving process).
-    local pre_restart_pid
-    pre_restart_pid=$(pgrep -x "claude" 2>/dev/null | head -1)
+    # is running after restart (not just the same surviving process), and so we
+    # can kill it directly if ExecStop fails silently.
+    #
+    # Read from the PID file written by claude-persistent.sh at launch time.
+    # This is unambiguous: pgrep -x "claude" could match an unrelated Claude
+    # process (e.g. a debug session), but the PID file records exactly the
+    # dispatcher that this health check is responsible for.
+    local pre_restart_pid=""
+    if [[ -f "$DISPATCHER_PID_FILE" ]]; then
+        pre_restart_pid=$(< "$DISPATCHER_PID_FILE")
+        # Sanity check: must be a non-empty integer
+        if [[ ! "$pre_restart_pid" =~ ^[0-9]+$ ]]; then
+            log_warn "dispatcher.pid contains non-numeric value '$pre_restart_pid' — ignoring"
+            pre_restart_pid=""
+        fi
+    fi
+    if [[ -n "$pre_restart_pid" ]]; then
+        log_info "Pre-restart dispatcher PID: $pre_restart_pid (from $DISPATCHER_PID_FILE)"
+    else
+        log_warn "No dispatcher.pid found — pre-restart PID unknown (new install or first restart)"
+    fi
 
     local tmux_uid
     tmux_uid=$(id -u)
@@ -1013,6 +1032,10 @@ Manual intervention required:
     # systemctl stop + tmux kill-server, kill it explicitly. SIGTERM first,
     # then SIGKILL after a brief wait. This handles cases where the process
     # detached from the tmux session before the kill-server ran.
+    #
+    # Unlike pgrep, targeting the PID file is unambiguous: it records exactly the
+    # dispatcher process launched by claude-persistent.sh, not any other Claude
+    # process that may be running (e.g. a debug session or subagent).
     if [[ -n "$pre_restart_pid" ]] && kill -0 "$pre_restart_pid" 2>/dev/null; then
         log_warn "Claude PID $pre_restart_pid survived systemctl stop + tmux kill-server — sending SIGTERM"
         kill -TERM "$pre_restart_pid" 2>/dev/null || true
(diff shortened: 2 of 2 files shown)


sayhar (commented):
PASS

**What this replaces:** PR #562 added the correct fix (kill block after ExecStop fails silently) but used `pgrep -x "claude" | head -1` to find the PID. That's ambiguous on any system where multiple Claude processes might be running (debug session, subagent, another user). This PR replaces it with a PID file written at launcher startup.

**PID file correctness:** The `$BASHPID` trick is correct. In bash, `$$` in a subshell still expands to the parent process PID. `$BASHPID` gives the actual subshell PID. After `exec claude`, the subshell is replaced — the running `claude` binary has the same OS PID that `$BASHPID` captured. Writing the file before `exec` (not after) is also correct: there's no window where claude is running but the file doesn't exist yet.

**Kill block logic:** The SIGTERM → sleep 2 → SIGKILL fallback is the standard double-tap pattern. Using `kill -0` for liveness checks (not `pgrep`) is correct — it checks the specific PID, not the process name.

**Post-restart verification:** Switching to PID file read here is correct and consistent. The retry loop handles the case where the new session hasn't written its PID file yet. One subtle point: if `pre_restart_pid` is empty (no PID file on first restart), the condition `[[ -z "$pre_restart_pid" || ... ]]` short-circuits to `true` immediately — so the verification loop exits on the first attempt without waiting for the new PID file. This is acceptable: if we had no pre-restart PID, we can't meaningfully compare, and the service/tmux checks still run.

**Cleanup:** PID file is removed after the kill block (not after `systemctl stop` returns 0) — correct, because we want it removed regardless of whether ExecStop succeeded or we had to kill manually.

**Not changed:** All other pgrep calls in the file (e.g. in `kill_orphaned_claude_processes`) are intentional name-based searches and are out of scope. This PR is narrowly scoped to `do_restart()`.

sayhar (commented):
## Review: PID file restart logic

**Verdict: PASS** with two minor notes worth considering.

### What this gets right

The core fix is sound. The `$BASHPID` / `exec` pattern in `claude-persistent.sh` is correct: the subshell writes its own PID, then exec-replaces itself with claude, so the running process retains that PID. Using `$$` would write the parent shell PID instead — the comment explains this clearly and the implementation is right.

The SIGTERM → sleep 2 → SIGKILL escalation sequence in `health-check-v3.sh` is well-structured. Targeting the specific PID from the file rather than `pgrep -x "claude" | head -1` eliminates the ambiguity problem (matching unrelated Claude processes) that motivated this PR.

Integer sanity check on the PID file read is a good defensive touch.

### Minor notes

**1. Pipe subshell — the written PID is not the process claude sees itself as.**

```bash
(
    echo "$BASHPID" > "$dispatcher_pid_file"
    exec claude ...
) 2>&1 | tee -a "$LOG_DIR/claude-session.log" || claude_exit_code=$?
```

When a subshell is part of a pipeline, bash forks a new child process for `tee` and a separate child for the subshell. The subshell's `$BASHPID` is the PID of that forked subprocess — and after `exec claude`, claude runs as that PID. So the PID file does contain the correct PID for the running claude process. The pattern works. But it is subtly reliant on bash's execution model for pipeline subshells; a comment noting "the subshell is a pipeline child, so $BASHPID is the actual forked PID that claude inherits" would help future readers.

**2. Empty `pre_restart_pid` short-circuits the post-restart PID check.**

```bash
if [[ -z "$pre_restart_pid" || ( -n "$post_restart_pid" && "$post_restart_pid" != "$pre_restart_pid" ) ]]; then
    break
fi
```

If `pre_restart_pid` is empty (first restart, no PID file yet), this condition is immediately true — the loop breaks at `pid_check_attempts=0` without waiting for the new PID file to appear. The PR description acknowledges this is acceptable because service/tmux checks still run. That's correct, but it means on first-ever restart the PID-changed check provides no signal at all. The log warn ("first restart or new install") is helpful context for operators. Acceptable tradeoff, just worth being aware of.

### PID file cleanup placement

`rm -f "$DISPATCHER_PID_FILE"` runs unconditionally after the kill block, which is correct — whether the process was already dead or we just killed it, the file is stale and should go. No issue here.

### Summary

Both changes are correct. The `$BASHPID` + exec approach solves the PID capture problem cleanly. The two notes above are informational, not blocking — neither represents a logic error or a regression risk.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR replaces an ambiguous `pgrep` match with a PID file for identifying the dispatcher Claude process during health-check restarts.

**What the code does:**
1. `scripts/claude-persistent.sh`: In `launch_claude()`, wraps the `claude` invocation in a subshell that writes `$BASHPID` to `$MESSAGES_DIR/config/dispatcher.pid` before exec-replacing itself with claude. After claude exits, the PID file is cleaned up. The comment explaining why `$BASHPID` is used instead of `$$` is excellent and necessary.

2. `scripts/health-check-v3.sh`: Reads the PID from `dispatcher.pid` instead of using `pgrep -x "claude" | head -1`. Adds:
   - Pre-restart: reads and validates the PID, then after `systemctl stop`, explicitly kills the old process with SIGTERM then SIGKILL if it survives (guards against silent ExecStop failures with stale tmux sockets).
   - Post-restart: polls the PID file for the new PID instead of polling `pgrep`. The comparison is now file-based (consistent source of truth).

**Why it matters:** `pgrep -x "claude" | head -1` is ambiguous — it could match an unrelated Claude process (a debug session, a user's local instance). The PID file records exactly the dispatcher process that `claude-persistent.sh` is responsible for. The SIGTERM/SIGKILL guard closes a real gap where ExecStop exits 0 but fails to kill the process.

**Observations:**
- The `$BASHPID` explanation is a valuable comment — this is a subtle bash scoping behavior that would otherwise be opaque.
- The PID file cleanup (`rm -f`) after exit is correct: health checks that run while claude is down won't find a stale PID file and mistakenly believe the old process is still running.
- The post-restart empty-PID short-circuit logic (skip comparison if `pre_restart_pid` is empty) is correct for first-ever restarts before the PID file has been written.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: use PID file in do_restart() to avoid ambiguous pgrep**

The approach is correct and the implementation is sound.

**Correctness of `$BASHPID` usage:**
The subshell writes `$BASHPID` then `exec claude`. In bash, `$$` inside a subshell evaluates to the *parent shell's* PID, not the subshell PID. `$BASHPID` correctly captures the subshell PID, and after `exec` the running claude process inherits that PID. This is the right tool for the job.

**Edge cases handled:**
- Non-numeric PID file contents are rejected (sanity check present)
- Missing PID file on first restart is handled gracefully with a warning, and the post-restart PID-changed check is skipped (not a false failure)
- Stale PID file (abnormal termination without trap): `kill -0` returns error, kill block is no-op, self-heals on next launch. Documented in comments.
- Cleanup of the PID file after `do_restart()` is explicit and runs before the post-restart verification loop

**Post-restart PID verification:**
The updated post-restart loop correctly reads from the same PID file source (not pgrep), waits for the file to appear, and requires it to be non-empty and numeric before comparing. The break condition `[[ -z "$pre_restart_pid" || ( -n "$post_restart_pid" && "$post_restart_pid" != "$pre_restart_pid" ) ]]` is correct: if pre-restart PID was unknown, treat any non-empty post-restart PID as success.

**One minor note:**
The PID file is cleaned up in `do_restart()` (line `rm -f "$DISPATCHER_PID_FILE"`), but the post-restart loop reads from the same file. Since the cleanup runs *before* `systemctl start`, and the new process writes the file after startup, the window where the file is absent is expected and handled (the retry loop covers it). The flow is correct.

**PR description accuracy:**
The description accurately describes the implementation. The assumption about `#!/bin/bash` is valid — the existing shebang confirms this.

No blocking concerns.

PR #577, opened 2026-03-17
fix(inbox): make write_result idempotent per task_id (CC 2.1.77 flood fix)

sayhar:
🤖🦞 Lobster (engineer):

## Problem

CC 2.1.77 introduced a regression in the SubagentStop hook: even when the hook returns `{"suppressOutput": true}`, the CC runtime still injects a "Stop hook feedback: [hook]: No stderr output" message into the subagent's conversation. Subagents that misread this as evidence that `write_result` failed call it again. The worst observed case: one subagent called `write_result` 19 times for the same task_id, flooding the dispatcher inbox with duplicate entries.

The existing send_reply deduplication (`TASK_REPLIED_DIR`, `SENT_REPLIES_DIR`) prevents duplicate *delivery* to the user, but it does not prevent duplicate inbox entries from being written. Each duplicate call still creates a new JSON file in `~/messages/inbox/`, and the dispatcher has to process all of them.

## Fix

**Primary fix (`src/mcp/inbox_server.py`):** `handle_write_result` is now idempotent per `task_id`. The first call records a marker in a new `WRITE_RESULT_DEDUP_DIR` directory (same filesystem-marker pattern as `TASK_REPLIED_DIR`). Subsequent calls with the same `task_id` return early with an explicit rejection message that names the CC platform noise so the model can recognize it and stop retrying. First write wins; markers expire after 1 hour.

The rejection response reads: *"write_result already recorded for task_id=... Your result was already delivered. This duplicate call was ignored. If you are seeing 'Stop hook feedback: No stderr output', that is CC platform noise — do NOT call write_result again."*

**Defense-in-depth (`.claude/sys.subagent.bootup.md`):** Added an explicit instruction explaining that the "Stop hook feedback: [hook]: No stderr output" message after a successful `write_result` is CC platform noise — not an indication that the result failed — and that calling `write_result` again will be rejected.

## What is not changed

- The existing send_reply → write_result relay deduplication (`_was_task_replied`, `_was_sent_directly`) is unchanged.
- The `TASK_REPLIED_DIR` mechanism (suppressing duplicate relay when subagent already sent via `send_reply`) is unchanged.
- The dispatcher's inbox processing logic is unchanged; the fix is entirely in the MCP layer.

## Tests

New file: `tests/unit/test_mcp_server/test_write_result_dedup.py`

The tests cover:
- First call succeeds and writes exactly one inbox file
- Second call with the same task_id is rejected with a message mentioning CC platform noise
- 19-duplicate-call scenario produces exactly 1 inbox file
- Different task_ids are independent (each writes its own file)
- Expired dedup marker allows retry (verifies the 1-hour expiry window works)
- Helper function unit tests (`_record_write_result`, `_was_write_result_called`)

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_write_result_dedup.py -v` — 8 passed, 0 failed
- [x] `uv run pytest tests/unit/test_mcp_server/ -v` — 239 passed, 16 failed. All 16 failures are pre-existing on `main` (cover `get_whisper_model`, `mark_processing`, `mark_processed`, auto-threading, user model dispatch in `test_write_observation`) — none introduced by this change.
- [ ] `sudo docker compose -f tests/docker/docker-compose.test.yml up unit-tests` — Docker image rebuild blocked (whisper.cpp compile step requires network access); ran tests directly via `uv run pytest` instead. The existing cached image (873 passed, pre-change) does not include the new test file. New tests confirmed passing via direct `uv run` above.

### DIFF
--- a/.claude/sys.subagent.bootup.md
+++ b/.claude/sys.subagent.bootup.md
@@ -90,6 +90,14 @@ Failing to pass `sent_reply_to_user=True` causes duplicate messages — the disp
 
 **If you were not given a `chat_id`:** do not call `send_reply` or `write_result` — your results will be returned directly to the caller.
 
+**CC platform noise — do NOT call `write_result` again after seeing this:** After you call `write_result` and your task ends, the CC runtime (≥ 2.1.77) may inject a message like:
+
+```
+Stop hook feedback: [hook-name]: No stderr output
+```
+
+This is platform noise from the SubagentStop hook and does NOT mean your `write_result` failed or was rejected. Your result was already delivered. Do NOT call `write_result` a second time — the server will ignore the duplicate and return an explicit error message to confirm the first call was recorded.
+
 ## Internal vs. User-Facing Tasks
 
 Not all subagent tasks are user-facing. Some are dispatched for internal analysis — log review, codebase research, pre-processing — where the result goes to the dispatcher only, not to the user.
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -401,6 +401,7 @@ AUDIO_DIR = BASE_DIR / "audio"
 SENT_DIR = BASE_DIR / "sent"
 SENT_REPLIES_DIR = BASE_DIR / "sent-replies"
 TASK_REPLIED_DIR = BASE_DIR / "task-replied"
+WRITE_RESULT_DEDUP_DIR = BASE_DIR / "write-result-dedup"
 TASKS_FILE = BASE_DIR / "tasks.json"
 TASK_OUTPUTS_DIR = BASE_DIR / "task-outputs"
 BISQUE_OUTBOX_DIR = BASE_DIR / "bisque-outbox"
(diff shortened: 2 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS** — changes are correct, minimal, and well-tested.

**What this does:** Adds per-task_id idempotency to `handle_write_result`. The guard uses the same filesystem-marker pattern already in use for `TASK_REPLIED_DIR`, so the approach is consistent with the existing codebase. Markers expire after 1 hour.

**Code path verified:**
1. `_was_write_result_called(task_id)` is checked after required-field validation, before any inbox write.
2. On a hit, returns a rejection message that names the CC platform noise — gives the model enough signal to stop retrying.
3. On first call, `_record_write_result(task_id)` is called immediately before the existing `sent_reply_to_user` dedup logic — so the marker is set even if dedup short-circuits later.
4. `WRITE_RESULT_DEDUP_DIR` is initialized in the startup `mkdir` loop; no migration needed for existing installs (directory is created on first server start).

**The sys.subagent.bootup.md addition** is accurate — it describes the CC 2.1.77 behavior correctly and tells subagents the server will reject duplicates with an explicit message, which is true.

**Tests:** 8 new tests cover the full surface area — happy path, 2-call dedup, 19-call stress case, independent task_ids, expiry, and helper function unit tests. All pass. The 16 pre-existing failures in `test_mcp_server/` are unrelated.

**One minor note:** The 1-hour dedup window (`_WRITE_RESULT_DEDUP_WINDOW_SECS = 3600`) is longer than the `_TASK_REPLIED_WINDOW_SECS = 300` window. This is intentional and correct — the dedup window should be generous enough to cover the full possible subagent lifetime. No action needed.

Safe to merge.

sayhar (commented):
PASS (with one minor observation)

## Review of write_result idempotency / dedup fix

### 1. Dedup key correctness and path traversal safety

`_write_result_dedup_key` hashes the raw `task_id` string through SHA-256 and uses only the first 24 hex chars, prefixed with `wr_`. The resulting filename is always `wr_[0-9a-f]{24}` — zero path components, no traversal possible. The key space is 2^96 which is collision-resistant for this use case. Clean.

### 2. 1-hour dedup window

The 1-hour window covers any realistic subagent lifetime by a large margin (typical subagent run: seconds to a few minutes). There is no scenario where the same `task_id` should legitimately produce two different `write_result` calls: task IDs are per-dispatch and each dispatch produces exactly one result. The 1-hour window vs 5-minute `sent_reply` window mismatch is fine — they guard different things. No concern here.

### 3. Marker write vs. inbox write ordering (atomicity)

**Observation (minor):** `_record_write_result(task_id)` is called BEFORE the inbox write. If the server crashes after the marker is written but before the inbox JSON is created, the task result is permanently lost — the dedup guard will reject any retry for up to 1 hour.

In practice this is a very narrow window and the CC 2.1.77 flooding problem is far worse than a lost result on a server crash, so the tradeoff is reasonable. However it is worth acknowledging: the "first write wins" semantics mean a crash-before-inbox-write silently drops the result rather than allowing a single retry. A post-write marker (marker written after a successful inbox write) would be safer, but at the cost of allowing one duplicate on a crash-before-marker scenario. Either way there is a window; the current choice optimises for the flood problem over the crash-loss problem.

### 4. WRITE_RESULT_DEDUP_DIR auto-creation

Added to the module-level `for d in [...]:  d.mkdir(parents=True, exist_ok=True)` block that runs at import time. This is the same pattern used by every other directory in the server. Reliable.

### 5. Rejection message

The rejection message is specific and actionable:

> `write_result already recorded for task_id=... Your result was already delivered. This duplicate call was ignored. If you are seeing 'Stop hook feedback: No stderr output', that is CC platform noise — do NOT call write_result again.`

It names the exact platform noise string, explains what happened, and gives an explicit directive. A looping agent reading this has enough information to stop. The explicit "do NOT call write_result again" is the right call.

One edge case: what if the model ignores the directive and calls `write_result` anyway? The dedup guard fires again and returns the same message, so it is self-limiting. No runaway loop path.

### 6. sys.subagent.bootup.md instruction

Clear and correct. It:
- Gives the verbatim CC noise string so the model can match it
- Explains causality (SubagentStop hook ≥ 2.1.77)
- States the directive unambiguously ("Do NOT call write_result a second time")
- Notes that the server will confirm the first call was recorded if the model does call again

Placed immediately after the `write_result` / `send_reply` pairing instructions, which is the right location.

### 7. Unit tests (8 tests)

All key cases are covered:

- `test_first_call_writes_inbox_file` — happy path, 1 file written
- `test_second_call_same_task_id_is_rejected` — dedup fires, 1 file total
- `test_second_call_response_mentions_platform_noise` — rejection message quality
- `test_19_duplicate_calls_produce_one_inbox_file` — worst-case flood scenario
- `test_different_task_ids_each_write_own_file` — independence of task IDs
- `test_expired_dedup_marker_allows_retry` — expiry window works
- `test_was_write_result_called_returns_false_before_any_call` — helper unit test
- `test_record_and_check_write_result_called` — helper round-trip

Coverage is solid. The 19-call flood test is a nice regression test for the exact bug. The expiry test correctly back-dates the marker rather than sleeping. No missing cases.

## Summary

The implementation is correct and well-structured. The one non-blocking note is the marker-before-inbox-write ordering, which is a deliberate tradeoff that favours flood prevention over single-crash-loss avoidance — appropriate given the problem being solved. All 7 review points pass.

PR #591, opened 2026-03-17
fix(hooks): is_dispatcher() transcript fallback broken for CC 2.1.76+

### LINKED ISSUE #589: fix(hooks): session_role.is_dispatcher() transcript fallback broken for CC 2.1.76+
## Bug

`is_dispatcher()` in `hooks/session_role.py` has a transcript fallback that scans for dispatcher-only tools (`wait_for_messages`, `check_inbox`). It reads `hook_input.get("transcript")` — but in CC 2.1.76+, Stop hooks pass `transcript_path` (a file path to a JSONL file) instead of an inline `transcript` list. So `hook_input.get("transcript")` returns `None`, the fallback never fires, and the function falls through to `return False` (treated as subagent).

## Effect

When the marker file check is inconclusive (e.g. duplicate sessions exist and session IDs don't match), the dispatcher is misidentified as a subagent and the `require-write-result.py` Stop hook blocks it from exiting.

## Root cause

`require-write-result.py` already handles both `transcript_path` (Stop) and `agent_transcript_path` (SubagentStop) via `_load_transcript_from_jsonl()`, but `session_role.py` has no equivalent logic — it only checks the legacy inline `transcript` field.

## Fix

In `is_dispatcher()`, after checking the inline `transcript` field, also try to load transcript from `transcript_path` (Stop hook) or `agent_transcript_path` (SubagentStop hook) using a `_load_transcript_from_jsonl()` helper (same pattern as `require-write-result.py`).

```python
# Try file-based transcript (CC 2.1.76+ Stop hook)
for key in ("transcript_path", "agent_transcript_path"):
    path = hook_input.get(key)
    if path:
        transcript = _load_transcript_from_jsonl(path)
        if transcript:
            return _transcript_has_dispatcher_tool(transcript)
```

The `_load_transcript_from_jsonl()` function needs to be added to `session_role.py` (copy or factor from `require-write-result.py`).

sayhar:
## What this fixes

When the dispatcher session marker file check is inconclusive (session ID mismatch or file absent), `is_dispatcher()` falls back to scanning the transcript for dispatcher-only tool calls. In CC 2.1.76+, Stop hooks no longer embed the transcript inline — they pass `transcript_path` (a JSONL file path) instead. `session_role.is_dispatcher()` only checked `hook_input.get("transcript")`, which returns `None` on 2.1.76+, so the fallback always failed silently and the function returned `False` (subagent). The practical effect: a dispatcher session was misidentified as a subagent, causing `require-write-result.py` to hard-block the dispatcher from exiting with `sys.exit(2)`.

`require-write-result.py` already handles both `transcript_path` and `agent_transcript_path` via `_load_transcript_from_jsonl()`. `session_role.py` had no equivalent — this PR closes that gap.

## Changes

`hooks/session_role.py`:

- **`_load_transcript_from_jsonl(path)`** — new helper, copied from `require-write-result.py`. Reads a JSONL transcript file and returns a list of parsed entries. Returns `[]` on any failure (never raises).
- **`is_dispatcher()`** — after the inline `transcript` check, now iterates `transcript_path` then `agent_transcript_path` and loads from file if present. This mirrors the logic already in `require-write-result.py`.
- **`_transcript_has_dispatcher_tool()`** — updated to handle both JSONL format (`entry["message"]["content"]`) and legacy inline format (`entry["content"]`). Previously only the legacy path was checked, so even a correctly-loaded JSONL transcript would not find any tool calls.

Nothing outside `session_role.py` is changed. `require-write-result.py` continues to do its own transcript loading independently — this change does not consolidate those paths (that would be a separate refactor).

## Functional pattern

Pure functions throughout: each helper is a total function with explicit fallback values (`[]`, `False`) rather than raising. The layered `is_dispatcher()` logic is a guard-clause pipeline — marker file first, then inline transcript, then file-based transcript, then conservative default.

## Tests run

- [x] Manual inspection: the three-path fallback chain in `is_dispatcher()` covers all known CC versions (pre-2.1.76 inline, 2.1.76+ Stop, 2.1.76+ SubagentStop).
- [ ] Automated unit tests — skipped: no test suite exists for `hooks/session_role.py` in the repo. Logic is straightforward enough to verify by inspection.

Closes #589

### DIFF
--- a/hooks/session_role.py
+++ b/hooks/session_role.py
@@ -13,9 +13,11 @@ dispatcher or a background subagent.
    compares to the `session_id` field in the hook JSON input.
    Match → dispatcher.  Mismatch or file absent → try fallback.
 
-2. **Transcript fallback (secondary)**: Scan the `transcript` field (present in
-   Stop hooks; absent in PreToolUse hooks) for tool_use blocks containing the
-   dispatcher-only tool `wait_for_messages` or `check_inbox`.
+2. **Transcript fallback (secondary)**: Scan the transcript for tool_use blocks
+   containing the dispatcher-only tools `wait_for_messages` or `check_inbox`.
+   CC 2.1.76+ passes a file path (`transcript_path` for Stop hooks,
+   `agent_transcript_path` for SubagentStop hooks) rather than an inline
+   `transcript` list. Both file-based and inline forms are tried in order.
    Found → dispatcher.  Not found → subagent.
 
 3. **Default**: If neither signal is available (e.g. a PreToolUse hook with no
@@ -67,9 +69,20 @@ def is_dispatcher(hook_input: dict) -> bool:
         return marker_result
 
     # --- Secondary: transcript scan ---
+    # Try inline transcript first (legacy CC < 2.1.76).
     transcript = hook_input.get("transcript")
     if transcript is not None:
         return _transcript_has_dispatcher_tool(transcript)
+
+    # Try file-based transcript (CC 2.1.76+):
+    #   Stop hook       → transcript_path
+    #   SubagentStop    → agent_transcript_path
+    for key in ("transcript_path", "agent_transcript_path"):
+        path = hook_input.get(key)
+        if path:
+            transcript = _load_transcript_from_jsonl(path)
+            if transcript:
+                return _transcript_has_dispatcher_tool(transcript)
 
     # --- Default: no signal → treat as subagent (conservative) ---
     return False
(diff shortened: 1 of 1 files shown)


sayhar (commented):
PASS: The fix is correct and complete.

The three-path fallback — inline transcript (legacy), `transcript_path` (Stop), `agent_transcript_path` (SubagentStop) — mirrors the pattern already in `require-write-result.py`. The `_load_transcript_from_jsonl()` implementation is identical to the one there.

The additional fix to `_transcript_has_dispatcher_tool()` is necessary and easy to miss: without it, a correctly-loaded JSONL transcript would still return `False` because the JSONL format nests content under `entry["message"]["content"]`, not `entry["content"]`. Both the file loading and the content traversal needed fixing.

One thing to verify: `require-write-result.py` and `session_role.py` now have duplicate `_load_transcript_from_jsonl()` implementations. This is fine for now (both are short and identical), but a future refactor could extract it to a shared `hooks/transcript_utils.py`. Not a blocker.

sayhar (commented):
PASS: changes are correct and safe to merge.

## What this fixes

The dispatcher vs subagent discrimination (`is_dispatcher()`) was broken for CC 2.1.76+, which delivers the session transcript as a JSONL file rather than an inline `transcript` list in the hook JSON. The Stop hook now passes `transcript_path` and SubagentStop passes `agent_transcript_path`. Without this fix, `is_dispatcher()` always fell through to the conservative default (→ False = subagent), causing the Stop hook to incorrectly treat the dispatcher session as a subagent.

## Review findings

**`is_dispatcher()` tries both file-based paths** — yes, the new `for key in ("transcript_path", "agent_transcript_path")` loop is placed after the inline check and before the default-False return. Order is correct: inline first (legacy), then file-based (CC 2.1.76+).

**`_transcript_has_dispatcher_tool()` handles JSONL nested format** — yes. The JSONL entries wrap message content under `msg["message"]["content"]` while the legacy inline format uses `msg["content"]` directly. The `nested_msg = msg.get("message")` branch handles this cleanly. Both formats are exercised in the updated docstring and in smoke tests.

**`_load_transcript_from_jsonl()` duplication** — the function is indeed duplicated from `require-write-result.py`. This is tech debt but not a blocker: `session_role.py` is a standalone shared utility intentionally designed to have no intra-repo imports (to avoid circular dependencies with the hook scripts). The duplication is therefore load-bearing by design, not laziness. A shared `_transcript_utils.py` would be the right long-term fix and warrants a follow-up issue.

**Tests** — no unit tests added for `session_role.py`. The module has no existing test suite in the repo. Not blocking, but a follow-up to add tests in `hooks/tests/` would reduce regression risk here.

## Smoke test run

Ran 9 local scenarios against the branch:
- `transcript_path` with `wait_for_messages` → correctly returns True
- `agent_transcript_path` with `check_inbox` → correctly returns True
- `transcript_path` with non-dispatcher tool → correctly returns False
- Legacy inline `transcript` field still works → correctly returns True
- Empty file → False, no crash
- Nonexistent file path → False, no crash
- Malformed JSONL lines are skipped gracefully
- JSONL nested `message.content` format detected correctly
- Legacy flat `content` format still detected correctly

All 9 passed.

PR #594, opened 2026-03-17
fix(hooks): remove require-write-result.py from Stop hook, keep only in SubagentStop

### LINKED ISSUE #588: bug: require-write-result.py Stop hook fires on dispatcher, not just subagents
## Problem

\`require-write-result.py\` is registered under **both** \`Stop\` and \`SubagentStop\` in \`.claude/settings.json\`. This means the hook fires when the dispatcher exits too, not only when subagents exit.

The hook has an \`is_dispatcher()\` exemption check at the top of \`main()\`, but that check relies on:
1. A marker file at \`~/messages/config/dispatcher-session-id\` matching the session ID, or
2. A transcript scan for dispatcher-only tools (\`wait_for_messages\`, \`check_inbox\`)

The transcript fallback (path 2) reads \`data.get(\"transcript\")\` — the inline field. But in CC 2.1.76+, Stop hooks pass \`transcript_path\` (a file path), not an inline \`transcript\`. So if the marker file is absent or mismatched (e.g. after a restart, or on a fresh boot before the first \`write-dispatcher-session-id\` hook runs), the \`is_dispatcher()\` check falls back to transcript scan, finds nothing in the inline \`transcript\` field, and incorrectly treats the dispatcher as a subagent — blocking it from exiting.

## Root cause

Two problems:
1. \`require-write-result.py\` should only be on \`SubagentStop\`, not \`Stop\`. The dispatcher never calls \`write_result\`.
2. The \`is_dispatcher()\` fallback in the hook does not read \`transcript_path\` when the inline \`transcript\` field is absent — it only checks \`data.get(\"transcript\")\`.

## Fix

**Part 1 (primary fix):** Remove \`require-write-result.py\` from the \`Stop\` hook in \`.claude/settings.json\`. Keep it only under \`SubagentStop\`.

**Part 2 (secondary safeguard):** In \`session_role.is_dispatcher()\`, when the inline \`transcript\` is absent but \`transcript_path\` is present, load and scan the JSONL file for dispatcher-only tools. This closes the gap where the marker file is unavailable.

## Affected file

\`.claude/settings.json\` (symlinked from \`~/lobster/.claude/settings.json\`)

sayhar:
Closes #588

## Summary

- Remove `require-write-result.py` from the `Stop` hook array in `install.sh` — new installs will no longer register it under `Stop`
- Add Migration 17 to `upgrade.sh` to strip the `Stop` entry from existing `settings.json` on upgrade
- The live `~/.claude/settings.json` was also updated directly (Stop now has an empty array)

`Stop` fires for the dispatcher main session; `SubagentStop` fires for Task-spawned subagents — these are mutually exclusive. The hook should only be under `SubagentStop`. The `is_dispatcher()` check in the hook was a band-aid for this misregistration and can remain as defensive code without harm.

Note: Part 2 from the issue (fixing the `is_dispatcher()` fallback to read `transcript_path`) is left as a follow-up — Part 1 is the root fix and makes Part 2 unnecessary for the Stop/dispatcher case.

## Test plan

- [ ] Fresh install: verify `settings.json` has `require-write-result.py` only under `SubagentStop`, not `Stop`
- [ ] Existing install upgrade: verify Migration 17 removes the `Stop` entry from `settings.json`
- [ ] Dispatcher session ends normally without hook interference
- [ ] Subagents without `write_result` are still blocked by the `SubagentStop` hook

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1719,27 +1719,6 @@ else
     info "Skipping inject-debug-bootup hook (settings.json not yet created)"
 fi
 
-# Set up Claude Code Stop hook to enforce write_result in subagent sessions
-chmod +x "$INSTALL_DIR/hooks/require-write-result.py" || true
-if [ -f "$CLAUDE_SETTINGS" ]; then
-    if ! jq -e '.hooks.Stop[]? | select(.matcher == "")' "$CLAUDE_SETTINGS" > /dev/null 2>&1; then
-        TMP_SETTINGS=$(mktemp)
-        jq '.hooks.Stop = (.hooks.Stop // []) + [{
-            "matcher": "",
-            "hooks": [{
-                "type": "command",
-                "command": "python3 '"$INSTALL_DIR"'/hooks/require-write-result.py",
-                "timeout": 10
-            }]
-        }]' "$CLAUDE_SETTINGS" > "$TMP_SETTINGS" && mv "$TMP_SETTINGS" "$CLAUDE_SETTINGS"
-        success "require-write-result Stop hook installed"
-    else
-        info "require-write-result Stop hook already configured in Claude Code settings"
-    fi
-else
-    info "Skipping require-write-result Stop hook (settings.json not yet created)"
-fi
-
 # Set up Claude Code SubagentStop hook to enforce write_result in subagent sessions
 # SubagentStop fires when a background sidechain session considers stopping — this is
 # the hook that actually catches subagents, whereas Stop only fires for the main session.
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1258,6 +1258,36 @@ with open(path, 'w') as f:
         migrated=$((migrated + 1))
     fi
 
+    # Migration 17: Remove require-write-result.py from the Stop hook in settings.json
+    # The Stop event fires for the dispatcher main session; SubagentStop fires for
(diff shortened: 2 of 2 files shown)


sayhar (commented):
PASS

**install.sh**
- `require-write-result.py` is correctly absent from the `Stop` section. The entire Stop hook block (21 lines) has been removed. The `SubagentStop` block immediately follows and is intact. No completeness issues.

**Migration 17 (upgrade.sh)**
- Numbered correctly — Migration 16 is the last entry on `main`, so 17 is right.
- Logic is sound: uses `jq` to filter out any Stop entry whose commands contain `require-write-result`, leaving other Stop entries untouched and `SubagentStop` completely unaffected.
- **Idempotent**: tested against the current live `settings.json` where `Stop` is already empty — the migration detects `length == 0` and skips entirely. No double-apply risk.
- **Removal logic verified**: tested against a synthetic settings.json with the bad Stop entry — jq correctly strips it to `[]` while preserving the SubagentStop entry.

**Smoke test (live system)**
- `~/.claude/settings.json` already has `"Stop": []` (the manual fix applied prior to this PR).
- `SubagentStop` is correctly configured with `require-write-result.py`.
- Running Migration 17 logic against current state: skips cleanly (idempotent confirmed).

Merging.

PR #596, opened 2026-03-17
fix(upgrade): define CLAUDE_SETTINGS in Migration 17 to prevent unbound variable crash

sayhar:
## Summary

Migration 17 (added in #594) crashes `upgrade.sh` on any install where `$CLAUDE_SETTINGS` has not been manually defined. With `set -euo pipefail` active at the top of the script, referencing an unbound variable is fatal — users running `lobster update` hit `CLAUDE_SETTINGS: unbound variable` and the upgrade aborts.

The fix is a single line: `local CLAUDE_SETTINGS="$HOME/.claude/settings.json"` added at the top of the Migration 17 block. This is the actual path to the Claude settings file. The rest of the migration logic is unchanged and correct.

The migration itself was a no-op on installs where the Stop hook array was already empty, which is why it wasn't caught before merge — but it would crash on any fresh install that has `require-write-result.py` registered under `Stop`.

**Functional patterns used:** none — this is a pure shell variable definition fix.

## Tests run

- [x] `bash -n scripts/upgrade.sh` — syntax check passes, no errors
- [x] Manual inspection of Migration 17 block — variable is now defined before first use, all three references resolve correctly
- [ ] Full `lobster update` run on an install with the hook registered — skipped: requires specific environment setup; fix is mechanical and verifiable by inspection

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1280,6 +1280,7 @@ with open(path, 'w') as f:
     # registered under Stop (which hit the dispatcher) as well as SubagentStop.
     # The is_dispatcher() guard in the hook was a band-aid for this misregistration.
     # Fix: remove the entry from Stop[], leave it only under SubagentStop.
+    local CLAUDE_SETTINGS="$HOME/.claude/settings.json"
     if [ -f "$CLAUDE_SETTINGS" ] && command -v jq >/dev/null 2>&1; then
         local stop_has_write_result
         stop_has_write_result=$(jq -r '


sayhar (commented):
Self-review note (cannot approve own PR): the fix is correct and minimal.

Migration 17 used $CLAUDE_SETTINGS in three places (file existence check, jq read, mv target) without ever defining it. With `set -euo pipefail`, the first reference aborts the upgrade immediately.

Adding `local CLAUDE_SETTINGS="$HOME/.claude/settings.json"` at the top of the block resolves all three references. The path is correct — it matches where Claude Code stores its settings. Using `local` is consistent with other variable scoping in this function. The jq filter logic from #594 is unchanged.

`bash -n scripts/upgrade.sh` passes. Safe to merge.

sayhar (commented):
## PR Review — Migration 17 fix (CLAUDE_SETTINGS unbound variable)

**Verdict: PASS**

### 1. `local CLAUDE_SETTINGS` — bash scoping is correct

The migration runs inside `run_migrations()` (line 935 of upgrade.sh), which is a regular bash function. `local` is valid here and correctly scopes the variable to that function. No issue.

### 2. Path matches install.sh

`$HOME/.claude/settings.json` exactly matches the canonical path used throughout the codebase. install.sh defines `CLAUDE_SETTINGS_DIR="$HOME/.claude"` and `CLAUDE_SETTINGS="$CLAUDE_SETTINGS_DIR/settings.json"` at lines 1429–1430. The fix uses the correct path.

### 3. jq detection query

```jq
[.hooks.Stop[]?.hooks[]?.command // empty]
| map(select(contains("require-write-result")))
| length
```

This correctly traverses the `Stop` array (each element is an object with a `.hooks[]` sub-array containing `.command` strings). The `//empty` and `[]?` guards handle missing or null values gracefully. Output is a count; the condition `!= "0"` is correct.

### 4. jq removal query

```jq
.hooks.Stop = (
    (.hooks.Stop // [])
    | map(select(
        (.hooks // [])
        | map(.command // "")
        | all(contains("require-write-result") | not)
    ))
)
```

This removes entire hook-group entries from `Stop[]` where any sub-hook command references `require-write-result`. The `all(... | not)` predicate is correct — it keeps only groups that have zero matching commands. The `(.hooks.Stop // [])` fallback handles installs where `Stop` is absent.

### 5. `TMP_SETTINGS` not declared `local` (minor, non-blocking)

`TMP_SETTINGS=$(mktemp)` at line 1292 is assigned without `local`. This is a minor style inconsistency — the same pattern appears in other migration blocks and in install.sh throughout. It is not a bug; `mktemp` always returns a valid path regardless of scoping. Not blocking merge.

### 6. No other migrations with similar broken patterns

The two other migrations that use locally-scoped uppercase variables (`HEALTH_MARKER` at line 1171, `SELFCHECK_MARKER` at line 1266) both declare them with `local` before use. Migration 17 was the only instance of this class of bug. The fix is complete.

### 7. Syntax check

`bash -n scripts/upgrade.sh` passes cleanly on the PR branch.

---

One-line summary: the fix is mechanical, correct, and complete. Safe to merge.

PR #603, opened 2026-03-17
fix(session-tracking): dynamic ghost-detector path and agent-status age expiry

### LINKED ISSUE #601: fix: ghost-detector.py uses hardcoded -home-admin- path, should be dynamic
## Bug

`scripts/ghost-detector.py` has a hardcoded path component in the filesystem scan glob:

```python
AGENT_OUTPUT_GLOB = "/tmp/claude-1000/-home-admin-lobster-workspace/*/tasks/"
```

On systems where the home directory is not `/home/admin` (e.g. `/home/lobster`), this glob matches nothing. The filesystem scan silently returns zero unregistered agents on every run, masking real problems.

## Impact

The \`--fs-scan\` path (and the default scan) has never found a ghost or unregistered agent on this system. Any agent that died without registering in the DB would be invisible.

## Fix

Replace the hardcoded \`-home-admin-\` segment with a value derived from the actual home directory at runtime:

\`\`\`python
import os
home = os.path.expanduser("~")
path_slug = home.strip("/").replace("/", "-")
AGENT_OUTPUT_GLOB = f"/tmp/claude-1000/-{path_slug}-lobster-workspace/*/tasks/"
\`\`\`

This makes the path portable across any install without requiring configuration.

### LINKED ISSUE #602: fix: agent-status.sh shows dead agents as running when stop_reason=tool_use with no age expiry
## Bug

`scripts/agent-status.sh` determines agent liveness solely from the `stop_reason` field in the JSONL output file:

- `end_turn` → agent is done (excluded from active display)
- `tool_use` (or empty) → agent is treated as "running"

There is no check on the file's last modification time. An agent that died mid-tool-call (crash, OOM kill, network disconnect) will have `stop_reason=tool_use` permanently frozen in the file. `scan_agent_status()` will report it as "running" forever.

## Impact

Dead agents accumulate in the status output indefinitely. The self-check message becomes cluttered with stale entries, and operators cannot distinguish genuinely running agents from long-dead ones.

## Fix

In `scan_agent_status()`, after detecting `stop_reason=tool_use`, add an age check on the output file's mtime. If the file hasn't been modified in more than N minutes (suggested: 30 minutes, matching the ghost-detector default), classify the entry as stale/dead rather than running.

\`\`\`bash
# Get file age in seconds
local file_mtime file_age_seconds now
now=$(date +%s)
file_mtime=$(stat -c %Y "$filepath" 2>/dev/null || stat -f %m "$filepath" 2>/dev/null)
file_age_seconds=$(( now - file_mtime ))
STALE_THRESHOLD_SECONDS=$(( 30 * 60 ))

if [ "$stop_reason" = "tool_use" ] && [ "$file_age_seconds" -gt "$STALE_THRESHOLD_SECONDS" ]; then
    continue  # treat as dead
fi
\`\`\`

sayhar:
## What problem this solves

Two bugs caused session monitoring to silently fail, making ghost agent detection unreliable on this system.

**Bug 1 (closes #601):** `ghost-detector.py` hardcoded `-home-admin-lobster-workspace` in the filesystem scan glob. On any system where `$HOME` is not `/home/admin` (here it is `/home/lobster`), the glob matched nothing. The unregistered-agent scan returned zero results on every run — real orphaned agents were invisible to monitoring.

**Bug 2 (closes #602):** `agent-status.sh` determined liveness solely from `stop_reason`. An agent that crashes mid-tool-call leaves `stop_reason=tool_use` frozen in its output file permanently. `scan_agent_status()` had no age check, so a crashed agent appeared as "running" forever, accumulating stale entries in every self-check message.

## What changed

**ghost-detector.py:** `AGENT_OUTPUT_GLOB` is now constructed at module load time via a pure helper `_default_agent_output_glob()` that derives the path slug from `os.path.expanduser("~")`. On `/home/lobster` this produces `-home-lobster-lobster-workspace`. No configuration required; portable to any install.

**agent-status.sh:** In `scan_agent_status()`, after detecting `stop_reason=tool_use`, an age check reads the file's mtime via `stat -c %Y` (GNU) with a `stat -f %m` (BSD) fallback. If the file is older than `STALE_THRESHOLD_SECONDS` (1800s / 30 min), the entry is skipped as dead. Agents with a fresh mtime continue to show as "running" normally.

Neither change affects `end_turn` handling or `scan_completed_tasks()`.

## Functional patterns used

- `_default_agent_output_glob()` is a pure function with no side effects; the constant it produces is captured once at module load.
- The bash age check uses only local variables and is isolated to the loop body.

## Tests run

- [x] `uv run pytest tests/unit/test_ghost_detector_filesystem.py -v` — 36 passed, 0 failed
- [x] `bash tests/test-agent-status.sh` — 10 passed, 8 failed (all 8 failures are pre-existing on `main`; confirmed identical failure set before and after this change)
- [x] Manual smoke test: file with `stop_reason=tool_use` + 35-minute-old mtime excluded from output; same file with current mtime shown as "running" — behaves correctly
- [x] `bash -n scripts/agent-status.sh` — syntax OK
- [x] `python3 -c "import ast; ast.parse(open('scripts/ghost-detector.py').read())"` — syntax OK

### DIFF
--- a/scripts/agent-status.sh
+++ b/scripts/agent-status.sh
@@ -124,11 +124,26 @@ scan_agent_status() {
             continue
         fi
 
+        # An agent with stop_reason=tool_use whose output file hasn't been modified
+        # in more than STALE_THRESHOLD_SECONDS is treated as dead (crashed mid-tool-call).
+        # Without this check a crashed agent would show as "running" forever.
+        local STALE_THRESHOLD_SECONDS=$(( 30 * 60 ))
+        if [ "$stop_reason" = "tool_use" ]; then
+            local now file_mtime file_age_seconds
+            now=$(date +%s)
+            # stat -c %Y is GNU coreutils; stat -f %m is BSD/macOS fallback
+            file_mtime=$(stat -c %Y "$filepath" 2>/dev/null || stat -f %m "$filepath" 2>/dev/null || echo "$now")
+            file_age_seconds=$(( now - file_mtime ))
+            if [ "$file_age_seconds" -gt "$STALE_THRESHOLD_SECONDS" ]; then
+                continue  # file too old — agent is dead, not running
+            fi
+        fi
+
         local status_text
         if [ -z "$stop_reason" ]; then
             status_text="starting"
         else
-            # "tool_use" or any other value = actively running
+            # "tool_use" (recently active) or any other value = actively running
             status_text="running"
         fi
 
--- a/scripts/ghost-detector.py
+++ b/scripts/ghost-detector.py
@@ -58,7 +58,17 @@ DB_PATH = Path.home() / "messages" / "config" / "agent_sessions.db"
 
 # Glob pattern for Claude Code session task directories.
 # The middle component is a session UUID that changes per Claude Code invocation.
-AGENT_OUTPUT_GLOB = "/tmp/claude-1000/-home-admin-lobster-workspace/*/tasks/"
+# The path component is derived dynamically from the actual home directory so the
+# script works on any install regardless of the username (e.g. /home/lobster vs
+# /home/admin).  Claude Code maps workspace paths to /tmp by replacing '/' with '-'.
+def _default_agent_output_glob() -> str:
+    home = os.path.expanduser("~")
+    # /home/lobster -> -home-lobster-
+    path_slug = home.strip("/").replace("/", "-")
+    return f"/tmp/claude-1000/-{path_slug}-lobster-workspace/*/tasks/"
+
+
+AGENT_OUTPUT_GLOB = _default_agent_output_glob()
 
 # Pattern for agent JSONL symlink filenames: agent-<hex_id>.jsonl
 AGENT_SYMLINK_PATTERN = re.compile(r"^agent-([0-9a-f]+)\.jsonl$")


sayhar (commented):
**PASS** — both fixes are correct. Specific findings below.

---

### ghost-detector.py: `_default_agent_output_glob()`

**Slug construction is correct.**

Sanity check run on this host (`/home/lobster`):

```
python3 -c "import os; home=os.path.expanduser('~'); print('-' + home.strip('/').replace('/', '-') + '-lobster-workspace')"
# Output: -home-lobster-lobster-workspace
```

Matches the actual directory: `/tmp/claude-1000/-home-lobster-lobster-workspace/` (confirmed to exist and contain session UUIDs).

`strip('/')` correctly handles both `/home/lobster` and `/home/lobster/` (trailing slash) identically.

**Module-load assignment is fine.** `_default_agent_output_glob()` only calls `os.path.expanduser("~")` — no I/O, no side effects, no exceptions possible under normal conditions. Module import is clean.

**Edge case — root user:** If run as root, `os.path.expanduser("~")` returns `/root`, producing `/tmp/claude-1000/-root-lobster-workspace/`. That is the correct Claude Code path for root. No issue.

**Remaining hardcoded UID `1000` in `/tmp/claude-1000/`:** The PR does not address this, but this host's user has UID 1000 (`id -u` → `1000`), which matches. This is a pre-existing limitation, not introduced here. Out of scope for this PR.

---

### agent-status.sh: age expiry check

**`stat` fallback chain is correct.** On this host (Linux/GNU coreutils 9.4):
- `stat -c %Y "$filepath"` succeeds and returns epoch seconds.
- The BSD fallback `stat -f %m` is skipped (returns error, shell continues to `|| echo "$now"`).
- The `|| echo "$now"` final fallback means a stat failure produces `file_age_seconds=0`, which is safe — the agent is NOT incorrectly skipped.

**`file_age_seconds=$(( now - file_mtime ))` produces correct positive values** when `file_mtime <= now` (which is always true for existing files). Clock skew could theoretically produce a negative value if the file was written in the future, but `[ "$file_age_seconds" -gt "$STALE_THRESHOLD_SECONDS" ]` would simply be false, so the agent would be reported as running — safe failure mode.

**`local STALE_THRESHOLD_SECONDS` inside the loop** — minor style note: this variable is constant per scan_agent_status() call; hoisting it before the loop would be cleaner, but it has no correctness impact since `local` in bash only scopes to the function, not the loop body.

**`local now file_mtime file_age_seconds`** — declaring these as `local` is correct and prevents leaking into the outer scope.

---

### Summary

| Check | Result |
|---|---|
| Slug from `/home/lobster` matches `/tmp` path | PASS |
| Module-load assignment side effects | PASS (none) |
| `stat -c %Y` / `stat -f %m` fallback chain | PASS |
| `file_age_seconds` arithmetic | PASS |
| Root user / unusual paths | PASS (handled correctly) |
| Clock skew | safe failure mode (agent stays "running") |

Ready to merge.

PR #613, opened 2026-03-17
fix: normalize spawned_at to UTC timezone-naive in auto-register hook

### LINKED ISSUE #612: fix: agent_sessions.db spawned_at timezone suffix breaks age-based stale row cleanup
## Problem

The `auto-register-agent.py` hook (PR #605) inserts `spawned_at` using Python's `datetime.now(timezone.utc).isoformat()`, producing a timezone-aware string like `2026-03-17T20:00:00+00:00`.

SQLite's `datetime('now', '-30 minutes')` produces timezone-naive output (`2026-03-17 20:00:00`). String comparisons between the two formats fail silently — age-based stale row cleanup never triggers for rows inserted by the auto-register hook.

## Fix options

**Option A (preferred):** Normalize `spawned_at` in the hook insert to UTC timezone-naive:
```python
spawned_at = datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S')
```

**Option B:** Update cleanup queries to handle both formats:
```sql
WHERE substr(spawned_at, 1, 19) < strftime('%Y-%m-%dT%H:%M:%S', datetime('now', '-30 minutes'))
```

Option A is simpler and prevents the issue at the source.

## Affected files
- `hooks/auto-register-agent.py` — insert site
- `scripts/agent-status.sh` — cleanup comparison

sayhar:
## Summary

Stale agent rows inserted by `hooks/auto-register-agent.py` were never cleaned up by the age-based logic in `scripts/agent-status.sh`. The root cause: the hook stored `spawned_at` using `isoformat()`, which produces `2026-03-17T20:00:00+00:00`. SQLite's `datetime('now', '-30 minutes')` produces `2026-03-17 20:00:00` — timezone-naive with a space separator. String comparison between these two formats fails silently, so the `spawned_at < datetime('now', '-30 minutes')` check never matched, and stale rows lingered indefinitely.

The fix changes the insert to use `strftime('%Y-%m-%d %H:%M:%S')`, which produces the exact format SQLite datetime functions compare against. The fix is applied at the insert site in the hook, preventing the mismatch at the source rather than patching every cleanup query.

Nothing in `scripts/agent-status.sh` is changed — it does not query `spawned_at` directly; cleanup is driven by file mtime and `stop_reason`. The `spawned_at` field is queried in `src/mcp/wire_server.py` and `src/mcp/inbox_server.py` (via `register_agent`), which already write their own timestamps — those insert paths are unaffected by this change.

## Changes

- `hooks/auto-register-agent.py`: one-line change, `isoformat()` → `strftime('%Y-%m-%d %H:%M:%S')`
- `tests/unit/test_hooks/test_auto_register_agent.py`: updated pre-seeded test fixture to use the correct format; added `test_spawned_at_is_sqlite_compatible_format` which asserts both the string pattern and that SQLite's `datetime('now', '-30 minutes')` comparison returns the expected result

## Functional patterns

Pure function tests for extraction logic are unchanged. The new test is a direct DB state assertion — no mocking, no implementation-detail coupling.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_auto_register_agent.py -x -q` — 29 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — all hook tests pass; failures in other unit test modules (daemon, bot) are pre-existing and unrelated to this change

Closes #612

### DIFF
--- a/hooks/auto-register-agent.py
+++ b/hooks/auto-register-agent.py
@@ -234,7 +234,7 @@ def insert_agent_session(
                 reply_message_ids   TEXT
             )
         """)
-        now = datetime.now(timezone.utc).isoformat()
+        now = datetime.now(timezone.utc).strftime('%Y-%m-%d %H:%M:%S')
         conn.execute(
             """
             INSERT OR IGNORE INTO agent_sessions
--- a/tests/unit/test_hooks/test_auto_register_agent.py
+++ b/tests/unit/test_hooks/test_auto_register_agent.py
@@ -278,7 +278,7 @@ class TestHookAgentWithAgentId:
         """)
         conn.execute(
             "INSERT INTO agent_sessions (id, description, chat_id, status, spawned_at)"
-            " VALUES ('agent-dup', 'richer description', '12345', 'running', '2026-01-01T00:00:00+00:00')"
+            " VALUES ('agent-dup', 'richer description', '12345', 'running', '2026-01-01 00:00:00')"
         )
         conn.commit()
         conn.close()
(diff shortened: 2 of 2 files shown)


sayhar (commented):
PASS

**Change verified:**  produces the correct SQLite-compatible format (e.g. `2026-03-17 20:00:00`) — no `T` separator, no timezone offset.

**SQLite compatibility confirmed:** `datetime('now', '-30 minutes')` also returns the `YYYY-MM-DD HH:MM:SS` format, so string comparison with `spawned_at` now works correctly for stale-row cleanup.

**Tests:** all 29 pass including the new `test_spawned_at_is_sqlite_compatible_format`, which:
1. Regex-checks the stored value matches `^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$`
2. Executes `spawned_at > datetime('now', '-30 minutes')` directly in SQLite and asserts the result is `1`

The fixture update to `test_insert_or_ignore_preserves_existing_row` is correct — old ISO-8601 value replaced with the new format.

One minor note: the branch checkout instruction in the task used `fix/spawned-at-timezone-format` but the actual branch is `fix/issue-612-spawned-at-timezone`. No issue with the PR itself — just a naming discrepancy in the review request.

sayhar (commented):
PASS

**Change verified:** `datetime.now(timezone.utc).strftime('%Y-%m-%d %H:%M:%S')` produces the correct SQLite-compatible format (e.g. `2026-03-17 20:00:00`) — no T separator, no timezone offset.

**SQLite compatibility confirmed:** `datetime('now', '-30 minutes')` also returns the `YYYY-MM-DD HH:MM:SS` format, so string comparison with `spawned_at` now works correctly for stale-row cleanup.

**Tests:** all 29 pass including the new `test_spawned_at_is_sqlite_compatible_format`, which:
1. Regex-checks the stored value matches `^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$`
2. Executes `spawned_at > datetime('now', '-30 minutes')` directly in SQLite and asserts the result is `1`

The fixture update to `test_insert_or_ignore_preserves_existing_row` is correct — old ISO-8601 value replaced with the new format.

One minor note: the branch name in the review request (`fix/spawned-at-timezone-format`) differs from the actual branch (`fix/issue-612-spawned-at-timezone`). No issue with the PR itself.

PR #616, opened 2026-03-17
fix: session_end closes 'starting' rows so write_result completes the agent lifecycle

### LINKED ISSUE #615: feat: write_result should mark agent_sessions row as completed
## Problem

The auto-register hook (PR #605) inserts a row with `status='starting'` when an agent is spawned. But when the agent calls `write_result`, the row is never updated to `status='completed'`. This means every subagent that ever ran leaves a permanent 'starting' row in agent_sessions.db.

## Expected

When `write_result` is called with a `task_id`, the MCP handler should do:
```sql
UPDATE agent_sessions SET status='completed', completed_at=datetime('now') WHERE task_id=?
```

## Why this is the right fix

`write_result` is the canonical 'I am done' signal for every subagent — it is mandatory (enforced by SubagentStop hook). Making it also close the registry row is the natural completion of the agent lifecycle. No extra calls needed from the subagent.

## Files to change

- `src/mcp/inbox_server.py` — the `write_result` tool handler
- May need a migration to add `completed_at` column if not present

sayhar:
## Summary

The `auto-register-agent` PostToolUse hook (landed in #605) inserts agent_sessions rows with `status='starting'` when an agent is spawned. When the agent later calls `write_result`, `handle_write_result` calls `session_store.session_end` — but `session_end`'s WHERE clause only matched `status='running'`. Rows inserted as `'starting'` were therefore never closed, accumulating as permanent ghost rows in `agent_sessions.db`.

The fix is a one-line change to `session_end`: expand the filter from `status = 'running'` to `status IN ('running', 'starting')`. This completes the agent lifecycle for both registration paths:
- Agents registered by the dispatcher via `session_start` (status `'running'`)
- Agents auto-registered by the PostToolUse hook via INSERT OR IGNORE (status `'starting'`)

No change to `inbox_server.py` was needed — `handle_write_result` already calls `session_store.session_end` correctly. The bug was entirely in the WHERE clause of the `session_end` update.

The `completed_at` column already exists in the schema (present since the original session_store DDL), so no migration is needed.

**Functional patterns:** pure SQL update with explicit status guard; all side effects remain isolated to `session_end`; idempotency preserved (completed/failed rows are unaffected by the expanded clause).

## Tests run

- [x] `uv run pytest tests/test_agent_sessions.py -v` — 33 passed, 0 failed (includes new `test_session_end_closes_starting_row` which directly reproduces the bug)
- [ ] Full unit suite (`tests/unit/`) — 70 pre-existing failures in `test_update_manager` and `test_memory` (wrong ELF class for sqlite_vec) unrelated to this change; not introduced here

Closes #615

### DIFF
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -316,12 +316,16 @@ def session_end(
     conn = _get_connection(resolved)
     now = datetime.now(timezone.utc).isoformat()
 
-    # Try matching on id first, then task_id
+    # Match on id or task_id.  Accept both 'running' (registered by the
+    # dispatcher via session_start) and 'starting' (auto-registered by the
+    # PostToolUse hook before the agent has fully initialised).  This closes
+    # the lifecycle gap where rows inserted as 'starting' were never updated
+    # because the old WHERE clause only matched 'running'.
     conn.execute(
         """
         UPDATE agent_sessions
         SET status = ?, completed_at = ?, result_summary = ?
-        WHERE (id = ? OR task_id = ?) AND status = 'running'
+        WHERE (id = ? OR task_id = ?) AND status IN ('running', 'starting')
         """,
         (status, now, result_summary, id_or_task_id, id_or_task_id),
     )
--- a/tests/test_agent_sessions.py
+++ b/tests/test_agent_sessions.py
@@ -161,6 +161,42 @@ def test_session_end_does_not_double_close(isolated_db):
     assert result["result_summary"] == "First close"
 
 
+def test_session_end_closes_starting_row(isolated_db):
+    """session_end closes rows with status='starting' (auto-registered by PostToolUse hook).
+
+    The auto-register-agent.py hook inserts rows as 'starting' before the agent
+    has fully initialised. Previously, session_end only matched 'running' rows,
+    so write_result calls on 'starting' rows were silently no-ops and the rows
+    accumulated indefinitely. This test verifies the fix: session_end now accepts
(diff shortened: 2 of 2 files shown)


sayhar (commented):
**PASS**

The fix is correct and properly wired.

**What the diff does:**
- `session_store.py`: expands `WHERE status = 'running'` to `WHERE status IN ('running', 'starting')` in `session_end()`. The comment accurately describes the lifecycle gap being closed.
- `test_agent_sessions.py`: adds `test_session_end_closes_starting_row` which directly mimics the PostToolUse hook by inserting a `status='starting'` row and verifying `session_end` closes it by task_id.

**Wiring check (inbox_server.py line ~4868):**
`handle_write_result` calls `_session_store.session_end(id_or_task_id=task_id, ...)` in a try/except. This is the correct code path — when a subagent calls `write_result`, the session is auto-unregistered via exactly this call. With the fix, rows inserted as `'starting'` by the hook are now caught here. The wiring is complete and correct.

**No issues found:**
- The SQL change is minimal and targeted.
- The test covers the exact failure scenario (task_id lookup on a 'starting' row).
- No migration required — schema unchanged.
- No other callers of `session_end` are affected by broadening the status match; rows already in `'completed'` or `'dead'` are excluded by the IN clause just as they were before.

One minor note: the test imports `sqlite3` inline inside the function body rather than at the top of the file. Not a blocker — just a style inconsistency with the rest of the test file.

PR #617, opened 2026-03-18
feat(hooks): add require-register-agent-task-id PreToolUse hook

sayhar:
## Summary

- Commits `hooks/require-register-agent-task-id.py` which was present on disk but never committed to the repo
- This hook blocks `register_agent` calls missing a `task_id` parameter, enabling reliable DB row matching in the SubagentStop hook
- The hook was already wired into `~/.claude/settings.json` (added during PR #520 rollout), so it was actively running but untracked — causing the `[WARN] Local changes detected` upgrade warning

## Test plan

- [ ] Verify `hooks/require-register-agent-task-id.py` is present in the repo after merge
- [ ] Confirm `lobster update` no longer reports local changes warning after merge + pull
- [ ] Verify the hook still blocks `register_agent` calls without `task_id` in a subagent session

### DIFF
--- a/hooks/require-register-agent-task-id.py
+++ b/hooks/require-register-agent-task-id.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+"""PreToolUse hook: blocks register_agent calls missing a task_id parameter.
+
+The task_id is required so the SubagentStop hook can reliably match the DB row
+by primary key when closing ghost agent entries.
+"""
+import json, sys
+
+data = json.load(sys.stdin)
+tool = data.get("tool_name", "")
+inp = data.get("tool_input", {})
+
+if tool != "mcp__lobster-inbox__register_agent":
+    sys.exit(0)
+
+task_id = inp.get("task_id")
+
+if not task_id:
+    print(
+        "BLOCKED: register_agent called without task_id. "
+        "Set task_id to the same value the subagent will use in write_result. "
+        "This enables reliable DB matching in the SubagentStop hook.",
+        file=sys.stderr,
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR adds a PreToolUse hook that blocks `register_agent` calls missing a `task_id` parameter.

**What the code does:** `hooks/require-register-agent-task-id.py` is a PreToolUse hook (25 lines). When the tool being called is `mcp__lobster-inbox__register_agent` and the `task_id` input field is missing or empty, it writes a blocking error message to stderr and exits with code 2 (which Claude Code interprets as a hard block). All other tools pass through immediately (`sys.exit(0)`).

**Why it matters:** The `task_id` parameter on `register_agent` is required for the SubagentStop hook to reliably match the DB row by primary key when closing ghost agent entries. Without `task_id`, the SubagentStop hook falls back to session_id matching, which is less reliable. Enforcing the requirement at hook time prevents the subtle bug of a registered agent that can never be properly closed.

**Observations:**
- The hook reads from stdin with `json.load(sys.stdin)` and accesses `tool_name` and `tool_input` — this matches Claude Code's PreToolUse hook input format.
- The error message is actionable: it explains what to pass and why.
- The hook is minimal and has no side effects — it only blocks, never modifies.
- The hook needs to be registered in `settings.json` (PreToolUse, no matcher or universal matcher) to take effect — this PR does not appear to include install.sh changes to register it. May need a follow-up or be handled by an existing registration path.

PR #618, opened 2026-03-18
feat(debug-mode): interactive mode parity with persistent mode startup

sayhar:
🤖🦞 Lobster (engineer):

## What problem this solves

When \`LOBSTER_DEBUG=true\`, Lobster uses \`claude-wrapper.exp\` instead of \`claude-persistent.sh\`. An audit found two gaps where the interactive/debug launcher diverged from the persistent launcher in ways that caused real operational problems:

**Gap 1 — no startup alert in debug mode.** \`claude-persistent.sh\` sends a Telegram "Lobster Starting" alert before spawning Claude so operators know a restart happened. \`claude-wrapper.exp\` had no equivalent, leaving debug-mode restarts invisible via Telegram.

**Gap 2 — stale dispatcher session ID on restart.** \`write-dispatcher-session-id.py\` (a \`SessionStart\` hook) uses \`~/messages/config/dispatcher-session-id\` to decide whether a new Claude session is the dispatcher or a subagent. On restart, the marker still held the old session's ID, causing the new dispatcher to be misclassified as a subagent for ~25 minutes — until transcript evidence of \`wait_for_messages\` calls accumulated. This affected all restart paths (both modes).

## What changed

**\`scripts/claude-wrapper.exp\`**: Added startup Telegram alert before \`spawn claude\`. Reads \`TELEGRAM_BOT_TOKEN\` and \`TELEGRAM_ALLOWED_USERS\` from \`~/lobster-config/config.env\` at runtime via \`grep\` (same approach as \`send_telegram_alert()\` in \`claude-persistent.sh\`) and fires a \`curl\` POST. Message is labeled "(debug mode)" to distinguish from production restarts.

**\`src/cli\` (the \`lobster\` CLI)**: \`cmd_stop\` now deletes \`~/messages/config/dispatcher-session-id\` after stopping services. The direct-stop block in \`cmd_restart\` does the same (since \`cmd_restart\` deliberately avoids calling \`cmd_stop\` to skip the maintenance flag). This is the primary defence against the stale-marker bug.

**\`hooks/write-dispatcher-session-id.py\`**: Added a secondary safety net (\`_stored_session_is_alive\`) for crash/OOM restarts that bypass \`lobster stop\`. If the stored session's \`.jsonl\` file no longer exists under \`~/.claude/projects/\`, the stored session has ended; the new session is classified as the replacement dispatcher rather than a subagent. Falls back conservatively (assume alive) if the projects directory is unreadable.

## What is not changed

- Normal dispatcher behaviour when the marker file is absent or matches the current session is unchanged.
- Subagent auto-registration logic is unchanged.
- \`claude-persistent.sh\` is unchanged — it already had the startup alert.
- No changes to \`session_role.py\` or any Stop/SubagentStop hooks.

## Functional patterns

Pure functions (\`_stored_session_is_alive\`, \`_is_dispatcher_session\`) with no side effects; side effects isolated to \`main()\`. Decision table replaces implicit boolean arithmetic.

## Tests run

- [x] \`python3 -c "import ast; ast.parse(open('hooks/write-dispatcher-session-id.py').read()); print('OK')"\` — syntax clean
- [ ] \`uv run pytest tests/unit/\` — skipped: no unit tests exist specifically for this hook; running the full suite requires the installed venv and MCP servers
- [ ] Live restart test — blocked: requires production session; safe to merge, behaviour is additive (alert fires on start, marker clears on stop)

**Blocked items needing attention before merge:** none — the primary fix (clearing the marker on stop) is a one-liner delete; the alert is fire-and-forget with \`catch {}\` so a curl failure cannot block startup.

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -66,3 +66,7 @@ connectors/whatsapp/.env.example:Phone number
 connectors/whatsapp/index.js:Email address
 scripts/install-whatsapp-connector.sh:Email address
 src/bot/whatsapp_bridge_adapter.py:Email address
+
+# claude-wrapper.exp reads TELEGRAM_BOT_TOKEN from config.env at runtime using
+# grep — this is a variable reference, not a hardcoded credential value
+scripts/claude-wrapper.exp:Hardcoded token
--- a/hooks/write-dispatcher-session-id.py
+++ b/hooks/write-dispatcher-session-id.py
@@ -8,21 +8,34 @@ LOBSTER_MAIN_SESSION=1 (the dispatcher and all subagents it spawns).
 
 ## Dispatcher path
 
-When the marker file is absent (fresh dispatcher start) or the current
-session_id matches the stored dispatcher ID, writes session_id to
-~/messages/config/dispatcher-session-id. This file is the primary signal
-that other hooks use to distinguish the dispatcher from subagents.
+When the marker file is absent (fresh dispatcher start), the current
+session_id matches the stored dispatcher ID, or the stored dispatcher session
+has ended (its .jsonl file is gone from ~/.claude/projects/), writes
+session_id to ~/messages/config/dispatcher-session-id. This file is the
+primary signal that other hooks use to distinguish the dispatcher from
+subagents.
 
 ## Subagent path
 
-When the marker file exists and the current session_id does NOT match the
-stored dispatcher ID, this session is a subagent. In that case, a minimal
-stub record is written to agent_sessions.db with status='running'. This
-ensures subagents are visible to the ghost detector even when the dispatcher
-forgets to call register_agent (e.g. after context compaction). The stub uses
-INSERT OR IGNORE so a racing register_agent call that writes a richer row
-first is preserved untouched. DB write failures are logged to stderr but
-never block session start (always exits 0).
+When the marker file exists, the current session_id does NOT match the stored
+dispatcher ID, AND the stored session is still alive (its .jsonl file exists),
+this session is a subagent. In that case, a minimal stub record is written to
+agent_sessions.db with status='running'. This ensures subagents are visible to
+the ghost detector even when the dispatcher forgets to call register_agent
+(e.g. after context compaction). The stub uses INSERT OR IGNORE so a racing
+register_agent call that writes a richer row first is preserved untouched. DB
+write failures are logged to stderr but never block session start (always
+exits 0).
+
+## Stale marker recovery
+
+If the dispatcher crashes or is killed without a clean `lobster stop`, the
+marker file retains the dead session's ID. On the next start, the new
+dispatcher's session_id won't match and would normally be misclassified as a
+subagent. This hook detects that the stored session's .jsonl file no longer
+exists and correctly classifies the new session as the replacement dispatcher.
+The primary defence against this scenario is `lobster stop` (and `restart`)
+clearing the marker file; this check is a secondary safety net.
 
 ## settings.json configuration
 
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Verified all three flagged concerns. No fixes required.

---

**1. Tcl/expect syntax in `claude-wrapper.exp`**

Both constructs are valid Tcl idioms:

- `catch { exec curl ... }` — `catch` takes a script body and suppresses non-zero exit exceptions from `exec`. The backslash-newline line continuations inside `{}` braces are valid Tcl (backslash-newline collapses to whitespace). This is the correct way to make a best-effort curl call in Expect.
- `string trim $bot_token` / `string trim $chat_id` — standard Tcl built-in, strips leading/trailing whitespace. Correct placement: called after `exec sh -c "grep ..."` captures the value, before the emptiness check.
- `[set config_file]` inside the double-quoted `exec sh -c` string — correct Tcl variable substitution within a string; equivalent to `$config_file` in this context.

**2. Bash variable scoping in `src/cli` restart path**

No scoping bug. `cmd_restart` (line 183) declares its own independent `local dispatcher_session_file="$MESSAGES_DIR/config/dispatcher-session-id"` — it does NOT reference `cmd_stop`'s local. Both functions independently construct the same path from `$MESSAGES_DIR`. The `rm -f` in `cmd_restart` correctly resolves the variable in its own scope.

**3. `_stored_session_is_alive` glob: `*/<session-id>.jsonl`**

Logic is correct. The glob searches for a file named exactly `<stored_session_id>.jsonl` (the dispatcher's UUID) in any immediate subdirectory of `~/.claude/projects/`. A subagent session file would have a different UUID as its filename; there is no mechanism by which the stored dispatcher session ID could collide with a subagent's session ID. The glob cannot produce a false positive match.

sayhar (commented):

## Architectural Assessment

**Is this the right approach?** Yes. The three-part design (CLI clears marker on stop, wrapper fires startup alert, hook adds crash-recovery) applies defense-in-depth correctly: the primary path (clean stop/restart) is a simple delete; the secondary path (crash recovery) is a read-only filesystem check with a conservative fallback. No new state is introduced — the hook inspects already-existing Claude Code `.jsonl` files.

**Scope:** Appropriately tight. Four files, no behavior changes to the happy path, no new persistent state.

---

## Edge Cases the Design Misses

**1. `lobster stop` when `$MESSAGES_DIR` is unset.**
`cmd_stop` does `if [[ -f "$dispatcher_session_file" ]]` before deleting — safe. But if `$MESSAGES_DIR` is empty/unset, `dispatcher_session_file` becomes `/config/dispatcher-session-id`, a path that almost certainly won't exist, so the check silently becomes a no-op. A `[[ -n "$MESSAGES_DIR" ]]` guard would make this explicit. Low risk.

**2. Crash-recovery race: two Claude sessions start simultaneously.**
If two instances start at exactly the same moment (e.g. a systemd double-restart), both see the dead JSONL gone and both conclude they are the dispatcher. The hook has no locking. In practice systemd serializes restarts; the prior behavior had the same race; this is not a regression.

**3. `~/.claude/projects/` is an internal Claude Code implementation detail.**
If a future Claude Code version moves session files, the fallback (`return True`) restores the old 25-minute regression rather than causing a crash. Acceptable conservative behavior.

**4. Tcl surrogate-pair emoji on non-UTF-8 locales.**
`\uD83D\uDD04` is correct UTF-16 surrogate notation; Tcl 8.5/8.6 on UTF-8 systems handles this fine. On non-UTF-8 locales (rare for a server install) the alert text may be garbled. The `catch {}` means it won't break startup. Low risk.

---

## New Failure Modes

None introduced on the critical path. All new code uses `catch {}` (Tcl), `|| true` (bash), or conservative fallbacks (Python). The hook always exits 0. On a partial `cmd_restart` failure (marker cleared, new session not yet started), the system is left in the correct state for the next start.

---

## PR Description Check

- Explains WHY: yes (25-minute misclassification window documented clearly)
- References issue number: implicit via branch name (`feature/issue-interactive-mode-parity`) — no explicit `Closes #NNN` line. Minor gap, does not block merge.
- Starts with `🤖🦞 Lobster (engineer):`: yes
- Notes what was tested: yes (syntax check documented; skipped items explained)

---

## Docker Install Test

Ran: `sudo docker compose -f ~/lobster/tests/docker/docker-compose.test.yml up install-test`
Result: **exited with code 0** — PASS

---

## Verdict

**PASS — ready to merge pending user approval.** Two minor follow-up items (unset `$MESSAGES_DIR` guard, explicit issue close reference) do not block merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): Follow-up — edge cases addressed. PASS ✓

- **`$MESSAGES_DIR` guard in `cmd_stop` and `cmd_restart`**: Added `[[ -n "$MESSAGES_DIR" ]]` guard in both locations (commit c6d47c5). The `cmd_stop` block now wraps the file check in the guard; `cmd_restart` uses the compact inline form.
- **Tcl surrogate-pair emoji on non-UTF-8 locales**: Added `encoding system utf-8` before the curl call in `claude-wrapper.exp` (same commit). Emoji in the startup alert will now encode correctly regardless of the system locale.

Docker install-test: exit 0. Unit tests: 1187 passed, 68 pre-existing failures (identical count on main, not introduced by this PR).

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR achieves interactive mode (debug/claude-wrapper.exp) parity with persistent mode startup, plus fixes a stale-marker crash scenario.

**What the code does (three changes):**

1. **`hooks/write-dispatcher-session-id.py` — stale marker recovery**: Adds `_stored_session_is_alive()` which checks whether the stored dispatcher session's JSONL file still exists under `~/.claude/projects/`. If the file is gone, the stored session ended (crash/OOM without clean stop), and the new session should be treated as the replacement dispatcher rather than a subagent. The fallback is conservative (assume alive) to avoid misclassifying legitimate subagents.

2. **`scripts/claude-wrapper.exp` — startup Telegram alert**: Adds a Telegram startup notification that fires before Claude is spawned, mirroring what `claude-persistent.sh` already does in persistent mode. Reads `TELEGRAM_BOT_TOKEN` and `TELEGRAM_ALLOWED_USERS` from `config.env` at runtime using `grep`. Wrapped in `catch` so a failure never blocks startup.

3. **`src/cli` — clear dispatcher session marker on stop/restart**: Both `cmd_stop` and `cmd_restart` now delete `dispatcher-session-id` on clean stop. This is the primary defence against the stale-marker scenario; the JSONL check in (1) is the secondary net.

**Why it matters:** Without the marker clear, a `lobster stop` + `lobster start` cycle left the old session ID in the marker file. The new dispatcher (different session ID) would be misclassified as a subagent for the ~25-minute window before the JSONL file was cleaned up by Claude Code. This caused subagent stub rows to be written for the dispatcher, polluting the agent_sessions DB.

**Observations:**
- The `_stored_session_is_alive` glob (`*/session_id.jsonl`) searches all workspace subdirectories. If a user has many workspaces this could be slow, but it is only called at session start (once).
- The security allowlist entry for `claude-wrapper.exp` correctly explains why the grep-based token read is not a hardcoded credential.
- The `[[ -n "$MESSAGES_DIR" ]]` guard in cli against `/config/dispatcher-session-id` is a sensible safety check.

PR #633, opened 2026-03-18
fix: remove truncated preview from write_result debug alerts

sayhar:
When \`LOBSTER_DEBUG=true\`, \`write_result\` emits a debug observation message to the Telegram inbox so the user sees subagent activity in real time. That message included a \`preview:\` line containing the first 60 characters of the result text. When a result begins with or contains a URL, this truncation cuts the URL mid-string — Telegram's link parser then renders the partial URL as a broken clickable link.

The preview adds no information the dispatcher doesn't already receive: the full result text lands in the inbox as the \`subagent_result\` payload immediately after. Removing the line keeps the debug alert to the five fields that matter: type, task, agent (if set), status (if set), and sent_reply.

No other debug observation paths are touched. The \`write_observation\` and \`memory_store\` debug alerts are unaffected — they don't include a preview field.

**Functional pattern:** pure deletion — the remaining alert construction is still a plain list-build-then-join with no replacement logic.

## Tests run
- [x] \`git diff HEAD~1\` — verified only the two lines (text_preview assignment + preview append) are removed, nothing else
- [ ] Live debug-mode Telegram test — skipped: requires a running Lobster instance with \`LOBSTER_DEBUG=true\`; safe to merge without it (no logic change, only omission of a field)

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4881,7 +4881,6 @@ async def handle_write_result(args: dict) -> list[TextContent]:
     # so the user sees the subagent message arrive in real time.
     # _emit_debug_observation is a no-op when debug alerts are disabled — single gate.
     agent_id = args.get("agent_id", "").strip() or None
-    text_preview = text[:60] + "…" if len(text) > 60 else text
     alert_lines = [
         f"\U0001f4e8 [subagent\u2192dispatcher] type: {msg_type}",
         f"task: {task_id}",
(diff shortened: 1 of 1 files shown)


sayhar (commented):

## Architectural approach

Pure deletion of a field that was adding noise and causing harm (truncated URLs rendered as broken Telegram links). The remaining debug alert retains all five fields that matter for dispatcher-side observability: type, task, agent, status, sent_reply. The `preview` field was genuinely redundant — the full result text arrives in the inbox payload immediately after.

This is the correct fix. Adding truncation logic (smarter truncation, URL-aware truncation, etc.) would be over-engineering a field that provides no value the dispatcher doesn't already have.

## Edge cases

None relevant. This is a pure omission in a debug-only code path gated by `LOBSTER_DEBUG=true`. The alert construction is a list-build-then-join with no conditional logic depending on the removed field. The deletion is clean.

## New failure modes

None. Removing a field from a debug observation message cannot affect production message handling.

## Scope

Minimal and correct. Two lines removed, nothing added. The diff is exactly as described in the PR body.

**Tests:** 65 pre-existing failures on main, 65 failures on this branch — no regressions.

PR #638, opened 2026-03-18
fix: resolve pre-existing failures in test_message_tools and test_transcription

sayhar:
## Summary

Three categories of drift had accumulated between tests and production code, causing 10 failures in these two files:

- **ValidationError propagation** (3 tests): `handle_send_reply` and `handle_mark_processed` raised `reliability.ValidationError` on bad input instead of returning a `TextContent` error response. MCP tool handlers must never raise — callers expect `list[TextContent]` back. Fixed by wrapping the validation calls in `try/except` in both handlers.

- **Auto-threading removed** (4 tests): `TestAutoThreading` expected `reply_to_message_id` to be auto-populated from `processing/` and `processed/` directories. That behaviour was intentionally removed in PRs #361 and #420 to prevent replies threading under the wrong message when multiple messages are in-flight. Updated tests document the removal and assert that `reply_to_message_id` is absent when not explicitly provided.

- **mark_processed guard for test chat_ids** (1 test): The soft-warning "No reply sent" return path was replaced with an auto-send-or-skip pattern. For fake/test chat_ids (≤ 1_000_000), Telegram rejects delivery so the guard skips the auto-reply fallback and still marks the message processed. Test updated to match.

- **Transcription API changed** (2 tests): Tests patched `get_whisper_model` (Python whisper library) but transcription now goes through `run_whisper_cpp`, a whisper.cpp CLI wrapper returning `(success: bool, text: str)`. Tests updated to patch `run_whisper_cpp` with an `AsyncMock`.

## Functional patterns

- Pure error-return pattern in handlers (no exceptions crossing handler boundaries)
- Immutable test fixtures — all assertions read from files written by the code under test

## Tests run

- [x] `uv run python -m pytest tests/unit/test_mcp_server/test_message_tools.py tests/unit/test_mcp_server/test_transcription.py -v` — 56 passed, 0 failed (was 10 failed)
- [x] `uv run python -m pytest tests/unit/ --tb=no -q` — confirmed no regressions; all failures in broader suite are pre-existing on main and outside this PR's scope

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -3193,8 +3193,12 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
 
 async def handle_send_reply(args: dict) -> list[TextContent]:
     """Send a reply to a message with input validation."""
-    # Validate inputs (raises ValidationError on bad data)
-    args = validate_send_reply_args(args)
+    # Validate inputs — return error TextContent instead of raising so callers
+    # (and tests) always receive a list[TextContent] back.
+    try:
+        args = validate_send_reply_args(args)
+    except ValidationError as e:
+        return [TextContent(type="text", text=f"Error: {e}")]
     chat_id = args["chat_id"]
     text = args["text"]
     source = args["source"]
@@ -3361,7 +3365,10 @@ async def handle_send_sms_reply(args: dict) -> list[TextContent]:
 
 async def handle_mark_processed(args: dict) -> list[TextContent]:
     """Mark a message as processed."""
-    message_id = validate_message_id(args.get("message_id", ""))
+    try:
+        message_id = validate_message_id(args.get("message_id", ""))
+    except ValidationError as e:
+        return [TextContent(type="text", text=f"Error: {e}")]
     force = args.get("force", False)
 
     # Check processing/ first, then inbox/ as fallback
--- a/tests/unit/test_mcp_server/test_message_tools.py
+++ b/tests/unit/test_mcp_server/test_message_tools.py
@@ -300,7 +300,13 @@ class TestAutoThreading:
         (processing_dir / "12345_msg.json").write_text(json.dumps(msg))
 
     def test_auto_threads_when_processing_message_exists(self, dirs):
-        """When no reply_to_message_id is given, threading uses the message in processing/."""
+        """Auto-threading was removed (PR #420): reply is standalone even when processing/ has a message.
+
+        Previously send_reply would look up telegram_message_id from processing/ and set
+        reply_to_message_id automatically.  That behaviour was removed to avoid threading
+        under the wrong message when multiple messages are in-flight.  Without an explicit
+        reply_to_message_id the reply is now sent standalone.
+        """
         outbox, processing, processed, sent = dirs
         self._make_processing_msg(processing, chat_id=123456, tg_msg_id=9001)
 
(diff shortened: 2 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — changes are correct and well-reasoned, with two minor notes.

---

## Overview

Three categories of real code drift are resolved cleanly. The test count went from 10 failed → 0, and the PR description accurately explains why each category changed.

---

## Per-category assessment

### 1. ValidationError → TextContent wrapping (`inbox_server.py`)

**Correct approach.** MCP handlers must return `list[TextContent]` — raising across the handler boundary is the wrong contract. The try/except is minimal and targeted. One note: the error string format `f"Error: {e}"` is fine for now, but if `ValidationError` ever carries a structured message with internal details, that will leak to callers. If `ValidationError` messages are already user-safe (no stack paths, no internal state), no action needed.

The concern raised in the brief — "callers that expected exceptions would be affected" — is a non-issue in practice: MCP callers never catch tool-level exceptions; they read `TextContent`. Any test that was `pytest.raises(ValidationError)` was testing the wrong contract.

### 2. Auto-threading test inversion (`TestAutoThreading`)

**Correctly inverted.** The four test docstrings are accurate and specific — they name the PR numbers (#361, #420) and explain the rationale (in-flight message ambiguity). The assertion flip from `== 9001` to `not in content` is the right way to document intentional removal. No issues.

### 3. mark_processed guard behavior (`test_blocks_human_message_without_reply`)

**Correct.** The new assertion (`"processed" in result[0].text.lower()` + file moved to `processed/`) matches the updated guard behavior for small/fake chat_ids. The inline comment `# fake/test chat_id — guard skips auto-reply` is helpful. One observation: the `<= 1_000_000` threshold is a magic number that appears to be defined somewhere in `inbox_server.py` — if that threshold ever changes, this test will silently use a chat_id that's now on the wrong side of the boundary. Worth a constant reference if one exists.

### 4. Transcription API migration (`test_transcription.py`)

**Clean.** Patching `run_whisper_cpp` with `AsyncMock(return_value=(True, "..."))` correctly mirrors the `(success: bool, text: str)` return type. The error path test uses `side_effect=Exception(...)` which covers the subprocess-crash scenario accurately.

---

## Summary

No blocking issues. The two minor points (ValidationError message contents, magic threshold constant) are low-risk observations, not blockers. The PR is ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR fixes pre-existing test failures in two test files by aligning them with behavioral changes that had already shipped but whose tests were never updated.

**What the code does:**

1. **test_message_tools.py**: Updates 6 tests that were asserting auto-threading behavior (auto-populating `reply_to_message_id` from `processing/` or `processed/` directories). That behavior was removed in PRs #361 and #420, but the tests still expected the old behavior. The tests are updated to assert that `reply_to_message_id` is NOT set. Also updates the 'blocks human message without reply' test to reflect the current behavior: for test/fake chat_ids (≤1_000_000), the guard skips the auto-reply but still marks the message processed.

2. **test_transcription.py**: Updates 2 tests that mocked `get_whisper_model` / `model.transcribe()`. The transcription backend was replaced with `run_whisper_cpp` (returning a `(success, text)` tuple from an async call), but the tests were still mocking the old API. Updated to mock `run_whisper_cpp` as `AsyncMock`.

3. **inbox_server.py (send_reply + mark_processed)**: Changes ValidationError raises to return error `TextContent` instead, so callers always get `list[TextContent]` back. This is what allows the tests to run without exception propagation.

**Why it matters:** Failing tests mask real regressions. These tests were failing silently (or being skipped), leaving the test suite in a misleading green state for the behaviors they're supposed to cover.

**Observations:** The ValidationError → TextContent change in inbox_server.py is a minor API behavior change at the tool layer. It's correct for MCP tools (which should always return `list[TextContent]`), but callers that previously relied on catching the exception will now get a text response instead. This should be fine since MCP callers don't typically catch tool exceptions.

PR #640, opened 2026-03-18
fix: write booted_at on hibernate wakeup to activate boot grace period

### LINKED ISSUE #639: Hibernate wakeup lacks boot grace period, causing false-positive stale-inbox restarts
## Problem

When Claude exits cleanly into hibernate mode, `claude-persistent.sh` waits for new inbox messages (`wait_for_wake()`), then relaunches Claude. However, it does **not** call `write_boot_stamp()` before relaunching.

The health check (`health-check-v3.sh`) only activates the `BOOT_GRACE_SECONDS=90` suppression window when `booted_at` is present in `lobster-state.json`. Without a fresh `booted_at`, the health check treats a hibernate wakeup the same as a system that's been running for hours — no grace period.

On a fresh startup, reading ~60KB of bootup files can take >4 minutes. During this time, the stale-inbox check fires and triggers an unnecessary restart.

## Root Cause

`write_boot_stamp()` is called once in `main()` at the very start:

```bash
preflight
write_boot_stamp  # called here on initial boot
```

But the main loop's hibernate-restart path does not call it:

```bash
if handle_exit "$exit_code"; then
    # Clean hibernation - wait for new messages
    wait_for_wake
    attempt=0  # Reset attempt counter after clean cycle
    # write_boot_stamp() NOT called here before next launch_claude()
```

## Fix

Call `write_boot_stamp()` after `wait_for_wake()` returns (and before the next `launch_claude()` runs) so the health check suppresses false-positive alerts during the wakeup initialization window.

## Files

- `scripts/claude-persistent.sh` — one-line fix in the main loop

sayhar:
## What this fixes

Hibernate wakeups had no boot grace period, so the health check's stale-inbox check could fire during the 60-90 seconds it takes a fresh Claude session to initialize. This caused unnecessary restarts immediately after every hibernate wakeup.

The existing `BOOT_GRACE_SECONDS=90` suppression in `health-check-v3.sh` works correctly for initial boots and health-check-initiated restarts, because both paths call `write_boot_stamp()` before launching Claude. The hibernate wakeup path was the only one that didn't.

## Root cause

`write_boot_stamp()` was called once in `main()` on the very first boot, but the main loop's hibernate branch — after `wait_for_wake()` returns — proceeded directly to the next `launch_claude()` call without writing a fresh `booted_at` timestamp. The health check's `is_boot_grace_period()` function reads `booted_at` from `lobster-state.json` and returns false (no grace) when the field is absent or stale, so wakeup launches got no protection.

## Change

One call added to `scripts/claude-persistent.sh` — `write_boot_stamp()` immediately after `wait_for_wake()` returns, before the loop continues to `launch_claude()`. The function already exists and does a read-modify-write, preserving all other fields in `lobster-state.json`. No changes to the health check or state file format.

The fix is intentionally minimal: the same function the initial boot uses, called in the one place that was missing it.

## Functional patterns

Pure function reuse — `write_boot_stamp()` is a side-effectful boundary function already tested by the initial boot path; the fix simply calls it in the missing context rather than duplicating logic.

## Tests run

- [x] `bash -n scripts/claude-persistent.sh` — syntax check passed
- [ ] Live hibernate/wakeup cycle — requires the running lobster-claude service; safe to verify post-merge by watching `lobster-state.json` for a fresh `booted_at` after the next hibernate wakeup

Closes #639

### DIFF
--- a/scripts/claude-persistent.sh
+++ b/scripts/claude-persistent.sh
@@ -521,6 +521,11 @@ Claude persistent session initializing."
         if handle_exit "$exit_code"; then
             # Clean hibernation - wait for new messages
             wait_for_wake
+            # Write a fresh boot stamp so the health-check's BOOT_GRACE_SECONDS
+            # suppression window activates for the upcoming wakeup launch.
+            # Without this, hibernate wakeups have no grace period and the
+            # stale-inbox check fires during the ~60-90s initialization window.
+            write_boot_stamp
             attempt=0  # Reset attempt counter after clean cycle
         else
             # Abnormal exit - brief pause before restart


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Placement is correct.** `write_boot_stamp()` is called immediately after `wait_for_wake()` returns and before the loop continues to `write_state("active"...)` + `launch_claude()`. This is exactly the right point in the flow: the health check's `is_boot_grace_period()` reads `booted_at` before it checks for a running process or drained inbox, so the stamp must exist before `launch_claude()` fires.

**State sequence is coherent.** The flow through `wait_for_wake()` already sets the state to `"waking"` on return. Then `write_boot_stamp()` updates `booted_at` (touching only that field), and the loop top immediately calls `write_state("active", ...)`. No state is clobbered or left inconsistent.

**Read-modify-write is safe.** The python3 block does `json.load`, sets `d['booted_at']`, and writes back — all other fields are preserved. The `except Exception: d = {}` fallback on a corrupted file is acceptable; it loses prior state but that's a tolerable edge case for an otherwise-corrupt file. The outer `|| true` suppresses errors silently, which is fine since `log "Boot stamp written..."` is only reached on the success path. One minor note: a silent failure here (e.g., disk full) would leave `booted_at` stale and the grace window inactive — but this is pre-existing behavior, not a regression introduced by this PR.

**No unexpected writes.** `write_boot_stamp()` is now called in exactly two places: once at initial boot in `main()` and once here after `wait_for_wake()`. There is no path through which it would be called on abnormal exits or rapid-restart backoffs.

**`install.sh` does not need updating.** The fix is a single new call to an existing function; there are no new files, directories, state keys, service definitions, or config fields that require migration.

The change is minimal, correct, and consistent with the existing boot path. No issues found.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR fixes a missing boot grace period on hibernate wakeup, which was causing the health check's stale-inbox detector to fire during Lobster's initialization window after waking from hibernation.

**What the code does:** In `scripts/claude-persistent.sh`, after `wait_for_wake` returns (i.e., the bot has been woken by a new message), `write_boot_stamp` is called before resetting the attempt counter. This writes a timestamp that activates the health check's `BOOT_GRACE_SECONDS` suppression window — the same window that protects a fresh startup from being immediately restarted due to inbox messages that accumulate during the ~60–90s initialization period.

**Why it matters:** Without this call, a hibernate wakeup has no grace period. Lobster wakes, begins re-initializing, and the health check (running every 2 minutes) may see a stale inbox message and trigger a restart before initialization completes. The fix ensures wakeups are treated identically to fresh starts for the purpose of the boot grace window.

**Observations:** The change is minimal and correctly placed — immediately after `wait_for_wake` and before `attempt=0` reset, which is the right ordering. The comment explains the rationale clearly. No tests changed since this is shell script behavior.

PR #642, opened 2026-03-18
fix: test isolation — prevent tests from corrupting production hibernate state and logs

### LINKED ISSUE #641: fix: test isolation bugs in inbox_server.py allow tests to corrupt production state
## Summary

Three related bugs allow pytest test runs to corrupt production state files and write to production log files.

### Bug 1 (HIGH): `handle_wait_for_messages` writes hibernate state without session guard

`handle_wait_for_messages(timeout, hibernate_on_timeout=True)` writes `mode=hibernate` to `LOBSTER_STATE_FILE` on timeout with no check against `_is_main_session()`. When pytest imports `inbox_server` and calls this function directly (not via the MCP socket), it writes to the real production state file.

**Fix:** Guard the `_write_lobster_state` call inside `handle_wait_for_messages` with `if _is_main_session()`.

### Bug 2 (HIGH): Module-level log handler setup writes to production `mcp-server.log` on import

`inbox_server.py` attaches a `RotatingFileHandler` at module level. Any test process that imports `inbox_server` also writes to the production log file.

**Fix:** Move file handler setup into a `setup_logging()` function called only from the `__main__` block.

### Bug 3 (MEDIUM): `patch.multiple` in `test_hibernation.py` doesn't reliably protect production

The `import inbox_server` inside the `with patch.multiple(...)` block runs after the patch context opens, but if the module is not yet cached the patch may not apply to the newly-imported module.

**Fix:** Move all `import inbox_server` calls outside the `patch.multiple` context so the module is cached before patching.

## Files

- `src/mcp/inbox_server.py`
- `tests/unit/test_hibernation.py`

sayhar:
## Problem

Running \`pytest tests/unit/test_hibernation.py\` on a production machine had two failure modes that could damage the live system:

1. The test put the production system into hibernate mode. \`handle_wait_for_messages\` wrote \`mode=hibernate\` to the real \`LOBSTER_STATE_FILE\` with no check for whether it was running in the main session. The \`patch.multiple\` in the test was supposed to redirect this write, but it didn't always fire (see #3 below).

2. Importing \`inbox_server\` in any test process attached a \`RotatingFileHandler\` at module level, writing test-generated log entries to \`~/lobster-workspace/logs/mcp-server.log\`.

3. \`patch.multiple("inbox_server", ...)\` opened before \`import inbox_server\` ran inside each test method. If \`inbox_server\` was not yet in \`sys.modules\` when the patch context opened, the patch targeted nothing — or targeted an object that was immediately replaced by the fresh import inside the \`with\` block.

## What changed

**\`src/mcp/inbox_server.py\`**

- \`handle_wait_for_messages\`: the \`_write_lobster_state(..., "hibernate")\` call is now gated on \`_is_main_session()\`. Test processes that import the module directly and call this function cannot set production state to hibernate. The existing session guard used by other tools is the right mechanism here; this change applies the same pattern to the hibernate write path.

- Log handler setup extracted into \`setup_logging()\`, called only from the \`__main__\` block. The \`logging.getLogger\` call stays at module level so callers get a usable logger — it just has no file handler attached unless running as the server.

**\`tests/unit/test_hibernation.py\`**

- \`inbox_server\` is now imported once at module level (after \`sys.path\` setup) so \`patch.multiple("inbox_server", ...)\` always targets the cached module object, not a freshly-loaded one.

- All inline \`sys.path.insert\` + \`from inbox_server import X\` patterns inside test methods are replaced with references to the module-level import.

- Tests in \`TestWaitForMessagesHibernation\` that exercise the hibernate state write now patch \`_is_main_session\` to return \`True\`, since the production guard that was missing before is now present and correct.

## Functional patterns

Pure function composition: \`setup_logging()\` is a side-effect–isolating wrapper with no return value. The session guard is a simple predicate. The test patches follow the same mock-at-boundary pattern already used throughout the suite.

## Tests run

- [x] \`uv run pytest tests/unit/test_hibernation.py -v\` — 28 passed, 0 failed
- [x] \`uv run pytest tests/unit/ -v\` — 1156 passed, 71 failed; all 71 failures are pre-existing on \`main\` (confirmed by running the same suite against main — same failures, same count)

Closes #641

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -623,14 +623,23 @@ LOG_DIR.mkdir(parents=True, exist_ok=True)
 
 log = logging.getLogger("lobster-mcp")
 log.setLevel(logging.INFO)
-_file_handler = RotatingFileHandler(
-    LOG_DIR / "mcp-server.log",
-    maxBytes=5 * 1024 * 1024,  # 5MB
-    backupCount=3,
-)
-_file_handler.setFormatter(logging.Formatter("%(asctime)s [%(levelname)s] %(message)s"))
-log.addHandler(_file_handler)
-log.addHandler(logging.StreamHandler())
+
+
+def setup_logging() -> None:
+    """Attach file and stream handlers to the module logger.
+
+    Called only when running as a server (from the __main__ block), not when
+    imported as a library by tests or other modules. This prevents test runs
+    from writing to the production mcp-server.log file.
+    """
+    _file_handler = RotatingFileHandler(
+        LOG_DIR / "mcp-server.log",
+        maxBytes=5 * 1024 * 1024,  # 5MB
+        backupCount=3,
+    )
+    _file_handler.setFormatter(logging.Formatter("%(asctime)s [%(levelname)s] %(message)s"))
+    log.addHandler(_file_handler)
+    log.addHandler(logging.StreamHandler())
 
 # Seed canonical templates on startup (idempotent — only copies missing files)
 def _seed_canonical_templates():
@@ -2773,9 +2782,15 @@ async def handle_wait_for_messages(args: dict) -> list[TextContent]:
             log.info(f"wait_for_messages timed out after {timeout}s")
 
             if hibernate_on_timeout:
-                # Write hibernate state so the bot knows to wake us on next message
-                _write_lobster_state(LOBSTER_STATE_FILE, "hibernate")
-                log.info("Hibernating: wrote state=hibernate, signalling graceful exit")
+                # Write hibernate state so the bot knows to wake us on next message.
+                # Only write when running in the main session — tests that import this
+                # module and call handle_wait_for_messages directly must not corrupt
+                # the production state file.
+                if _is_main_session():
+                    _write_lobster_state(LOBSTER_STATE_FILE, "hibernate")
+                    log.info("Hibernating: wrote state=hibernate, signalling graceful exit")
+                else:
+                    log.info("Hibernating: skipping state write (not main session)")
                 return [TextContent(
                     type="text",
                     text=(
@@ -6899,4 +6914,5 @@ async def main():
 
 if __name__ == "__main__":
     import asyncio
+    setup_logging()
     asyncio.run(main())
--- a/tests/unit/test_hibernation.py
+++ b/tests/unit/test_hibernation.py
@@ -20,6 +20,14 @@ from unittest.mock import MagicMock, patch, call
 
 import pytest
 
+# Ensure inbox_server is importable. Must happen before any test class references
+# the module so that patch.multiple("inbox_server", ...) always targets the cached
+# module object rather than a freshly-imported one (which would not yet be patched).
+_MCP_SRC = str(Path(__file__).parent.parent.parent / "src" / "mcp")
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with two minor notes

## Overall verdict

The three bugs are real, the fixes are correct, and the test changes are clean. Good to merge as-is; the notes below are low-priority follow-ups, not blockers.

---

## 1. `_is_main_session()` as the hibernate guard — is it the right choice?

Yes. The function uses tmux ancestry as its primary check (unforgeable by a test process) and falls back to `LOBSTER_MAIN_SESSION=1`. Both failure modes fail closed (returns `False`), which is the right default for a state-mutating write: a test process that can't confirm it is the main session must not write production hibernate state. There are no known unintended consequences of failing closed here — the only side-effect of skipping the write in a non-main session is that the log says "skipping state write (not main session)", which is exactly what you want.

The guard is consistent with how `_SESSION_GUARDED_TOOLS` already protects other tools (line 2453), so this change follows an established pattern rather than introducing a novel one.

One thing worth noting: the cache (`_main_session_cache`) is module-level and survives across test runs within the same process. If a future test ever sets `LOBSTER_MAIN_SESSION=1` in the environment and then imports `inbox_server`, the cache will lock in `True` for every subsequent test in that process. The current test file correctly patches `_is_main_session` at the call site rather than relying on the env var, which avoids this trap — but the fragility is worth keeping in mind if new test authors try a different approach.

## 2. `setup_logging()` coverage

**The `__main__` path is covered.** The systemd unit for the main lobster instance runs `claude-wrapper.exp` → `python inbox_server.py`, which hits `__main__`, so production is fine.

**One gap: `inbox_server_http.py`** imports `inbox_server` as a library (`from inbox_server import server as _full_server, ...`) and is started by its own `__main__` block with `logging.basicConfig(level=logging.INFO)`. It does not call `inbox_server.setup_logging()`, which means the `lobster-mcp` logger has no file handler attached when running in HTTP mode. Whether that is intentional (HTTP mode uses systemd journal via `logging.basicConfig`, so journald captures everything) or an oversight depends on operational intent. At minimum, a comment in `inbox_server_http.py` clarifying "we intentionally don't attach the file handler here" would help future readers.

This is not a regression introduced by this PR (the HTTP server existed before), but the PR makes it easier to call `setup_logging()` where needed, so it is worth noting.

## 3. Conftest / isolation story

The existing `mcp_directories` fixture in `tests/conftest.py` redirects all the message inbox/outbox/processing dirs via `patch.multiple`, so production inbox dirs are already protected for tests that use that fixture. The hibernation test class uses its own `dirs` fixture that redirects the relevant paths directly. This is complete for the paths exercised by `test_hibernation.py`.

One residual gap that predates this PR: the conftest `mcp_directories` fixture patches `"mcp.inbox_server"` (with a package prefix) while `test_hibernation.py` patches `"inbox_server"` (bare). These refer to the same module object only if it was imported under the bare name first (which the module-level import in the test file ensures). This is fine now, but the inconsistency between fixture namespaces could confuse future contributors. Not a blocker.

## 4. The 71 pre-existing failures

The PR description says these were confirmed against `main` with the same count. That is sufficient due diligence. No action needed.

---

**Summary:** merge when ready.

PR #647, opened 2026-03-18
fix: stop startup cleanup from killing agents that survive server restarts

### LINKED ISSUE #645: Reconciler falsely kills healthy running agents after server restart
## Summary

Two related bugs cause healthy agents to be killed and reported as dead.

## Bug 1: Startup cleanup kills agents that survive server restarts

`cleanup_stale_running_sessions` (in `session_store.py`) assumes that any output file with an mtime predating the server start time cannot belong to a live agent. The docstring says:

> **Assumption:** subagents cannot outlive a server restart. Claude Code subagents run as child processes of the MCP server; when the server is killed, all subagents are killed with it.

**This assumption is wrong.** Claude Code agents run in their own processes (tmux panes, spawned via the `Agent` tool). They are NOT child processes of the MCP server. When the server is restarted (e.g. `systemctl restart lobster-inbox`), the agents keep running. Their output files keep getting updated.

The mtime check at startup (`file_mtime < server_start_time → mark dead`) incorrectly kills every running agent after every server restart.

### Trace

1. Agent runs, output file is `/tmp/claude-1000/.../tasks/<agent-id>.output` with mtime=T1
2. Server restarts at T2 > T1
3. `cleanup_stale_running_sessions` sees `file_mtime(T1) < server_start_time(T2)` → marks dead
4. `_startup_sweep()` finds the newly-dead session → enqueues `[SUBAGENT ERROR]` notification
5. Agent was still alive and completes successfully minutes later

## Bug 2: "0m" in notification message

When `_startup_sweep()` runs, it calls `get_unnotified_completed()` which returns rows via `_row_to_dict` — a plain dict with no `elapsed_seconds` field. When `_enqueue_reconciler_notification` reads `session.get("elapsed_seconds")`, it gets `None`, defaults to 0, and formats `elapsed_min = 0`. This produces the "no output file after 0m — marked dead" text that was the observable symptom.

## Fix

**For Bug 1:** The startup cleanup should not use mtime to infer liveness. After a server restart, the correct check is to read the `stop_reason` in the output file:
- If `stop_reason=end_turn` → mark completed (agent finished before or during restart)
- If `stop_reason=tool_use` or file is empty → leave as `running` (may still be alive)
- If output file is missing entirely → mark dead (file was cleaned up)

**For Bug 2:** Compute `elapsed_seconds` from `spawned_at` in `_enqueue_reconciler_notification` when `elapsed_seconds` is absent from the session dict.

## Observed symptoms

- Self-check shows agent at 50–100 turns, status "running"
- Minutes later: `[SUBAGENT ERROR]` — "Agent timed out / disappeared: ... (no output file after 0m — marked dead)"
- Agent completes successfully after being "killed"

sayhar:
Closes #645

## The problem

Every server restart was killing healthy running agents and sending false \`[SUBAGENT ERROR]\` notifications. Two bugs combined to produce the observed "no output file after 0m — marked dead" symptom.

**Bug 1 — Wrong liveness model in startup cleanup**

\`cleanup_stale_running_sessions\` assumed Claude Code agents are child processes of the MCP server and therefore cannot outlive a restart. That assumption is incorrect. Claude Code agents run independently (in their own process/tmux pane) and keep running when \`systemctl restart lobster-inbox\` is issued.

The function used an mtime heuristic: if \`output_file.mtime < server_start_time\`, mark dead. This fired for *every running agent* on *every restart*, since their output files were last written before the server came back up.

**Bug 2 — "0m" in notification text**

The startup sweep calls \`get_unnotified_completed()\`, which returns rows via \`_row_to_dict\` — a plain dict with no \`elapsed_seconds\` field. When \`_enqueue_reconciler_notification\` fell back to \`elapsed_seconds=None → 0\`, the notification read "no output file after 0m — marked dead" regardless of actual elapsed time. This "0m" was the diagnostic clue.

## The fix

**Startup cleanup** now reads \`stop_reason\` from the output file instead of checking mtime:
- \`stop_reason=tool_use\` (or empty file) → **leave running** — agent may still be alive
- \`stop_reason=end_turn\` → **mark completed** — agent finished before/during restart
- output file missing entirely → **mark dead** — file was cleaned up, no live agent can use it

The mtime check is removed entirely. The \`server_start_time\` parameter is retained (used for \`completed_at\` timestamps) but no longer used for liveness inference.

**Notification elapsed time** now falls back to computing from \`spawned_at\` when \`elapsed_seconds\` is absent from the session dict, so the "Xm elapsed" value is accurate regardless of which query path produced the session row.

## Functional patterns

Pure function \`_read_stop_reason_from_path\` (already existing) is reused for the startup check — the same deterministic file-read used by the reconciler loop. No new judgment paths, just a correct liveness signal at startup.

## Tests run

- [x] \`uv run pytest tests/test_agent_sessions.py -v\` — 36 passed, 0 failed
- [x] \`uv run pytest tests/unit/test_mcp_server/ -v\` — 18 pre-existing failures (confirmed identical on main), 0 new failures introduced
- [x] Pre-existing failure baseline confirmed via \`git stash\` + rerun — same 18 failures on unmodified main

**Tests replaced:** \`test_cleanup_stale_running_output_old_mtime\` (expected mtime-based kill — now wrong behavior) and \`test_cleanup_stale_running_skips_fresh_file\` (relied on mtime > server_start logic). Replaced with \`test_cleanup_stale_running_output_tool_use_left_running\`, \`test_cleanup_stale_running_output_end_turn_marked_completed\`, \`test_cleanup_stale_does_not_kill_agents_surviving_restart\`, and \`test_cleanup_stale_marks_end_turn_as_completed_not_dead\`.

### DIFF
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -457,42 +457,39 @@ def cleanup_stale_running_sessions(
     server_start_time: datetime,
     path: Path | None = None,
 ) -> list[str]:
-    """Mark pre-existing 'running' rows as 'dead' on server startup.
+    """Reconcile pre-existing 'running' rows on server startup.
 
-    After a force-restart, agents that were mid-execution have their last
-    ``stop_reason`` stuck at ``tool_use``.  The reconciler's liveness check
-    returns ``"running"`` for those files, so the normal dead-threshold logic
-    never fires.  This function closes that gap by inspecting every
-    ``status='running'`` row **before** the reconciler loop begins.
+    After a restart, agents whose output files have ``stop_reason=end_turn``
+    are marked completed.  Agents whose output files are missing are marked
+    dead.  Agents whose output files exist with ``stop_reason=tool_use`` (or
+    no stop_reason yet) are left running — Claude Code agents are independent
+    processes that survive MCP server restarts.
 
-    A session is declared dead at startup if ANY of these is true:
+    Previous versions of this function used an mtime heuristic: if the output
+    file's mtime predated the server start time, the agent was assumed dead.
+    That assumption is incorrect — Claude Code agents are NOT child processes
+    of the MCP server and can outlive a restart.  Using mtime caused false
+    positives where healthy agents were killed immediately after every server
+    restart.
 
-    1. Its ``output_file`` does not exist on disk — the file was cleaned up,
-       so no live agent can be writing to it.
-    2. Its ``output_file`` exists but its mtime predates ``server_start_time``
-       — the file was last touched before this server instance started, so it
-       cannot belong to an agent from the current run.
-    3. (Fallback) ``output_file`` is absent from the DB row AND elapsed time
-       since ``spawned_at`` exceeds ``timeout_minutes`` (or a generous default
-       of 120 minutes) — best-effort cleanup for unregistered output files.
+    Decision rules for each ``status='running'`` row:
 
-    **Assumption:** subagents cannot outlive a server restart. Claude Code
-    subagents run as child processes of the MCP server; when the server is
-    killed (e.g. ``systemctl restart``), all subagents are killed with it.
-    This means any ``status='running'`` row found at startup time cannot
-    belong to a genuinely live agent — it is always safe to mark it dead.
-
-    All matched rows are marked ``status='dead'`` with a ``completed_at``
-    timestamp of ``server_start_time`` so callers know when the cleanup ran.
+    1. ``output_file`` is set and target JSONL is missing on disk → dead
+       (file was cleaned up or never created; no live agent can be writing it)
+    2. ``output_file`` is set and ``stop_reason=end_turn`` → completed
+       (agent finished before or during the restart window)
+    3. ``output_file`` is set and ``stop_reason=tool_use`` (or empty) → leave running
+       (agent may still be alive; normal reconciler loop will time it out if needed)
+    4. ``output_file`` absent, elapsed > timeout → dead (fallback heuristic)
+    5. ``output_file`` absent, elapsed ≤ timeout → leave running
 
     Args:
-        server_start_time: The UTC datetime when the current server process
-                           started.  Any output file with mtime < this value
-                           cannot belong to a live agent from this session.
+        server_start_time: UTC datetime when the current server process started
+                           (kept for logging; no longer used for mtime comparison).
         path:              DB path override (for tests).
 
     Returns:
-        List of agent IDs that were marked dead.
+        List of agent IDs that were marked dead or completed.
     """
     resolved = path if path is not None else _DEFAULT_DB_PATH
     conn = _get_connection(resolved)
@@ -506,7 +503,7 @@ def cleanup_stale_running_sessions(
     )
     rows = cursor.fetchall()
 
-    dead_ids: list[str] = []
+    changed_ids: list[str] = []
     completed_at = server_start_time.isoformat()
 
     for row in rows:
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with minor observations

## Summary

The fix is correct and well-motivated. The mtime heuristic was fundamentally wrong: Claude Code agents are independent processes, not children of the MCP server, so mtime < server_start_time fires for *every* running agent on every restart. Replacing it with `_read_stop_reason_from_path` is the right approach.

---

## Bug 1: Startup cleanup liveness model

**Is `_read_stop_reason_from_path` reliable for distinguishing running vs stopped agents?**

Yes, with one important nuance. The function reads the last 4 KB of the JSONL file and regexes for the last `stop_reason`. The three-way return (`"running"` / `"done"` / `"missing"`) maps cleanly onto the three cases we care about.

**What if the file doesn't exist yet (agent just started)?**

`_read_stop_reason_from_path` returns `"missing"` when the symlink target doesn't exist. This correctly maps to the `file_status == "missing"` branch → **marked dead**. This is a pre-existing behavior (the old code also marked missing-file sessions dead) and is correct: if there's no output file, we have no evidence the agent is alive.

However, there's a narrow window during startup: an agent that was just spawned (before the restart) might not yet have written its output file. Under both the old and new logic, it gets marked dead. This is an acceptable trade-off given the grace period in the reconciler loop — but worth noting as a known limitation.

**Race condition: startup sweep vs active agent writing to its output file**

The startup sweep reads a file the agent is actively writing. This is safe because:
1. The function seeks to the last 4 KB and scans for `stop_reason`.
2. An active agent writes `stop_reason=tool_use` lines continuously; these return `"running"` → left alone.
3. There is no exclusive lock needed — a partial write at the tail of the file would either find `tool_use` (→ leave running) or find no match (→ return `"running"` by default). The only dangerous case would be a partial write of `end_turn` bytes, but `_STOP_REASON_RE` requires the full string `end_turn` — a partial write would fail to match and fall through to `"running"`. This is race-safe by construction.

**Edge case: malformed stop_reason in output file**

If the JSON is malformed or the stop_reason value is something other than `"end_turn"`, `_read_stop_reason_from_path` returns `"running"`. The startup cleanup leaves the session running and the reconciler's normal loop handles it. This is the safe fallback. No concern here.

---

## Bug 2: elapsed_seconds fallback

**Does `elapsed_seconds` fallback correctly flow through the notification pipeline?**

Yes. The fix adds a `spawned_at`-based fallback when `elapsed_seconds` is `None`. The `get_unnotified_completed()` path returns `_row_to_dict` rows which lack `elapsed_seconds`, so the fallback fires. The elapsed is computed as `(now - spawned_dt).total_seconds()` — this gives the total agent lifetime, which is the correct value for the "Xm elapsed" message.

One subtle point: for sessions marked completed at startup (the `end_turn` branch), `completed_at` is set to `server_start_time` rather than the actual agent completion time. The elapsed shown in the notification will therefore be slightly inflated (agent lifetime + server-down time). This is a pre-existing minor imprecision, not introduced by this PR.

---

## `completed` status flowing through the notification pipeline

Checking `_startup_sweep` → `get_unnotified_completed` → `_enqueue_reconciler_notification`:

```python
outcome = session.get("status", "completed")
_enqueue_reconciler_notification(session, outcome=outcome)
```

When the startup cleanup marks a session `status='completed'`, `get_unnotified_completed` returns it, `outcome` is read from `session["status"]` → `"completed"`, and `_enqueue_reconciler_notification` routes to the `subagent_result` / `"Agent completed"` branch. This is correct — end_turn agents now get a "completed" notification rather than a "dead" notification.

---

## Test coverage

The new tests are thorough and directly target the regression scenario (mtime-predating-server-start with tool_use → should NOT kill). The four new test cases cover:

- `tool_use` left running (new primary regression test)
- `end_turn` marked completed (new positive case)
- `surviving-agent` with backdated mtime (explicit regression for the startup-kills-running-agents bug)
- `completed-at-restart` end_turn → completed status (double-coverage)

The pre-existing `test_cleanup_stale_running_output_missing` test still covers the missing-file path.

One gap: there's no test for the `_enqueue_reconciler_notification` elapsed fallback (`spawned_at`-based computation when `elapsed_seconds` is absent). This is a unit-testable path. Not blocking, but worth adding.

---

1. **`changed_ids` semantics shift**: The function now returns both `dead` and `completed` agent IDs in `changed_ids`. Any callers that logged or asserted on the return value expecting only dead agents should be checked. (The caller in `inbox_server.py` only logs the count, so no issue there.)

2. **`server_start_time` parameter retained but unused**: The docstring correctly notes this. It's used only for `completed_at` timestamps. The parameter is kept for API stability — fine.

3. **No test for malformed JSON in output file**: `_read_stop_reason_from_path` handles this correctly (regex fails to match → returns `"running"`), but a test would make the guarantee explicit.

---

## Verdict

PASS. The fix correctly addresses the root cause. The `_read_stop_reason_from_path` approach is race-safe, handles edge cases gracefully (malformed stop_reason → running, no file → dead), and the notification pipeline flows correctly for both `completed` and `dead` outcomes. The one minor test gap (elapsed fallback) is not blocking.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR fixes a critical false-positive in the startup cleanup that was killing live agents after every server restart.

**What the code does:** Replaces the mtime-based heuristic in `cleanup_stale_running_sessions` with a stop_reason-based heuristic. Previously, any running session whose output file had an mtime predating the server start time was marked dead — which is incorrect because Claude Code agents are independent processes that survive MCP server restarts. The new logic reads the `stop_reason` field from the output JSONL file:
- Missing file → mark dead (file was cleaned up, agent is gone)
- `stop_reason=end_turn` → mark completed (agent finished cleanly)
- `stop_reason=tool_use` or empty → leave running (agent may still be alive)

Also fixes a related bug in `_enqueue_reconciler_notification` where `elapsed_seconds` was missing from startup-sweep rows (coming from `get_unnotified_completed` instead of `get_active_sessions`), causing '0m' in notification messages.

**Why it matters:** Before this fix, every server restart would immediately kill all running agents (because every live agent's output file has mtime < server_start_time). This was causing spurious 'agent failed' notifications and potentially orphaning work.

**Observations:**
- The test suite is comprehensive: the renamed test (`test_cleanup_stale_running_output_old_mtime` → `test_cleanup_stale_running_output_tool_use_left_running`) directly documents the behavioral inversion.
- The `changed_ids` rename from `dead_ids` correctly reflects that the return value now includes both dead and completed sessions.
- The elapsed-time fallback (`spawned_at` computation) for the notification path is a clean fix.

PR #655, opened 2026-03-18
fix: apply age threshold to empty stop_reason sessions in agent-status.sh

### LINKED ISSUE #614: fix: agent-status.sh should apply age threshold for empty stop_reason
## Problem

agent-status.sh currently applies a 30-min stale threshold only for `stop_reason=tool_use` (added in PR #603). For agents with empty `stop_reason` (i.e. still 'running'), there is no age cutoff — so stale /tmp output files from old agents persist indefinitely and appear as 'starting' in self-check output.

## Expected

Agents with empty stop_reason and output files older than N minutes (e.g. 60 min) should be treated as stale and excluded from the self-check running agent list.

## Root cause

Diagnosed in session 2026-03-17: bban36a5n, biw463jin, bmc5razqg, b19jn46z8, b983h8l76 all had stale /tmp output files with no stop_reason, causing them to appear as 'starting' in every self-check indefinitely.

sayhar:
## What problem this solves

Sessions with empty `stop_reason` (the "starting" state) were never evicted from the self-check agent list. The stale threshold check added in #603 only guarded `stop_reason=tool_use`. As a result, output files from agents that crashed before writing any stop_reason would appear as "starting" in every self-check indefinitely — exactly what happened with sessions bban36a5n, biw463jin, bmc5razqg, b19jn46z8, b983h8l76 diagnosed on 2026-03-17.

## What changed

The stale check in `scan_agent_status()` now covers two cases:

- `stop_reason=tool_use` — excluded after **30 minutes** (unchanged, agent crashed mid-tool-call)
- empty `stop_reason` — excluded after **60 minutes** (new, agent never progressed past startup)

The 60-minute threshold for "starting" is intentionally longer than the 30-minute threshold for "running" to avoid falsely evicting legitimately slow-starting agents.

The implementation is a pure conditional extension of the existing pattern: one outer `if` covering both cases, selecting the appropriate threshold by a nested `if`. No new state, no new files.

## Functional patterns

Pure function — the stale check is a stateless predicate over (stop_reason, file_age) with no side effects.

## Tests run

- [x] `bash -n scripts/agent-status.sh` — syntax clean
- [x] Manual functional test: 61-min old empty-stop_reason file excluded, 5-min old file shown as "starting" — both correct
- [x] `bash tests/test-agent-status.sh` — 10 passed, 8 failed. The 8 failures are pre-existing and test an old output format ("last activity Xs ago", "stale Xh") that was removed in an earlier refactor. Pass/fail count is identical to pre-fix baseline.

Closes #614

### DIFF
--- a/scripts/agent-status.sh
+++ b/scripts/agent-status.sh
@@ -124,17 +124,25 @@ scan_agent_status() {
             continue
         fi
 
-        # An agent with stop_reason=tool_use whose output file hasn't been modified
-        # in more than STALE_THRESHOLD_SECONDS is treated as dead (crashed mid-tool-call).
-        # Without this check a crashed agent would show as "running" forever.
-        local STALE_THRESHOLD_SECONDS=$(( 30 * 60 ))
-        if [ "$stop_reason" = "tool_use" ]; then
+        # Agents whose output files are stale are treated as dead.
+        # - stop_reason=tool_use: stale after 30 min (crashed mid-tool-call)
+        # - empty stop_reason:    stale after 60 min (never progressed past "starting")
+        # Without these checks a crashed agent would show as "running"/"starting" forever.
+        local STALE_TOOL_USE_SECONDS=$(( 30 * 60 ))
+        local STALE_STARTING_SECONDS=$(( 60 * 60 ))
+        if [ "$stop_reason" = "tool_use" ] || [ -z "$stop_reason" ]; then
             local now file_mtime file_age_seconds
             now=$(date +%s)
             # stat -c %Y is GNU coreutils; stat -f %m is BSD/macOS fallback
             file_mtime=$(stat -c %Y "$filepath" 2>/dev/null || stat -f %m "$filepath" 2>/dev/null || echo "$now")
             file_age_seconds=$(( now - file_mtime ))
-            if [ "$file_age_seconds" -gt "$STALE_THRESHOLD_SECONDS" ]; then
+            local threshold
+            if [ "$stop_reason" = "tool_use" ]; then
+                threshold=$STALE_TOOL_USE_SECONDS
+            else
+                threshold=$STALE_STARTING_SECONDS
+            fi
+            if [ "$file_age_seconds" -gt "$threshold" ]; then
                 continue  # file too old — agent is dead, not running
             fi
         fi


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: Approve**

The fix is clean, well-reasoned, and addresses the root cause precisely.

**Threshold asymmetry (60 min vs 30 min) — makes sense**

`tool_use` stale agents have already been actively running and crashed mid-call; 30 min is a tight, appropriate window. Empty `stop_reason` agents are in early startup and may legitimately take longer to produce first output, so 60 min is a reasonable conservative buffer. The PR description explains this explicitly and the logic holds.

**`[ -z "$stop_reason" ]` — correctly catches early-phase crashed agents**

`_get_stop_reason` returns empty string when the file is empty, not yet written, or has no `stop_reason` field. This exactly describes an agent that crashed before writing anything meaningful. The check is correct.

**Edge case: legitimate slow-starting agents**

The main risk is a genuinely slow-starting agent (e.g., one running a long bash build without intermediate tool calls) being falsely excluded from the summary after 60 min. Two mitigations apply:

1. The consequence is exclusion from the self-check display, not termination — low-stakes false positive.
2. Any actively-running agent that makes tool calls will write `tool_use` to its output and be covered by the 30-min path instead, so only agents with truly no output at all hit the 60-min path.

The 60-min threshold is a reasonable engineering tradeoff given the observed real-world failures (5 zombie sessions on 2026-03-17 that triggered this PR).

**Shell implementation — clean**

Outer `if ... || [ -z "$stop_reason" ]` with a nested threshold selector is the minimal, readable extension of the existing pattern. No new state, no new files, correct quoting throughout. `bash -n` clean per the PR's own test checklist.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR extends the agent-status stale-detection logic to cover agents that never progressed past the 'starting' state.

**What the code does:** Previously, `agent-status.sh` only applied a stale threshold to sessions with `stop_reason=tool_use` (30 minutes). Sessions with an empty `stop_reason` — meaning the agent started but never set a stop_reason yet — were left running indefinitely. This PR adds a second threshold: empty `stop_reason` sessions are considered stale (dead) after 60 minutes. The conditional is restructured to compute the right threshold depending on which case applies.

**Why it matters:** An agent that is stuck at startup with no stop_reason would previously show as 'running' or 'starting' forever in the status output. The 60-minute threshold is a reasonable upper bound — a legitimate agent should have produced output within an hour. This prevents ghost 'starting' entries from polluting the status display.

**Observations:**
- The threshold variables (`STALE_TOOL_USE_SECONDS`, `STALE_STARTING_SECONDS`) are declared with `local` inside the loop on every iteration, which is correct bash but slightly wasteful. Not a correctness issue.
- The 60-minute threshold for empty stop_reason is generous. If agents are regularly timing out at this boundary it may warrant tuning, but it is a safe initial value.

PR #659, opened 2026-03-18
fix(health-check): pre-emptively mark near-threshold messages before restart

### LINKED ISSUE #555: fix(health-check): post-restart verification fires false-positive alert on pre-restart stale messages
## Problem

After a health-check-triggered restart, the post-restart verification check (`check_post_restart_inbox`) scans the inbox for stale messages. If a message was already stale *before* the restart began, the check reports "new stale messages detected post-restart" — even though the restart succeeded and the message predates it.

This is a false positive: the alert implies the restart failed to resolve the issue, but the restart completed successfully (PID changed, system GREEN).

## Observed (2026-03-17 03:16 UTC)

- 03:16 — message 10331 at 187s stale. Restart triggered.
- 03:16:07 — post-restart check finds a *different* message already older than 180s (predated the restart).
- Alert fires: "Restarted, but new stale messages detected post-restart."
- 03:18 — new PID confirmed, system GREEN. Restart was successful.

## Fix

The post-restart inbox check should filter out messages whose timestamps predate the restart initiation time. Only messages that became stale *after* the restart started are evidence that the restart didn't help.

Alternatively, add a grace period: if the restart succeeded (PID changed, tmux alive), don't fire the "still stale" alert for messages younger than N seconds from restart completion.

## Impact

Causes unnecessary alarm for operators when the restart was actually successful.

sayhar:
## Problem

When `do_restart()` fires for a stale-inbox condition, `record_stale_inbox_markers()` marks inbox messages as circuit-breakered so the post-restart `check_inbox_drain()` call doesn't re-alarm on the same files. The bug: only messages already over `STALE_THRESHOLD_SECONDS` (240s) were marked. Messages in the 120–240s range at restart time are not marked — and during the 10–60s restart sequence, they cross the threshold and appear as "new stale" to the post-restart check. The restart succeeded (PID changed, system GREEN), but the alert fires anyway.

This is the root cause of the 2026-03-17 03:16 UTC incident in #555: a message predating the restart was found stale post-restart, generating a false "Restarted, but new stale messages detected" Telegram alert.

## Fix

Add `RESTART_WINDOW_BUFFER_SECONDS=120` — a conservative buffer above the worst-case restart duration (~60s). In `record_stale_inbox_markers()`, the mark threshold is lowered from `STALE_THRESHOLD_SECONDS` to `STALE_THRESHOLD_SECONDS - RESTART_WINDOW_BUFFER_SECONDS` (240 - 120 = 120s). Any message older than 120s at restart time is pre-emptively circuit-breakered, ensuring it won't be treated as "new stale" by the post-restart check even if it crosses 240s during the restart sequence.

The log message now distinguishes between genuinely stale messages and pre-emptively marked near-threshold ones, making operator logs clearer.

This fix is complementary to PR #504 (post-restart cooldown for restart loops). That PR addresses a separate problem (loop prevention); this one addresses the false-positive alert on successful restarts.

## Call sequence (verified)

In `do_restart()`:
1. `record_stale_inbox_markers()` — now marks messages > 120s old
2. Restart (stop + kill-server + start + up to ~30s wait + retries)
3. `check_inbox_drain()` — skips circuit-breakered files, only alarms on genuinely new stale messages

The fix sits entirely in step 1 — earlier pre-emption of the circuit breaker.

## Functional patterns

The fix replaces the hardcoded threshold with a derived value (`mark_threshold = STALE_THRESHOLD - BUFFER`), making the relationship between constants explicit and the intent clear in logs.

Closes #555

## Tests run

- [ ] Live restart test — blocked: requires production environment with real inbox files and systemd. No automated test harness for health-check-v3.sh exists in the repo.
- [x] `bash -n scripts/health-check-v3.sh` — syntax check passed, no errors.

**Note for reviewer:** The threshold arithmetic (240 - 120 = 120s mark threshold) should be verified against the worst-case restart duration. The script header documents ~60s worst-case; the 120s buffer is intentionally 2x that.

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -63,6 +63,7 @@ MAINTENANCE_FLAG="$MESSAGES_DIR/config/lobster-maintenance"
 LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$MESSAGES_DIR/config/lobster-state.json}"
 STALE_THRESHOLD_SECONDS=240          # 4 minutes - RED if any message older (watchdog handles soft recovery at 90s)
 YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW warning
+RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of the stale threshold before a restart
 
 MAINTENANCE_EXPIRY_SECONDS=3600      # 1 hour - stale maintenance flag is auto-cleared and checks resume
 
@@ -876,9 +877,14 @@ record_stale_inbox_markers() {
         [[ -z "$file_time" ]] && continue
 
         local age=$((now - file_time))
-        if [[ $age -gt $STALE_THRESHOLD_SECONDS ]]; then
+        local mark_threshold=$(( STALE_THRESHOLD_SECONDS - RESTART_WINDOW_BUFFER_SECONDS ))
+        if [[ $age -gt $mark_threshold ]]; then
             touch "$STALE_INBOX_MARKER_DIR/$basename_f"
-            log_info "Circuit breaker: marked $basename_f as restart-triggering"
+            if [[ $age -gt $STALE_THRESHOLD_SECONDS ]]; then
+                log_info "Circuit breaker: marked $basename_f as restart-triggering (stale: ${age}s)"
+            else
+                log_info "Circuit breaker: pre-emptively marked $basename_f (near-threshold: ${age}s, within ${RESTART_WINDOW_BUFFER_SECONDS}s restart window)"
+            fi
         fi
     done < <(find "$INBOX_DIR" -maxdepth 1 -name "*.json" -print0 2>/dev/null)
 }


sayhar (commented):

## Summary

The fix correctly addresses the root cause of the #555 false-positive alert by extending the circuit-breaker pre-emption window from 0s to 120s before the stale threshold. The logic is sound, the arithmetic is correct, and the call sequence is unchanged in all the right ways.

---

## Threshold Arithmetic

**STALE_THRESHOLD_SECONDS=240, RESTART_WINDOW_BUFFER_SECONDS=120, mark_threshold=120**

- The mark threshold (120s) gives a 120s window before a message reaches the 240s stale threshold.
- Worst-case restart duration documented in the script header is ~60s.
- The buffer is therefore 2x the documented worst case — conservative and appropriate.
- Any message already 120s old at the moment `record_stale_inbox_markers()` runs will be marked before the stop sequence begins. Even if the restart takes the full ~60s, that message would be at most 180s old post-restart, still 60s below the 240s alert threshold. The math holds.

One minor note: `mark_threshold` is recomputed on every loop iteration (`local mark_threshold=$(( ... ))` inside the `while` loop). This is harmless — the value is constant — but it's a trivially wasteful pattern. Moving it above the loop would be cleaner. Not a blocker.

## Call Sequence

The `do_restart()` sequence remains:
1. `record_stale_inbox_markers()` — **now marks messages >120s old** (was >240s)
2. Stop (systemctl stop + tmux kill-server + optional SIGTERM/SIGKILL)
3. Start (systemctl start)
4. Wait and verify (PID change, service/tmux readiness, up to ~15s of retries)
5. `check_inbox_drain()` — skips circuit-breakered files, alarms only on genuinely new stale messages

The fix is correctly placed at step 1. No other logic in `do_restart()` was changed.

The conditional guard `if [[ "$reason" == *"stale inbox"* ]]` before `record_stale_inbox_markers` is unchanged and still correct — the pre-emptive marking only fires for stale-inbox restarts, not process/tmux failures.

## Log Branching

The new `if/else` inside the marker loop only changes which log message is emitted — the `touch` call executes unconditionally whenever `age -gt mark_threshold`. The branching cannot affect marker creation. The log messages are clear and operationally useful: operators can now distinguish "this message was already over threshold when we restarted" from "this message was approaching threshold and we marked it proactively."

## Test Coverage

As noted in the PR, there is no automated test harness for this script. `bash -n` confirms syntax is clean. The change is small (one constant, one threshold substitution, one log branch) and straightforward to verify by inspection. The absence of tests is a pre-existing gap, not introduced by this PR.

## Minor Findings (non-blocking)

1. **`mark_threshold` inside the loop**: Cosmetically inefficient — computed every iteration but constant. Consider hoisting it above the `while` loop in a follow-up.
2. **120s buffer vs. worst-case restart**: The 2x safety factor is good. If restart durations ever grow (e.g., slower systemd activation or more PID-check retries are added), revisit whether 120s remains sufficient.
3. **No interaction with `clear_stale_inbox_markers()`**: Confirmed that `clear_stale_inbox_markers()` is only called on healthy-inbox paths — it will not interfere with the pre-emptively written markers during a restart sequence.

---

**Verdict: PASS.** The fix is correct, conservative, and well-scoped. The threshold arithmetic is sound, the call sequence is correct, and the log branching does not affect behavior. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS**

This PR addresses a race condition in the health-check circuit breaker where messages could be marked for restart-triggering only after they exceed the full stale threshold, leaving a gap where a restart could begin with messages that will immediately look stale again post-restart.

**What the code does:** Introduces a `RESTART_WINDOW_BUFFER_SECONDS=120` constant (2 minutes). The mark threshold is lowered from `STALE_THRESHOLD_SECONDS` (240s) to `STALE_THRESHOLD_SECONDS - RESTART_WINDOW_BUFFER_SECONDS` (120s). Messages in the 120–240s age band are now pre-emptively marked before a restart occurs. Logging distinguishes pre-emptive marks from confirmed-stale marks.

**Why it matters:** Without pre-emptive marking, when the health check restarts at t=240s, the restart takes ~60–90s, and the pre-existing messages (now ~300–330s old) still look stale on the first post-restart check — triggering another restart. Pre-marking at t=120s ensures circuit-breaker markers exist before the restart window, so the cooldown suppression logic (PR #504) can suppress the false-positive on the next run.

**Concerns:** The 120s buffer is hardcoded relative to the 240s threshold. If STALE_THRESHOLD_SECONDS is ever tuned, the subtraction will still work, but the buffer as a fraction of threshold (50%) is large — worth noting. Minor: `mark_threshold` is computed inside the loop on every iteration; moving it outside would be cleaner but is not a correctness issue.

Issue #209, opened 2026-03-13
fix: replace source-as-type-proxy guards with type-based classification (inbox_server.py + lobster_bot.py)


sayhar:
## Summary

A codebase audit identified two places where `source` (the messaging platform: `"telegram"`, `"slack"`) is used as a proxy for message type. This is incorrect because `subagent_result` and similar system messages can share the same `source` value as genuine user messages, causing these guards to misfire.

---

## Bug 1 (Real): `inbox_server.py` — `mark_processed` guard

**File:** `src/mcp/inbox_server.py`, around line 2124–2129

**Problem:**

The condition `if source in _HUMAN_SOURCES` is used to decide whether to trigger the "did we reply to this user?" fallback check after marking a message processed. A `subagent_result` message with `source="telegram"` passes this filter even though it is not a user-originated message. This can cause spurious fallback behavior (e.g. generating an unsolicited reply or logging a false missed-reply warning) when processing subagent results.

**Fix:**

Replace the primary gate:
```python
# Before
if source in _HUMAN_SOURCES:
    ...

# After
if msg_type in USER_FACING_TYPES:
    ...
```

`USER_FACING_TYPES` should enumerate the message types that represent real user input (e.g. `"text"`, `"photo"`, `"voice"`, `"callback"`) and explicitly exclude `"subagent_result"`, `"subagent_error"`, and `"system"`.

---

## Bug 2 (Minor): `lobster_bot.py` — typing indicator loop

**File:** `src/bot/lobster_bot.py`, around line 413–416

**Problem:**

`typing_refresh_loop` sends Telegram typing indicators for any message currently in the `processing/` state. A `subagent_result` message that the dispatcher is currently handling will incorrectly trigger a typing indicator, making it look like Lobster is composing a new message when it is just forwarding an existing one.

**Fix:**

Add a type check inside `typing_refresh_loop`: skip messages whose type is not in `USER_FACING_TYPES`.

---

Resolved by PR #213 and PR #207.

PR #661, opened 2026-03-18
fix: use USER_FACING_TYPES allowlist for tier-1 signal extraction and context injection

sayhar:
## Problem

Two blocks in `mark_processing` used manually-maintained denylist tuples (`_SKIP_OBSERVATION_TYPES` and `_SKIP_CONTEXT_TYPES`) to decide whether to fire tier-1 signal extraction and user model context injection. Both tuples were missing system message types — `subagent_notification`, `cron_reminder`, `compact_group`, `update_notification`, and others — so those system messages would trigger tier-1 observations and user context injection even though they contain no user-originated content. This is both wasteful and incorrect: tier-1 extraction is supposed to learn from what the user writes, not from system notifications.

## Fix

Replace both denylist tuples with a membership check against `USER_FACING_TYPES` (already imported from `message_types.py`, already used in `check_inbox` for the same purpose). The allowlist approach is closed-world: only message types explicitly recognized as user-originated (text, voice, photo, callback, etc.) pass through; any new system type added in the future is blocked by default rather than requiring a manual denylist update.

Both blocks are changed:
- Tier-1 signal extraction (`_queue_observation` call)
- User model context injection (`_user_model.get_context()` call)

The fix removes 11 lines and adds 2. No new symbols are introduced; `USER_FACING_TYPES` was already imported at line 560.

This is the same pattern established by issues #209 / PRs #207 and #213.

## Tests run
- [x] `uv run pytest tests/unit/` — 71 failed, 1184 passed (all 71 failures are pre-existing on `main`; zero new failures introduced by this change, confirmed by running the same suite against `main` with `git stash`)
- [x] `uv run pytest tests/unit/test_mcp_server/` — 18 failed, 260 passed (same 18 pre-existing failures on both `main` and this branch)

**Blocked items needing attention before merge:** none — all failing tests are pre-existing and unrelated to this change.

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -3480,12 +3480,7 @@ async def handle_mark_processing(args: dict) -> list[TextContent]:
 
     # Queue background observation (non-blocking, best-effort)
     msg_text = msg_data.get("text", "") or msg_data.get("transcription", "")
-    _SKIP_OBSERVATION_TYPES = (
-        "subagent_result", "subagent_error", "self_check", "subagent_observation",
-        "subagent_recovered",  # recovery events are system-internal; salvaged text is not user content
-        "debug_observation",  # never run tier 1 on our own debug output (would loop)
-    )
-    if msg_text and msg_type not in _SKIP_OBSERVATION_TYPES:
+    if msg_text and msg_type in USER_FACING_TYPES:
         _queue_observation(
             msg_text, message_id,
             source=msg_data.get("source"),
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — Correct and tidy. Replacing manually-maintained denylist tuples with a membership check against USER_FACING_TYPES is the right call: the allowlist is closed-world, already established in the codebase, and prevents future system types from accidentally passing through to tier-1 extraction or user model context injection. Net -9 lines. No new symbols introduced. Test results confirm zero regressions against pre-existing failures on main.

PR #665, opened 2026-03-18
fix: remove REACTION_SIGNALS allowlist — deliver all emoji to dispatcher

### LINKED ISSUE #664: Remove REACTION_SIGNALS mapping from lobster_bot.py — pass raw emoji to dispatcher
## Background

PR #490 (merged 2026-03-17) added a reaction-handling pipeline to `lobster_bot.py`. Part of that implementation includes a hard-coded `REACTION_SIGNALS` dict that maps specific emoji to semantic string labels (`yes`, `no`, `cancel`) before the message is written to the inbox:

```python
REACTION_SIGNALS: dict[str, str] = {
    "\U0001f44d": "yes",    # 👍 thumbs up
    "\U0001f44e": "no",     # 👎 thumbs down
    "\u2705": "yes",        # ✅ check mark button
    "\U0001f44c": "yes",    # 👌 OK hand
    "\u274c": "no",         # ❌ cross mark
    "\U0001f6ab": "cancel", # 🚫 no entry sign
}
```

Any emoji **not** in this dict is silently dropped — the reaction never reaches the inbox.

## Problem

This pre-mapping is the wrong layer for emoji interpretation:

1. **Silent drops**: any emoji the user reacts with that isn't in the allowlist is discarded, so the dispatcher never sees it.
2. **Hardcoded semantics**: the "meaning" of an emoji is context-dependent. The dispatcher is better placed to interpret whether 👍 means "yes", "looks good", "acknowledged", etc., based on what the reacted-to message was about.
3. **Inflexible**: adding or changing the meaning of a reaction requires modifying bot code rather than dispatcher logic or user preferences.

## Desired change

- Remove the `REACTION_SIGNALS` dict entirely from `lobster_bot.py`.
- Remove the `signal` parameter from `_emit_reaction_signal` and from the inbox JSON payload.
- Remove the guard `if signal is None: continue` — all emoji reactions should be delivered (subject to the undo window).
- Update the `text` field to drop the `(signal)` interpolation, e.g. `[Reaction: 👍 on message 1234]`.
- Update log lines accordingly.
- Keep everything else: the 5-second `REACTION_UNDO_WINDOW_SECS` undo window, `_pending_reactions` cancellation, `_sent_message_buffer` / `reacted_to_text` context.

## Resulting inbox JSON schema

```json
{
  "id": "...",
  "source": "telegram",
  "type": "reaction",
  "chat_id": 123456,
  "telegram_message_id": 789,
  "emoji": "👍",
  "reacted_to_text": "Here is the summary you asked for...",
  "text": "[Reaction: 👍 on message 789]",
  "timestamp": "2026-03-17T10:00:00Z"
}
```

The `signal` field is removed. The dispatcher receives the raw emoji and can interpret it with full context.

## Files affected

- `src/bot/lobster_bot.py` — remove `REACTION_SIGNALS`, update `_emit_reaction_signal`, `handle_reaction`
- Any tests that assert on the `signal` field in reaction inbox entries

sayhar:
## Summary

Before this change, any Telegram emoji reaction not in a hard-coded six-entry allowlist (`REACTION_SIGNALS`) was silently dropped — the dispatcher never saw it. This meant reactions like 🎉, 😬, or any custom emoji were permanently lost. The allowlist also pre-translated emoji to semantic strings (`yes`/`no`/`cancel`), baking context-dependent interpretation into the wrong layer.

This PR removes the allowlist entirely. All emoji reactions now pass through the 5-second undo window and land in the inbox as-is. The dispatcher receives the raw `emoji` and `reacted_to_text` fields and interprets intent in context — the same way it handles any other ambiguous input.

## What changed

- **`lobster_bot.py`**: `REACTION_SIGNALS` dict removed. `_emit_reaction_signal` loses the `signal` parameter. The guard `if signal is None: continue` is removed so all emoji are buffered. Inbox payload is `{emoji, reacted_to_text, telegram_message_id}` — no `signal` field.
- **`message_types.py`**: `"reaction"` added to `INBOX_USER_TYPES` (suppresses spurious unknown-type warnings when reactions arrive).
- **`inbox_server.py`**: Auto-reply suppression extended to `type: "reaction"` — reactions must not trigger the "Noted." fallback.
- **`sys.dispatcher.bootup.md`**: Reaction handling section updated — `signal` field removed from key-fields list, processing rules rewritten to describe contextual emoji interpretation instead of a signal lookup table.
- **`tests/unit/test_bot/test_reaction_handling.py`**: Tests updated to assert `REACTION_SIGNALS` is absent, `signal` is not in inbox JSON, and previously-unknown emoji (😀) now creates a pending task.

## What is NOT changed

- The 5-second `REACTION_UNDO_WINDOW_SECS` undo window — untouched.
- `_pending_reactions` cancellation on removal — untouched.
- `_sent_message_buffer` / `reacted_to_text` context — untouched.
- All other message types — untouched.

## Functional patterns

Pure functions: `_lookup_reacted_to_text` is a pure lookup with no side effects. `_emit_reaction_signal` is pure except for the final atomic inbox write at the system boundary. Side effects (I/O) are isolated to the outermost call.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_reaction_handling.py -v` — 13 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 1165 passed (71 pre-existing failures in test_update_manager and test_mcp_server unrelated to this change; confirmed present on origin/main)

Closes #664

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -419,31 +419,31 @@ When replying, always pass the correct `source` parameter to `send_reply` — Te
 
 **Handling edited messages:** When a message has `_edit_of_telegram_id` set, it is the user's edited version of a previously sent message. Process it as a normal message. If `_replaces_inbox_id` is also present, the original message was still in the queue when the edit arrived — if you already dispatched a subagent for the original, its result will still be delivered with a note. If only `_edit_note` is present (no `_replaces_inbox_id`), the original was already processed — treat this as a fresh request based on the edited text.
 
-**Handling reaction messages:** When a message has `type: "reaction"`, the user reacted to one of your sent messages. Reactions are signals, not conversational turns — they confirm or reject something you asked.
+**Handling reaction messages:** When a message has `type: "reaction"`, the user reacted to one of your sent messages. All emoji reactions are delivered — interpret them in context.
 
 Key fields:
 - `telegram_message_id` — Telegram ID of the message that was reacted to
 - `reacted_to_text` — snippet of what that message said (populated from the bot's sent-message buffer)
-- `emoji` — the raw emoji character
-- `signal` — structured intent: `"yes"`, `"no"`, or `"cancel"`
+- `emoji` — the raw emoji character (e.g. `"👍"`, `"❌"`, `"🎉"`)
 
 **Processing rules:**
 
 ```
 1. mark_processing(message_id)
-2. Look at signal:
-   - "yes"    → treat as affirmative confirmation (user approved whatever you asked)
-   - "no"     → treat as rejection (user declined)
-   - "cancel" → treat as cancellation (abort the pending action)
-3. Use reacted_to_text to identify which pending decision this refers to
-4. Act directly on the signal — no need to ask "did you mean yes?"
+2. Interpret emoji in context of reacted_to_text:
+   - 👍 / ✅ / 👌 → likely affirmative (but consider what was said)
+   - 👎 / ❌     → likely rejection or disagreement
+   - 🚫          → likely cancellation
+   - Any other emoji → interpret based on the message content and conversation history
+3. Use reacted_to_text to identify which pending decision or message this refers to
+4. Act on the interpreted intent — no need to ask "did you mean yes?"
 5. mark_processed(message_id)
    # Do NOT send_reply unless your response adds real value.
    # Reactions are signals; the user expects action, not conversation.
 ```
 
 **When to reply vs. stay silent:**
-- If the reaction resolves a pending question (yes/no to "should I merge?"), act on it and reply with what you did.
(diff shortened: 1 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — Right move. The REACTION_SIGNALS allowlist was lossy (emoji like 🎉, 😬, custom emoji silently dropped) and baked interpretation into the wrong layer. All reactions now flow through the undo window and land in the inbox as raw emoji + reacted_to_text, letting the dispatcher interpret intent in context — the same approach used for all other ambiguous input. The auto-ack sends immediate feedback to the user on reaction receipt. The auto-reply suppression for 'reaction' in mark_processed prevents the spurious 'Noted.' fallback. Tests updated correctly: REACTION_SIGNALS absence asserted, previously-ignored 😀 now creates a pending task. The dispatcher bootup docs are updated consistently.

PR #672, opened 2026-03-18
fix: load transcript from agent_transcript_path in require-auditor-context-update.py

### LINKED ISSUE #670: fix: require-auditor-context-update.py broken in CC 2.1.76+ — reads transcript inline but only agent_transcript_path is provided
## Bug

`hooks/require-auditor-context-update.py` (SubagentStop hook) is silently broken since CC 2.1.76+.

The hook calls `hook_input.get("transcript", [])` to read the agent transcript and check whether the auditor called `write_observation`. But SubagentStop events in CC 2.1.76+ no longer include the transcript inline — they only provide `agent_transcript_path` (a file path to the JSONL transcript).

Since `transcript` is always an empty list, `_is_auditor_session()` always returns `False`, and the hook exits 0 silently without enforcing anything. The auditor enforcement gate is completely bypassed.

## Fix

Load the transcript from `hook_input.get("agent_transcript_path")` instead of (or in addition to) the inline `transcript` key. `require-write-result.py` already does this correctly — use the same pattern.

## Tests needed

Add a unit test that:
1. Creates a mock SubagentStop event with only `agent_transcript_path` (no inline `transcript`)
2. Writes a fake JSONL transcript to that path (with and without `write_observation` calls)
3. Asserts the hook correctly detects compliance/non-compliance

This would have caught the CC 2.1.76+ regression immediately.

## Related
- [Issue #623](https://github.com/SiderealPress/lobster/issues/623) — 9 hooks with no test coverage

sayhar:
Closes #670

## What was broken

`require-auditor-context-update.py` was silently broken since CC 2.1.76+. The hook read `hook_input.get("transcript", [])` to inspect the agent's tool calls, but SubagentStop events no longer embed the transcript inline — they only provide `agent_transcript_path` (a JSONL file path). Since `transcript` was always an empty list, `_is_auditor_session()` always returned `False`, and the entire auditor enforcement gate was silently bypassed: auditor sessions could exit without updating `system-audit.context.md` and the hook let them through.

## What changed

Three fixes, mirroring the pattern already used by `require-write-result.py`:

1. **Transcript loading**: `main()` now reads `agent_transcript_path` from the hook input and loads the JSONL file via a new `_load_transcript_from_jsonl()` helper. The legacy inline `transcript` key is kept as a fallback for older CC versions.

2. **JSONL format support in `_extract_tool_calls()`**: In CC 2.1.76+ JSONL format, tool use items are nested under `entry["message"]["content"]`, not `entry["content"]`. The function now checks both locations, matching the approach in `_collect_tool_use_and_text()` in `require-write-result.py`.

3. **`_session_start_time()` signature**: The function previously fell back to scanning `hook_input.get("transcript", [])` for timestamps, which was also always empty. It now accepts the already-loaded transcript as a parameter so both code paths use the same data.

No behavioral change for non-auditor sessions (still exits 0 immediately) or for the two compliant-exit conditions (context file updated, or safe word present).

## Functional patterns

Pure functions throughout: `_load_transcript_from_jsonl`, `_extract_tool_calls`, `_is_auditor_session`, `_safe_word_in_transcript`, and `_session_start_time` are all stateless transforms with explicit inputs and outputs. `main()` is the sole side-effecting boundary.

## Tests run

- [x] `uv run pytest tests/test_require_auditor_context_update.py -v` — 27 passed, 0 failed
  - Covers: JSONL loading, both transcript formats (JSONL and legacy inline), auditor detection via Read/Bash, safe-word detection, `_session_start_time` timestamp sourcing, compliant exits (safe word, file updated), non-compliant exits (exits 2), non-auditor pass-through, invalid stdin handling
- [x] `uv run pytest tests/test_require_write_result.py -v` — 30 passed, 0 failed (no regression in the sibling hook)

### DIFF
--- a/hooks/require-auditor-context-update.py
+++ b/hooks/require-auditor-context-update.py
@@ -17,6 +17,23 @@ Detection strategy for "is this an auditor session":
   Scan the transcript for a ReadFile tool call whose path contains
   "system-audit.context.md". The auditor definition requires reading this
   file at session start, so its presence is a reliable signal.
+
+## SubagentStop transcript handling (CC 2.1.76+)
+
+SubagentStop events in CC 2.1.76+ no longer include an inline `transcript`
+field. They only provide `agent_transcript_path` — a path to a JSONL file
+containing the subagent's conversation. This hook loads the transcript from
+that file path, falling back to the legacy inline `transcript` key for older
+CC versions.
+
+## JSONL message format
+
+Each line of the JSONL transcript file has the structure:
+    {"type": "assistant", "message": {"role": "assistant", "content": [...]}, ...}
+
+Tool use items are nested under entry["message"]["content"], NOT entry["content"].
+`_extract_tool_calls` handles both the JSONL format and the legacy inline
+format where content is directly on the message dict.
 """
 import json
 import os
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The fix is correct and complete. Three independent issues were all caused by the same root: the hook never read from `agent_transcript_path`. All three are addressed:

- `_load_transcript_from_jsonl()` is a faithful copy of the pattern from `require-write-result.py`, including blank-line skipping and per-line error tolerance.
- `_extract_tool_calls()` now handles both the JSONL nested format (`entry["message"]["content"]`) and the legacy inline format (`entry["content"]`), matching `_collect_tool_use_and_text()` in the sibling hook.
- `_session_start_time()` now receives the already-loaded transcript rather than re-fetching the always-empty inline key.

The fallback to `hook_input.get("transcript", [])` is retained for older CC versions, which is the right defensive posture.

Test coverage is thorough: 27 tests, covering both compliant and non-compliant auditor sessions via `agent_transcript_path`, the legacy inline fallback path, the empty/missing path cases, all the pure helper functions individually, and the invalid-stdin guard. The existing `test_require_write_result.py` suite is unaffected (30 tests, 0 failures).

No concerns. Safe to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**What the fix does:**

CC 2.1.76 changed how SubagentStop hook events deliver transcript data. Previously the hook input included an inline `transcript` list; from 2.1.76 onward it only provides `agent_transcript_path` — a path to a JSONL file. This PR updates `require-auditor-context-update.py` to read the transcript from that file when present, falling back to the inline format for older CC versions.

There are two secondary changes that are required to make this work:

1. `_extract_tool_calls` now handles both the JSONL entry format (`entry["message"]["content"]`) and the legacy inline format (`entry["content"]`), since the nesting structure differs between the two.
2. `_session_start_time` now accepts the already-loaded transcript list as a parameter rather than re-reading it from `hook_input["transcript"]`, which is always empty in CC 2.1.76+.

**Why it matters:**

Without this fix, the auditor enforcement hook silently fails on CC 2.1.76+ — it sees an empty transcript, never detects auditor sessions, and exits 0 every time. That means non-compliant auditor subagents can exit without updating the context file and the hook won't catch it. This restores correct enforcement behavior.

**Test coverage:**

All 27 tests pass. The new `TestMainSubagentStop` class exercises both the new JSONL path and the legacy fallback path, including the edge cases (no path key, empty path string, bad JSON stdin). The logic coverage looks complete.

The fallback to inline transcript is clean and the error handling is defensive throughout (`try/except` on both file open and per-line parsing, returning `[]` on any failure). No concerns.

PR #677, opened 2026-03-18
fix: remove periodic self-check inbox injection

sayhar:
## What this fixes

The `scripts/periodic-self-check.sh` cron job (every 3 minutes) was injecting `status? (Self-check)` and `[Task Completed]` messages into the dispatcher inbox. These were immediately marked processed with no user-visible effect — generating ~20 no-op inbox entries per hour.

Subagent results are delivered directly to the inbox via `write_result`. The self-check injection was a legacy mechanism from before that path existed. It is now pure noise.

## What changed

- **Deleted** `scripts/periodic-self-check.sh` — the script that wrote the messages
- **Deleted** `tests/test-self-check.sh` — the test suite for it
- **install.sh**: replaced the self-check setup block with a one-liner that removes the cron entry on fresh installs
- **scripts/upgrade.sh**: superseded Migration 18 (which added the cron entry) with a no-op comment; added **Migration 21** which removes the `# LOBSTER-SELF-CHECK` cron entry on existing installs
- **.claude/sys.dispatcher.bootup.md**: removed the Self-Check Reminders section and updated the system message type description

## What is NOT changed

- Reconciler (5s polling that delivers real `subagent_result` / `subagent_notification` messages)
- `health-check-v3.sh` and its `# LOBSTER-HEALTH` cron entry
- Ghost-detector scheduled job
- `whatsapp-health-check.sh` (standalone script; its comment referencing periodic-self-check is stale but harmless)

## Functional patterns

Pure removal — no new logic introduced. Migration 21 follows the same idempotent crontab filter pattern used in prior migrations.

## Tests run

- [x] `git diff --stat` — 5 files, 574 deletions, 15 insertions; no unintended changes
- [x] `crontab -l` manually verified on live system: `# LOBSTER-SELF-CHECK` entry present before; will be removed by Migration 21 on next `upgrade.sh` run
- [ ] Full test suite — skipped: requires Docker; no code logic changed, only deletion + cron management

**Blocked items needing attention before merge:** none — Migration 21 handles live cleanup on next upgrade run.

Sahar has explicitly approved removing the self-check.


sayhar (commented):
Lobster (reviewer): PASS

The change does exactly what it says — deletes the self-check injection script, removes its cron setup from install.sh, and adds Migration 21 to clean up the cron entry on existing installs. Migration 18 is correctly superseded with a no-op stub to preserve numbering continuity.

Key checks:
- The reconciler loop in inbox_server.py is untouched (verified by grep)
- health-check-v3.sh cron entry is untouched
- ghost-detector is untouched
- install.sh cleanup line uses the same `grep -v "# LOBSTER-SELF-CHECK"` filter as the old setup, so it's safe on fresh installs that never had the entry
- Migration 21 is idempotent: if the entry doesn't exist, the crontab write is a no-op

No concerns. Safe to merge.

PR #685, opened 2026-03-18
fix: exactly one Telegram notification per compaction

### LINKED ISSUE #684: fix: exactly one Telegram notification per compaction
## Problem

When context compaction occurs, the user receives 0 or 2 Telegram notifications instead of exactly 1.

**In production (LOBSTER_DEBUG=false):** 0 notifications from `on-compact.py` (gated behind debug flag), then 1-2 from health-check (`do_restart` deferred restart alert + "System recovered automatically"), totaling 0, 1, or 2 depending on timing.

**In debug mode:** 1 from `on-compact.py` + potentially 1-2 from health-check = 2-3 total.

## Root Cause

Three contributing factors:

1. **`on-compact.py` gates its Telegram notification on `LOBSTER_DEBUG=true`** — so in production, no notification fires when compaction starts.
2. **The 1-second `wait_for_messages` timeout post-compaction** — the compact-reminder lands in inbox, but the dispatcher's `wait_for_messages(timeout=1, hibernate_on_timeout=True)` may time out immediately if the inbox is momentarily empty when polled, causing the dispatcher to hibernate.
3. **Health-check misreads the situation** — `do_restart()` sends "Health check deferred restart..." when subagents are active, then "System recovered automatically" after a successful restart.

## Fix (approach)

1. **Remove the `LOBSTER_DEBUG` gate in `on-compact.py`**: Always send the Telegram notification on compaction (unconditionally). Change the message from dev-flavored `[DEV] Context compacted...` to a clean user-facing message.
2. **Suppress health-check Telegram alerts during active compaction window**: The health-check already suppresses *stale-inbox restarts* via `is_compaction_recent()`. Extend this suppression to the Telegram alerts in `do_restart()` — if compaction was recent, skip the alert (the user already got notified via step 1).

## Plan

- [ ] Update `on-compact.py`: remove `LOBSTER_DEBUG` gate, change message to user-facing text, always send when credentials available
- [ ] Update `health-check-v3.sh`: pass compaction-recent context into `do_restart()`, suppress Telegram alert (but still restart if needed) when compaction is recent
- [ ] Update tests for `on-compact.py`
- [ ] Verify no double-notification path remains

sayhar:
## Problem

Before this fix, context compaction produced **0 or 2 Telegram notifications** depending on the deployment mode — never exactly 1.

**Root cause:** two independent notification paths that were never coordinated.

| Mode | on-compact.py | health-check | Total alerts |
|------|---------------|--------------|--------------|
| Production (LOBSTER_DEBUG=false) | 0 (gated behind debug flag) | 1-2 (deferred restart + recovered) | 1-2 confusing alerts with no mention of compaction |
| Debug (LOBSTER_DEBUG=true) | 1 (dev-flavored message) | 1-2 (same alerts) | 2-3 total; messages contradict each other |

The health-check was already suppressing stale-inbox **restarts** during the compaction window (via `is_compaction_recent()`), but it was still sending the Telegram alerts that accompany those restarts.

---

### Before

```
Compaction event
|
+-> on-compact.py runs
|     +-- writes compacted_at to lobster-state.json   OK
|     +-- writes compact-reminder to inbox            OK
|     +-- LOBSTER_DEBUG=true?
|           +-- yes -> send "[DEV] Context compacted..."
|           +-- no  -> (silence)
|
+-> health-check runs (~60s later)
      +-- is_compaction_recent()?
            +-- yes -> skip restart   <- suppresses restart action
            |         BUT still sends Telegram alerts:
            |           "Health check deferred restart: N subagents in flight"
            |           "System recovered automatically. Reason: <stale-inbox>"
            +-- no  -> normal restart + alert
```

**Result:** 0 notifications in production. 2-3 in debug mode. Neither path explains "compaction" to the user.

### After

```
Compaction event
|
+-> on-compact.py runs
|     +-- writes compacted_at to lobster-state.json   OK
|     +-- writes compact-reminder to inbox            OK
|     +-- send_compaction_notify() -- ALWAYS fires
|           +-- sends "Context compacted. Re-orienting..."
|
+-> health-check runs (~60s later)
      +-- is_compaction_recent()?  -> compaction_recent=true/false
            |
            +-- yes (compaction_recent=true)
            |     +-- do_restart(reason, suppress_alert=true)
            |           +-- subagent guard triggered? -> log only, no alert  [suppressed]
            |           +-- BLACK state reached?      -> ALERT always fires  [never suppressed]
            |           +-- restart failed?           -> ALERT always fires  [never suppressed]
            |           +-- restart succeeded?        -> log only, no alert  [suppressed]
            |
            +-- no (compaction_recent=false)
                  +-- do_restart(reason, suppress_alert=false)
                        +-- all alerts fire normally
```

**Result:** exactly 1 notification per compaction event, always. The health-check's restart mechanics still run; only the Telegram alerts are suppressed.

---

### `on-compact.py`

- `maybe_send_dev_telegram_notify()` (gated on `LOBSTER_DEBUG`) replaced by `send_compaction_notify()` (always fires when Telegram credentials are present).
- `_is_debug_mode()` helper removed — no longer needed.
- Message text changes from `[DEV] Context compacted. Re-orienting from CLAUDE.md + handoff.` to `Context compacted. Re-orienting...`
- This function is now the **single canonical notification** for a compaction event.

### `health-check-v3.sh`

- `do_restart()` gains a `suppress_alert` second parameter (default `false`).
- `main()` passes `$compaction_recent` (already computed by `is_compaction_recent()`) as `suppress_alert` when transitioning to RED.
- Two alert call sites inside `do_restart()` are conditionally suppressed: the "deferred restart — subagents in flight" alert and the "system recovered automatically" alert.
- Two alert call sites are **never suppressed**: BLACK state (max restarts exceeded) and restart failed. Both require user action regardless of what triggered the restart.

---

## Failure paths: what is and is not suppressed

| Alert | Suppressed during compaction window? | Reason |
|-------|--------------------------------------|--------|
| Deferred restart (subagents in flight) | Yes | Informational; compaction already notified |
| System recovered automatically | Yes | Informational; compaction already notified |
| BLACK: max restart attempts exceeded | **No** | Requires user intervention |
| Restart failed / service timeout | **No** | Requires user intervention |

---

## Flow diagram: compaction event to notification

```
Claude context compaction triggered
           |
           v
   on-compact.py (hook)
   +-----------------------------------+
   | 1. write compacted_at             |
   | 2. write compact-reminder msg     |
   | 3. send_compaction_notify()       |-----> Telegram: "Context compacted."
   +-----------------------------------+         (always, unconditionally)
           |
           | (lobster restarts, re-orients from CLAUDE.md + handoff)
           |
           v  ~60s later
   health-check-v3.sh
   +--------------------------------------------------------------+
   | is_compaction_recent()?                                      |
   |   yes -> compaction_recent=true                              |
   |   no  -> compaction_recent=false                             |
   +--------------------------------------------------------------+
           |
    +------+------------------------------------+
    |                                           |
    v                                           v
   do_restart(reason, suppress=true)      do_restart(reason, suppress=false)
   +--------------------------------+     +--------------------------------+
   | subagent guard? -> log only    |     | subagent guard? -> alert sent  |
   | restarted OK?  -> log only     |     | restarted OK?  -> alert sent   |
   | BLACK state?   -> ALERT fires  |     | BLACK state?   -> alert sent   |
   | restart failed?-> ALERT fires  |     | restart failed?-> alert sent   |
   +--------------------------------+     +--------------------------------+
         0 extra notifications                  normal alert behaviour
```

---

## Tests

- `uv run pytest tests/smoke/test_on_compact.py -v` — 5 passed, including new **A5** test that pins the always-on notification behaviour: asserts `send_compaction_notify()` is called exactly once regardless of `LOBSTER_DEBUG`.
- `uv run pytest tests/smoke/ -v` — 36 passed, 7 failed (all 7 pre-existing failures in `test_require_background_agent.py` and `test_require_write_result.py`, unrelated to this PR).

Closes #684

### DIFF
--- a/hooks/on-compact.py
+++ b/hooks/on-compact.py
@@ -9,9 +9,10 @@ reminder to re-read CLAUDE.md and re-orient from handoff/memory context.
 The script is idempotent: if a compact-reminder message already exists in the
 inbox it skips writing a duplicate.
 
-Dev mode: if LOBSTER_DEBUG=true (or set in config.env), also sends a Telegram
-message directly to the owner's chat ID so the developer is immediately notified
-that a compaction occurred.
+Notification: always sends a Telegram message directly to the owner's chat ID
+so the user is immediately notified that a compaction occurred.  The health
+check suppresses its own alerts during the compaction window so exactly one
+notification reaches the user per compaction event.
 
 State: always writes compacted_at to lobster-state.json so that the health
 check can suppress stale-inbox false-positives during the compaction pause.
@@ -59,7 +60,7 @@ REMINDER_TEXT = (
 
 SENTINEL_FILE = Path(os.path.expanduser("~/messages/config/compact-pending"))
 
-DEV_TELEGRAM_MESSAGE = "\u26a0\ufe0f [DEV] Context compacted. Re-orienting from CLAUDE.md + handoff."
+COMPACTION_TELEGRAM_MESSAGE = "\u267b\ufe0f Context compacted. Re-orienting..."
 
 
 def already_pending() -> bool:
@@ -172,24 +173,15 @@ def _parse_config_env() -> dict:
     return config
 
 
-def _is_debug_mode(config: dict) -> bool:
-    """Return True if LOBSTER_DEBUG is 'true' in the environment or config.env."""
-    env_val = os.environ.get("LOBSTER_DEBUG", "").lower()
-    if env_val == "true":
-        return True
-    config_val = config.get("LOBSTER_DEBUG", "").lower()
-    return config_val == "true"
-
-
-def _send_telegram_dev_notify(bot_token: str, chat_id: str) -> None:
+def _send_telegram_notify(bot_token: str, chat_id: str, text: str) -> None:
     """
-    Send DEV_TELEGRAM_MESSAGE to chat_id via the Telegram Bot API.
+    Send text to chat_id via the Telegram Bot API.
     Logs to stderr on failure so the cause is visible in Claude hook output.
     Never raises — must not crash the hook.
     """
     try:
         url = f"https://api.telegram.org/bot{bot_token}/sendMessage"
-        payload = json.dumps({"chat_id": chat_id, "text": DEV_TELEGRAM_MESSAGE}).encode()
+        payload = json.dumps({"chat_id": chat_id, "text": text}).encode()
         req = urllib.request.Request(
             url,
             data=payload,
@@ -217,15 +209,16 @@ def _send_telegram_dev_notify(bot_token: str, chat_id: str) -> None:
         print(f"[on-compact] Telegram notify failed: {type(exc).__name__}: {exc}", file=sys.stderr)
 
 
-def maybe_send_dev_telegram_notify() -> None:
+def send_compaction_notify() -> None:
     """
-    If LOBSTER_DEBUG is true and credentials are available, send a Telegram
-    notification to the owner that a context compaction occurred.
+    Send a Telegram notification to the owner that a context compaction occurred.
+
+    Always fires when credentials are available — not gated on LOBSTER_DEBUG.
+    This is the single canonical notification for a compaction event; the
+    health-check suppresses its own alerts during the compaction window so
+    exactly one notification reaches the user per compaction.
     """
     config = _parse_config_env()
-
-    if not _is_debug_mode(config):
-        return
 
     bot_token = config.get("TELEGRAM_BOT_TOKEN", "").strip()
     allowed_users = config.get("TELEGRAM_ALLOWED_USERS", "").strip()
@@ -236,7 +229,7 @@ def maybe_send_dev_telegram_notify() -> None:
     # Take the first user ID from a comma- or space-separated list.
     first_chat_id = allowed_users.replace(",", " ").split()[0]
 
-    _send_telegram_dev_notify(bot_token, first_chat_id)
+    _send_telegram_notify(bot_token, first_chat_id, COMPACTION_TELEGRAM_MESSAGE)
 
 
 def main() -> None:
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

All five review points check out. No blocking issues found.

---

## 1. suppress_alert — correct suppression scope

`do_restart()` has 7 `send_telegram_alert` call sites total. The PR suppresses exactly two:

- **Suppressed** (compaction noise): deferred-restart subagent-guard alert (line ~937), recovered-successfully alert (line ~1119).
- **Always fires** regardless of flag: BLACK/unrecoverable (line ~953), ABORT can't-kill-PID (line ~1019), post-restart inbox still stale (line ~1109), restart-failed PID unchanged (line ~1129), service/tmux not ready (line ~1144).

Every genuine failure path fires. The two suppressed alerts are correctly identified as health-check noise that duplicates the on-compact.py notification.

## 2. on-compact.py — exactly one notification

`send_compaction_notify()` is called unconditionally once in `main()`. The `_is_debug_mode` gate is fully removed. Call count is correct.

## 3. 0-notification edge case

If `TELEGRAM_BOT_TOKEN` or `TELEGRAM_ALLOWED_USERS` is absent, `send_compaction_notify()` returns early and no notification is sent — same silent-fail behavior as before. This is acceptable: users without Telegram configured don't expect Telegram notifications. No regression here.

## 4. Bash string comparison

`[[ "$suppress_alert" != "true" ]]` is correct. `compaction_recent` is set as a bare `true`/`false` bash string in `main()` and passed positionally. The `${2:-false}` default also handles missing-arg correctly.

## 5. install.sh completeness

`COMPACTION_SUPPRESS_SECONDS` is a script-internal constant hardcoded at line 69 of `health-check-v3.sh` (`=300`). It is not a `config.env` user variable and does not need an `install.sh` entry. No config.env additions in this PR.

---

## Minor notes (non-blocking)

- The "post-restart inbox still stale" alert at line ~1109 fires unconditionally — this is correct since it represents a new stale inbox after restart, not the original compaction-triggered restart.
- Test A5 uses a tracking stub that replaces the no-op lambda from `_load_hook`, so the call count assertion is valid and the `LOBSTER_DEBUG` env cleanup in the `finally` block is thorough.

LGTM.

PR #691, opened 2026-03-18
fix: subagents default to dispatcher routing when unsure about user notification

### LINKED ISSUE #690: fix: subagents default to dispatcher routing when unsure about user notification
## Problem

Subagents currently treat all tasks as user-facing by default, calling \`send_reply\` directly before \`write_result\`. But some tasks — most notably opening a PR — should route through the dispatcher first so a reviewer can be spawned before the user is notified. There's no guidance in the subagent bootup for this case.

## Goal

Add a rule near the \`write_result\` / \`send_reply\` behavior section of \`sys.subagent.bootup.md\` that captures:

- When it's clear that a code review (or similar dispatcher action) should happen before user notification (e.g., you opened a PR), do NOT call \`send_reply\` — call \`write_result\` only and let the dispatcher route.
- When in doubt about whether to reply directly or route through the dispatcher, default to routing through the dispatcher (\`write_result\` without \`sent_reply_to_user=True\`, no \`send_reply\`).

## Out of scope

No changes to dispatcher behavior or \`write_result\` semantics — this is purely a guidance addition to the subagent bootup doc.

sayhar:
🤖🦞 Lobster (engineer):

Closes #690. Replaces the intent of (now-closed) PR #519 with a simpler, general rule.

## What changed and why

Before this change, subagents had a simple binary: if the task prompt says "do NOT call send_reply", treat it as internal; otherwise treat it as user-facing and call `send_reply` directly. This created a gap — subagents that open PRs were calling `send_reply` immediately, bypassing any chance for the dispatcher to spawn a reviewer first.

This PR adds a qualification to that default: when a task produces output that the dispatcher should act on before the user sees it (the clearest example: you just opened a PR and a code review should happen first), skip `send_reply` and call `write_result` only. And when it's not clear which applies, the rule is to route through the dispatcher rather than reply directly. The dispatcher can always relay; a premature `send_reply` cannot be un-sent.

The change is a single sentence addition to the delivery-pattern summary at the top of `sys.subagent.bootup.md`, immediately where subagents learn the user-facing vs. internal distinction.

## Scope

One line changed in `.claude/sys.subagent.bootup.md`. No code changes, no behavior changes to existing tools or infrastructure.

## Tests run

- [x] Manual diff review — change is exactly one sentence added to the summary paragraph, no other lines touched
- [ ] Runtime behavior — doc-only change; no automated test coverage for prompt behavior

### DIFF
--- a/.claude/sys.subagent.bootup.md
+++ b/.claude/sys.subagent.bootup.md
@@ -16,7 +16,7 @@ When your task is complete, choose the right delivery pattern based on your task
 - **User-facing tasks (default):** call `send_reply` directly, then `write_result`. This is the crash-safe delivery pattern — the user gets their reply even if the dispatcher has restarted.
 - **Internal tasks (dispatcher-only):** skip `send_reply`. Call `write_result` only. The dispatcher reads your result and decides what to relay.
 
-Your task prompt will say "do NOT call send_reply" or "Use write_result only" for internal tasks. If it says nothing, treat it as user-facing.
+Your task prompt will say "do NOT call send_reply" or "Use write_result only" for internal tasks. If it says nothing, treat it as user-facing — with one important exception: **when your task produces output the dispatcher should review before notifying the user (e.g., you opened a PR), do NOT call `send_reply`. Call `write_result` only and let the dispatcher route.** When in doubt about whether to reply directly or route through the dispatcher, default to routing through the dispatcher (`write_result` without `sent_reply_to_user=True`, no `send_reply`). The dispatcher will decide what to relay and when.
 
 See the **"Internal vs. User-Facing Tasks"** section below for full patterns and code examples.
 
--- a/.claude/sys.subagent.bootup.md
+++ b/.claude/sys.subagent.bootup.md
@@ -16,7 +16,7 @@ When your task is complete, choose the right delivery pattern based on your task
 - **User-facing tasks (default):** call `send_reply` directly, then `write_result`. This is the crash-safe delivery pattern — the user gets their reply even if the dispatcher has restarted.
 - **Internal tasks (dispatcher-only):** skip `send_reply`. Call `write_result` only. The dispatcher reads your result and decides what to relay.
 
-Your task prompt will say "do NOT call send_reply" or "Use write_result only" for internal tasks. If it says nothing, treat it as user-facing — with one important exception: **when your task produces output the dispatcher should review before notifying the user (e.g., you opened a PR), do NOT call `send_reply`. Call `write_result` only and let the dispatcher route.** When in doubt about whether to reply directly or route through the dispatcher, default to routing through the dispatcher (`write_result` without `sent_reply_to_user=True`, no `send_reply`). The dispatcher will decide what to relay and when.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Lobster (reviewer): NEEDS-WORK

**Summary:** The intent is correct, but the sentence is too long and introduces ambiguity at the edges.

**What works:**
- The PR example ("e.g., you opened a PR") directly maps to Sahar's "spin up a code review before telling the user" intent
- "When in doubt, route through the dispatcher" is a sound default
- The explicit clarification that `write_result` without `sent_reply_to_user=True` is the routing mechanism is useful

**Concerns:**

1. **Sentence length and parsing.** The added text is one run-on sentence split across two conceptual rules: (a) the exception to "treat as user-facing" and (b) the tie-breaker for ambiguous cases. These are distinct enough to warrant two sentences or a sub-bullet.

2. **Ambiguity loophole.** "When your task produces output the dispatcher should review before notifying the user" is agent-interpreted. A subagent could reason: "I wrote a comment, not a PR — the dispatcher doesn't need to review *that* first" and rationalize using `send_reply`. The original wording ("when in doubt, do not reply") is a stronger gate because it places the burden on the agent to be *certain* it's user-facing, not just that it *thinks* the dispatcher doesn't need first look.

   Sahar's framing ("ideally this will be clear to you, but when in doubt — do not reply") implies the default should tilt toward `write_result` for anything that touches external systems (GitHub, calendar, code, etc.), not just when the agent infers the dispatcher wants first dibs.

3. **The "with one important exception" framing** weakens the existing rule without replacing it. Currently: "If it says nothing, treat it as user-facing." The new text appends an exception that's vague in scope. A cleaner structure would be:
   - Rule: default to user-facing *unless* the task touches external systems or produces artifacts (PR, commit, file change, report) — those always route through the dispatcher.
   - Tie-breaker: when genuinely uncertain, `write_result` only.

**Suggested rework (two sentences to replace the appended block):**

> If it says nothing, treat it as user-facing — unless your task produced an artifact or side-effect the dispatcher should review first (e.g., you opened a PR, committed code, or created a report). When in doubt, do not call `send_reply`; call `write_result` only and let the dispatcher route.

This mirrors Sahar's exact phrasing more closely, removes the agent's interpretive wiggle room, and separates the two rules cleanly.

**Verdict:** The insertion point (mid-paragraph after the existing "if it says nothing" rule) is appropriate. The logic is right but the current wording leaves a rationalization gap. A small rewrite tightens it.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Checklist:**

1. **Explicit categories** — PASS. "PR opened, commit pushed, report generated" are concrete examples, not a vague judgment.

2. **Hard tie-breaker** — PASS. "When in doubt, default to `write_result` without `send_reply`" is a hard rule, not discretion.

3. **Conciseness** — PASS (marginal). The sentence is denser than the original, but every clause is load-bearing: the category list, the rule for those categories, the tie-breaker, and the rationale. The rationale ("The dispatcher can always relay; a premature reply cannot be un-sent") could be cut to save ~12 words, but it does earn its keep by making the rule memorable. No fluff.

4. **No regression** — PASS. The user-facing default ("If it says nothing, treat it as user-facing") is preserved. The explicit-prompt detection ("do NOT call send_reply" / "Use write_result only") is unchanged. The new clause is additive.

One minor observation: the sentence is now long enough that a subagent reading fast might miss the tie-breaker buried at the end. A future pass could split it into two sentences for scannability, but it's not a blocker.

PR #704, opened 2026-03-19
fix: suppress system_context debug observations from Telegram delivery

### LINKED ISSUE #703: fix: suppress system_context debug observations from Telegram delivery
## Problem

When `LOBSTER_DEBUG=true`, users see debug messages in Telegram like:

```
[debug|mcp-only] system_context from unknown
🔍 [tier 1 fired] msg=... extracted 1 signal(s): topic='learning' (0.60)
```

These come from `_emit_debug_observation()` in `inbox_server.py`, which writes directly to the bot outbox (bypassing the dispatcher) for all categories including `system_context`.

## Expected behavior

Per `sys.dispatcher.bootup.md` (subagent_observation routing table):

> `system_context` | `memory_store` silently (no user message) | same as debug-off — do NOT send_reply

The doc is explicit: `system_context` observations must never be forwarded to the user, even in debug mode.

## Root cause

`_emit_debug_observation()` does not filter on `category`. All callers that pass `category="system_context"` (or use the default) still deliver to Telegram:

- `_observation_worker()` line ~136: tier 1 signal summaries (uses default `system_context`)
- `handle_write_result()` line ~5197: subagent→dispatcher routing events
- `handle_memory_store()` line ~5966: memory write debug events
- `handle_memory_search()` line ~6003: memory search debug events

Note: `handle_write_observation()` already correctly suppresses `system_context` at line 5285 (`if _DEBUG_MODE and category != "system_context"`). The fix brings `_emit_debug_observation()` in line with that same rule.

## Fix

Add an early return in `_emit_debug_observation()` when `category == "system_context"`, making it consistent with the existing suppression logic in `handle_write_observation`.

## Related

- PR #351 (added this debug delivery path)
- Issue #396 (observation pipeline architecture)
- Issue #634 (event bus to replace hardcoded debug delivery)

sayhar:
## Summary

Closes #703.

- `_emit_debug_observation()` was sending all debug observations to Telegram, including `system_context` ones (tier-1 signal summaries, memory writes, subagent routing events)
- Per `sys.dispatcher.bootup.md`, `system_context` must be stored silently and **never** forwarded to the user, even in debug mode
- Added an early return in `_emit_debug_observation()` when `category == "system_context"`, consistent with the suppression already present in `handle_write_observation()` (line 5285)

**Before:** With `LOBSTER_DEBUG=true`, users see noisy messages like:
```
[debug|mcp-only] system_context from unknown
🔍 [tier 1 fired] msg=... extracted 1 signal(s): topic='learning' (0.60)
```

**After:** Only `system_error` and `user_context` debug observations reach Telegram. `system_context` is always suppressed.

## Affected call sites (all now suppressed)

| Location | Category | Description |
|---|---|---|
| `_observation_worker()` ~L136 | `system_context` (default) | Tier-1 signal summaries |
| `handle_write_result()` ~L5197 | `system_context` | Subagent→dispatcher routing events |
| `handle_memory_store()` ~L5966 | `system_context` | Memory write debug events |
| `handle_memory_search()` ~L6003 | `system_context` | Memory search debug events |

## Test plan

- [ ] With `LOBSTER_DEBUG=true`, send a message and confirm no `[debug|mcp-only] system_context` messages appear in Telegram
- [ ] Confirm `system_error` debug observations (e.g. from observation worker exceptions) still deliver to Telegram
- [ ] Confirm `user_context` debug observations still deliver to Telegram

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -300,10 +300,13 @@ def _emit_debug_observation(
     never sees these messages and no dispatcher tokens are burned.
 
     When debug mode is off, this function is a no-op.
+    When category is "system_context", this function is always a no-op — per the
+    dispatcher bootup doc, system_context observations must be stored silently and
+    never forwarded to the user, even in debug mode.
 
     Args:
         text: The observation body text.
-        category: "system_context", "system_error", or "user_context".
+        category: "system_context" (always suppressed), "system_error", or "user_context".
         visibility: "mcp-only" if the MCP layer is emitting this directly (dispatcher
             has not seen it yet), or "dispatcher" if the dispatcher's main loop is
             emitting this after processing the message through its inbox.
@@ -311,10 +314,17 @@ def _emit_debug_observation(
             Falls back to "unknown" if not provided.
 
     Label format: [debug|{visibility}] {category} from {emitter}
-    Example: [debug|mcp-only] system_context from task:linear-digest
+    Example: [debug|mcp-only] system_error from task:linear-digest
 
     Never raises — must be safe to call from any context including threads.
     """
+    # system_context observations are internal routing decisions — never forward
+    # to Telegram. The dispatcher bootup doc is explicit: system_context must be
+    # stored silently (memory_store), do NOT send_reply, even in debug mode.
+    # This mirrors the suppression already present in handle_write_observation().
+    if category == "system_context":
+        return
+
     # Fast path: skip I/O if debug alerts have been resolved and are disabled.
     if _DEBUG_RESOLVED and not _DEBUG_ALERTS_ENABLED:
         return


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The fix is correct and minimal. The early return for `category == "system_context"` at line 323 is exactly the right approach — it mirrors the existing suppression guard already in `handle_write_observation()` at line 5285, making the behavior consistent across both paths.

**What I verified:**

1. **Call sites are correctly handled.** All four call sites noted in the PR description pass `category="system_context"` (either explicitly or via the default). After the fix, none of them will reach the outbox write. The two call sites that pass `category="system_error"` (observation worker exception at ~L144, context inject error at ~L3739) correctly still pass through.

2. **Default parameter alignment.** The function signature has `category: str = "system_context"` as default. Callers that rely on the default (e.g., tier-1 signal summary at ~L136, context injection debug at ~L3726, L3733) were previously leaking to Telegram in debug mode — this is exactly the bug described in the PR. The fix suppresses all of them.

3. **Docstring and comment updates are accurate.** The updated docstring ("always suppressed"), inline comment citing the dispatcher bootup doc, and updated example label (changed from `system_context` to `system_error`) are all consistent with the new behavior.

4. **No behavior change for non-system_context categories.** `system_error` and `user_context` observations still flow through normally.

One minor observation (not blocking): the PR description's table lists `handle_write_result()` ~L5197 as a call site, and that call at line 5197 passes `category="system_context"` explicitly. However, `handle_write_observation()` at L5285 already suppresses `system_context` before calling `_emit_debug_observation`, so that path was already safe. The new early return adds defense-in-depth there but isn't strictly necessary for that specific call. This is fine — defense-in-depth is correct behavior here.

No issues. Ready to merge.

PR #707, opened 2026-03-19
🤖🦞 Lobster (engineer): add get_message_by_telegram_id MCP tool

### LINKED ISSUE #706: Add get_message_by_telegram_id tool to MCP inbox server
## Problem

The dispatcher uses \`get_conversation_history\` to retrieve messages by recency/sequence. There's no way to fetch a specific message by its Telegram message ID (the integer ID like 12824). This is needed when the dispatcher needs to look up the original content of any message by its Telegram ID — for example, to reconstruct context for a reply chain.

## What to build

Add a new \`get_message_by_telegram_id\` MCP tool that:
- Accepts a \`telegram_message_id\` (integer) and optional \`chat_id\` (for disambiguation when the same TG message ID could appear across chats — unlikely but safe)
- Searches \`processed/\`, \`inbox/\`, \`processing/\`, and \`failed/\` directories for a message JSON file containing a matching \`telegram_message_id\` field
- Returns the full message object (text, timestamp, reply_to, type, source, etc.)
- Also expose the new tool in the read-only HTTP allowlist (\`inbox_server_http.py\`)

## Why a new tool (not extending get_conversation_history)

\`get_conversation_history\` is a bulk listing tool that loads, filters, and sorts all messages. Adding a single-message lookup by ID requires a different access pattern (direct field match, stop-on-first-hit). A focused tool keeps the interface cleaner and intent clear.

## Acceptance criteria

- [ ] New tool \`get_message_by_telegram_id\` registered in the tools list
- [ ] Handler searches across all relevant message directories
- [ ] Returns full message JSON including text, timestamp, type, reply_to, telegram_message_id
- [ ] Graceful "not found" response when no match exists
- [ ] Tool added to \`READONLY_TOOLS\` in \`inbox_server_http.py\`
- [ ] Tests pass (ruff, mypy if applicable)

sayhar:
Closes #706

## Why this exists

The dispatcher has no way to retrieve a specific message when it knows only a Telegram message ID (the integer like 12824). `get_conversation_history` loads all messages, filters, and paginates — the wrong tool for a point lookup. This gap matters whenever the dispatcher needs to reconstruct the original content of a message it holds only a TG ID reference for (e.g. from a reply chain, a subagent notification, or a stale check).

## What changed

A new `get_message_by_telegram_id(telegram_message_id, chat_id?)` tool is added to the lobster-inbox MCP server.

**Access pattern:** pure point lookup — searches `processed/`, `inbox/`, `processing/`, and `failed/` in that priority order, stops on the first match. No sorting, no pagination, no full directory scans beyond what's needed to find the hit.

**Return value:** the full message object rendered as text — Telegram message ID, Lobster internal message ID, source, chat ID, type, from user, timestamp, full text, reply-to context (truncated at 300 chars), and any attached file paths. "Not found" returns a clear message naming which directories were searched.

**Optional `chat_id` parameter:** disambiguates when the same TG message ID could theoretically exist across chats. Not required in practice since TG message IDs are unique per chat, but safe to pass.

**HTTP read-only bridge:** `get_message_by_telegram_id` is also added to `READONLY_TOOLS` in `inbox_server_http.py` so remote Claude Code instances connected over the HTTP bridge can use it.

## What is NOT changed

`get_conversation_history` is untouched — no behavior changes to any existing tool. Only `inbox_server.py` (tool registration + new handler) and `inbox_server_http.py` (allowlist entry) are modified.

## Flow note

No state machine changes. This is a pure read-only query path — no message moves between directories, no marks, no side effects.

## Functional patterns used

Pure function: `handle_get_message_by_telegram_id` has no side effects, reads only from the filesystem, and returns a deterministic result given the same inputs. The search is expressed as a lazy scan that short-circuits on first match (early return inside the loop) rather than loading all messages into memory.

## Tests run

- [x] `uv run ruff check src/mcp/inbox_server.py src/mcp/inbox_server_http.py` — pre-existing errors only (45 errors all present on `main`; zero errors in the new code)
- [x] Live functional test: called `handle_get_message_by_telegram_id({'telegram_message_id': 9666})` against the real `~/messages/` directory — correctly returned the matching message from `processed/` including text, timestamp, and reply-to context
- [x] Not-found case: `telegram_message_id=99999999` — returned expected "No message found" message naming all searched directories
- [ ] Full test suite (`pytest`) — no test suite exists for this file in the repo; no tests to run

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1312,6 +1312,28 @@ async def list_tools() -> list[Tool]:
                 },
             },
         ),
+        # Telegram Message Lookup Tool
+        Tool(
+            name="get_message_by_telegram_id",
+            description="Look up a specific message by its Telegram message ID (the integer ID like 12824). Searches across all message directories (processed, inbox, processing, failed) and returns the full message object. Useful when the dispatcher needs to retrieve original message content for a known Telegram message ID.",
+            inputSchema={
+                "type": "object",
+                "properties": {
+                    "telegram_message_id": {
+                        "type": "integer",
+                        "description": "The Telegram message ID (integer) to look up.",
+                    },
+                    "chat_id": {
+                        "oneOf": [
+                            {"type": "integer"},
+                            {"type": "string"},
+                        ],
+                        "description": "Optional: filter by chat ID to disambiguate when the same Telegram message ID might appear in multiple chats.",
+                    },
+                },
+                "required": ["telegram_message_id"],
+            },
+        ),
         # Task Management Tools
         Tool(
             name="list_tasks",
@@ -2601,6 +2623,8 @@ async def _dispatch_tool(name: str, arguments: dict[str, Any]) -> list[TextConte
         return await handle_get_stats(arguments)
     elif name == "get_conversation_history":
         return await handle_get_conversation_history(arguments)
+    elif name == "get_message_by_telegram_id":
+        return await handle_get_message_by_telegram_id(arguments)
     elif name == "list_tasks":
         return await handle_list_tasks(arguments)
     elif name == "create_task":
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Summary:** Clean, well-scoped addition. The implementation is correct, has no side effects, and follows existing patterns (json-load-with-except, same directory constants, TextContent return type). The "not found" message is clear and actionable — it names all four searched directories by label, which gives the dispatcher enough context to act.

**Filename suffix optimization (noted, not blocking):** The file naming convention is \`{timestamp_ms}_{telegram_message_id}.json\`. A single \`f.stem.endswith(f"_{tg_id}")\` check before \`open()\` would skip the JSON parse for every non-matching file, making the scan O(matches) in parse overhead rather than O(all files). For a large \`processed/\` directory this could be a meaningful speedup. Suggested addition inside the loop, before the try/open block:

\`\`\`python
if not f.stem.endswith(f"_{tg_id}"):
    continue
\`\`\`

Not a correctness issue — just a cheap win for directories with many files.

**Minor observations (informational):**
- \`async\` declaration with synchronous file I/O is consistent with the existing pattern in this file — fine.
- \`chat_id\` string-coercion comparison is correct and safe.
- \`reply_to\` field access uses \`.get()\` with fallbacks throughout — no KeyError risk.
- HTTP allowlist placement is correct (inside READONLY_TOOLS, alphabetically ordered with existing entries).

**No issues found in:** tool schema (oneOf for integer|string is valid JSON Schema), dispatch routing, error handling on JSON parse, timestamp parsing for both Z-suffix and +offset formats, or image_files list handling.

PR #711, opened 2026-03-19
fix: prevent dispatcher stall on large subagent results (#705)

### LINKED ISSUE #705: Dispatcher violates 7-second rule when composing reply to large subagent results
## What happened

During the 03:28 UTC health check restart, the dispatcher received a large subagent result (1,265 lines) and spent approximately 8 minutes composing an inline response on the main thread. This directly violates the 7-second rule.

## Impact

- 4 user messages queued up unanswered during the inline composition
- The health check detected the stall and triggered a session restart
- Messages were delayed and session state was disrupted

## Root cause

The dispatcher currently handles subagent results inline — reading the result and composing a user-facing reply in the same main-loop iteration. For small results this is fine, but for large or complex results, the LLM composition time can far exceed the 7-second threshold.

## Fix

When a subagent result requires a substantial reply to the user, the dispatcher should spawn a reply-writing subagent instead of composing inline. This is the same pattern already used for incoming user messages that require non-trivial work.

Concretely:
- After receiving a subagent result, the dispatcher should estimate or detect whether composing the reply will be fast (short, templated) or slow (long result, complex synthesis)
- If slow: spawn a `reply-writer` subagent, pass it the subagent result and the original user context, and return to the main loop immediately
- The reply-writer subagent calls `send_reply` + `write_result` when done, same as any other subagent

## Notes

The 7-second rule currently reads as applying to user messages. This issue clarifies that it applies equally to subagent result handling — any main-thread work that could take more than ~2 seconds should be offloaded to a subagent.

sayhar:
🤖🦞 Lobster (engineer):

## Problem

The dispatcher main loop was receiving subagent results inline and composing user-facing replies on the main thread. For large results (1,265 lines in the 2026-03-19 incident), this LLM composition time reached 8 minutes — far exceeding the 7-second rule — and triggered a health-check restart with 4 messages queued unanswered.

The 7-second rule was documented as applying to *incoming user messages*, but the same constraint applies equally to subagent result handling: any main-thread work that can take more than ~2 seconds must be offloaded.

## What changed

Three coordinated layers of defence, from innermost (enforcement) to outermost (authoring guidance):

**1. Server-side truncation (`inbox_server.py` — `handle_write_result`)**

`write_result` text is now hard-truncated at 2,000 chars with the suffix `\n[Result truncated at 2,000 chars. Use artifacts for full content.]`. This is a safety net: a 1,000-line result can no longer reach the dispatcher at all. The truncation is logged as a warning so operators can identify subagents that need updating.

**2. Dispatcher offload rule (`.claude/sys.dispatcher.bootup.md`)**

The `subagent_result` handler's `else` branch (non-PR, non-notification results) now checks `len(reply_text) > 800`. When true, it spawns a `lobster-generalist` relay subagent to compose and send the reply, then immediately returns to `wait_for_messages()`. Results ≤800 chars are still sent inline — the fast path is preserved for typical short replies.

**Before (both paths inline):**
```
subagent_result arrives
  → if PR URL: spawn reviewer (already async) ✓
  → else: build reply_text, send_reply inline  ← blocks main loop for large results
```

**After:**
```
subagent_result arrives
  → if PR URL: spawn reviewer (async) ✓
  → else if len(reply_text) > 800: spawn relay subagent (async) ✓
  → else: send_reply inline (fast path, ≤800 chars only)
```

**3. Subagent authoring guidance (`.claude/sys.subagent.bootup.md` + `write_result` tool description)**

Added a bolded rule near the top of the long-reports section:

> **Keep `write_result` text short.** The `text` field should be ≤2,000 chars — a summary or verdict, not the full output. Put full content in `artifacts` (file paths). A 1,000-line report in `text` will stall the main loop and may trigger a health-check restart.

The `write_result` tool `text` field description in the MCP schema now also states the limit and its enforcement, so the constraint is visible at the point of use.

## What is not changed

- The PR-URL branch in the `subagent_result` handler is unchanged — it was already async (spawns a reviewer subagent).
- The `subagent_notification` and `subagent_error` paths are unchanged.
- No changes to the Telegram bot, session store, or any runtime data paths.

## Tests run

- [ ] No unit tests exist for `handle_write_result` — skipped: no test infrastructure for this module
- [ ] Live integration test — blocked: requires running dispatcher session; safe to merge as the change is a pure safety net with no behaviour change for results ≤2,000 chars

**Blocked items needing attention before merge:** none — the truncation only fires for results that already caused the 8-minute stall.

Closes #705

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -332,14 +332,36 @@ Check the `sent_reply_to_user` field first, then check for engineer → reviewer
                        reply_text += f"\n\n---\n{content}"
                    except:
                        pass  # skip unreadable files silently
-           send_reply(
-               chat_id=msg["chat_id"],
-               text=reply_text,
-               source=msg.get("source", "telegram"),
-               thread_ts=msg.get("thread_ts"),            # Slack thread
-               reply_to_message_id=msg.get("telegram_message_id")  # Telegram threading
-           )
-           mark_processed(message_id)
+
+           # Large result text (>~800 chars): composing and relaying a long result inline
+           # can exceed the 7-second rule and stall the main loop (issue #705).
+           # Offload to a reply-writer subagent; return to the loop immediately.
+           if len(reply_text) > 800:
+               Task(
+                   subagent_type="lobster-generalist",
+                   run_in_background=True,
+                   prompt=(
+                       f"---\n"
+                       f"task_id: relay-{msg.get('task_id', 'unknown')}\n"
+                       f"chat_id: {msg['chat_id']}\n"
+                       f"source: {msg.get('source', 'telegram')}\n"
+                       f"---\n\n"
+                       f"Relay the following subagent result to the user. "
+                       f"Send the full content via send_reply, then call write_result with sent_reply_to_user=True.\n\n"
+                       f"{reply_text}"
+                   ),
+               )
+               mark_processed(message_id)
+               # Return to wait_for_messages() immediately — reply-writer delivers async
+           else:
+               send_reply(
+                   chat_id=msg["chat_id"],
+                   text=reply_text,
+                   source=msg.get("source", "telegram"),
+                   thread_ts=msg.get("thread_ts"),            # Slack thread
+                   reply_to_message_id=msg.get("telegram_message_id")  # Telegram threading
+               )
+               mark_processed(message_id)
 ```
 
 **IMPORTANT — never relay raw file paths to the user.** File paths like `~/lobster-workspace/reports/foo.md` are server-side references that are useless on mobile. When a `subagent_result` contains `artifacts`, read the files and include their content inline in `send_reply`. Do not mention the path in the reply.
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -323,12 +323,14 @@ Check the `sent_reply_to_user` field first, then check for engineer → reviewer
                mark_processed(message_id)
                # Return to wait_for_messages() — reviewer's write_result arrives separately
        else:
-           # Large result text (>~4,000 chars): the dispatcher would need significant
-           # LLM time to compose and relay inline, risking the 7-second rule (issue #705).
+           # Large result text: the dispatcher would need significant LLM time to compose
+           # and relay inline, risking the 7-second rule (issue #705).
            # Check size BEFORE inlining artifacts so the relay subagent receives only
            # msg["text"] (already bounded at 2,000 chars by inbox_server truncation)
            # and handles artifact inlining itself.
-           if len(msg["text"]) > 4000:
+           # Note: write_result text is server-truncated at 2,000 chars; this threshold
+           # must stay below that cap to be reachable.
+           if len(msg["text"]) > 1200:
                Task(
                    subagent_type="lobster-generalist",
                    run_in_background=True,
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -323,20 +323,12 @@ Check the `sent_reply_to_user` field first, then check for engineer → reviewer
                mark_processed(message_id)
                # Return to wait_for_messages() — reviewer's write_result arrives separately
        else:
-           # Build reply text: inline artifact content when present
-           reply_text = msg["text"]
-           if msg.get("artifacts"):
-               for artifact_path in msg["artifacts"]:
-                   try:
-                       content = Read(artifact_path)   # read the file
-                       reply_text += f"\n\n---\n{content}"
-                   except:
-                       pass  # skip unreadable files silently
-
-           # Large result text (>~800 chars): composing and relaying a long result inline
-           # can exceed the 7-second rule and stall the main loop (issue #705).
-           # Offload to a reply-writer subagent; return to the loop immediately.
-           if len(reply_text) > 800:
+           # Large result text (>~4,000 chars): the dispatcher would need significant
+           # LLM time to compose and relay inline, risking the 7-second rule (issue #705).
+           # Check size BEFORE inlining artifacts so the relay subagent receives only
+           # msg["text"] (already bounded at 2,000 chars by inbox_server truncation)
+           # and handles artifact inlining itself.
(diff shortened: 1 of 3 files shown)


sayhar (commented):

## Overall

The three-layer approach is sound and well-motivated. The server-side truncation (layer 1) and documentation updates (layer 3) are ready. The dispatcher offload rule (layer 2) has one correctness issue and one design concern worth addressing before merge.

---

## Layer 1 — `inbox_server.py` truncation (LGTM)

- Truncation at 2,000 chars is a reasonable safety net.
- Warning log is appropriate for operator visibility.
- The suffix `\n[Result truncated at 2,000 chars. Use artifacts for full content.]` is clear.
- One minor note: `_WRITE_RESULT_TEXT_LIMIT` is defined inside the function body on every call. Consider moving it to module level (e.g. `WRITE_RESULT_TEXT_LIMIT = 2_000`) for easier configuration later. Not a blocker.

---

## Layer 2 — Dispatcher relay subagent (NEEDS-WORK)

### Issue 1: relay subagent prompt contains the full reply_text verbatim

The prompt passed to the relay subagent is:

```
f"Relay the following subagent result to the user. ...\n\n{reply_text}"
```

`reply_text` at this point already has artifacts inlined (the loop above reads artifact files and appends them). If a subagent returns a 1,500-char text + several large artifact files, `reply_text` could be many thousands of characters — and the relay subagent receives all of it as its prompt. This doesn't fix the root stall; it just moves the large payload construction to a different thread. The relay subagent also has to call `send_reply` with that content, which itself may be truncated by Telegram/Slack limits.

**Suggested fix:** pass `reply_text` as-is (already bounded at 2,000 chars by the server), and pass artifact paths separately so the relay subagent can decide how to format them — or read and inline them itself. Alternatively, cap the artifact inlining loop at a character budget.

### Issue 2: 800-char threshold may be too aggressive (engineer-flagged)

The engineer flagged this: "Could it over-eagerly spawn relay subagents for 400–800 char results?" The 800-char threshold will fire for many routine replies (a 10-sentence summary is easily 600–900 chars). Spawning a full subagent for a 850-char result adds latency and token cost without clear benefit — a `send_reply` call is nearly instantaneous.

The 7-second rule concern is about *LLM composition time*, not about reply length per se. A short reply_text that needs inline formatting/composition still takes time; a long reply_text passed to `send_reply` directly does not (send_reply is a fast tool call, not an LLM completion).

**Suggested fix:** The threshold should guard against LLM composition time, not byte length. If `reply_text` is already fully composed (which it is at this point — it's a string ready to send), `send_reply` inline is safe regardless of length. The relay subagent path adds value only when there is additional LLM work to do (e.g. summarising, formatting). Consider raising the threshold significantly (e.g. 4,000–8,000 chars) or removing the relay path entirely and relying on the 2,000-char server truncation + artifacts pattern to keep results short upstream.

---

## Layer 3 — Documentation (LGTM)

- Bold rule in `sys.subagent.bootup.md` is well-placed and clearly worded.
- Updated `write_result` tool description accurately describes enforcement.
- The note that "dispatcher reads artifacts and inlines them" is correct and helpful.

---

## Missing tests (pre-existing, acknowledged)

No new tests — acknowledged in PR. The truncation path is a one-liner and low risk; the dispatcher pseudocode change is harder to unit test. Acceptable for now given the incident severity.

---

## Summary of required changes before merge

1. **Fix the relay subagent prompt**: do not pass artifact-inlined `reply_text` verbatim — it can still be large. Pass the bounded text (pre-artifact-inlining) and handle artifacts separately.
2. **Re-evaluate the 800-char threshold**: if the goal is to avoid LLM composition time on the main thread, `send_reply(text=reply_text)` is already a tool call (not an LLM turn) and is fast. The threshold as written spawns unnecessary relay subagents for typical mid-length replies.

sayhar (commented):

Fixes 2 and 3 are correct and well-structured. Fix 1 introduces a logical dead zone that undermines the whole mechanism.

**Fix 2 — size check before artifact inlining** ✓
Moving the length check to `msg["text"]` before any artifact inlining is the right structural fix. The inline path now only runs for genuinely short results.

**Fix 3 — relay subagent prompt** ✓
Passing `msg["text"]` and artifact paths separately is clean. The relay subagent can inline artifacts on its own thread, which is exactly the intent.

**Fix 1 — threshold at 4,000 chars** ✗ (creates dead code)

`inbox_server.py` hard-truncates `write_result` text at 2,000 chars before storing. That means `len(msg["text"])` can never exceed 2,000. The `> 4000` condition is therefore unreachable — every result takes the `else` (inline) path, no matter how large the original was.

The inline `else` path still does artifact inlining: a result with a short `msg["text"]` but a large artifact file produces a long `reply_text` on the main thread. The offload branch exists precisely to prevent that, but it never fires.

To fix: either lower the threshold to something below the server cap (e.g. `len(msg["text"]) > 1500`), or raise the server-side truncation limit above 4,000. The threshold should be below the cap, not above it.

No other issues found.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The 1,200-char threshold is sound and the fix is correct end-to-end.

**On the threshold value:**
The previous 4,000-char threshold was dead code: inbox_server.py truncates write_result text at 2,000 chars before it reaches the dispatcher, so len(msg["text"]) could never exceed 2,000. Setting the relay threshold to 1,200 makes the branch reachable and leaves a 800-char buffer below the server cap — reasonable headroom. At 1,200 chars (~200-240 words), a result is large enough to justify async relay; shorter replies still take the fast inline path.

**On the comment clarity:**
The dispatcher doc comment — "Note: write_result text is server-truncated at 2,000 chars; this threshold must stay below that cap to be reachable" — directly explains the invariant. Anyone who later considers raising the threshold will see immediately why it cannot exceed 2,000. Clear and correct.

For results in the 1,200–2,000 char range, the relay spawns a full lobster-generalist subagent whose only job is to call send_reply once. That's slightly heavy for what is essentially a passthrough. A lightweight relay or a direct send_reply inline up to the 2,000-char server cap would also work. But given the stated goal (never block the main loop), the current approach is safe and consistent with the pattern already used for PR reviews. No change needed.

**Three-layer coherence:** server truncates at 2,000 → dispatcher offloads at 1,200 (reachable) → subagent docs say ≤2,000. The layers are consistent and the invariants are documented. Ready to merge.

PR #718, opened 2026-03-19
fix: make startup catchup internal-only (no send_reply to user)

sayhar:
The startup compact-catchup subagent was sending its recovery summary directly to the user via `send_reply` on every Lobster restart. This was inconsistent with the post-compaction catchup, which has always been internal-only. The startup catchup is a dispatcher orientation tool, not a user notification — the user doesn't need to see a raw activity summary every time the process restarts.

## What changed

- `chat_id` in the startup catchup prompt: `8305714125` → `0`
- `source` in the startup catchup prompt: `telegram` → `system`
- Removed the instruction telling the subagent to call `send_reply` and set `sent_reply_to_user=True`
- Updated step 3 description to state the startup catchup is now internal-only
- Updated the comparison table: both startup and post-compaction catchup are now identical on `chat_id` and Delivery rows
- Updated the "when result arrives" handler: now reads `msg["text"]` for situational awareness and updates `handoff.md` if anything notable changed, rather than just marking processed (which was the subagent_notification path)

## What is unchanged

The subagent itself (`compact-catchup`) and the compaction-state.json update logic are untouched. The post-compaction catchup flow is untouched.

## Before / after flow

```
Before:
  Dispatcher starts
       │
       ▼
  spawn compact-catchup (chat_id: 8305714125)
       │
       ▼
  subagent calls send_reply → user sees raw activity summary
  subagent calls write_result(sent_reply_to_user=True)
       │
       ▼
  subagent_notification arrives → mark_processed (no action)

After:
  Dispatcher starts
       │
       ▼
  spawn compact-catchup (chat_id: 0, source: system)
       │
       ▼
  subagent calls write_result (no send_reply)
       │
       ▼
  subagent_result arrives with chat_id: 0
  → dispatcher reads text for context
  → updates handoff.md if notable (failed agents, open threads)
  → mark_processed, silent
```

## Functional patterns

Pure data transformation: the prompt is a static string with no side-effect injection. The dispatcher result handler is a deterministic branch on `chat_id == 0` — no judgment call needed.

## Tests run

- [x] Manual diff review — 7 insertions, 9 deletions, all in `sys.dispatcher.bootup.md`
- [ ] Live dispatcher restart test — not run; this is a documentation/instruction change with no code execution path to test in isolation

🤖🦞 Lobster (engineer):


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with minor note

## Core mechanic: correct

- `chat_id: 0` / `source: system` is the established internal-only signal used by compact-catchup, agent-monitor, oom-check, and scheduled jobs. Consistent.
- Removing the `send_reply` instruction and the `sent_reply_to_user=True` note is correct. The subagent will now call only `write_result`, producing a `subagent_result` (not `subagent_notification`). The type correction in the handler comment is accurate.
- The dedicated handler block for `startup-catchup` mirrors the `compact-catchup` handler and takes priority over the generic `subagent_result` path — same pattern already proven for post-compaction.

## No regression risk from the generic handler

The generic `subagent_result` handler has no `chat_id: 0` guard and would call `send_reply(chat_id=0, ...)` if it ran. But this is pre-existing behavior shared with `compact-catchup` — not introduced here. The dedicated handler instructions for `task_id: "startup-catchup"` provide the override, same as for `compact-catchup`.

## Minor: table says "identical" but handlers have a new difference

The comparison table now shows startup and post-compaction as identical on `chat_id` and Delivery rows, which is correct. But the new startup handler adds `update handoff.md if anything notable changed` — a behavior the post-compaction handler does NOT have. The table's Purpose row partially reflects this ("after restart gap" vs "after compaction") but doesn't call out the handoff-update step. Not a bug, just a documentation gap — a reader could miss that startup catchup has a side effect the post-compaction one lacks. Consider a footnote or an explicit "Side effect" row if the table is ever revised.

## Verdict

Change is correct, safe to merge. The type change (`subagent_notification` → `subagent_result`), the `chat_id`/`source` change, and the removal of `send_reply` are all consistent and mutually correct. The handoff-update behavior is a net improvement.

PR #720, opened 2026-03-19
feat(hooks): add MAX_HOOK_FIRES=3 circuit breaker to require-auditor-context-update

### LINKED ISSUE #719: bug: stop-hook feedback loop can trap the auditor agent indefinitely
## What is the bug

\`require-auditor-context-update.py\` is a SubagentStop hook that blocks the auditor agent from exiting (exit 2) until one of two conditions is met:

1. \`system-audit.context.md\` was modified during the session (mtime check), OR
2. The transcript contains the safe word \`AUDIT_CONTEXT_UNCHANGED\` in a \`write_result\` call.

When neither condition is met, Claude Code re-injects the error message and re-runs the agent. If the agent still cannot satisfy the condition — due to turn exhaustion, a crash loop, or because the condition itself is never triggerable in its current form — the hook fires again, blocking exit again, forever. The agent never terminates.

This is a stop-hook feedback loop. The hook's exit-2 re-runs the agent, the agent fails the same condition, the hook fires again. The session can spin indefinitely consuming tokens with no way out.

## How the circuit breaker prevents it

PR #718 adds \`MAX_HOOK_FIRES = 3\` tracking (via \`/tmp/lobster-auditor-hook-fires-{agent_key}\`). After 3 fires without the exit condition being met, the hook:

1. Writes a \`system_error\` JSON entry to \`~/lobster-workspace/logs/observations.log\` with a \`CIRCUIT BREAKER TRIPPED\` prefix — loud enough to surface in monitoring.
2. Allows the stop (\`exit 0\`) unconditionally.

The threshold is 3, not 5, because exhaustion here is a more serious signal than in \`require-write-result.py\`: the auditor *ran* and *read* the context file, but never completed its post-condition. This points to something structurally wrong with the agent or its environment.

## Open question: WHY is the condition never met?

The circuit breaker prevents the runaway session, but it is a safety valve, not a root-cause fix.

The underlying question — **why does the auditor agent repeatedly fail to either update \`system-audit.context.md\` or emit the safe word?** — is being investigated separately. Possible explanations include:

- The mtime check uses the session start time, but the session start timestamp may be unavailable or wrong in CC 2.1.76+, making \`_context_file_updated_since()\` always return False even when the file was updated.
- The auditor agent updates the file but exits before \`write_result\` is called (so \`require-write-result.py\` blocks first and the auditor hook never gets a clean run).
- The auditor agent reads the context file but encounters an error before it can write, leaving the file unmodified.
- The safe word is emitted as text output instead of as a tool call input, so \`_safe_word_in_transcript()\` misses it.

Tracing which path trips first (this hook or \`require-write-result\`) and whether the context file mtime is actually advancing during auditor sessions would narrow the root cause.

## Related

- \`hooks/require-write-result.py\` — same feedback-loop risk, already has \`MAX_HOOK_FIRES = 5\`
- \`hooks/require-auditor-context-update.py\`

sayhar:
🤖🦞 Lobster (engineer):

## Problem

`require-auditor-context-update.py` blocks the auditor agent from exiting (exit 2) until one of two conditions is met: the context file was modified, or the safe word `AUDIT_CONTEXT_UNCHANGED` appears in a `write_result` call. If neither condition is ever satisfied — due to turn exhaustion, a crash loop, or a structural bug in the auditor — the hook fires again every time Claude Code re-runs the agent. The session never terminates. This is a stop-hook feedback loop identical to the one already solved in `require-write-result.py`.

## What this changes

Adds a `MAX_HOOK_FIRES = 3` circuit breaker, mirroring the pattern in `require-write-result.py` but with a lower threshold. After 3 fires without the exit condition being met:

1. The hook appends a `system_error` JSON entry to `~/lobster-workspace/logs/observations.log` with a `CIRCUIT BREAKER TRIPPED` prefix (loud, structured, monitorable).
2. The hook exits 0 — unconditionally allowing the stop rather than blocking forever.

The threshold is 3 (not the 5 used in `require-write-result.py`) because exhaustion here is a more serious signal: the auditor ran, read the context file, but still failed its post-condition. That's a structural problem, not a recoverable mistake.

On success (either condition met), the hook now also cleans up any fire-count temp file, so the counter resets correctly between sessions that reuse the same agent slot.

## Flow change

```
Before:
  hook fires → exit 2 → agent re-runs → hook fires → exit 2 → ... (unbounded)

After:
  hook fires (1) → exit 2 → agent re-runs
  hook fires (2) → exit 2 → agent re-runs
  hook fires (3) → log system_error → exit 0 → session terminates
```

No change to the two success paths (mtime check, safe word check) — those exit exactly as before, now also clearing fire-count state.

## What is out of scope

This PR does not fix *why* the auditor fails to satisfy the condition. That root-cause investigation is tracked in issue #719. The circuit breaker is a safety valve, not a diagnosis.

## Functional patterns

Pure functions throughout (fire-count read/write/cleanup are each single-responsibility, side-effect-isolated helpers). The main function remains a linear decision tree. No mutable shared state; all state is in temp files to survive process restarts.

## Tests run

- [x] Manual read of both files — logic trace confirmed: fire_count increments on each block, trips at >= MAX_HOOK_FIRES, logs error entry, exits 0
- [ ] Live hook invocation test — skipped: requires a running CC session with auditor agent; no automated test harness exists for hooks

Closes #719

### DIFF
--- a/hooks/require-auditor-context-update.py
+++ b/hooks/require-auditor-context-update.py
@@ -55,9 +55,11 @@ the circuit breaker trips.
 
 ## suppressOutput
 
-On all exit-0 paths the hook prints {"suppressOutput": true} so CC does not
-inject a "Stop hook feedback: No stderr output" system message that would
-trigger a spurious extra agent turn.
+This hook exits silently (no stdout/stderr) on all exit-0 paths.
+require-write-result.py (which runs first for the same SubagentStop event)
+already emits {"suppressOutput": true} and covers the event. Emitting a
+second suppressOutput here causes CC to surface it as feedback content
+rather than suppress it.
 """
 import json
 import os
@@ -66,15 +68,16 @@ import time
 from datetime import datetime, timezone
 from pathlib import Path
 
-# JSON to emit on every successful (allow) exit — suppresses the
-# "Stop hook feedback: No stderr output" injection that CC 2.1.76+ produces
-# even when the hook exits 0 with no output.
-_SILENT_OK = json.dumps({"suppressOutput": True})
-
-
 def _exit_ok() -> None:
-    """Exit 0 with JSON that suppresses CC feedback injection."""
-    print(_SILENT_OK)
+    """Exit 0 silently.
+
+    Intentionally produces no stdout or stderr output. For SubagentStop hooks,
+    suppressOutput in stdout is not reliably honoured by CC when multiple hooks
+    are registered for the same event — the first hook (require-write-result.py)
+    already emits suppressOutput and covers the event. Adding a second
+    suppressOutput from this hook causes CC to report it as feedback rather than
+    suppress it. Silence is the correct behaviour here.
+    """
     sys.exit(0)
 
 CONTEXT_FILE = Path(os.path.expanduser(
--- a/hooks/require-auditor-context-update.py
+++ b/hooks/require-auditor-context-update.py
@@ -52,6 +52,12 @@ Fire count is tracked in /tmp/lobster-auditor-hook-fires-{agent_key} as JSON:
 
 The file is cleaned up after a successful exit (either condition met) or after
 the circuit breaker trips.
+
+## suppressOutput
+
+On all exit-0 paths the hook prints {"suppressOutput": true} so CC does not
+inject a "Stop hook feedback: No stderr output" system message that would
+trigger a spurious extra agent turn.
 """
 import json
 import os
@@ -59,6 +65,17 @@ import sys
 import time
 from datetime import datetime, timezone
 from pathlib import Path
+
+# JSON to emit on every successful (allow) exit — suppresses the
+# "Stop hook feedback: No stderr output" injection that CC 2.1.76+ produces
+# even when the hook exits 0 with no output.
+_SILENT_OK = json.dumps({"suppressOutput": True})
+
+
+def _exit_ok() -> None:
+    """Exit 0 with JSON that suppresses CC feedback injection."""
+    print(_SILENT_OK)
+    sys.exit(0)
 
 CONTEXT_FILE = Path(os.path.expanduser(
     "~/lobster-user-config/agents/system-audit.context.md"
@@ -326,7 +343,7 @@ def main() -> None:
     try:
         hook_input = json.load(sys.stdin)
     except Exception:
-        sys.exit(0)  # Unreadable input — never block
+        _exit_ok()  # Unreadable input — never block
 
     # CC 2.1.76+: SubagentStop passes the transcript as a JSONL file at
     # agent_transcript_path rather than inline. Load from the file path,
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The circuit breaker implementation is correct and well-structured. The core logic is sound: fire count increments only on the failure path, threshold check uses `>= MAX_HOOK_FIRES` so it trips at exactly 3 fires (correct), cleanup happens on all exit paths (both success paths and the breaker trip), and the temp file approach survives process restarts correctly. The pattern mirrors `require-write-result.py` closely enough that maintenance should be straightforward.

## Findings

**No blocking issues.**

### Threshold and logic (correct)
- `fire_count >= MAX_HOOK_FIRES` trips at count == 3, which is the intended behaviour. The briefing called this out as a concern — it is fine.
- The increment happens _after_ both success checks, so a successful exit on fire 3 would clean up and exit 0 without tripping the breaker. This is correct: the breaker only trips when neither condition was ever met in 3 consecutive attempts.

### observations.log write is best-effort — tradeoff is acceptable
The `_write_circuit_breaker_error` function swallows all exceptions. The concern raised in the briefing is valid but the tradeoff is correct here: the hook's job is to control exit, not to guarantee logging. Blocking or crashing the hook on a log failure would be worse than a silent log miss. The breaker _does_ exit 0 regardless of whether the log write succeeds, which is the right priority ordering.

One minor note: if `~/lobster-workspace/logs/` does not exist and `mkdir` also fails (e.g. disk full), the exception is silently swallowed. This is already covered by the best-effort contract, but worth knowing.

### _agent_key fallback to "unknown"
If both `agent_id` and `session_id` are absent, all agents share a single temp file at `/tmp/lobster-auditor-hook-fires-unknown`. This is an unlikely edge case for a hook that only fires for auditor sessions, but it could cause cross-contamination if two auditor sessions ran concurrently with missing IDs. The sanitiser on the key is correct and necessary. This is not a blocker — just documenting the degenerate case.

### No `suppressOutput` on success paths
`require-write-result.py` emits `{"suppressOutput": True}` on exit 0 to prevent CC from injecting a "Stop hook feedback" system message into the next agent turn. This hook does not do that. Whether that matters depends on whether the auditor hook fires as a SubagentStop or Stop hook and whether CC's feedback injection would cause a spurious re-run. This is a pre-existing gap (not introduced by this PR), but worth tracking if spurious re-fires are observed after a successful audit exit.

### No automated test
Noted and acceptable. The PR documents it explicitly. Issue #719 is the right place for root-cause work. The circuit breaker itself is simple enough that a logic trace is sufficient for review confidence.

## Verdict

Logic is correct, cleanup is complete, the best-effort logging tradeoff is reasonable, and the PR description is accurate and complete. Ready to merge.

Issue #715, opened 2026-03-19
fix: camofox-browser install.sh should require/install Node.js 22+


sayhar:
## Problem

The current `install.sh` only checks for Node.js >= 18 and points users to install Node 20 via NodeSource. However, the camofox-browser server's dependencies use `import ... with` syntax (import assertions/attributes), which requires Node.js 21+. Running on Node 18 or 20 causes startup failures.

## Proposed Fix

Update `lobster-shop/camofox-browser/install.sh` to:

1. Check if Node.js >= 21 is installed (`node --version`)
2. If not installed or version is too old, automatically install Node 22 LTS via NodeSource (rather than erroring out and asking the user to do it manually)
3. Add a comment explaining why 21+ is required (import-with syntax in dependencies)

## Tasks

- [ ] Update version check threshold from 18 to 21
- [ ] Add auto-install of Node 22 LTS via NodeSource when Node is missing or too old
- [ ] Update the NodeSource URL from `setup_20.x` to `setup_22.x` in the error message/auto-install
- [ ] Add explanatory comment about the import-with syntax requirement

PR #737, opened 2026-03-20
fix: rebuild better-sqlite3 from source in camofox install.sh

sayhar:
## Problem

After PR #716 upgraded Node to v22 (from NodeSource), the camofox browser pre-warm fails silently at every startup:

```
{"level":"error","msg":"browser pre-warm failed (will retry on first request)","error":"libnode.so.109: cannot open shared object file: No such file or directory"}
```

The server stays up but browserConnected is permanently false, meaning the first actual browser request always pays a cold-start penalty (and may fail).

## Root cause

better-sqlite3 ships pre-built native binaries via prebuild-install. The pre-built binary for linux/arm64 was compiled against Ubuntu's system Node package (libnode109 = Node 18), which installs Node as a shared library (libnode.so.109). NodeSource's Node 22 package installs Node as a monolithic static binary - it does not provide libnode.so at all. So the pre-built addon's dlopen fails.

Verified with ldd before fix:
  libnode.so.109 => not found

After npm rebuild better-sqlite3 --build-from-source, the addon compiles against the current Node headers with static linkage - no libnode.so referenced.

## Fix

Add one line after npm install in install.sh:

  npm rebuild better-sqlite3 --build-from-source

This is idempotent and fast (~30s).

## Verification

- Applied the rebuild manually on the running install
- /health returns {"ok":true} with no libnode error in logs
- New startup log shows no error line (just the expected yt-dlp warning)

Companion to #716 / closes follow-up on #715.

---
🤖🦞 Lobster (ops)

### DIFF
--- a/lobster-shop/camofox-browser/install.sh
+++ b/lobster-shop/camofox-browser/install.sh
@@ -136,6 +136,15 @@ step "Installing Node.js dependencies"
 
 cd "$SERVER_DIR"
 npm install --production 2>&1 | tail -5
+
+# Rebuild native addons (better-sqlite3) from source to match the current Node
+# binary. Pre-built binaries for better-sqlite3 are compiled against the Ubuntu
+# system libnode (libnode.so.109, i.e. Node 18), which does not exist when Node
+# is installed from NodeSource as a static binary. Building from source compiles
+# against the installed node headers with static linkage, so no libnode.so is
+# required at runtime.
+info "Rebuilding native addons from source for Node $(node --version)..."
+npm rebuild better-sqlite3 --build-from-source 2>&1 | tail -5
 success "Node.js dependencies installed"
 
 info "Fetching Camoufox browser engine (this may take a minute on first run)..."


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — merging.

**Placement**: Correct. The `npm rebuild` is placed immediately after `npm install --production` and before `npx camoufox-js fetch`. Node is guaranteed installed at that point (the top-of-script prereq check installs it if absent), so node headers are available and the build will target the right binary.

**Build tools availability**: Not a concern for this skill's installer. The camofox install.sh is a standalone skill installer, not called from main install.sh. However, `build-essential` (gcc, make, python3) is installed by the main lobster install.sh for both apt and dnf systems — so on any properly bootstrapped Lobster host these are present. The comment in the diff correctly explains *why* `--build-from-source` is needed (NodeSource Node 22 = static binary, no libnode.so, prebuild-install's pre-built arm64 binary was linked against libnode.so.109 from the Ubuntu system Node 18 package). Root cause analysis is accurate and well-documented.

**upgrade.sh migration**: Not needed. The fix is in camofox's own install.sh. To apply it on an existing install, the user just reruns `bash ~/lobster/lobster-shop/camofox-browser/install.sh`, which is the standard skill re-install path. The rebuild step is idempotent (~30s), causes no harm if run repeatedly. No state file or sentinel is needed.

**Node version fragility**: Low risk. `--build-from-source` compiles against whatever Node headers are present at build time. If the Node version changes in the future, rerunning install.sh picks up the new headers automatically. The only non-idempotent risk would be if `node-gyp` itself has a version incompatibility, but that's inherent to any native addon and not introduced by this change.

**Overall**: Minimal diff, well-commented, correctly targeted, addresses a real regression from PR #716. No issues.

PR #738, opened 2026-03-20
fix: Docker test build — whisper examples target + add psutil test dep

sayhar:
## What this fixes

The Docker integration test build was failing at the whisper.cpp cmake step. The root cause: `-DWHISPER_BUILD_EXAMPLES=OFF` suppresses the `whisper-cli` target entirely, so `cmake --build --target whisper-cli` has nothing to build and exits with an error.

The fix builds examples but skips the HTTP server, which is the only binary we don't need in CI:

```
-DWHISPER_BUILD_EXAMPLES=ON \
-DWHISPER_BUILD_SERVER=OFF
```

This produces `whisper-cli` (required by the transcription tests) without pulling in the server binary.

## Second change: psutil in test requirements

`test_memory.py` does a top-level `import psutil`. The package was not listed in `tests/requirements-test.txt`, so the Docker test container failed on import with `ModuleNotFoundError: No module named 'psutil'`. Added `psutil>=5.9.0` under a "System utilities" comment.

## Scope

No behavior changes to production code. Both changes are test infrastructure only.

## Tests run

- [ ] `docker compose -f tests/docker/docker-compose.test.yml up --build` — not run locally; this PR exists precisely to unblock that build

**Blocked items needing attention before merge:** none — the changes are mechanical and verifiable from the diff.

Relates to #738 (test suite drift tracking issue)

### DIFF
--- a/tests/docker/Dockerfile.integration
+++ b/tests/docker/Dockerfile.integration
@@ -58,7 +58,8 @@ RUN git clone --depth 1 https://github.com/ggml-org/whisper.cpp.git \
           -S /home/testuser/lobster-workspace/whisper.cpp \
           -DBUILD_SHARED_LIBS=OFF \
           -DWHISPER_BUILD_TESTS=OFF \
-          -DWHISPER_BUILD_EXAMPLES=OFF && \
+          -DWHISPER_BUILD_EXAMPLES=ON \
+          -DWHISPER_BUILD_SERVER=OFF && \
     cmake --build /home/testuser/lobster-workspace/whisper.cpp/build \
           --config Release -j$(nproc) --target whisper-cli && \
     curl -L --retry 3 \
--- a/tests/requirements-test.txt
+++ b/tests/requirements-test.txt
@@ -17,6 +17,9 @@ docker>=7.0.0
 aioresponses>=0.7.6
 responses>=0.24.0
 
+# System utilities (required by test_memory.py)
+psutil>=5.9.0
+
 # General utilities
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): LGTM. Two focused fixes:

1. **Dockerfile.integration cmake flags**: The previous  meant  (an example binary) could never be built, making the subsequent  a dead target. Flipping to  +  correctly enables whisper-cli while skipping the heavier server binary. This is the minimal correct fix.

2. **psutil>=5.9.0**: Appropriate lower bound (5.9.0 is from 2022, stable and widely tested). No conflicts with existing test deps (pytest, docker SDK, faker, freezegun, etc.).

Change is minimal and focused. Approving for squash merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): LGTM. Two focused fixes:

1. **Dockerfile.integration cmake flags**: The previous `WHISPER_BUILD_EXAMPLES=OFF` meant whisper-cli (an example binary) could never be built, making the subsequent `--target whisper-cli` a dead target. Flipping to `EXAMPLES=ON` + `SERVER=OFF` correctly enables whisper-cli while skipping the heavier server binary. This is the minimal correct fix.

2. **psutil>=5.9.0**: Appropriate lower bound (5.9.0 is from 2022, stable and widely tested). No conflicts with existing test deps (pytest, docker SDK, faker, freezegun, etc.).

Change is minimal and focused. Approving for squash merge.

PR #743, opened 2026-03-20
fix: add __init__.py to tests/smoke to resolve collection error

sayhar:
🤖🦞 Lobster (engineer):

## Problem

Running the full test suite with `pytest tests/` failed at collection:

```
ERROR collecting test_require_write_result.py
import file mismatch:
  imported module 'test_require_write_result' has this __file__ attribute:
    tests/smoke/test_require_write_result.py
  which is not the same as the test file we want to collect:
    tests/test_require_write_result.py
```

Both `tests/smoke/` and `tests/` contain a file named `test_require_write_result.py`. Without an `__init__.py`, pytest treats them as the same Python module name — the first one collected gets cached, and the second one fails with an import mismatch.

## Fix

Add an empty `tests/smoke/__init__.py`. This namespaces the smoke package so pytest can distinguish `tests.smoke.test_require_write_result` from `tests.test_require_write_result`.

## Result

1865 tests collected (was 1835 + 1 error). No tests changed.


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — single empty __init__.py, resolves module name collision between tests/smoke/ and tests/ directories. No logic changes.

PR #744, opened 2026-03-20
fix: remove sys.modules pollution in test_vps_integration

sayhar:
🤖🦞 Lobster (engineer):

## What problem this solves

`test_vps_integration.py` was poisoning the test suite by injecting MagicMock objects into `sys.modules` at collection time. The file used a `sys.path.insert` hack to import `inbox_server` as a bare module, and to avoid import errors it pre-populated `sys.modules["reliability"]` (and `mcp`, `watchdog`, etc.) with MagicMocks before the import ran.

The consequence: by the time the conftest autouse fixture (`isolate_inbox_server_paths`) ran, `sys.modules["reliability"]` was already a MagicMock. The fixture imports `src.mcp.inbox_server` via the canonical path, which does `from reliability import validate_send_reply_args` — and got the mock. Every subsequent call to `_record_direct_send` in other test modules then failed with `TypeError: object supporting the buffer API required`, causing 193 tests to fail in the full suite.

## What changed

The module-level `sys.path.insert` and the `_MOCK_MODULES` block (60+ lines of manual sys.modules population) are replaced with a single clean import:

```python
from src.mcp.inbox_server import (
    load_sync_repos, parse_branch_info, parse_compare_info, format_sync_status,
)
```

All `patch("inbox_server.XXX")` targets are updated to `patch("src.mcp.inbox_server.XXX")` so they refer to the same module object the conftest fixture patches. Inline `from inbox_server import ...` calls inside autouse fixtures are updated to `from src.mcp.inbox_server import ...` for the same reason.

No logic changes — every test exercises the same functions with the same assertions.

The real deps (`mcp`, `watchdog`, `reliability`) are available in the test venv, so no mocking at import time is needed.

## Functional patterns

Pure functions (`load_sync_repos`, `parse_branch_info`, `parse_compare_info`, `format_sync_status`) are tested without any patching — they are invoked directly. The async handler tests use `patch` to swap out `run_gh_command` at the boundary, keeping the functions under test pure and deterministic.

## Tests run

- [x] `uv run pytest tests/test_vps_integration.py -v --tb=short` — 26 passed, 0 failed
- [x] `uv run pytest tests/integration/test_full_flow.py tests/test_vps_integration.py -v --tb=short` — 32 passed, 0 failed (cross-contamination gone)
- [x] `uv run pytest tests/ --ignore=tests/docker -q --tb=no 2>/dev/null | tail -3` — 28 failed (down from 193; remaining failures are pre-existing and unrelated to this change)

### DIFF
--- a/tests/test_vps_integration.py
+++ b/tests/test_vps_integration.py
@@ -12,12 +12,10 @@ Covers:
 """
 
 import asyncio
-import importlib.util
 import json
 import os
 import shutil
 import subprocess
-import sys
 import tempfile
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Verified the following:
- `sys.path.insert` removed
- `_MOCK_MODULES` block and all module-level `sys.modules` mutations removed
- Import updated to canonical `from src.mcp.inbox_server import ...`
- All `patch("inbox_server.XXX")` calls updated to `patch("src.mcp.inbox_server.XXX")`
- No new sys.modules pollution introduced

Clean fix. Approving for squash merge.

PR #751, opened 2026-03-20
fix: restore gws OAuth client secret from lobster-config on install

### LINKED ISSUE #750: install.sh: restore gws OAuth client secret from lobster-config on fresh install
## Problem

`install.sh` sets up the `gws` credential sync cron job but does **not** restore the OAuth client secret (`~/.config/gws/client_secret.json`). After a fresh install or reinstall, `gws auth login` fails immediately with:

```
No OAuth client configured.
```

The client secret is created once via the Google Cloud Console and must exist before `gws auth login` can run. There is no way to recover it automatically — the user must re-download or restore it manually every time they reinstall.

## Expected behaviour

The convention for sensitive credentials that can't go in the repo is to store them in `~/lobster-config/` and have `install.sh` copy them into place. The gws client secret should follow this convention: if `~/lobster-config/gws-client-secret.json` exists, install.sh should copy it to `~/.config/gws/client_secret.json` during the GWS Credential Sync setup step.

If the file is missing, a clear warning should be printed so the user knows what to do instead of hitting a cryptic error later.

## Fix

In `install.sh`, around line 1399 (the `step "Setting up gws credential sync..."` block), add a restore step before the cron setup.

A matching migration (Migration 29) should be added to `scripts/upgrade.sh` for existing installs that already ran `install.sh`.

## Impact

Without this, any reinstall silently breaks Gmail access until the user manually places the client secret. With this, reinstalls are fully self-healing for users who have stored the file in `~/lobster-config/`.

- [ ] Add client secret restore block to `install.sh` (GWS Credential Sync section)
- [ ] Add Migration 29 to `scripts/upgrade.sh`

sayhar:
🤖🦞 Lobster (engineer):

## What this does

\`install.sh\` now restores the gws OAuth client secret during install/reinstall.

**Background:** \`gws\` is the Google Workspace CLI used for Gmail access. Before it can authenticate, it needs a \`client_secret.json\` file from Google Cloud Console. Previously this file wasn't restored on reinstall, so \`gws auth login\` would fail after any fresh install.

**The fix:** If \`~/lobster-config/gws-client-secret.json\` exists, install.sh copies it to \`~/.config/gws/client_secret.json\`. If it doesn't exist (which is the current state — gws isn't configured yet), install warns and continues. No harm either way.

**Also:** \`scripts/upgrade.sh\` migration 29 does the same copy for existing installs, guarded so it doesn't overwrite an already-working file.

## One-time manual setup still required

To actually use gws, someone needs to:
1. Create a GCP project, enable Gmail API, create OAuth 2.0 credentials (Desktop app type)
2. Download \`client_secret.json\` → place at \`~/lobster-config/gws-client-secret.json\`
3. Run \`gws auth login\` and authorize the target account

After that, future reinstalls automatically restore auth capability.

Closes #750

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1398,6 +1398,19 @@ success "Nightly consolidation configured (runs at 03:00 nightly)"
 
 step "Setting up gws credential sync..."
 
+# Restore gws OAuth client secret from user config if present.
+# The client_secret.json is created once via Google Cloud Console and stored in
+# ~/lobster-config/ so reinstalls automatically restore gws auth capability.
+GWS_CLIENT_SECRET_SRC="$HOME/lobster-config/gws-client-secret.json"
+GWS_CONFIG_DIR="$HOME/.config/gws"
+if [ -f "$GWS_CLIENT_SECRET_SRC" ]; then
+    mkdir -p "$GWS_CONFIG_DIR"
+    cp "$GWS_CLIENT_SECRET_SRC" "$GWS_CONFIG_DIR/client_secret.json"
+    success "gws OAuth client secret restored from lobster-config"
+else
+    warn "No gws-client-secret.json found in ~/lobster-config/ — run 'gws auth setup' or copy client_secret.json manually"
+fi
+
 chmod +x "$INSTALL_DIR/scripts/sync-gws-credentials.py" || true
 
 # Add gws credential sync to crontab (runs daily at 04:00)
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1489,6 +1489,21 @@ with open(path, 'w') as f:
         migrated=$((migrated + 1))
     fi
 
+    # Migration 29: Restore gws OAuth client secret from lobster-config
+    # install.sh previously set up the gws credential sync cron job but did not
+    # restore the client_secret.json. Without it, gws auth login fails with
+    # "No OAuth client configured." on any reinstall.
+    # If the user has stored the secret at ~/lobster-config/gws-client-secret.json
+    # (the canonical location for private credentials), copy it into place now.
+    local GWS_SECRET_SRC="$HOME/lobster-config/gws-client-secret.json"
+    local GWS_SECRET_DEST="$HOME/.config/gws/client_secret.json"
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Summary:** Both changes are correct and well-structured. The install.sh block and upgrade.sh migration work together cleanly to handle the gws OAuth client secret restore gap.

**1. install.sh bash correctness**
The block is valid bash. The `[ -f "$GWS_CLIENT_SECRET_SRC" ]` guard ensures it fails gracefully if the source doesn't exist — it falls through to the `else` branch and emits a `warn` message rather than erroring. The `mkdir -p` before `cp` correctly handles the case where `~/.config/gws/` doesn't exist yet. No issues.

**2. Migration numbering**
Current highest is Migration 28 (daily log-export cron entry). Migration 29 is the correct next number.

**3. Upgrade migration guard (`[ ! -f dest ]`)**
The "don't overwrite if dest exists" behavior is intentional and reasonable. The working credential is already in place, so overwriting it would risk breaking an authenticated install. The "corrupt dest" edge case is not worth handling here: if `client_secret.json` is corrupt, `gws auth login` will fail with an explicit error and the user can manually replace it. Silently overwriting on every upgrade would be the worse default. The guard is correct.

**4. install.sh always-copies behavior (no dest-exists guard)**
This is the right call for reinstalls. install.sh is a full reinstall path where you want a clean, known-good state. Copying unconditionally from the canonical source (`~/lobster-config/gws-client-secret.json`) ensures the credential is always restored from the authoritative location. If the user had a working credential in `~/.config/gws/` that differs from the source, the source is the ground truth on reinstall. Behavior is intentional and correct.

**5. Security**
No concerns. The file is being copied within the same user's home directory (`~/lobster-config/` → `~/.config/gws/`), both paths are user-owned. The `cp` doesn't change permissions — `client_secret.json` will inherit the umask. If you want to be extra cautious, `chmod 600 "$GWS_CONFIG_DIR/client_secret.json"` after the copy would be a minor hardening improvement, but it's not a blocker given same-user scope.

**Minor optional suggestion:** `chmod 600` on the copied credential in install.sh (not required for PASS).

PR #759, opened 2026-03-20
Strengthen bootup wording: catchup must always be a background subagent

### LINKED ISSUE #758: Dispatcher ran startup catchup inline instead of as background subagent (3 occurrences)
## Bug Description

The startup catchup (and potentially post-compaction catchup) was run inline on the dispatcher main thread instead of being spawned as a background subagent. This occurred 3 times in recent sessions.

## Root Cause

Prompt compliance failure. The bootup instructions describe spawning a \`compact-catchup\` subagent with \`run_in_background=True\`, but the wording was not strong enough to prevent the dispatcher from running the catchup inline — reading files and calling \`check_inbox\` directly on the main thread.

This violates the 7-second rule: catchup involves file I/O, inbox scanning, and summarization — all of which block the main loop for tens of seconds or more, causing new user messages to pile up unanswered.

## Impact

- New messages queue up unanswered during the catchup window
- Health check may restart the dispatcher mid-catchup
- System appears unresponsive to the user

## Fix

Strengthen the wording in \`sys.dispatcher.bootup.md\` for both the startup catchup section and the compact-reminder handler section:
- Add an explicit WARNING callout (matching the style of the 7-second rule) that catchup MUST be a background subagent
- Make it unambiguous that the dispatcher must NOT read files or call check_inbox inline for catchup
- Add the violation pattern to the "DO NOT DO THIS" examples

## Resolution

See PR for the prompt-level fix. No code changes needed — this is entirely a bootup instruction clarity issue.

sayhar:
## What went wrong

The dispatcher ran startup catchup inline 3 times instead of delegating to a background subagent. In each case, it called `check_inbox` and read `compaction-state.json` directly on the main thread — blocking all incoming messages for 30–60+ seconds. This is a 7-second rule violation: new user messages pile up unanswered, and the health check may restart the dispatcher mid-catchup.

The root cause was prompt compliance, not code. The bootup instructions described spawning a `compact-catchup` subagent but did not clearly prohibit the inline alternative, leaving room for a compliant-looking violation.

## What changed

`sys.dispatcher.bootup.md` — two sections strengthened:

**compact-reminder handler** (post-compaction): A WARNING blockquote (matching the 7-second rule style) is added before the procedure steps. It names the prohibited tools (`check_inbox`, `Read`), states the duration cost, and shows the violation pattern with `# VIOLATION` annotations — the same format used in the 7-second rule section.

**Startup Behavior step 3**: An inline WARNING is added immediately after the subagent spawn instruction. It makes explicit that `run_in_background=True` is mandatory, states the blocking cost, and clarifies the expected flow (spawn → wait_for_messages → result arrives later).

No code changes. This is a documentation-only fix to a prompt compliance gap.

## Functional patterns

Pure documentation edit — no code logic involved. The change follows the existing callout style established by the 7-second rule warning block.

## Tests run

- [x] Manual diff review — 14 insertions, 0 deletions, no prose removed or reworded beyond additions
- [ ] Live session test — blocked: requires restarting the dispatcher with the new bootup file (safe to merge; the symlink at `~/lobster-workspace/.claude` means this takes effect on next dispatcher boot)

Closes #758


sayhar (commented):
🤖🦞 Lobster (reviewer): This is a documentation-only change with no code touched. The two WARNING callouts are correctly placed — immediately before the procedure they guard in the compact-reminder section, and inline with the specific step in the Startup Behavior list. Both use the established style (blockquote, bold WARNING header, specific prohibited tools named, violation pattern shown). The duration estimate (30–60+ seconds) is accurate for a catchup that reads a JSON file and scans up to 50 inbox messages. No existing prose was removed or modified. The commit message and PR description are accurate. Safe to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with notes

**Overall:** The fix is well-targeted and follows the established callout style. Both callouts gate the right procedures, are clearly worded, and should reduce recurrence. Some minor concerns noted below.

---

**Placement — compact-reminder handler (line ~176)**

The WARNING block is placed *before* the procedure steps, which is exactly right. A dispatcher reading the section for the first time hits the prohibition before encountering the tool calls it would be tempted to make. No issue here.

**Placement — Startup Behavior step 3 (line ~676)**

The inline WARNING is appended as a blockquote on the same line as the instruction it annotates. This is functional, but the visual hierarchy is slightly awkward — the warning appears subordinate to the step rather than co-equal. If the dispatcher skims step 3 as "spawn the agent, noted," it may not register the nested blockquote. A stronger placement would be a standalone blockquote *after* step 3 rather than a trailing annotation on the same bullet. That said, this is a style preference and not a blocking concern.

**Accuracy of stated blocking duration**

The PR body says "10–12 minutes" (from the issue briefing), but the callouts say "30–60+ seconds." These are inconsistent. The issue description says each stall caused a health-check restart and lasted long enough to matter — so the 30–60 second floor may be accurate for simple cases, while the 10–12 minute figure may reflect worst-case summarization. The callout should be internally consistent with the reported incident data. Recommend reconciling: either "several minutes" or "up to 10+ minutes depending on inbox depth." As written, "30–60+ seconds" technically includes those longer durations but undersells the severity.

**Violation examples**

The two-line `# WRONG` block (`check_inbox` + `Read`) is concrete and scannable — good. It mirrors the 7-second rule style effectively. One small gap: the compact-reminder section shows what NOT to do but doesn't show the correct pattern (spawn-and-return) in code form. The startup section does describe the correct flow in prose. Parity between the two sections would help, but this is not blocking.

**Sufficiency of prompt-only fix**

Legitimate concern. Three stalls in two days suggests the dispatcher chose the inline path despite existing instructions. A documentation-only fix raises the barrier but does not eliminate the failure mode. Structural safeguards worth tracking (not blocking this PR, but worth an issue):
- A health-check or watchdog that detects `wait_for_messages` absence for >30s during startup
- A subagent that validates the dispatcher's first N tool calls after boot to catch inline catchup early
- A compaction-state read interceptor that logs a warning when called on the main thread

None of these are required to merge — the callouts are a legitimate and incremental improvement. But the issue should note that this is a partial mitigation, not a structural fix.

**Summary:** Callouts are accurate, well-placed, and follow convention. Minor issues: duration figure should match incident data, startup callout could be a standalone block rather than an inline annotation. Prompt-only fix is acceptable as a first mitigation; recommend a follow-up issue for structural enforcement.

sayhar (commented):
🤖🦞 Lobster (reviewer): Duration figure corrected to 10–15 min. PASS — no remaining issues.

PR #760, opened 2026-03-20
fix: suppress WFM health-check during catchup to prevent false-positive restarts

### LINKED ISSUE #757: fix: dispatcher stalls 10-12 min during catchup, triggers health-check restart
🤖🦞 Lobster (engineer):

## Bug

The Lobster dispatcher stalls for 10–12 minutes during post-compaction or startup catchup generation. This causes the health-check WFM (wait_for_messages) freshness check to fire RED and trigger an unnecessary restart. This has happened 3 times in 2 days:

- Mar 19 03:28 UTC
- Mar 20 06:24 UTC
- Mar 20 23:12 UTC

## Root Cause

The health check (`health-check-v3.sh`) has a `WFM_STALE_SECONDS=600` (10-minute) threshold. If `wait_for_messages` hasn't been called in 600s, it restarts the dispatcher.

The `COMPACTION_SUPPRESS_SECONDS=300` (5-minute) window suppresses the WFM check after a compaction event — but the catchup subagent can take 10–12 minutes. The 5-minute suppression expires before the catchup finishes, and the health check fires.

There are two possible sub-causes:
1. The dispatcher is NOT following the `run_in_background=True` instruction and waits inline for the catchup result, blocking `wait_for_messages` entirely.
2. The dispatcher IS following the instruction, but the catchup runs long enough that the 5-minute `COMPACTION_SUPPRESS_SECONDS` window expires before catchup finishes, exposing the WFM check.

Both cases converge on the same fix: the health check needs to know that a catchup is actively in progress and suppress the WFM freshness check for longer.

## Fix Plan

**Option A (code-level, preferred):** Add a `catchup_started_at` field in `lobster-state.json` written by the dispatcher (via a small helper script) before spawning the catchup subagent. The health check reads this field and suppresses the WFM freshness check for `CATCHUP_SUPPRESS_SECONDS` (e.g., 900s). The dispatcher also writes `catchup_finished_at` when the subagent result arrives, so suppression ends cleanly.

**Option C (prompt-level, secondary):** Strengthen the bootup file warning against inline waiting during catchup to reduce the chance of regression.

## Acceptance Criteria

- [ ] Implement `catchup_started_at` / `catchup_finished_at` write helper script
- [ ] Health check reads `catchup_started_at` and suppresses WFM check during active catchup window
- [ ] Dispatcher bootup instructions updated to call write helper before spawning catchup subagent
- [ ] Tests cover the new suppression path in health-check
- [ ] Upgrade script entry if needed

sayhar:

The Lobster dispatcher was being restarted unnecessarily — 3 times in 2 days (Mar 19 03:28, Mar 20 06:24, Mar 20 23:12 UTC) — because the health check's WFM (wait_for_messages) freshness check fired RED while a catchup subagent was running.

After a compaction or restart, the dispatcher spawns a `compact-catchup` or `startup-catchup` subagent to recover situational awareness. This subagent can take 10–12 minutes. The health check's `WFM_STALE_SECONDS=600` threshold fires when `wait_for_messages` hasn't been called in 10 minutes. The existing `COMPACTION_SUPPRESS_SECONDS=300` only suppressed for 5 minutes — not long enough to cover the full catchup window.

## What changed

**New: `record-catchup-state.sh`**  
A shell helper that writes `catchup_started_at` or `catchup_finished_at` to `lobster-state.json`. The dispatcher calls `record-catchup-state.sh start` before spawning the catchup subagent, and `record-catchup-state.sh finish` when the result arrives. Both are atomic (written via `os.replace()`).

**Updated: `health-check-v3.sh`**  
- New `CATCHUP_SUPPRESS_SECONDS=900` (15 min) configuration constant
- New `is_catchup_active()` function: reads `catchup_started_at` and `catchup_finished_at` from `lobster-state.json`; returns true (suppress) if started < 15 min ago AND `catchup_finished_at` is absent or older than `catchup_started_at`
- The WFM freshness check now also suppresses when `is_catchup_active()` returns true, alongside the existing hibernation / compaction / boot-grace suppressions
- Hard 15-min cap prevents permanent suppression if the subagent crashes and never writes `catchup_finished_at`

**Updated: `sys.dispatcher.bootup.md`**  
- Both startup and post-compaction catchup handlers now call `record-catchup-state.sh start` before spawning the subagent and `record-catchup-state.sh finish` when the result arrives
- Added explicit CRITICAL warning against waiting inline for catchup results (to reduce regression risk)

## Flow: before vs. after

**Before** (broken):
```
Compaction → on-compact.py writes compacted_at (suppresses health check for 5 min)
           → dispatcher spawns catchup subagent (background, 10-12 min)
           → calls wait_for_messages()
           → ... 5 min pass ... compaction suppression expires ...
           → ... 5 more min ... WFM threshold (600s) fires → RESTART (false positive)
```

**After** (fixed):
```
Compaction → on-compact.py writes compacted_at (suppresses inbox checks for 5 min)
           → dispatcher calls record-catchup-state.sh start (sets catchup_started_at)
           → dispatcher spawns catchup subagent (background, 10-12 min)
           → calls wait_for_messages()
           → health check sees is_catchup_active() → suppresses WFM check (up to 15 min)
           → catchup result arrives → dispatcher calls record-catchup-state.sh finish
           → health check sees catchup_finished_at >= catchup_started_at → lifts suppression
```

## What is NOT changed

- The `COMPACTION_SUPPRESS_SECONDS=300` (inbox drain suppression) is unchanged — this is a separate check for stale user messages, not the WFM heartbeat
- `WFM_STALE_SECONDS=600` is unchanged — the threshold itself is correct; only the suppression window during catchup was missing
- Subagent crash recovery: if `catchup_finished_at` is never written, suppression expires after 15 min (hard cap), not indefinitely

## Functional patterns used

Pure state reads (`is_catchup_active` reads lobster-state.json without side effects), atomic writes via `os.replace()`, hard timeout cap to prevent indefinite suppression.

## Tests run

- [x] `bash /tmp/test-is-catchup-active.sh` — 4 test cases (no state file, started only, started+finished, expired) all pass
- [x] `bash -n scripts/health-check-v3.sh` — shell syntax check passes
- [x] `bash -n scripts/record-catchup-state.sh` — shell syntax check passes
- [ ] Full health-check integration test — not run; requires live Lobster environment with systemd, tmux, and inbox messages

Closes #757


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Overall: looks correct and safe to merge. One question worth answering before merge, one minor nit.**

---

**Correctness of suppression logic**

The `is_catchup_active()` function correctly handles all four cases:
- No state file → not active (correct)
- `catchup_started_at` set, `catchup_finished_at` absent → active (correct)
- Both set, finished >= started → not active (correct)
- Started > 15 min ago with no finished → not active, hard cap applied (correct)

The function is pure (no side effects, reads state file only) and fails safely — if the state file is malformed or missing, it returns 1 (no suppression), which means the health check continues to protect the system.

---

**Question: does the dispatcher actually call wait_for_messages during catchup?**

The description says "the dispatcher IS calling wait_for_messages()" during catchup, but the issue title says the stall is because it is NOT calling wait_for_messages. If the real bug is that the dispatcher waits inline (blocking the loop entirely), then the WFM heartbeat file would go stale because `wait_for_messages` is never called — and suppressing the WFM check would fix the symptom (health-check restart) without fixing the underlying behavioral regression (dispatcher stalls user messages for 10-12 min).

The PR description is correct that this fix is needed regardless: even if the dispatcher is following the run_in_background=True instruction, the 5-min compaction suppression is insufficient for a 12-min catchup. But it would be worth confirming via logs whether the dispatcher was actually blocking inline or correctly backgrounding. If it was blocking inline, the prompt-level CRITICAL warning added to bootup.md is the mechanism that fixes the root cause.

This doesn't block merge — the code change is correct either way — but worth noting for monitoring after deploy.

---

**Nit: record-catchup-state.sh uses python3 not uv**

The script comment says "Uses python3 (not uv) because this script runs outside the Claude/uv context" — this is accurate and intentional (the script is called as a Bash tool call during dispatcher startup, before uv is necessarily in scope). No change needed, but worth confirming on the target host that `python3` is available in PATH when called from Claude Code's tool executor. Based on the rest of the codebase (health-check-v3.sh uses `python3` directly throughout), this is fine.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: This is a well-structured fix with correct logic and good defensive design. Notes below.

---

## 1. Correctness of suppression logic — PASS

`is_catchup_active()` is logically correct:
- Suppresses when `catchup_started_at` is present AND within the 900s window AND `catchup_finished_at` is absent OR older than `catchup_started_at`
- Lifts immediately when `catchup_finished_at >= catchup_started_at`
- Hard-caps at 900s even if `catchup_finished_at` is never written

One subtle edge: if `date -d "$finished_at" +%s` fails (malformed timestamp), `finished_epoch` stays empty via `|| true`, so the code falls through and *continues suppressing*. That's the right fail-safe here (prefer false-negative over false-positive restart), but it's worth a comment.

The two separate `python3` calls to read `started_at` and `finished_at` are fine for a health-check script, but note they each re-read the file. A tiny TOCTOU window exists where a `finish` write occurs between the two reads: the script would read `started_at`, then read a `finished_at` that is newer. In practice this only causes a one-cycle early lift — benign and acceptable.

## 2. Shell script syntax and atomic write — PASS

`bash -n` passes per the PR. The atomic write via `os.replace()` is correct. The heredoc `<<PYEOF` inside the shell script is clean.

One note: `STATE_FILE` in `record-catchup-state.sh` is constructed from `$LOBSTER_MESSAGES_DIR` / `config/lobster-state.json`, while `health-check-v3.sh` uses `$LOBSTER_STATE_FILE`. Worth confirming these resolve to the same path in the production environment — if they ever diverge the two sides talk past each other. Could add a comment or a `LOBSTER_STATE_FILE_OVERRIDE` doc note.

## 3. Bootup placement — PASS

`record-catchup-state.sh start` is placed before `spawn` in both the compact-catchup and startup-catchup handlers. That is the correct ordering — the health check is armed before the subagent is in flight, closing the race where the subagent starts but the state file hasn't been written yet.

`record-catchup-state.sh finish` in the compact-catchup handler is placed correctly (step 4, before `mark_processed`). The startup-catchup `finish` call is inserted inline in the result-arrival prose — correct but slightly less prominent than the numbered list for the compact path. Not a bug.

The new CRITICAL warning against inline waiting is well-placed and accurate.

## 4. 900s suppression window — APPROPRIATE

900s (15 min) vs 600s WFM threshold is the right margin: gives 5 min of headroom above the observed 10-12 min catchup duration. If catchup ever exceeds 15 min, the health check will correctly fire — that would be a genuine stall, not a false positive.

## 5. Risk of suppression masking real stalls — LOW, ACCEPTABLE

The suppression is conditioned on an explicit `catchup_started_at` write, which only happens in two well-defined code paths. It is not a blanket WFM suppression. The hard 900s cap ensures a crashed or hung catchup subagent doesn't suppress indefinitely. The only scenario where a real stall could be masked is if the dispatcher hangs *after* writing `catchup_started_at` but *before* the catchup finishes — but in that case the 900s cap still fires, so the window is at most 15 min of missed detection, not permanent blindness. This is an acceptable trade-off given the 3 false-positive restarts it prevents.

---

## Open question: inline vs. background stall

The engineer's open question (blocking inline vs. compaction suppression expiring before 12-min catchup) is answered by the timing data in the PR description:

- `COMPACTION_SUPPRESS_SECONDS=300` (5 min) expired while the catchup was still running (10-12 min)
- After 5 min the compaction suppression lifts, then 10 min later the WFM threshold (600s from the last `wait_for_messages` call) fires

This points to the **suppression-window-too-short** scenario, not an inline-blocking bug. The dispatcher was correctly calling `wait_for_messages()` but the health check wasn't giving it enough cover. The fix correctly addresses this root cause.

---

## Minor nits (non-blocking)

- The `|| true` on `finished_epoch` parse failure silently continues suppressing; a `log_info "Catchup: could not parse finished_at=$finished_at, continuing suppression"` would aid debugging.
- `record-catchup-state.sh` uses bare `python3` (correct per the comment — outside uv context), but a stray `uv` environment wouldn't break anything here since it's a bare Python stdlib call.
- Consider adding `catchup_started_at` and `catchup_finished_at` to whatever schema docs or lobster-state.json documentation exists, so future maintainers know these fields are health-check-critical.

Overall this is a clean, targeted fix. No blocking issues.

sayhar (commented):
🤖🦞 Lobster (reviewer): Nits resolved — log line added, STATE_FILE paths verified. PASS — no remaining issues.

Issue #781, opened 2026-03-23
fix: ghost-session agent_failed messages stall dispatcher; SessionStart hook creates ghost sessions on restart


sayhar:
## Summary

Two bugs cause a cascade that triggers repeated WFM stale restarts:

1. `agent_failed` messages with `chat_id=0` (ghost sessions) cause the dispatcher to deliberate for up to 12 minutes — consuming the WFM freshness budget and triggering a restart.
2. The `SessionStart` hook (`write-dispatcher-session-id.py`) registers the dispatcher itself as a **subagent** on crash-restart, creating a ghost session that the reconciler marks dead 30 minutes later (feeding Bug 1).

---

## Bug 1: `agent_failed` ghost messages stall the dispatcher

**Root cause:** When the reconciler marks a session dead (including sessions with `chat_id=0` — i.e., sessions that were auto-registered by the `SessionStart` hook with no real task context), it enqueues an `agent_failed` message to the inbox. The dispatcher receives this message and deliberates on it — inspecting `original_chat_id`, `task_id`, `original_prompt`, deciding whether to re-queue or escalate — even when all of those fields indicate there is nothing to do.

The decision heuristic in `sys.dispatcher.bootup.md` documents the drop path (`original_chat_id == 0` → drop silently), but the dispatcher is an LLM and can still spend 10–12 minutes reaching that conclusion. During that time, WFM is blocked, heartbeat staleness accumulates, and the health check restarts the system.

**Fix:** In `_build_reconciler_message()` in `inbox_server.py`, populate the `original_chat_id` field on `agent_failed` messages when the session's `chat_id` is `"0"` or `""`. This is already done — but additionally, add a new field `needs_deliberation: false` to `agent_failed` messages whose `original_chat_id` is `0`/`""`. The dispatcher bootup doc is updated to short-circuit on this field: `mark_processed` immediately, no LLM deliberation.

**Alternative approach (stronger):** Extend `_build_reconciler_message()` to include a `should_drop` boolean derived purely from the session data. The dispatcher prompt is updated with a single rule: if `msg["should_drop"] == True`, call `mark_processing` + `mark_processed` and skip all deliberation. This is a pure function decision made in Python, not by the LLM — consistent with the principle that deterministic checks belong in code.

**Proposed design:**

```python
# In _build_reconciler_message() for the 'dead' branch:
original_chat_id = session.get("chat_id", "0")
should_drop = not original_chat_id or original_chat_id in ("0", "", "None")

return {
[...]

PR #783, opened 2026-03-23
docs(dispatcher): add fast-exit rule for agent_failed ghost-session messages

sayhar:
## Summary

Updates the dispatcher bootup doc to reflect that ghost sessions are now suppressed at the reconciler level — no `agent_failed` message is emitted for them — so the dispatcher's fast-exit rule is simpler: every `agent_failed` it receives comes from a session that had a real user.

## What changed

**`.claude/sys.dispatcher.bootup.md`** — The "Fast-exit: agent_failed for ghost sessions" section previously stated that messages with `should_drop: True` should be dropped without deliberation. That field has been removed (see PR #782). The section now:
- States upfront that ghost sessions are suppressed at the reconciler, so the dispatcher only ever sees `agent_failed` messages from real-user sessions.
- Describes the dispatcher's responsibility as deciding re-queue/escalate/drop rather than checking a flag.

No other sections changed. The "Handling Agent Failures" section and the full `agent_failed` response pseudocode below it are untouched.

## Tests run
- [ ] Doc-only change, no automated tests. Verified doc text matches the behavior in PR #782.

Relates to #781

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -408,6 +408,13 @@ Check the `sent_reply_to_user` field first, then check for engineer → reviewer
 
 The reconciler and agent-monitor route dead/failed agent events to `chat_id=0` with `type: "agent_failed"`. These are **system-internal** — never relay them to the user's Telegram directly. The dispatcher reads the context and decides the right action.
 
+## Fast-exit: agent_failed for ghost sessions
+
+When a message has `type: "agent_failed"` (or contains `should_drop: True`) AND `chat_id == 0`:
+- `mark_processed` immediately — no deliberation, no subagent spawn
+- These are reconciler notifications about dead ghost sessions. There is no user to reply to and no action to take.
+- Handling time must be <1 second. If you find yourself thinking about it, stop — just drop it.
+
 **When `wait_for_messages` returns a message with `type: "agent_failed"`:**
 
 ```
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -410,12 +410,15 @@ The reconciler and agent-monitor route dead/failed agent events to `chat_id=0` w
 
 ## Fast-exit: agent_failed for ghost sessions
 
-Ghost sessions (sessions whose original `chat_id` is 0, "", or None) are suppressed by the reconciler before they reach the inbox — no `agent_failed` message is emitted for them. You will therefore only receive `agent_failed` messages for sessions that had a real user associated with them.
+Ghost session suppression works in two layers:
+
+**Layer 1 — reconciler filter:** Sessions registered with `agent_type='dispatcher'` are skipped by the reconciler entirely. These never produce any inbox message — not even an `agent_failed`. This covers the dispatcher's own internal sessions.
+
+**Layer 2 — dispatcher fast-exit:** Other internal/background sessions (e.g. cron subagents, scheduled job workers, system monitors) may fail with no real user attached. The reconciler does emit `agent_failed` for these — they arrive in the inbox with `chat_id=0`. When the dispatcher receives an `agent_failed` with `chat_id == 0`, drop it immediately: there is no user to notify and no action to take. This fast-exit prevents the dispatcher from stalling on internal bookkeeping events.
 
 When a message has `type: "agent_failed"` AND `chat_id == 0`:
 - `mark_processed` immediately — no deliberation, no subagent spawn
-- These are reconciler notifications about dead real-user sessions where the dispatcher is responsible for deciding the follow-up action (re-queue, escalate, or drop).
-- Handling time must be <1 second unless you decide to re-queue or escalate. If you find yourself deliberating with no clear action, just drop it.
+- Handling time must be <1 second. There is no user to notify. If you find yourself deliberating, just drop it.
 
 **When `wait_for_messages` returns a message with `type: "agent_failed"`:**
 
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -410,10 +410,12 @@ The reconciler and agent-monitor route dead/failed agent events to `chat_id=0` w
 
 ## Fast-exit: agent_failed for ghost sessions
 
-When a message has `type: "agent_failed"` (or contains `should_drop: True`) AND `chat_id == 0`:
+Ghost sessions (sessions whose original `chat_id` is 0, "", or None) are suppressed by the reconciler before they reach the inbox — no `agent_failed` message is emitted for them. You will therefore only receive `agent_failed` messages for sessions that had a real user associated with them.
+
+When a message has `type: "agent_failed"` AND `chat_id == 0`:
 - `mark_processed` immediately — no deliberation, no subagent spawn
-- These are reconciler notifications about dead ghost sessions. There is no user to reply to and no action to take.
-- Handling time must be <1 second. If you find yourself thinking about it, stop — just drop it.
+- These are reconciler notifications about dead real-user sessions where the dispatcher is responsible for deciding the follow-up action (re-queue, escalate, or drop).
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — doc-only change, adds missing fast-exit rule that complements the code fix in #782. No functional change.

sayhar (commented):
🤖🦞 Lobster (reviewer): Reviewed against PR #782's stated behavior.

**Accuracy check:** PR #782 implements ghost-session suppression at the reconciler level — no inbox file is written for dead sessions with `chat_id` 0/""/None. The `should_drop` field is removed entirely. This PR's doc reflects that correctly: "Ghost sessions … are suppressed by the reconciler before they reach the inbox — no `agent_failed` message is emitted for them."

**Internal consistency issue:** The Fast-exit section then says "When a message has `type: "agent_failed"` AND `chat_id == 0`" and instructs the dispatcher to `mark_processed` immediately. But the preceding sentence just explained that ghost sessions produce no messages at all — so _why_ would a real-user session message arrive with `chat_id == 0`? The section conflates two different concepts:

1. Ghost sessions (suppressed at source — no message emitted) — correctly described in the first paragraph.
2. `agent_failed` messages for real sessions that happen to have `chat_id == 0` — described in the bullet block, but a real-user session shouldn't have `chat_id == 0` by definition.

The logic is circular: if ghost sessions are suppressed, the `chat_id == 0` fast-exit rule is either (a) dead code / unreachable, or (b) a defense against legacy messages that predate PR #782. The doc doesn't clarify which, and the bullet text ("reconciler notifications about dead real-user sessions") contradicts itself — real-user sessions don't have `chat_id == 0`.

**Suggested clarification:** Either note explicitly that this is a defense-in-depth guard for pre-#782 inbox files, or remove the `chat_id == 0` case since it's now unreachable by construction.

No code changes in this PR — the risk is doc confusion only, not runtime breakage.

Verdict: NEEDS-WORK (minor — doc is internally contradictory on the chat_id==0 case)

sayhar (commented):
🤖🦞 Lobster (reviewer): [updated after doc fix] Verdict: PASS

The fast-exit section now correctly explains both suppression layers:

- **Layer 1 (reconciler):** Sessions with `agent_type='dispatcher'` are skipped entirely — these never produce an `agent_failed` message.
- **Layer 2 (dispatcher fast-exit):** Other internal sessions (cron subagents, system monitors) can fail with `chat_id=0`. The reconciler does emit `agent_failed` for these. The fast-exit rule drops them immediately — correct behavior, valid rule, no contradiction.

The original doc was self-contradictory because it claimed ghost sessions never reach the inbox, then gave a rule for when they do. The updated text removes that contradiction by naming the two layers distinctly and explaining why both are necessary.

No behavior change — doc only.

sayhar (commented):
🤖🦞 Lobster (reviewer): [updated to reflect final 3-layer design] Verdict: PASS

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The documentation accurately reflects the new behavior introduced by PR #782. The three-layer model is clear, technically correct, and removes the now-obsolete `should_drop: True` handling.

## Findings

**Accuracy** — The doc correctly describes all three suppression layers as implemented in PR #782:
- Layer 1 (reconciler skips `agent_type='dispatcher'` sessions entirely) matches the PR #782 description
- Layer 2 (dead sessions with no user suppressed at `_enqueue_reconciler_notification()` via the `chat_id == 0/""/None` check) matches the PR #782 implementation
- Layer 3 (dispatcher defense-in-depth fast-exit for any `agent_failed` with `chat_id == 0`) is unambiguous: mark_processed immediately, <1 second, no deliberation

**Clarity** — The fast-exit rule is explicit and unambiguous. The "<1 second / if you find yourself deliberating, just drop it" instruction is strong behavioral guidance for the LLM dispatcher. No interpretation is required.

**Correctness of removal** — The old `should_drop: True` flag check is correctly absent. The new rule is keyed on `chat_id == 0` which is the right predicate. No stale references remain.

**Minor observation** — The new section is inserted before the `agent_failed` pseudocode block, which is good placement. The layered explanation may be more detail than needed for the dispatcher's day-to-day operation, but the extra context helps maintainers and does no harm.

## Merge readiness

Ready for Sahar's sign-off. This PR should merge together with or after PR #782 — the docs reference behavior that only exists once #782 is deployed. Merging #783 before #782 would make the "ghost sessions are suppressed at the reconciler" claim false for any running instance.

sayhar (commented):
Lobster (reviewer): PR #783 review

**Verdict: NEEDS-WORK**

## Summary of findings

The added documentation is factually accurate in describing what PR #782 implemented, but there is a mismatch between what the PR description claims the section says and what the diff actually adds. The PR description states the doc now reflects that "the dispatcher only ever sees `agent_failed` messages from real-user sessions" — but the actual diff still includes a Layer 3 dispatcher fast-exit rule for `chat_id == 0`. This is not a minor discrepancy: it affects whether the dispatcher follows the old cognitive model (check and drop) or the new one (every message is real, decide re-queue/escalate/drop).

## Specific issues

**1. PR description vs. actual diff (blocking)**

The PR description says:
> The section now... states upfront that ghost sessions are suppressed at the reconciler, so the dispatcher only ever sees `agent_failed` messages from real-user sessions.

But the diff adds a "Layer 3 (dispatcher) — defense-in-depth fast-exit" block that still instructs the dispatcher:
> When a message has `type: "agent_failed"` AND `chat_id == 0`: mark_processed immediately

This contradicts the stated intent. If the dispatcher "only ever sees real-user sessions," the fast-exit rule for `chat_id == 0` either (a) should not exist, or (b) should be framed as a paranoia guard with a clear note that this path should never trigger post-#782/#809.

**2. Post-#809 compatibility**

Layer 3's defense-in-depth rationale mentions "inbox files from before the reconciler fix was deployed" as a surviving trigger. That's a valid transitional case. However, now that #809 has also been merged (removing `_auto_register_subagent()` stub creation entirely), the doc should acknowledge that the pool of ghost sessions is now draining to zero — new stubs are no longer being created. This strengthens the argument that Layer 3 is purely historical, which is fine to document, but it should be stated clearly.

**3. Layer 2 nuance is correct**

The claim that "Completed sessions with `chat_id=0` are NOT suppressed" is accurate: PR #782's early-return guard is conditioned on `outcome == "dead"`, so completed/succeeded sessions with `chat_id=0` still write to inbox. This is correctly documented.

## Recommended fix

Either:

**Option A (simpler, matches PR description intent):** Remove Layer 3 entirely and add a note that `chat_id == 0` messages are fully suppressed upstream. The dispatcher no longer needs a fast-exit rule.

**Option B (keep defense-in-depth, fix the framing):** Keep Layer 3 but revise the PR description and the section intro. Remove the claim that "the dispatcher only ever sees real-user sessions" and instead state "ghost sessions are suppressed upstream in the common case; Layer 3 is a never-should-trigger paranoia guard for legacy inbox files pre-deploy." Also add a note that post-#809, no new ghost stubs are created.

The doc is not wrong about what the code does, but the internal narrative is self-contradictory. A dispatcher reading this section will get mixed signals about whether `chat_id == 0` is possible or not.

sayhar (commented):
🤖🦞 Lobster (reviewer): LGTM. The three-layer architecture description is accurate and maps correctly to the implementation in PR #782.

**Layer 1** (dispatcher sessions skipped entirely) and **Layer 2** (dead sessions with no real user suppressed at source) correctly describe the two independent guards in `inbox_server.py`. **Layer 3** (dispatcher fast-exit as defense-in-depth) is a good reframe — it drops the now-removed `should_drop` field approach and replaces it with a simpler invariant: "if `chat_id == 0` reaches the inbox, drop it, no deliberation needed."

The distinction between `outcome == 'dead'` (suppressed) vs completed sessions with `chat_id=0` (not suppressed) is correctly called out — that's a subtle but important boundary. The PR description is accurate about scope: only the fast-exit section changed, and the full `agent_failed` pseudocode below it is untouched.

No concerns. Ready to merge.

PR #785, opened 2026-03-23
fix: resolve false Claude post-install failure on fresh Ubuntu

### LINKED ISSUE #784: install.sh: PATH not updated before claude post-install check causes false failure
## Bug

When `~/.local/bin` is not in the system's default PATH (common on fresh Ubuntu installs), the Claude Code binary installs successfully to `~/.local/bin/claude` but the post-install check (`command -v claude`) fails because the `export PATH=...` line runs too late or doesn't propagate correctly.

## Symptom
```
✅ Installation complete!   ← from Anthropic's installer (success)
[ERROR] Claude Code installation failed  ← from Lobster's post-install check (false failure)
```

## Fix
Move the `export PATH="$HOME/.local/bin:$PATH"` line in `install.sh` to *before* the `command -v claude` post-install check, or source it in a way that affects the current shell context.

## Workaround
```bash
export PATH="$HOME/.local/bin:$PATH"
bash ~/lobster/install.sh  # re-run; detects claude is already installed, continues
```

sayhar:
## Summary

On a fresh Ubuntu system where `~/.local/bin` is not in the default PATH, the Anthropic installer writes the `claude` binary successfully, but the post-install check immediately after (`command -v claude`) can return not-found despite the `export PATH` line running first. This causes install.sh to exit with `[ERROR] Claude Code installation failed` even though the binary is present and working.

Two mechanisms combine to produce the bug:

1. **Bash command hash table.** Bash caches negative `command -v` lookups. After `export PATH` updates the variable, `command -v` may still return the cached "not found" result from before the install ran, because the hash table is not automatically invalidated when PATH changes.

2. **No direct fallback.** Even with a correct PATH, if the hash table returns stale data there is no secondary check against the known install location.

## What changed

Two additions to the Claude Code install block (`install.sh` lines 810–826):

- `hash -r 2>/dev/null || true` — flushes bash's command hash table immediately after the PATH export, so `command -v` re-resolves from the updated PATH. The `|| true` makes this a no-op on non-bash shells that don't support `hash`.
- `|| [ -x "$HOME/.local/bin/claude" ]` appended to the post-install check — direct executable test against the known install path as a belt-and-suspenders fallback if `command -v` still misbehaves.

No surrounding code was touched. The PATH export line and its position are unchanged.

## What is out of scope

This does not change how the pre-install check at line 526 sets `CLAUDE_INSTALLED`, which uses the same `command -v` pattern. That check runs before the installer, so a stale hash there is not a problem (the binary doesn't exist yet). If the user re-runs install.sh on a system where the binary was installed in a prior run but PATH is still stale, they'd hit the same false-skip at that earlier check — that is a separate issue.

## Functional patterns

Pure sequential logic with no new state introduced. The `hash -r` call is a side-effectful shell builtin scoped to this shell session, isolated at the boundary where it is needed.

## Tests run

- [x] Static code review of the diff — logic is correct: `hash -r` fires before `command -v`, fallback `[ -x ... ]` covers the residual case.
- [ ] Live Docker test (fresh Ubuntu, no `~/.local/bin` in PATH) — skipped: destructive network install not safe to run in this environment, and the fix is a two-line ordering/fallback change where code review is sufficient.

Closes #784

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -807,8 +807,10 @@ if [ "$CLAUDE_INSTALLED" = false ]; then
 
     curl -fsSL https://claude.ai/install.sh | bash
 
-    # Add to PATH for current session
+    # Add to PATH for current session and clear bash's command hash table so
+    # command -v picks up the newly installed binary immediately.
     export PATH="$HOME/.local/bin:$PATH"
+    hash -r 2>/dev/null || true
 
     # Persist ~/.local/bin to PATH in shell config files
     PATH_LINE="export PATH=\"\$HOME/.local/bin:\$PATH\""
@@ -821,7 +823,7 @@ if [ "$CLAUDE_INSTALLED" = false ]; then
         fi
     done
 
-    if command -v claude &>/dev/null; then
+    if command -v claude &>/dev/null || [ -x "$HOME/.local/bin/claude" ]; then
         success "Claude Code installed"
     else
         error "Claude Code installation failed"


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**What the diff does:** Adds `hash -r 2>/dev/null || true` after the PATH export and widens the post-install check to `command -v claude || [ -x "$HOME/.local/bin/claude" ]`.

**Is the fix mechanically correct?** Yes.
- `hash -r` is the standard bash mechanism for flushing the command lookup cache after a PATH change. The `|| true` guard makes it safe on non-bash POSIX shells.
- `[ -x "$HOME/.local/bin/claude" ]` is a direct, cache-independent existence check against the known install path. It is correct to use `-x` (executable) rather than `-e` (exists).
- Both additions are inside the `if [ "$CLAUDE_INSTALLED" = false ]` block, so they do not affect re-runs where Claude was already detected.

**Does it touch anything it shouldn't?** No. The surrounding PATH-persist loop and the pre-install check at line 526 are untouched.

**Known gap noted in PR description:** The pre-install `CLAUDE_INSTALLED` check at line 526 has the same `command -v` stale-hash exposure on re-run after a failed install, but that is a different scenario and intentionally out of scope here.

**Verdict:** Safe to merge.

PR #806, opened 2026-03-24
feat: LOBSTER_ENV mode gate + scoped staging credentials

sayhar:
## What this solves

Two problems caused a Telegram bot conflict: the production persistent session had no way to stand down during SSH dev work, and the staging Docker container inherited production tokens because it read loose env vars sourced from the shared `config.env`.

**Problem 1 — no lifecycle gate:** When doing interactive SSH work, `claude-persistent.sh` kept running and `health-check-v3.sh` kept monitoring and restarting. There was no single switch to make the production session inert without stopping services manually.

**Problem 2 — no credential boundary:** `config.env` is a single file containing all credentials. The staging compose file read `${TELEGRAM_BOT_TOKEN}` from the host environment, which was already sourced from `config.env`. Staging silently got the production token with no enforcement boundary.

### LOBSTER_ENV gate

A single `LOBSTER_ENV=production|dev|test` variable (default: `production`) gates both lifecycle components:

- `claude-persistent.sh`: exits cleanly (exit 0) when `LOBSTER_ENV != production`. systemd's `RemainAfterExit=yes` holds the service as "active" so no restart loop fires.
- `health-check-v3.sh`: reads `LOBSTER_ENV` from `config.env` (same pattern as `LOBSTER_DEBUG` already uses) and exits 0 immediately when not `production`. The cron entry keeps firing, so flipping back to production takes effect within 4 minutes with no manual step. Resource checks (disk/memory) also skip — the service is intentionally idle and alerting on it would be noise.

To pause the session for dev work:
```bash
# In ~/lobster-config/config.env:
LOBSTER_ENV=dev
sudo systemctl restart lobster-claude  # script exits cleanly
```

To resume:
```bash
LOBSTER_ENV=production
sudo systemctl restart lobster-claude
```

### Credential boundary

`docker-compose.staging.yml` now uses `env_file:` pointing to `~/lobster-config/config.staging.env` instead of reading loose host env vars that inherited from `config.env`. The staging file is structurally isolated from the production file — there is no code path that causes staging to read `config.env`.

A new repo template `config/config.staging.env.example` documents which credentials belong in staging scope, with an explicit prohibition on the production bot token.

The flow before vs. after:

```
BEFORE:
  host shell → source config.env → TELEGRAM_BOT_TOKEN=<PRODUCTION>
  docker-compose → environment: TELEGRAM_BOT_TOKEN: "${TELEGRAM_BOT_TOKEN}"
  container → gets production token

AFTER:
  docker-compose → env_file: config.staging.env (separate file)
  container → gets only staging-scoped credentials
  config.env → never read by staging container
```

### Migration

`upgrade.sh` Migration 32 appends `LOBSTER_ENV=production` to existing `config.env` files that don't already have it. No behavior change — both scripts already default to `production` when the variable is absent. The migration just makes the knob visible and editable.

`install.sh` writes `LOBSTER_ENV=production` into new config.env files during setup.

## What is NOT changed

- No changes to dispatcher logic, message routing, or any MCP servers
- Slack, SMS, and other router services are unaffected
- The health check's locking, logging, and escalation ladder are unchanged — only the early-exit guard is new
- The `lobster-claude.service.template` comment addition is documentation only; the variable is already loaded via `EnvironmentFile={{CONFIG_DIR}}/config.env`

## Tests run

- [x] `bash -n scripts/claude-persistent.sh` — syntax OK
- [x] `bash -n install.sh` — syntax OK
- [ ] Live `LOBSTER_ENV=dev` end-to-end (service starts, script exits, health check skips, service resumes on flip back) — blocked: requires production restart; safe to merge as the default is `production` and no existing behavior changes unless the variable is explicitly set

### DIFF
--- a/config/config.staging.env.example
+++ b/config/config.staging.env.example
@@ -1,0 +1,30 @@
+# Lobster Staging Credentials
+#
+# Copy to ~/lobster-config/config.staging.env and fill in values.
+# Used by docker/staging/docker-compose.staging.yml.
+#
+# RULE: Never put production credentials in this file.
+#   - Production TELEGRAM_BOT_TOKEN stays in config.env only.
+#   - Use a dedicated test bot created via @BotFather for staging.
+#   - All other tokens (Linear, Vercel, etc.) should use staging/test equivalents
+#     if they exist, or be omitted entirely if not needed for staging work.
+#
+# LOBSTER_ENV must be set to a non-production value here so that lifecycle
+# automation (health check, auto-restart) does not activate inside the container.
+
+# --- Required for staging Telegram bot ---
+TELEGRAM_BOT_TOKEN=<your-test-bot-token>       # NEVER use production token here
+TELEGRAM_ALLOWED_USERS=<your-chat-id>
+
+# --- Environment mode ---
+LOBSTER_ENV=staging
+
+# --- Logging ---
+LOBSTER_DEBUG=true
+
+# --- GitHub (optional: only if staging work requires gh CLI access) ---
+# GITHUB_TOKEN=<personal-access-token-for-staging>
+
+# --- MCP HTTP auth (regenerate; do not reuse production secret) ---
+# LOBSTER_INTERNAL_SECRET=<new-random-secret>
+# MCP_HTTP_TOKEN=<new-random-token>
--- a/docker/staging/docker-compose.staging.yml
+++ b/docker/staging/docker-compose.staging.yml
@@ -4,6 +4,11 @@
 # dedicated test bot token. Isolated from production via a separate
 # messages volume and workspace.
 #
+# Credential setup (one-time):
+#   cp ~/lobster/config/config.staging.env.example ~/lobster-config/config.staging.env
+#   # Fill in TELEGRAM_BOT_TOKEN (test bot), TELEGRAM_ALLOWED_USERS, etc.
+#   # NEVER put the production TELEGRAM_BOT_TOKEN in config.staging.env.
+#
 # Usage:
 #   cd ~/lobster
 #   sudo docker compose -f docker/staging/docker-compose.staging.yml up -d
@@ -24,11 +29,17 @@ services:
       dockerfile: docker/staging/Dockerfile.staging
     container_name: lobster-staging
     restart: unless-stopped
+    # Scoped credentials: reads ONLY from config.staging.env, never from config.env.
+    # This is the credential boundary that prevents staging from inheriting production
+    # tokens. config.staging.env must NOT contain the production TELEGRAM_BOT_TOKEN.
+    # See config/config.staging.env.example for the template.
+    env_file:
+      - ${LOBSTER_CONFIG_DIR:-/home/lobster/lobster-config}/config.staging.env
     environment:
-      # Test bot token - safe to use in staging
-      TELEGRAM_BOT_TOKEN: "${TELEGRAM_BOT_TOKEN}"
-      TELEGRAM_ALLOWED_USERS: "${TELEGRAM_ALLOWED_USERS}"
-      LOBSTER_DEBUG: "false"
+      # GitHub PAT for gh CLI access inside staging container.
+      # Sourced from the host environment so it doesn't need to live in
+      # config.staging.env (which is committed-adjacent and more exposed).
+      GITHUB_TOKEN: "${GITHUB_TOKEN:-}"
     volumes:
       # Mount host Claude credentials (read-only auth, but settings.json needs write)
       # We mount the whole .claude dir so credentials are available
--- a/docker/staging/docker-compose.staging.yml
+++ b/docker/staging/docker-compose.staging.yml
@@ -34,6 +34,9 @@ services:
     # tokens. config.staging.env must NOT contain the production TELEGRAM_BOT_TOKEN.
     # See config/config.staging.env.example for the template.
     env_file:
+      # LOBSTER_CONFIG_DIR defaults to /home/lobster/lobster-config — the standard
+      # single-user install path. Override this env var before running compose if
+      # your config lives elsewhere (e.g. export LOBSTER_CONFIG_DIR=/custom/path).
       - ${LOBSTER_CONFIG_DIR:-/home/lobster/lobster-config}/config.staging.env
     environment:
       # GitHub PAT for gh CLI access inside staging container.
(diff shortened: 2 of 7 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Overall

Solid implementation. The two problems (no lifecycle gate, no credential boundary) are addressed cleanly with minimal blast radius. The default of `production` preserves existing behavior for all current installs. I have a few observations but nothing that blocks merge.

---

## claude-persistent.sh

**Gate placement: correct.** The early-exit fires after the log header block but before `preflight`. This means `preflight` (which does the heavy lifting: auth checks, waiting for network, etc.) never runs in non-production mode — exactly what you want.

**Edge case — log line written to stdout, not \$LOG_FILE.** The gate uses `log "..."` which writes to a log file that hasn't been created yet (log dir is created inside `setup_log_rotation`, called from `main` after this point). On a fresh install or after log rotation, the log dir may not exist. The log call will silently fail (bash redirect to a nonexistent dir returns an error but doesn't abort because errexit isn't set at that point). This is acknowledged in the engineer's briefing as acceptable. Worth noting: the exit 0 still fires correctly, so the behavior is correct even if the log line is lost.

---

## health-check-v3.sh

**Gate placement: correct.** LOBSTER_ENV is read and the gate fires before the log dir is created (`mkdir -p` comes after). Same silent-fail risk on the `echo >> "$LOG_FILE"` line inside the gate — but again, exit 0 still fires correctly.

**LOBSTER_ENV parsing is consistent with LOBSTER_DEBUG pattern.** The grep/cut/tr pipeline mirrors the existing pattern in the same file. Good.

---

## docker-compose.staging.yml

**Credential boundary is structurally sound.** `env_file` pointing to a scoped file is the right mechanism — no code path exists that reads `config.env` inside the container.

**GITHUB_TOKEN handling:** Using `${GITHUB_TOKEN:-}` (empty default) correctly prevents compose from erroring on unset vars. Sourcing it from host env rather than `config.staging.env` is a reasonable tradeoff given the exposure concern noted in the comment.

**Fallback path:** `${LOBSTER_CONFIG_DIR:-/home/lobster/lobster-config}` — this hardcoded fallback is fine for the typical single-user install but will silently use the wrong path on non-standard installs where `LOBSTER_CONFIG_DIR` is not exported to the environment before running compose. Acceptable for now, but worth documenting in the README/usage comment if staging gets wider use.

---

## upgrade.sh Migration 32

**Idempotent and safe.** The `! grep -q '^LOBSTER_ENV='` guard prevents double-writes. Appending with heredoc is clean. Migration number 32 appears to be in sequence.

---

## install.sh

Two insertion points (one for the "already configured" path, one for the fresh-config path) — both get the new block. This looks correct.

---

## services/lobster-claude.service.template

Comment-only change, accurate, no functional impact.

---

## Summary

- No behavioral regression risk: default is `production`, existing installs unaffected until variable is set.
- Credential boundary works structurally: staging container cannot inherit `config.env` values.
- Silent log-line failure on early-exit in fresh installs is documented and benign.
- Live end-to-end test is the only gap; safe to defer given the safe default.

Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — nit fixes confirmed clean

Reviewed commit 356fc6b against the two requested nit fixes:

**health-check-v3.sh** — `mkdir -p "$(dirname "$LOG_FILE")"` correctly added as the first statement inside the `LOBSTER_ENV != production` early-exit block. The `echo >> $LOG_FILE` that follows can no longer silently fail on a fresh install where the log directory hasn't been created yet. The outer `mkdir -p` (a few lines earlier, before the gate) still runs on the happy path; this inner one handles only the non-production early-exit path. Correct and tidy.

**docker-compose.staging.yml** — Comment placed directly above the `env_file` path explains the `/home/lobster/lobster-config` default and the `LOBSTER_CONFIG_DIR` override pattern. Exactly the right location; a future operator reading the compose file will see the explanation at the point of use.

Both fixes are additive, non-breaking, and address the original nits without introducing new issues. The rest of the PR (LOBSTER_ENV gate logic, credential boundary, Migration 32, service template comment, install.sh backfill) was already sound.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — both problems are cleanly solved with minimal surface area.

## What it does

Adds a `LOBSTER_ENV=production|dev|test` gate to `claude-persistent.sh` and `health-check-v3.sh` so both scripts exit cleanly when not in production mode. Separately, scopes the staging Docker container to read credentials from `config.staging.env` instead of inheriting the host environment that already has `config.env` sourced.

## Findings

**LOBSTER_ENV gate — correct pattern:**
- `claude-persistent.sh`: exits 0 early, before `preflight`. systemd's `RemainAfterExit=yes` holds the service as "active" without restart loops. Clean.
- `health-check-v3.sh`: reads `LOBSTER_ENV` from `config.env` using the same `grep | cut | tr` pattern already used for `LOBSTER_DEBUG`. Then exits 0 if not production. The cron entry keeps firing, so restoring production mode takes effect within 4 minutes with no manual intervention. Good design.
- Both default to `production` when the variable is absent — no behavior change for existing installs that don't run migration 32 immediately.

**Double `mkdir -p` in health-check:** The gate block has `mkdir -p "."` before the log write, and there's already a `mkdir -p` at the top of the file. This is harmless but slightly redundant. Not a blocking issue.

**Credential boundary — correct:**
- Replacing `environment: TELEGRAM_BOT_TOKEN: "${TELEGRAM_BOT_TOKEN}"` with `env_file: config.staging.env` severs the path from `config.env` to the staging container. The host environment is no longer the source.
- `GITHUB_TOKEN` remains sourced from the host environment (not `config.staging.env`). This is intentional per the comment: "Sourced from the host environment so it doesn't need to live in config.staging.env." Acceptable — `GITHUB_TOKEN` is a PAT, not a bot token, and not the credential the PR is trying to isolate.

**Migration 32:** Appends `LOBSTER_ENV=production` only when the variable is absent. Idempotent, non-destructive. Correct.

**`install.sh`:** Two separate heredoc blocks both get the `LOBSTER_ENV=production` addition. Both paths (interactive and non-interactive) are covered.

**`config.staging.env.example`:** Has `LOBSTER_ENV=staging` and a clear prohibition on the production bot token. Good documentation.

## Verdict

PASS. The two problems (lifecycle gate, credential boundary) are solved cleanly. The double `mkdir -p` is a minor nit. Sahar's earlier nit (mkdir before log write in the gate block) was already addressed in 356fc6b.

PR #815, opened 2026-03-24
Add Stop hook to nudge dispatcher to call wait_for_messages

### LINKED ISSUE #814: Add Stop hook to nudge dispatcher to call wait_for_messages
## Background

The dispatcher runs in an infinite loop: process messages → call `wait_for_messages` → repeat. When it stalls after processing a batch of results, it ends its turn without calling `wait_for_messages`. The health check catches this after ~12 min and restarts.

The existing `hooks/require-write-result.py` Stop hook does exactly this pattern for subagents — it checks if `write_result` was called and exits with code 2 + a message if not, which injects a reminder into the next Claude turn.

## Task

Create a new Stop hook for the dispatcher at `hooks/require-wait-for-messages.py` that:

1. **Detects if this is the dispatcher session** — via `session_role.is_dispatcher()`. If NOT the dispatcher, exit 0 (no-op).
2. **Reads the transcript** (from `transcript_path` or inline) to check if `mcp__lobster-inbox__wait_for_messages` was called in the most recent turn.
3. **If WFM was NOT called** — exit with code 2 and output a reminder message.
4. **If WFM WAS called** — exit 0 (all good).

Register the hook in `~/.claude/settings.json` under the `stop` hooks array.

## Implementation plan
- [ ] Create `hooks/require-wait-for-messages.py`
- [ ] Register in `~/.claude/settings.json` under Stop hooks

sayhar:
Closes #814

## Why this exists

The dispatcher's main loop is: process messages → call `wait_for_messages` → repeat. After processing a batch of subagent results, it can end a turn without calling `wait_for_messages`. The health check catches this stall after ~12 minutes and restarts — a long window with missed messages and unnecessary restarts.

The existing `require-write-result.py` SubagentStop hook solves the analogous problem for subagents. This PR applies the same pattern to the dispatcher's main loop — with a warn-only policy rather than blocking.

## What changed

A new `hooks/require-wait-for-messages.py` Stop hook fires on every Stop event. For the dispatcher session, it scans the transcript: if `mcp__lobster-inbox__wait_for_messages` was not called, it **prints a warning to stderr and exits 0**. The stop is never blocked — this is purely advisory.

Three correctness fixes were made after initial review:

1. **Warn-only (not blocking):** The hook exits 0 in all cases. Previously it exited 2 (blocking). A blocked stop on the dispatcher is disruptive; a warning gives the same signal with no downside.

2. **Transcript read failure fix:** `_load_transcript_from_jsonl` now returns `None` on I/O error (vs `[]` for a successfully-read empty file). Previously a read failure would return `[]`, which the hook treated identically to a transcript where `wait_for_messages` was not called — triggering a false-positive block. Now the hook detects the failure, prints a distinct warning, and exits 0.

3. **`LOBSTER_MAIN_SESSION=1` guard:** Added an early `_exit_ok()` for sessions that don't carry this env var — matching the pattern used by `write-dispatcher-session-id.py`. Sessions started outside Lobster (e.g. a developer's personal Claude Code) exit immediately without checking the dispatcher marker file.

**Subagent sessions are exempted** via `session_role.is_dispatcher()` — the hook reads the dispatcher marker file and falls back to transcript scanning. If the session is not the dispatcher, the hook exits 0 silently before reading the transcript.

The hook is pure and stateless: no retry counter, no fallback emit, no DB calls.

## Functional patterns

Pure functions for each concern (load transcript, collect tool names, classify session, decide outcome) composed in a linear `main()`. All I/O is isolated at the boundaries: stdin for hook input, stderr for the warning message, stdout for the suppressOutput JSON.

## Registration

- `install.sh`: adds the Stop hook entry during fresh installs
- `scripts/upgrade.sh` Migration 33: idempotently registers the hook on existing installs
- `~/.claude/settings.json` on this host: updated in-place (verified via `jq`)

## Before / after

```
Before (original PR, blocking):
  Dispatcher stalls (no WFM call)
        |
  Stop hook fires (same turn, exit 2)
        |
  Stop blocked — reminder injected into next turn
        |
  Dispatcher calls WFM and resumes

After (this update, warn-only):
  Dispatcher stalls (no WFM call)
        |
  Stop hook fires (same turn, exit 0)
        |
  Warning printed to stderr
        |
  Stop proceeds normally
```

The health check is unchanged and remains the backstop for genuinely stuck sessions.

## Tests run

- [x] `echo '{"hook_event_name": "Stop", "session_id": "not-a-dispatcher"}' | LOBSTER_MAIN_SESSION=1 python3 hooks/require-wait-for-messages.py` — exits 0, outputs `{"suppressOutput": true}`
- [x] `echo '{"hook_event_name": "Stop", "session_id": "any"}' | python3 hooks/require-wait-for-messages.py` — exits 0 immediately (no LOBSTER_MAIN_SESSION guard)
- [x] Dispatcher session with empty transcript: piped live dispatcher session_id with `"transcript": []` and `LOBSTER_MAIN_SESSION=1` — exits 0, stderr contains the reminder message (warn, not block)
- [x] Dispatcher session, transcript_path pointing to nonexistent file: exits 0, stderr contains `WARNING: could not read transcript` (no false-positive block)
- [x] Dispatcher session, transcript containing `wait_for_messages` tool_use block: exits 0, outputs `{"suppressOutput": true}` (no warning)

### DIFF
--- a/hooks/require-wait-for-messages.py
+++ b/hooks/require-wait-for-messages.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-Stop hook: ensure the dispatcher calls wait_for_messages before ending a turn.
+Stop hook: warn the dispatcher if it ends a turn without calling wait_for_messages.
 
 The dispatcher's main loop is: process messages → call wait_for_messages →
 repeat. When the dispatcher stalls — typically after processing a batch of
@@ -10,8 +10,8 @@ window with missed messages.
 
 This hook fires on every Stop event (dispatcher or subagent). Subagent sessions
 are immediately exempted via session_role.is_dispatcher(). For the dispatcher,
-the hook scans the transcript: if wait_for_messages was not called, it exits
-with code 2 and injects a reminder into the next Claude turn.
+the hook scans the transcript: if wait_for_messages was not called, it prints
+a warning to stderr and exits 0 (warn-only — the stop is never blocked).
 
 ## Transcript handling
 
@@ -34,10 +34,12 @@ the hook outputs JSON with {"suppressOutput": true} on all success paths.
 ## Exemptions
 
 The hook does NOT fire for:
+- Sessions without LOBSTER_MAIN_SESSION=1 (non-Lobster Claude Code sessions)
 - Subagent sessions (is_dispatcher() returns False)
 - Any session where wait_for_messages was called at least once in the transcript
 """
 import json
+import os
 import sys
 from pathlib import Path
 
@@ -64,8 +66,12 @@ def _exit_ok() -> None:
     sys.exit(0)
 
 
-def _load_transcript_from_jsonl(path: str) -> list:
-    """Load transcript entries from a JSONL file. Returns [] on any error."""
+def _load_transcript_from_jsonl(path: str) -> list | None:
+    """Load transcript entries from a JSONL file.
+
+    Returns a list of entries on success (may be empty if the file is empty),
+    or None on any I/O or OS error (distinguishes read failure from empty file).
+    """
     try:
         messages = []
         with open(path) as fh:
@@ -78,7 +84,7 @@ def _load_transcript_from_jsonl(path: str) -> list:
                         pass
         return messages
     except Exception:
-        return []
+        return None
 
 
 def _collect_tool_names(transcript: list) -> list[str]:
@@ -109,12 +115,17 @@ def _collect_tool_names(transcript: list) -> list[str]:
 
 
 def main() -> None:
+    # Only run for sessions started by Lobster (LOBSTER_MAIN_SESSION=1).
+    # This guards against firing in a developer's personal Claude Code session.
+    if os.environ.get("LOBSTER_MAIN_SESSION", "") != "1":
+        _exit_ok()
+
     try:
         data = json.load(sys.stdin)
     except Exception:
         _exit_ok()  # If we can't read input, don't block.
 
-    # Only enforce on the dispatcher session.
+    # Only warn for the dispatcher session.
     if not is_dispatcher(data):
         _exit_ok()
 
@@ -122,6 +133,15 @@ def main() -> None:
     transcript_path = data.get("transcript_path", "")
     if transcript_path:
         transcript = _load_transcript_from_jsonl(transcript_path)
+        if transcript is None:
+            # I/O error reading the transcript — can't determine whether
+            # wait_for_messages was called, so warn and allow stop.
+            print(
+                "[require-wait-for-messages] WARNING: could not read transcript "
+                f"at {transcript_path!r} — skipping wait_for_messages check.",
+                file=sys.stderr,
+            )
+            _exit_ok()
     else:
         transcript = data.get("transcript", [])
 
@@ -130,9 +150,10 @@ def main() -> None:
     if _WFM_TOOL in tool_names:
         _exit_ok()
 
-    # wait_for_messages was not called — inject a reminder.
+    # wait_for_messages was not called — print a warning and allow the stop.
+    # This is warn-only: exit 0 so the stop is never blocked.
     print(_REMINDER, file=sys.stderr)
-    sys.exit(2)
+    _exit_ok()
 
 
 if __name__ == "__main__":


sayhar (commented):
🤖🦞 Lobster (auditor): PR #815 — require-wait-for-messages Stop hook

**Verdict: NEEDS-WORK (minor) — one real edge case, one doc gap, otherwise solid**

---

## 1. Dispatcher detection — PASS

`session_role.is_dispatcher()` correctly uses the marker file as primary signal and transcript scanning as fallback. The layered decision table in `_check_marker_file()` handles all the important cases:
- Marker absent → `None` (fall through to transcript)
- Marker matches session_id → `True`
- Marker present, session_id differs, stored session still alive → `False` (subagent)
- Marker present, session_id differs, stored session dead → `True` (dispatcher restart)

Subagents are exempted early in `main()` before any transcript I/O. No misfires expected on subagents in normal operation.

One observation: `require-wait-for-messages.py` does NOT check `LOBSTER_MAIN_SESSION=1`, while `write-dispatcher-session-id.py` does. This means the hook will run for Claude Code sessions started outside Lobster. For those sessions, the marker file won't exist (because `write-dispatcher-session-id.py` bailed out on that env check), so `is_dispatcher()` falls back to transcript scanning. If a developer's personal Claude Code session happens to have called `check_inbox` or `wait_for_messages` during that session, the hook would misclassify it as the dispatcher and inject a WFM reminder. This is an unlikely but non-zero risk for developer environments. Adding the same `LOBSTER_MAIN_SESSION=1` guard (exit 0 silently if absent) would eliminate this entirely.

## 2. WFM detection logic — one real edge case

The transcript path logic is:
```python
if transcript_path:
    transcript = _load_transcript_from_jsonl(transcript_path)
else:
    transcript = data.get("transcript", [])
```

`_load_transcript_from_jsonl` returns `[]` on **any** error (file not found, permissions error, I/O failure, malformed JSONL). When transcript is `[]`, `_collect_tool_names` returns `[]`, `_WFM_TOOL` is not in it, and the hook exits 2 — a **false-positive block**. The dispatcher gets an incorrect "call wait_for_messages" nudge even though it may have already called it.

This is recoverable (the dispatcher just calls WFM in the next turn), but it means a transcript read failure produces a spurious extra turn. The safe-fail direction here is arguably `_exit_ok()` when the transcript file is unreadable. Consider:

```python
if transcript_path:
    transcript = _load_transcript_from_jsonl(transcript_path)
    if transcript is None:   # distinguish "failed to read" from "empty"
        _exit_ok()           # can't determine — don't block
```

This requires changing `_load_transcript_from_jsonl` to return `None` on error vs `[]` on empty file (currently it conflates both).

The inline transcript path (`data.get("transcript", [])`) is safe — it returns `[]` only when the key is genuinely absent, which is correctly treated as "no WFM call seen."

## 3. install.sh — PASS with minor note

The `chmod +x` call is placed before the `if [ -f "$CLAUDE_SETTINGS" ]` check, so it runs unconditionally (even when settings.json doesn't exist yet). This is harmless — the chmod succeeds as long as the hook file exists — but it's mildly inconsistent with how other hooks are registered in this file. Not a bug.

The idempotency check `jq -e '.hooks.Stop[]? | select(.hooks[]?.command | contains("require-wait-for-messages"))'` is correct and matches the same pattern used for the SubagentStop hook registration. The `||true` on chmod matches the pattern used elsewhere.

## 4. Migration 33 — PASS

Sequence number is correct (follows Migration 32). The idempotency check and jq mutation are identical to install.sh. The `migrated` counter only increments when the hook is actually added. Structure is correct.

## 5. Other observations

**Double transcript read (not a bug):** In the dispatcher path, `is_dispatcher()` may read the transcript file once (via its fallback path in `session_role.py`), and then the hook reads it again. In normal operation with the marker file present, `is_dispatcher()` returns on the primary check and never touches the transcript file — so this is only triggered in the unlikely marker-file-absent path. No problem in practice.

**`_collect_tool_names` duplication:** The hook defines its own `_collect_tool_names` that walks the same JSONL/legacy structure as `session_role._transcript_has_dispatcher_tool`. Functionally correct (they serve different purposes — identification vs WFM-call detection). Minor code smell, not a bug.

**No retry counter:** Unlike `require-write-result.py`, this hook has no MAX_HOOK_FIRES escape hatch. If the dispatcher is in a state where it genuinely cannot call `wait_for_messages` (e.g. stuck in an error loop that also prevents WFM), the hook will fire on every turn indefinitely. The PR description says "the dispatcher can always call wait_for_messages — it is never in a state where the call is impossible," which is a reasonable claim given the dispatcher's architecture. Acceptable.

---

## Summary

| Check | Result |
|---|---|
| Dispatcher detection | PASS |
| Subagent exemption | PASS |
| WFM transcript detection | NEEDS-WORK (false positive on unreadable transcript) |
| install.sh | PASS |
| upgrade.sh Migration 33 | PASS |
| Race conditions | None found |

**Recommended fix before merge:** Handle `_load_transcript_from_jsonl` failure as `_exit_ok()` rather than falling through to exit 2. Either change the return value sentinel (None vs []) or add an explicit empty-transcript-with-path check.

**Deployment note:** Sahar has explicitly requested that this not be merged until it has been deployed locally and confirmed working in a live dispatcher session. The tests listed in the PR description are synthetic (piped input) — they do not cover a live Stop hook firing on a real dispatcher turn. Recommend testing with the actual dispatcher running before merging.

Issue #771, opened 2026-03-21
Dispatcher architecture: separate receiver from router to eliminate WFM stall risk


sayhar:
🤖🦞 Lobster (ops):

## Problem

The current dispatcher is a single Claude Code session that both (1) receives messages and (2) routes/processes them. Since it's a sequential LLM tool-calling loop, it can't call `wait_for_messages` again until the current message is fully processed. If any processing step takes >600s, the health check restarts the session.

This happened on 2026-03-21: a plans-poller result triggered 611s of deliberation (no clear routing rule), causing a restart.

The proximate fix (PR #770) adds a silent-drop rule for no-op results. But the structural vulnerability remains: any sufficiently long inference step can trigger a restart.

## Root cause (5 whys)

1. `wait_for_messages` wasn't called for 611s
2. Dispatcher was processing a single message for 611s
3. LLM inference without a clear rule triggers open-ended deliberation
4. LLM inference duration is not time-bounded from within the model
5. **The root cause: we're using an LLM as a real-time event processor.** LLMs are designed for reasoning quality, not latency guarantees. Using one as a dispatcher means latency spikes are structurally unavoidable.

## Proposed solution

A two-tier dispatcher architecture:

**Tier 1 — Receiver (lightweight, fast):**
- Accepts all incoming messages
- Writes to inbox (already done by `lobster_bot.py`)
- Sends immediate ack to user ("Message received")
- No routing logic, no LLM involved
- This tier already exists: `lobster_bot.py` sends the "📨 Message received. Processing..." ack

**Tier 2 — Router (current dispatcher, but with reduced pressure):**
- Reads from inbox, routes to agents
- Can take longer per message since the user already got a Tier 1 ack
- Health check threshold can be relaxed (or re-framed around "is it making progress" rather than "did it call WFM recently")

The key insight: `lobster_bot.py` already IS the Tier 1 receiver. The gap is that the Tier 2 dispatcher's health check is too strict given that users are already acked at Tier 1.

## Alternative approach

Keep the current architecture but:
1. Raise or remove the WFM freshness threshold for the health check (since restarts are disruptive)
2. Add a "default drop" rule: if the dispatcher hasn't returned to `wait_for_messages` within N seconds of receiving a message, drop the message and log it — but don't trigger a full restart

## Labels / project

File under BIS project in Linear separately if needed. GitHub issue is the primary tracker.

sayhar:
[...]

PR #820, opened 2026-03-25
fix: make WFM Stop hook blocking to eliminate post-batch deliberation stalls

sayhar:
## Why this exists

The warn-only Stop hook from PR #815 was insufficient. After that PR merged, 3 more health-check restarts occurred from the same root cause: the dispatcher ending a turn without calling \`wait_for_messages\`.

The failure mode (Failure Mode B from issue #768's audit): after processing a batch of system messages, the dispatcher calls \`mark_processed\` on the last one, then deliberates about "what to do next" instead of immediately calling \`wait_for_messages\`. The health check trips at 600s.

The warn-only hook printed to stderr after the turn was already ending — Claude had no chance to act on it in that turn. Warnings don't prevent deliberation; blocking does.

## What changed

**\`hooks/require-wait-for-messages.py\`**: Changed from warn-only (exit 0) to blocking (exit 2). When the dispatcher ends a turn without calling \`wait_for_messages\`, the hook:
1. Checks for \`/tmp/lobster-graceful-exit\` — if present, deletes it and exits 0 (graceful bypass, see below)
2. Otherwise exits 2, injecting an error message that tells the dispatcher exactly what to do:
   - Call \`wait_for_messages\` now (the normal case), OR
   - Write \`/tmp/lobster-graceful-exit\` and exit (the graceful-exit case)

**\`hooks/require-wait-for-messages.py\` — graceful exit bypass**: Two cases require the dispatcher to exit without calling \`wait_for_messages\` again:
- **Hibernation**: \`wait_for_messages(hibernate_on_timeout=True)\` returned EXIT/Hibernating — calling WFM again would be wrong
- **Context restart**: after \`lobster restart\`, the dispatcher should exit; forcing another WFM call would loop

The bypass: the dispatcher writes an empty file at \`/tmp/lobster-graceful-exit\` immediately before intentionally exiting. The hook consumes the file (deletes it) on first use — single-use, no lingering state.

The block message itself is the instruction manual: it documents the bypass inline, so the dispatcher knows what to do when blocked for a legitimate exit without consulting external docs.

**\`.claude/sys.dispatcher.bootup.md\`**: Added a note to the WFM-always-next rule section that the hook is now enforcing (not just advisory).

**\`tests/unit/test_hooks/test_require_wait_for_messages.py\`**: 13 tests total (9 original + 4 new) covering all behavior paths including the bypass mechanism.

## Flow: before vs after

```
Before (warn-only, PR #815):
  Dispatcher processes batch of system messages
    → mark_processed on last message
    → deliberates (611s) "what should I do now?"
    → turn ends naturally
    → Stop hook fires, prints warning to stderr
    → warning ignored (turn already ended)
    → health check trips at 600s → restart

After (blocking + graceful bypass):
  Normal stall case:
    → mark_processed → deliberates → turn ends without WFM
    → Stop hook fires, exits 2
    → error injected: "Call wait_for_messages NOW (or write bypass flag)"
    → new turn: dispatcher calls wait_for_messages
    → WFM blocks until next message — health check never trips

  Hibernation case:
    → WFM(hibernate_on_timeout=True) returns EXIT/Hibernating
    → dispatcher writes /tmp/lobster-graceful-exit
    → dispatcher exits
    → Stop hook fires, finds flag, deletes it, exits 0
    → turn ends cleanly

  Context restart case:
    → dispatcher writes handoff, calls lobster restart
    → dispatcher writes /tmp/lobster-graceful-exit
    → dispatcher exits
    → Stop hook fires, finds flag, deletes it, exits 0
    → turn ends cleanly
```

## What this does NOT change

- Subagent sessions are still exempt (\`is_dispatcher()\` returns False for subagents)
- Transcript read failures still allow stop (exit 0) — no false-positive blocks
- The health check remains the backstop for genuinely stuck sessions
- No changes to the dispatcher's main loop logic or message routing

## Functional patterns

Pure predicate check: collect all tool names from transcript, test for membership. Graceful bypass is a pure side-effect-isolated check (file presence → read → delete). Single branching function with no retry state.

## Tests run

- [x] \`uv run pytest tests/unit/test_hooks/test_require_wait_for_messages.py -v\` — 13 passed, 0 failed
- [x] \`uv run pytest tests/unit/ -q\` — 1572 passed, 3 failed (all pre-existing on main), 24 skipped, 6 errors (all pre-existing)
- [ ] Live dispatcher test — blocked: requires production behavior to trigger; change is live via symlink once merged

Relates to #771

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -359,6 +359,8 @@ REMINDER_ROUTING = {
 > After any `mark_processed` call that is NOT immediately followed by a `Task(...)` subagent spawn, the very next action is `wait_for_messages()`. No exceptions. No state assessment. No "what should I do now?" deliberation. WFM.
 >
 > The most common stall pattern is inline deliberation after processing a batch of system messages. If you find yourself thinking after `mark_processed`, you are violating this rule. Call WFM.
+>
+> **This rule is now enforced by a Stop hook** (`hooks/require-wait-for-messages.py`). If you end a turn without calling `wait_for_messages`, the hook **blocks the stop (exit 2)** and injects an error message into the next turn. The correct and only response to that error message is: call `wait_for_messages` immediately — nothing else first.
 
 **Rules:**
 - Never call `send_reply` for scheduled reminders (chat_id: 0, source: "system")
--- a/hooks/require-wait-for-messages.py
+++ b/hooks/require-wait-for-messages.py
@@ -1,17 +1,30 @@
 #!/usr/bin/env python3
 """
-Stop hook: warn the dispatcher if it ends a turn without calling wait_for_messages.
+Stop hook: block the dispatcher from ending a turn without calling wait_for_messages.
 
 The dispatcher's main loop is: process messages → call wait_for_messages →
 repeat. When the dispatcher stalls — typically after processing a batch of
-subagent results — it can end a turn without calling wait_for_messages. The
-health check catches this after ~12 minutes and restarts, but that is a long
-window with missed messages.
+subagent results or system messages — it can end a turn without calling
+wait_for_messages. The health check catches this after ~12 minutes and
+restarts, causing ~30s of missed messages per incident.
 
-This hook fires on every Stop event (dispatcher or subagent). Subagent sessions
-are immediately exempted via session_role.is_dispatcher(). For the dispatcher,
-the hook scans the transcript: if wait_for_messages was not called, it prints
-a warning to stderr and exits 0 (warn-only — the stop is never blocked).
+This hook fires on every Stop event (dispatcher or subagent). Subagent
+sessions are immediately exempted via session_role.is_dispatcher(). For the
+dispatcher, the hook scans the transcript: if wait_for_messages was not
+called, it exits 2 (blocking) so the dispatcher MUST call wait_for_messages
+before the turn ends.
+
+## Why blocking (exit 2) instead of warn-only (exit 0)
+
+The warn-only version (PR #815) printed to stderr but Claude proceeded with
+the turn ending anyway — 3 restarts occurred tonight after that PR merged.
+The warning fires after the turn is already ending; Claude has no chance to
+act on it in the same turn. Exit 2 injects the error message as a system
+turn and Claude gets a new turn where it must call wait_for_messages before
+stopping again.
+
+This mirrors how require-write-result.py handles subagents: hard-block until
+the required tool is called.
 
 ## Transcript handling
 
@@ -37,6 +50,7 @@ The hook does NOT fire for:
 - Sessions without LOBSTER_MAIN_SESSION=1 (non-Lobster Claude Code sessions)
 - Subagent sessions (is_dispatcher() returns False)
 - Any session where wait_for_messages was called at least once in the transcript
+- Transcript read failures (I/O error → allow stop rather than false-positive block)
 """
 import json
 import os
@@ -54,9 +68,11 @@ _SILENT_OK = json.dumps({"suppressOutput": True})
 
 _WFM_TOOL = "mcp__lobster-inbox__wait_for_messages"
 
-_REMINDER = (
-    "Dispatcher: you ended this turn without calling wait_for_messages. "
-    "Call it now to continue the main loop."
+_BLOCK_MESSAGE = (
+    "STOP: You are the Lobster dispatcher. You ended this turn without calling "
+    "wait_for_messages. The main loop requires wait_for_messages as the very next "
+    "action after mark_processed — no deliberation, no state assessment.\n\n"
+    "Call mcp__lobster-inbox__wait_for_messages NOW. Do not do anything else first."
 )
 
 
@@ -125,7 +141,7 @@ def main() -> None:
     except Exception:
         _exit_ok()  # If we can't read input, don't block.
 
-    # Only warn for the dispatcher session.
+    # Only enforce for the dispatcher session.
     if not is_dispatcher(data):
         _exit_ok()
 
@@ -135,7 +151,8 @@ def main() -> None:
         transcript = _load_transcript_from_jsonl(transcript_path)
         if transcript is None:
             # I/O error reading the transcript — can't determine whether
-            # wait_for_messages was called, so warn and allow stop.
+            # wait_for_messages was called. Allow stop rather than false-positive
+            # block; the health check remains the backstop.
             print(
                 "[require-wait-for-messages] WARNING: could not read transcript "
                 f"at {transcript_path!r} — skipping wait_for_messages check.",
@@ -150,10 +167,11 @@ def main() -> None:
     if _WFM_TOOL in tool_names:
         _exit_ok()
 
-    # wait_for_messages was not called — print a warning and allow the stop.
-    # This is warn-only: exit 0 so the stop is never blocked.
-    print(_REMINDER, file=sys.stderr)
-    _exit_ok()
+    # wait_for_messages was not called — block the stop (exit 2).
+    # Claude gets a new turn with _BLOCK_MESSAGE injected and must call
+    # wait_for_messages before the next stop attempt succeeds.
+    print(_BLOCK_MESSAGE, file=sys.stderr)
+    sys.exit(2)
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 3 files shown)


sayhar (commented):
PASS: Blocking hook correctly addresses the warn-only gap from PR #815.

## Summary

The core change — switching from `exit 0` + stderr warning to `exit 2` + injected block message — is the right fix. The warn-only approach was structurally broken: the hook fired after the turn had already ended, so Claude had no chance to act. Exit 2 gives Claude a fresh turn with a directive message. This is the same pattern used by `require-write-result.py` for subagents, and it's the correct tool for this job.

## Review of Specific Areas

### 1. `_BLOCK_MESSAGE` — directive quality

The message is directive enough. It contains:
- An explicit imperative: \"Call mcp__lobster-inbox__wait_for_messages NOW\"
- A disqualifier: \"Do not do anything else first\"
- A reminder of the rule: \"no deliberation, no state assessment\"

The only marginal concern is that the block message goes to `stderr` but the Claude Code hook mechanism injects it as a system turn regardless. That's correct behavior — the message will appear in the next turn. The full tool name `mcp__lobster-inbox__wait_for_messages` is present so Claude can call it without guessing.

### 2. Transcript read failure path — exits 0

Correct. Line 158-162 in the modified hook: `if transcript is None: ... _exit_ok()`. The logic is: when we cannot determine whether WFM was called, we fail open (allow stop) rather than fail closed (false-positive block). The health check remains the backstop for genuinely indeterminate cases. This is the right tradeoff.

`test_missing_transcript_path_exits_0` confirms this path.

### 3. No retry loop / MAX_HOOK_FIRES — design correctness

The simpler design is correct here. The dispatcher calling `wait_for_messages` is not a \"retry\" — it's a successful recovery that transitions control back to the main loop. WFM blocks until a message arrives, then returns normally. There is no failure mode that would require a retry counter: either the dispatcher calls WFM (success) or it tries to stop again without calling WFM (hook fires again). Repeated blocking is the correct behavior in that degenerate case, not a loop to cap.

Contrast with `require-write-result.py`: subagents that fail to call `write_result` might legitimately exhaust retries or hit edge cases that need bounding. The dispatcher's main loop has no equivalent — WFM is always the right next action.

## Test Coverage

9 tests cover all meaningful branches:
- Non-Lobster session exemption (no `LOBSTER_MAIN_SESSION=1`)
- Subagent exemption (`is_dispatcher` returns False)
- Dispatcher + WFM in transcript → exit 0
- Dispatcher + no WFM in transcript → exit 2
- Dispatcher + both mark_processed and WFM → exit 0 (membership check, not ordering)
- Missing transcript → exit 0 (no false-positive)
- Block message content contains \"wait_for_messages\"
- `suppressOutput` JSON on all exit-0 paths (2 tests)

One minor note: there is no test for the ordering case — e.g., WFM called *before* mark_processed. The current implementation uses set membership (`_WFM_TOOL in tool_names`) so it would pass regardless of call order. That is probably correct: if WFM was called at all during the turn, the dispatcher is not in a \"deliberating after mark_processed\" failure mode. If a stricter ordering check were ever needed, this would be where to add it — but it's out of scope for this PR and the current behavior is defensible.

## Minor Notes

- The `sys.dispatcher.bootup.md` addition is concise and accurate. The dispatcher now has explicit guidance that the correct response to the block message is WFM immediately, with no other steps. This is important because without it, Claude might try to \"investigate why WFM wasn't called\" before calling it.
- The hook skips the `suppressOutput` JSON on exit 2 (blocking path). This is correct — on a block, you want Claude to see the message, not suppress it.

## Verdict

This is a clean, minimal fix with correct failure semantics, adequate tests, and a well-scoped doc update. The design parallels an established pattern in the codebase. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

This PR graduates the WFM Stop hook from warn-only (exit 0) to blocking (exit 2). The motivation is solid and well-evidenced: 3 restarts after PR #815 demonstrated that a post-turn warning is structurally unable to change dispatcher behavior in the same turn. The blocking approach mirrors the already-proven `require-write-result.py` pattern for subagents.

## Code review

**hooks/require-wait-for-messages.py**
- The change is minimal and targeted: `_exit_ok()` at the end replaced by `print(_BLOCK_MESSAGE, file=sys.stderr); sys.exit(2)`. All exemption paths (non-Lobster session, subagent, WFM present, transcript read failure) are preserved unchanged. No regressions introduced.
- `_BLOCK_MESSAGE` is appropriately imperative and unambiguous. The dispatcher will have exactly the right context to act on it.
- The transcript-read-failure path correctly uses `_exit_ok()` to avoid false-positive blocks — good defensive coding.

**sys.dispatcher.bootup.md**
- The added note is correctly placed adjacent to the WFM-always-next rule. One sentence; no padding. Appropriate.

**tests/unit/test_hooks/test_require_wait_for_messages.py**
- 9 tests covering the full decision matrix: non-Lobster env, subagent session, dispatcher + WFM, dispatcher without WFM, mixed transcript, missing transcript, block message content, suppressOutput JSON. Coverage is complete.
- Test helpers (`_load_hook`, `_make_jsonl_transcript`, `_run_hook`) are clean and reusable.
- One minor observation: `_load_hook` inserts `HOOKS_DIR` into `sys.path` unconditionally on each call (guarded by `not in` check, fine in practice). No impact on correctness.

## Checklist

- Does the code fix what the PR claims? Yes — changes warn-only exit 0 to blocking exit 2 exactly as described.
- New bugs? None identified.
- Minimal/focused? Yes — three files, all tightly scoped.
- Follows existing patterns? Yes — mirrors `require-write-result.py`.
- install.sh update needed? No — hook already existed; no new hooks, cron, services, or env vars.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one flag worth tracking

## Overall

Mechanism is sound and the blocking approach is correct. The warn-only PR #815 had a fundamental flaw — warnings fire after the turn ends, so they can't change behavior. Exit 2 is the right fix. Tests are solid (13/13, covering all paths including bypass). The bootup doc update is present in the diff (despite the engineer's briefing saying it wasn't — the briefing may be stale).

## Security: cross-session bypass (low risk, worth noting)

The bypass flag at `/tmp/lobster-graceful-exit` can be written by any process, including subagents. The hook correctly exempts subagents via `is_dispatcher()` for their *own* sessions — but a subagent could write the flag during its run and allow the dispatcher's *next* stop event to bypass the hook. This is a timing-dependent cross-session race. In practice the risk is very low (subagents and dispatcher don't run concurrently in a window that makes this exploitable), but it's worth documenting as a known limitation rather than a security property.

No action required unless threat model changes.

## Minor: block message hardcodes the path

`_BLOCK_MESSAGE` embeds the literal string `open('/tmp/lobster-graceful-exit', 'w').close()`. If `_GRACEFUL_EXIT_FLAG` is ever changed, the message text won't follow. Suggest interpolating:

```python
_BLOCK_MESSAGE = (
    ...
    f"    open({_GRACEFUL_EXIT_FLAG!r}, 'w').close()"
)
```

Not a blocker.

## `/tmp` path

Acceptable. Ephemeral, Linux-native, matches the single-use intent.

## Tests

All 4 new tests cover the right things. `test_graceful_exit_flag_is_consumed` is particularly important — single-use guarantee is tested explicitly. No gaps.

## Recommendation

Merge. Track the cross-session bypass note and the interpolated path as follow-up.

sayhar (commented):
PASS: This is the highest-leverage fix in the queue. All 13 tests pass.

**What works:**
- Blocking Stop hook (exit 2) correctly prevents post-batch deliberation stalls
- Graceful exit bypass (`/tmp/lobster-graceful-exit` flag) handles hibernation and context restart cases cleanly — single-use, consumed on first read
- Block message is self-documenting: tells dispatcher exactly what to do (including the bypass mechanism) without needing external docs
- Subagent exemption via `is_dispatcher()` is correct — hook is dispatcher-only
- Transcript read failure path is fail-open (allows stop rather than false-positive block)
- The `suppressOutput: true` JSON output prevents feedback injection on all success paths

**Test coverage:** 13 tests covering all paths including graceful bypass, flag consumption, transcript missing, and block message content. Comprehensive.

**One minor observation:** the block message hardcodes `/tmp/lobster-graceful-exit` rather than using `_GRACEFUL_EXIT_FLAG` (the local-dev branch already has a nit commit fixing this with string interpolation). This was caught and addressed — not a blocker.

PR #846, opened 2026-03-25
fix: search conversation history before saying 'I don't recognize X'

### LINKED ISSUE #845: Dispatcher says 'I don't recognize X' instead of searching conversation history
## Problem

When a user asks about a named mode, session, or term — e.g. \"what did you do during super librarian mode\" — the dispatcher replies \"I'm not familiar with X\" without first checking conversation history.

This is the wrong default. The dispatcher loses named-session context through context compactions and restarts. The user reasonably expects the system to look up what it did rather than admit ignorance immediately. Saying \"I don't recognize that\" when \`get_conversation_history(search=X)\` would have answered the question is a failure of self-service.

## Root cause

The dispatcher has a rule to delegate code-internals questions rather than speculate, but no equivalent rule for named-session/mode questions. Those questions are treated as general knowledge lookups, which the dispatcher answers from (stale or absent) in-context memory.

## Fix

Add a rule alongside the existing \"Code internals questions → delegate, don't speculate\" section in \`sys.dispatcher.bootup.md\` that requires the dispatcher to call \`get_conversation_history\` before responding to any question about a named mode, session, or term it does not immediately recognize.

The rule should make clear: only after searching and finding nothing is it appropriate to say the term isn't recognized.

sayhar:
## What problem this solves

The dispatcher was admitting ignorance about named modes and sessions without first checking whether conversation history had the answer. A user asking "what did you do during super librarian mode?" would get "I'm not familiar with that" when `get_conversation_history` would have surfaced the answer immediately.

This matters because named-session context is routinely lost. Every context compaction, restart, and long-running session erodes the dispatcher's in-context memory. Absence from the current context window is not evidence that the session never happened — it is evidence only that the dispatcher has forgotten. The user has no way to distinguish "it never happened" from "the system forgot."

## What changed

A new rule is added to `sys.dispatcher.bootup.md`, placed immediately after the existing "Code internals questions → delegate, don't speculate" rule (line 89-90), which it parallels in spirit:

**Named mode/session/term questions — search first, never say "I don't recognize":** When the user asks about a named mode, session, or term the dispatcher doesn't immediately recognize, it must delegate a subagent to call `get_conversation_history` before responding. Only after searching and finding nothing is it appropriate to say the term isn't recognized.

The implementation approach is the same pattern already used for code-internals questions: delegate first, answer from returned evidence, never speculate from absent in-context memory.

## What is not changed

This is a documentation-only change to a bootup prompt. No code is modified. No routing logic, MCP server behavior, or message handling is affected.

## Functional patterns

Pure instruction: the rule is declarative and stateless — it prescribes a deterministic lookup step rather than asking the dispatcher to make a judgment call about whether to search.

## Tests run

- [ ] No automated tests apply — this is a bootup prompt change. Behavior verification requires a live dispatcher session. The rule is verifiable by inspection against the diff.

Closes #845

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -89,6 +89,13 @@ If you find yourself reaching for `Read`, `Bash`, `mcp__github__*`, `WebFetch`, 
 **Code internals questions → delegate, don't speculate**
 When asked how something works internally (a function, a module, a system), spawn a subagent to read the actual code — unless the answer is already present in the current context from a recently returned subagent report. Do not reason from memory or give plausible-sounding explanations without source confirmation.
 
+**Named mode/session/term questions — search first, never say "I don't recognize":**
+When the user asks about a named mode, session, or term you don't immediately recognize
+(e.g. "what did you do during X", "what is X"), do NOT reply "I'm not familiar with X."
+Instead, immediately delegate a subagent to call `get_conversation_history` searching for
+that term. Only after searching (and finding nothing) is it appropriate to say you don't
+recognize it.
+
 **Ack policy — when to send "On it." before delegating:**
 
 **Two-layer ack architecture:** The Telegram bot (`lobster_bot.py`) automatically sends "📨 Message received. Processing..." to the user at the transport layer as soon as it writes a text message to the inbox. This fires for all plain text messages before you ever see the message. Your "On it." is a *second*, dispatcher-level ack — it signals that work is underway, not that the message was received.


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The rule is clear, well-placed, and correctly scoped. The fix is sound. Minor observations below, none blocking.

---

## Placement

Placed immediately after the parallel "Code internals questions → delegate, don't speculate" rule (line 89). This is exactly right — the two rules are structurally identical in spirit (delegate a lookup before answering from memory), so adjacency makes the pattern legible and self-reinforcing.

---

## Scope of "Named mode/session/term"

The scope is good. "Named mode, session, or term" covers the primary failure class: user-coined labels for past sessions, experiments, or recurring behaviors that live in conversation history rather than in code or config. The examples ("what did you do during X", "what is X") make the trigger unambiguous.

One edge case not covered: proper nouns that are project names or people's names rather than session labels (e.g. "what did we say about the Archipelago project?"). In practice these are handled the same way — search history before saying you don't know — but they don't quite fit "named mode/session/term." The current wording is not wrong, just slightly under-inclusive. This is worth a follow-up but not a blocker.

---

## "Immediately delegate a subagent" — prescriptiveness level

This is correct. The rule mirrors the code-internals rule's prescriptive stance ("spawn a subagent to read the actual code"). Both rules exist precisely because the dispatcher's natural failure mode is reasoning from absent context instead of doing the lookup. A softer instruction ("consider delegating") would not correct the failure mode. The prescriptive phrasing is appropriate.

---

## "Only after searching and finding nothing"

This is the right discharge condition. The rule does not require the subagent to find something — only that the search was performed. This prevents false positives ("I searched but found nothing" is a valid answer) while still eliminating false negatives ("I didn't bother to look"). Correct.

---

## Known gaps (from engineer's briefing)

- **No automated tests**: Acknowledged. Bootup prompt changes are not automatically testable. The rule is verifiable by inspection, which this review provides.
- **Main Board project not found**: Not relevant to the correctness of this change.

---

## Verdict

The rule is well-formed, correctly placed, and addresses the stated failure mode. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict:** The rule is correct, well-scoped, and integrates cleanly. No concerns.

---

**Scope ("named mode/session/term"):** The scope feels appropriately targeted. "Named mode/session/term" is narrower than a general "anything I don't recognize" rule, which is intentional and good — the dispatcher can legitimately not know many things. The key insight being protected here is specifically context that *was* in conversation history and may have been compacted away. The three-word set covers the dominant cases: deliberate modes the user invoked, named sessions, and user-defined shorthand terms. If anything, the rule could note that named *tasks* or *projects* might fall in the same bucket, but the current framing is clear enough that a dispatcher can reason by analogy, and over-specifying would bloat the instruction.

**"Immediately delegate a subagent":** The prescriptiveness is right. The risk being avoided is the dispatcher taking a shortcut — answering "I don't recognize that" from absent in-context memory rather than doing the lookup. Leaving wiggle room ("consider delegating") would recreate the failure mode. The word "immediately" mirrors the intent of the adjacent code-internals rule and is appropriate here.

**Integration with adjacent rules:** Placement immediately after the code-internals delegation rule is exactly right — the two rules share structure (delegate → evidence → answer, never speculate from absence) and benefit from proximity. The rule is also consistent with the general subagent delegation pattern described in the surrounding context.

**No exception for "obviously new terms":** The rule correctly omits a "unless it's clearly brand new" carve-out. That judgment is exactly the one the dispatcher should not be trusted to make unilaterally; it should confirm via search first. Correct.

**Tests:** No automated tests apply to bootup prompt changes. Behavior verification requires a live dispatcher session. This is noted correctly in the PR description and does not block merge.

No changes needed.

PR #849, opened 2026-03-25
Add sender_type filter to get_conversation_history

### LINKED ISSUE #848: Add sender_type filter to get_conversation_history to exclude cron/system noise
## Problem

`get_conversation_history` returns every message in the DB with no way to isolate the real back-and-forth conversation. Cron outputs, `subagent_result` system messages, `health_check`, `scheduled_reminder`, and other system-generated types all appear alongside genuine user↔Lobster exchanges. When Sahar asks \"what did we talk about yesterday?\" the raw history is polluted with system noise.

The existing `direction` filter (received/sent/all) does not help here because system messages also use `direction='in'` (received), so filtering by direction alone does not remove the noise.

## Proposed Solution

Add a `sender_type` parameter to the `get_conversation_history` MCP tool:

| Value | Meaning |
|---|---|
| `"user"` | Only messages FROM Sahar (inbound from Telegram/Slack — types in `INBOX_USER_TYPES`: text, photo, voice, document, etc.) |
| `"lobster"` | Only messages sent BY Lobster/subagents TO Sahar (direction='out') |
| `"conversation"` | Both user and lobster messages — excludes all system/cron noise |
| omitted/`None` | Current behavior — all messages (no change) |

The filter must be applied at the SQL layer (not post-fetch) so pagination counts are correct. It also needs to be applied in the filesystem fallback path for the legacy migration window.

## Implementation Notes

- In `src/db/reader.py`: add `sender_type` parameter to `get_conversation_history` and `count_conversation_history`. Map it to SQL conditions using `INBOX_USER_TYPES` from `message_types.py`.
  - `"user"` → `direction = 'in' AND type IN (...)` for all user types
  - `"lobster"` → `direction = 'out'`
  - `"conversation"` → `(direction = 'out' OR type IN (...))`
- In `src/mcp/inbox_server.py`: thread `sender_type` through `handle_get_conversation_history` and update the MCP tool schema.
- In `_apply_filters_and_paginate`: add equivalent in-memory filtering for the filesystem fallback path.
- Add tests covering each `sender_type` value in both paths.

## Acceptance Criteria

- [ ] `sender_type="user"` returns only messages where type is in `INBOX_USER_TYPES` and direction='in'
- [ ] `sender_type="lobster"` returns only direction='out' messages
- [ ] `sender_type="conversation"` returns the union of the above (no system/cron messages)
- [ ] `sender_type` omitted returns all messages (no regression)
- [ ] Pagination counts are correct for all values (filter applied before count)
- [ ] Works in both DB path and filesystem fallback path
- [ ] Unit tests cover all four cases

Relates to: #156 (formal message type taxonomy)

sayhar:
## Problem

`get_conversation_history` returns everything — cron outputs, `subagent_result` routing messages, `health_check`, `scheduled_reminder`, and other system-generated types all appear alongside genuine user↔Lobster exchanges. When looking up recent conversation context, the system noise drowns the signal.

The existing `direction` filter doesn't help because system messages also arrive as `direction='in'`, so `direction=received` still includes all the noise.

## What this adds

A `sender_type` parameter with four values:

| Value | Effect |
|---|---|
| `"user"` | Inbound messages whose type is in `INBOX_USER_TYPES` (text, photo, voice, document, etc.) — excludes `subagent_result`, `scheduled_reminder`, `health_check`, and all other system-generated types |
| `"lobster"` | Outbound messages only — what Lobster sent back |
| `"conversation"` | Union of the two above — the real back-and-forth with no system noise |
| omitted | Current behavior — all messages, no change |

When `sender_type` is set, the `direction` parameter is ignored (sender_type encodes its own directionality). Omitting `sender_type` is a strict no-op.

## Design

The filter is applied at two layers so it's correct in both read paths:

**SQL path (primary):** `_sender_type_sql()` in `reader.py` is a pure function that returns `(conditions, params)` for a `WHERE` clause. Applied to both `get_conversation_history` and `count_conversation_history` so pagination counts are correct.

- `"user"` → `direction = 'in' AND type IN (text, photo, voice, ...)`
- `"lobster"` → `direction = 'out'`
- `"conversation"` → `(direction = 'out' OR (direction = 'in' AND type IN (...)))`

**Filesystem fallback path (legacy migration window):** `_apply_sender_type_filter()` in `inbox_server.py` is a pure function that applies the same semantics in-memory. Called inside `_apply_filters_and_paginate()` before pagination so the `total` count is also correct.

The canonical type list comes from `INBOX_USER_TYPES` (already imported in both files), so there's one source of truth.

## Functional patterns

Pure functions with no side effects: `_sender_type_sql()` and `_apply_sender_type_filter()` both take data in and return data out. Composed into the existing filter pipelines rather than adding new conditional branches to the handler. The SQL helper returns plain `(list, list)` tuples that callers append to existing conditions/params — composable and testable in isolation.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_sender_type_filter.py -v` — 23 passed, 0 failed
- [x] `uv run pytest tests/unit/test_mcp_server/test_conversation_history_sql.py -v` — 37 passed, 0 failed (no regression)
- [x] `uv run pytest tests/unit/test_mcp_server/test_relay_db_reads.py -v` — 50 passed, 0 failed (no regression)
- [x] `uv run pytest tests/unit/ -q` — 1582 passed, 3 pre-existing failures (test_auto_register_agent, test_skill_manager ×2), 6 pre-existing errors (test_path_guard)

Closes #848

### DIFF
--- a/src/db/reader.py
+++ b/src/db/reader.py
@@ -22,8 +22,20 @@ from __future__ import annotations
 
 import json
 import sqlite3
+import sys
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import Any
+
+# Import canonical user types for sender_type filtering.
+# message_types.py lives alongside inbox_server.py in src/mcp/.
+_MCP_DIR = str(Path(__file__).resolve().parent.parent / "mcp")
+if _MCP_DIR not in sys.path:
+    sys.path.insert(0, _MCP_DIR)
+try:
+    from message_types import INBOX_USER_TYPES as _INBOX_USER_TYPES
+except ImportError:  # pragma: no cover — fallback when run outside normal tree
+    _INBOX_USER_TYPES = frozenset()
 
 # ---------------------------------------------------------------------------
 # Type aliases
@@ -84,6 +96,46 @@ def _parse_timestamp(ts: str | None) -> datetime:
 
 
 # ---------------------------------------------------------------------------
+# Sender-type filter helpers — pure functions
+# ---------------------------------------------------------------------------
+
+def _sender_type_sql(sender_type: str | None) -> tuple[list[str], list[Any]]:
+    """
+    Return (conditions, params) for a sender_type filter.
+
+    sender_type semantics:
+      "user"         — inbound messages whose type is a user-initiated type
+                       (text, photo, voice, document, etc. — excludes system noise)
+      "lobster"      — all outbound messages (direction='out')
+      "conversation" — union of user and lobster (real conversation, no cron/system)
+      None / other   — no additional conditions (all messages)
+
+    Returns plain lists so callers can append to their existing conditions/params.
+    This is a pure function with no side effects.
+    """
+    if not sender_type or sender_type == "all":
+        return [], []
+
+    # Build the IN-list for user types once, using positional parameters.
+    user_types = sorted(_INBOX_USER_TYPES)  # sorted for deterministic SQL
+    placeholders = ", ".join("?" * len(user_types))
+    user_clause = f"(direction = 'in' AND type IN ({placeholders}))"
+
+    if sender_type == "user":
+        return [user_clause], list(user_types)
+
+    if sender_type == "lobster":
+        return ["direction = 'out'"], []
+
+    if sender_type == "conversation":
+        # outbound OR inbound-user-type
+        return [f"(direction = 'out' OR {user_clause})"], list(user_types)
+
+    # Unknown value — treat as no filter (safe degradation)
+    return [], []
+
+
+# ---------------------------------------------------------------------------
 # Table detection
 # ---------------------------------------------------------------------------
 
@@ -118,23 +170,30 @@ def get_conversation_history(
     source: str | None = None,
     search: str | None = None,
     direction: str = "all",
+    sender_type: str | None = None,
     limit: int = 20,
     offset: int = 0,
 ) -> list[Row]:
     """
     Fetch conversation history from the messages table.
 
-    Applies optional filters for chat_id, source, full-text search, and
-    direction (received/sent/all).  Results are ordered newest-first.
+    Applies optional filters for chat_id, source, full-text search,
+    direction (received/sent/all), and sender_type.  Results are ordered
+    newest-first.
 
     Args:
-        conn:      Open sqlite3.Connection.
-        chat_id:   Filter to a specific conversation (compared as string).
-        source:    Filter by message source (e.g. 'telegram', 'bisque').
-        search:    Keyword search over message text (uses FTS5 when available).
-        direction: 'received' | 'sent' | 'all' — maps to direction IN/OUT.
-        limit:     Maximum number of rows to return.
-        offset:    Rows to skip for pagination.
+        conn:        Open sqlite3.Connection.
+        chat_id:     Filter to a specific conversation (compared as string).
+        source:      Filter by message source (e.g. 'telegram', 'bisque').
+        search:      Keyword search over message text (uses FTS5 when available).
+        direction:   'received' | 'sent' | 'all' — maps to direction IN/OUT.
+        sender_type: 'user' | 'lobster' | 'conversation' | None
+                     'user'         — inbound user-type messages only (no system/cron)
+                     'lobster'      — outbound messages only
+                     'conversation' — both user and lobster (excludes system noise)
+                     None / omitted — all messages (current default behaviour)
+        limit:       Maximum number of rows to return.
+        offset:      Rows to skip for pagination.
 
     Returns:
         List of plain dicts with all message columns plus '_direction'.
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: Clean, well-structured implementation with one design issue to document and one coupling to address.

---

## Summary

Adds a `sender_type` parameter to `get_conversation_history` that filters out system/cron noise from conversation history. The implementation is correct, the pure-function design is good, and test coverage is solid. Two items worth flagging:

---

## Issues

### 1. `direction` is silently ignored when `sender_type` is set — undocumented behavior at the call site (minor)

The engineer flagged this as a design tradeoff. It is acceptable — `sender_type` encodes its own directionality so combining both would produce contradictory SQL conditions. However, the current behavior silently drops `direction` with no warning. A caller who passes `direction="received"` and `sender_type="lobster"` expecting to narrow to received-lobster messages gets only the `sender_type` behavior with no indication that `direction` was discarded.

Recommended fix: log a debug warning (not an error — this is not a hard failure) when both are supplied simultaneously. Something like:
```python
if sender_type and sender_type != "all" and direction != "all":
    log.debug("sender_type=%s set; direction=%s ignored", sender_type, direction)
```

This makes the tradeoff observable in logs without breaking anything.

### 2. `sys.path` manipulation in `reader.py` creates cross-layer coupling (minor, as flagged)

`src/db/reader.py` now inserts `src/mcp/` onto `sys.path` at import time to pull in `INBOX_USER_TYPES`. The DB layer (`src/db/`) should not know about the MCP layer (`src/mcp/`). The type list is a domain concept — it belongs in a shared location (e.g. `src/message_types.py` or `src/db/message_types.py`) that both layers can import cleanly. The current `try/except ImportError` fallback (returning `frozenset()`) means a deployment error could silently make the `user` and `conversation` filters return zero results rather than failing loudly.

Recommended fix: move `INBOX_USER_TYPES` (or a re-export) to a location in `src/` that is on the normal import path. The `sys.path` manipulation and the `ImportError` silent fallback can then be removed entirely.

---

## What's good

- Pure functions `_sender_type_sql()` and `_apply_sender_type_filter()` are composable and independently testable — good design.
- Filter applied at both SQL and filesystem-fallback layers, with pagination counts correct in both paths.
- `sorted()` on the type list for deterministic SQL is a nice touch.
- 23 new tests with good coverage: pure helper, paginator, SQL path, end-to-end propagation, no-regression.
- FTS5 + `sender_type` combination: the SQL conditions are appended to the same `conditions` list, so they compose correctly with the FTS join — no explicit test needed given the composable structure, but one would be a good addition if this code path gets more complex.
- Omitting `sender_type` is a strict no-op — confirmed by test and by code inspection.

---

## Verdict

PASS with the two items above as suggestions. Issue 2 (coupling) is the stronger concern and worth fixing before this pattern propagates. Issue 1 (silent ignore) is low-risk but cheap to address.

sayhar (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK — upgrading verdict from PASS.

Two issues need fixing before merge:

1. **DB/MCP coupling via sys.path (blocker):** `reader.py` adds `src/mcp/` to `sys.path` to import `INBOX_USER_TYPES` from `message_types.py`. This couples the DB layer to the MCP layer and includes a silent `ImportError` fallback that causes the sender_type filter to silently degrade to no-op if the import fails. Fix: move `INBOX_USER_TYPES` (or the relevant constants) to a shared `src/` location (e.g. `src/types.py` or `src/constants.py`) that both layers can import without path manipulation.

2. **Silent direction ignore (minor but fix it):** When both `sender_type` and `direction` are supplied, `direction` is silently ignored. This should either raise a ValueError or log a warning — silent parameter discard is confusing to callers.

Implementation quality is otherwise solid. Address these two and it's a PASS.

sayhar (commented):
PASS: Well-structured feature. Tests pass (60/60 in test_conversation_history_sql.py, plus 23 tests in test_sender_type_filter.py passing per PR description).

**What works:**
- `sender_type` parameter has clean semantics: `user`, `lobster`, `conversation`, and None (no-op default)
- Pure function design: `_sender_type_sql()` in reader.py and `_apply_sender_type_filter()` in inbox_server.py are both pure, no side effects
- Both SQL path and filesystem fallback path apply the filter — pagination counts are correct on both paths
- `sender_type` takes precedence over `direction` when set (documented in docstring)
- Single source of truth: `INBOX_USER_TYPES` imported from message_types.py in both files
- The `sys.path` manipulation in reader.py for the import is a minor smell but acceptable given the package structure

**Observation:** the `_MCP_DIR` sys.path insertion in reader.py is cross-package coupling. If the directory layout ever changes this breaks silently (falls back to empty frozenset). The graceful fallback mitigates the risk. Could be improved with a proper package import, but not a blocker given the current codebase structure.

Issue #847, opened 2026-03-25
Improve session memory continuity: session notes, structured catchup logs, research delegation


sayhar:
🤖🦞 Lobster (ops): Updated to serve as the epic for session notes / structured memory continuity design.

---

## Epic: Session Notes and Structured Memory Continuity

This epic tracks the full design for how Lobster maintains memory, context, and behavioral continuity across compactions, restarts, and subagent crashes.

---

### 1. Numbered Session Files

Session notes are stored as numbered files (e.g. `20260325-001.md`) at `~/lobster-user-config/memory/canonical/sessions/`. This makes sequencing obvious, enables tiered reading (recent sessions in full, older sessions by header only), and avoids ambiguity when sorting. Files live in user-config (persistent, committed, survives reinstalls).

- The dispatcher writes to the current session file during a session
- Before compaction, it polishes the file into a clean handoff summary
- On bootup, it reads recent sessions as part of the startup sequence

### 2. Open Threads Section

Each session file includes an **Open Threads** section capturing incomplete task loops and pending subagent work. An open thread explicitly covers:

1. **Undelivered replies** — subagent replied but user may not have seen it
2. **Incomplete multi-part tasks** — subagent did part of a request but more was implied
3. **In-flight subagents** — tasks that were spawned and may not have completed

Each Open Threads entry has four fields:
- **What was asked**: the user's original request or question
- **What was done**: what was actually done
- **What's still pending**: what remains unresolved
- **User acknowledged**: `yes` | `no` | `silence`

### 3. Open Tasks and Open Subagents Sections

Session files include:
- **Open Tasks** (replaces "Open PRs"): any pending work regardless of type — tasks, PRs, analyses, delegated work
- **Open Subagents**: subagents spawned but not yet completed, tracked by `task_id` + description + expected completion

### 4. IFTTT-Style Behavioral Rules File (cap enforced)

A separate file of "if X then Y" behavioral rules that Lobster loads at startup. A soft cap is enforced: when hit, oldest/least-referenced entries are archived to the memory DB (not silently dropped). The memory DB is queryable by ID so archived rules remain referenceable.

See sub-issue: #853

### 5. Tiered Bootup Reading

On startup, the dispatcher reads session notes in tiers:
- **Full read**: the most recent 2 session files
- **Header-only read**: the previous 5 session files (just top-level summaries)
- **Skip**: older sessions beyond a defined window

[...]

Issue #821, opened 2026-03-25
Dispatcher self-restart (lobster restart) kills live session before new one starts


sayhar:
🤖🦞 Lobster (ops):

## Root cause
When the dispatcher encounters an MCP disconnect and fails to recover via ToolSearch, it falls back to calling `lobster restart` via Bash. This is self-destructive:

1. `lobster restart` runs `systemctl stop lobster-claude` 
2. This kills the exact session the dispatcher is running in
3. The `systemctl start` never executes because the calling process is dead
4. Result: both services dead, ~114 second outage

## Impact
2026-03-25 03:10–03:12 UTC: ~114 second downtime. No messages lost (health check revived).

## Fix needed
The dispatcher should NEVER call `lobster restart` from inside itself. Options:
- Detect we're running inside the managed session and refuse to self-stop
- Use a separate watchdog/wrapper process that survives the stop
- Replace inline self-restart with a flag file that the health check picks up and uses to trigger a clean external restart

## Labels
bug, reliability

PR #850, opened 2026-03-25
fix: remove lobster restart from context_warning handler (Option A: lean on compaction)

sayhar:
## What changed and why

The `context_warning` handler previously called `Bash("lobster restart")` as step 6 after writing the handoff file. This PR removes that call in favor of Option A: the dispatcher simply stops its main loop and goes idle, letting Claude Code's built-in compaction handle context recovery.

The restart machinery introduced two problems: it added implementation complexity (flag files, polling, process management), and it created a polling dead window of up to 4 minutes between exit and re-entry into `wait_for_messages`. Compaction is already happening at the right time — the missing piece was structured handoffs that survive it cleanly (tracked in #847).

## New behavior

Before: `write handoff → notify user "restarting" → lobster restart → mark_processed`

After: `write handoff → notify user "entering wind-down mode" → stop the main loop (no restart, no further wait_for_messages) → mark_processed`

Claude Code compacts naturally. The compact-reminder handler recovers context on resume. The health check restarts the session if it goes fully dead.

## What is not changed

- The handoff file format and location are unchanged
- Steps 1–4 (mark_processing, wind-down mode, agent drain, handoff write) are unchanged
- The compact-reminder handler itself is not modified here
- The health check restart mechanism is not modified here

## Functional patterns

Pure instructional change — the modification removes a side-effecting command and replaces it with a stop instruction. No new state is introduced.

## Relates to

Relates to #821. PR #823 was closed in favor of this approach.

## Tests run

- [x] Manual diff review — changes are confined to steps 5–6 and the Rules block of the `context_warning` handler section
- [ ] Live context-pressure test — blocked: requires the dispatcher running under real context load; safe to merge as the change is purely dispatcher instruction prose, not executable code

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -837,8 +837,8 @@ When a scheduled job finishes, `run-job.sh` calls `scheduled-tasks/post-reminder
    - Set internal flag: WIND_DOWN_MODE = True
    - Do NOT spawn new non-trivial subagents
    - For any new user messages: ack the user, call create_task to record the
-     request, and tell the user "I'm restarting shortly — will pick this up
-     immediately after." Do NOT delegate to a background subagent.
+     request, and tell the user "I'm compacting context shortly — will pick
+     this up immediately after." Do NOT delegate to a background subagent.
    - Quick inline responses (no subagent) are still OK.
 
 3. Drain in-flight agents:
@@ -853,15 +853,18 @@ When a scheduled job finishes, `run-job.sh` calls `scheduled-tasks/post-reminder
      "context_pct": <used_percentage from the message>,
      "pending_tasks": <list_tasks(status="pending") output>,
      "last_user_message": "<text of the last user-sourced message you processed>",
-     "note": "Graceful restart due to context pressure"
+     "note": "Graceful wind-down due to context pressure — compaction will recover"
    }
    (Create ~/lobster-workspace/data/ if it does not exist.)
 
 5. Send user (use the admin chat_id from your config / context):
-   "Context at {used_percentage}% — restarting gracefully. Back in a moment."
+   "Context at {used_percentage}% — entering wind-down mode. Handing off cleanly."
    (Substitute the `used_percentage` value from the `context_warning` message.)
 
-6. Bash("lobster restart")
+6. Stop the main loop — do NOT call `wait_for_messages()` again. Do NOT call
+   `lobster restart`. Write the handoff and go idle. Claude Code will compact
+   naturally; the compact-reminder handler will recover context. The health
+   check will restart the session if it goes fully dead.
 
 7. mark_processed(message_id)
 ```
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Overall assessment

The change is clean and correct. Removing `Bash("lobster restart")` from the `context_warning` handler directly addresses the root cause documented in #821: the dispatcher was killing its own session before the new one could start. Leaning on compaction + the health check is the right architecture.

---

## Review questions

**1. Does "stop the main loop / go idle" give enough guidance to stop calling `wait_for_messages()`?**

Mostly yes. Step 6 now reads:

> "Stop the main loop — do NOT call `wait_for_messages()` again. Do NOT call `lobster restart`. Write the handoff and go idle."

The phrasing "do NOT call `wait_for_messages()` again" is explicit and unambiguous. One edge case worth noting: step 7 is `mark_processed(message_id)`, which runs *after* step 6's "go idle" instruction. That ordering is intentional (you need to mark the message before stopping), but it means the dispatcher must make one more MCP call after being told to stop. The prose doesn't flag this, and a strict reading of "go idle" could lead a model to skip `mark_processed`. Consider adding a parenthetical to step 6 clarifying that `mark_processed` in step 7 is the *final* permitted call, or reordering step 7 before step 6.

**2. Is "the health check will restart the session if it goes fully dead" accurate?**

Yes — and the qualifier "fully dead" is doing important work here. From `health-check-v3.sh`:

- In `hibernate` mode (clean exit, wrapper polling), the health check escalates to RED only if: the tmux session is missing, the wrapper process dies, or user messages go stale past the 4-minute threshold.
- The health check does NOT restart on a clean compaction pause — there is active suppression via `COMPACTION_SUPPRESS_SECONDS=300` and `is_compaction_recent()`.

So the claim is accurate: the health check handles *genuine death* (dead tmux/wrapper/stuck inbox) but does not interfere with a clean compaction pause. The wording is appropriately calibrated.

**3. Is the user message wording ("entering wind-down mode. Handing off cleanly.") the right tone?**

Yes. It is more accurate than "restarting shortly" (which was both technically wrong and slightly alarming), and "Handing off cleanly" implies continuity without making a promise about timing. One small note: the phrase "entering wind-down mode" is internal jargon that may be opaque to users. "Context window is full — pausing to compact. I'll pick up where I left off." would be more user-legible, but this is cosmetic and should not block merge.

---

## Findings summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | Minor | Step 7 (`mark_processed`) ordering vs. step 6 ("go idle") is correct but could cause a model to skip it; consider a clarifying note |
| 2 | Cosmetic | "entering wind-down mode" is internal terminology; consider user-facing language |

No blocking issues. The core change — remove self-restart, lean on compaction — is architecturally sound and well-documented.

sayhar (commented):
PASS: Correct, minimal change. Removes the `lobster restart` call and replaces it with an idle stop, letting Claude Code's built-in compaction handle recovery.

**What changed:**
- Step 5 user message updated from "restarting" to "entering wind-down mode" — accurate
- Step 6 changed from `Bash("lobster restart")` to an idle stop instruction
- Handoff note updated: "Graceful restart" → "Graceful wind-down due to context pressure — compaction will recover"
- Rules block updated: `lobster restart` failure handling removed, replaced with "Do NOT call lobster restart" guidance

**Why this is correct:** The restart machinery introduced a polling dead window and implementation complexity. Compaction is already the right mechanism — the only gap was structured handoffs (addressed in #847/#862). This PR removes the wrong thing cleanly.

PR #852, opened 2026-03-25
feat: adaptive polling for bot-talk-poller (hot mode, issue #851)

### LINKED ISSUE #851: feat: adaptive polling for bot-talk-poller (hot mode on activity)
## Problem

The bot-talk-poller currently runs hourly (jobs.json schedule: \`0 * * * *\`). When AlbertLobster is actively replying, an hourly cadence means new messages can sit unread for up to 60 minutes. But running every 2 minutes all the time would launch a full Claude instance constantly, which is expensive and noisy.

## Proposed Solution

Adaptive polling via a "hot mode" state file and a companion fast-poller job:

1. **State file** at \`~/lobster-workspace/data/bot-talk-state.json\`:
   - \`last_message_ts\`: ISO timestamp of last seen AlbertLobster message
   - \`hot_mode\`: bool — true when activity was detected in the last poll
   - \`consecutive_empty_polls\`: counter for cooldown
   - \`hot_polls_count\`: counter for escalation (optional: go from 2min → 30sec if activity continues)

2. **bot-talk-poller** (hourly, \`0 * * * *\`) — existing job:
   - Reads and writes state file
   - Sets \`hot_mode=true\` when new messages seen
   - Clears \`hot_mode\` after N consecutive empty fast polls (e.g. 3)
   - Resets \`consecutive_empty_polls\` on each hourly run

3. **bot-talk-poller-fast** (every 2 min, \`*/2 * * * *\`) — new job:
   - Reads state file first
   - **Fast-exits** if \`hot_mode=false\` (no work, no Claude cost, no noise)
   - If \`hot_mode=true\`: runs full poll, updates state, increments \`consecutive_empty_polls\`
   - After 3 consecutive empty polls: sets \`hot_mode=false\`

## Implementation Steps

- [ ] Update \`bot-talk-poller.md\` workspace task to write/read state file
- [ ] Create \`bot-talk-poller-fast.md\` workspace task with fast-exit guard
- [ ] Add \`bot-talk-poller-fast.md\` to \`scheduled-tasks/tasks/\` in repo as canonical template
- [ ] Register \`bot-talk-poller-fast\` scheduled job (\`*/2 * * * *\`)
- [ ] Update \`bot-talk-poller\` schedule in jobs.json to hourly (\`0 * * * *\`) — already correct

## State File Schema

\`\`\`json
{
  "last_message_ts": "2026-03-25T15:00:00Z",
  "hot_mode": false,
  "consecutive_empty_polls": 0,
  "hot_mode_activated_at": null
}
\`\`\`

## Behavior Flow

```
Activity detected
      ↓
bot-talk-poller sets hot_mode=true
      ↓
bot-talk-poller-fast runs every 2 min
  ├─ hot_mode=true → full poll
  │   ├─ new messages → reset consecutive_empty_polls, notify Sahar
  │   └─ no messages → increment consecutive_empty_polls
  │       └─ >= 3 → set hot_mode=false
  └─ hot_mode=false → fast-exit (no work done)
```

sayhar:
## What and why

The bot-talk-poller was stuck at hourly cadence, meaning new AlbertLobster messages
could sit unread for up to 60 minutes. Running at 2-minute intervals all the time
would launch a full Claude instance every 2 minutes, burning cost and producing noise
even when there is nothing new to process.

This PR implements a two-job split with a shared state file so the system self-tunes:
fast when active, silent when not.

## How it works

A state file at `~/lobster-workspace/data/bot-talk-state.json` carries:
- `hot_mode`: bool — true when AlbertLobster activity was seen in the last poll
- `consecutive_empty_polls`: cooldown counter
- `last_message_ts`: timestamp of last seen message

**bot-talk-poller** (hourly) is the baseline. When it finds new messages it sets
`hot_mode=true`. When it finds silence and `consecutive_empty_polls >= 3`, it clears
`hot_mode`.

**bot-talk-poller-fast** (*/2 * * * *) is the companion. It reads the state file
first. If `hot_mode=false` — the common case — it calls `write_task_output` and
exits immediately, no API calls, no network traffic. If `hot_mode=true`, it runs the
full poll, manages the cooldown counter, and eventually self-cools after 6 minutes
of silence (3 x 2-minute polls).

```
AlbertLobster posts a message
      |
bot-talk-poller (hourly) detects it -> hot_mode=true, notifies Sahar
      |
bot-talk-poller-fast runs every 2 min:
  |-- hot_mode=true  -> full poll
  |   |-- new messages -> reset counter, notify
  |   +-- no messages -> increment consecutive_empty_polls
  |       +-- >= 3 (after 6 min) -> hot_mode=false
  +-- hot_mode=false -> fast-exit (no work done)
```

## What is in this PR

Two migrations added to `scripts/upgrade.sh`:

**Migration 34** — seeds `bot-talk-poller-fast.md` to the workspace tasks directory
(inline heredoc, since task files are workspace-only and blocked from the repo by
the pre-commit hook) and registers the */2 cron job in `jobs.json`, then syncs
crontab. Idempotent: skips if the task file already exists.

**Migration 35** — updates `bot-talk-poller`'s schedule from `*/2 * * * *` to
`0 * * * *` on installs where the pre-split schedule is still present. Also idempotent.

The workspace task files on this install were updated directly (not via upgrade.sh),
so migrations 34 and 35 are no-ops here and will run on any fresh `upgrade.sh` on
another install.

The `.githooks/security-allowlist.txt` entry covers the bot-talk server IP embedded
in the migration heredoc — it is a hostname, not a credential.

## Functional patterns

Pure state reads/writes (the state file is read-only on fast-exit path), immutable
updates via write-then-rename for atomicity, and fast-exit as a guard clause before
any side effects (the fast poller has zero network calls on the cold path).

## Tests run

- [x] `git commit` — pre-commit hook passes (task files blocked as expected; upgrade.sh allowed)
- [x] `git push` — pre-push security scanner passes after allowlist entry for server IP
- [ ] Live end-to-end (fast poller triggered by hot_mode) — job is registered and crontab
  is live; functional verification requires AlbertLobster to post. No code path changed
  for the existing hourly baseline.

**Blocked items needing attention before merge:** none — the live install already has the
workspace changes applied; migrations only run on other installs via `upgrade.sh`.

Closes #851

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -75,3 +75,8 @@ scripts/claude-wrapper.exp:Hardcoded token
 # throughout the codebase (scheduled-tasks/tasks/bot-talk-poller*.md). The default
 # value is a fallback when BOT_TALK_API_URL env var is unset — not a secret.
 scheduled-tasks/bot-talk-check-dispatch.sh:IP address
+
+# upgrade.sh Migration 34 embeds the bot-talk server address (46.224.41.108) inline
+# as task file content. This is a server hostname (not a credential); the same address
+# is already present in the workspace task file bot-talk-poller.md (not committed).
+scripts/upgrade.sh:IP address
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1826,6 +1826,191 @@ EOF
             fi
         fi
     fi
+    # Migration 50: Seed bot-talk-poller-fast task and register the */2 cron job
+    # The adaptive polling feature (issue #851) splits bot-talk polling into two jobs:
+    #   - bot-talk-poller (hourly): baseline poll + hot_mode state management
+    #   - bot-talk-poller-fast (*/2 * * * *): fast-exit unless hot_mode=true
+    # This migration seeds the task file and registers the fast job in jobs.json.
+    local fast_task_file="$WORKSPACE_DIR/scheduled-jobs/tasks/bot-talk-poller-fast.md"
+    local btp_jobs_file="$WORKSPACE_DIR/scheduled-jobs/jobs.json"
+    if [ ! -f "$fast_task_file" ] && [ -f "$btp_jobs_file" ]; then
+        cat > "$fast_task_file" << 'TASK_EOF'
+# Bot Talk Poller (Fast / Hot Mode)
+
+**Job**: bot-talk-poller-fast
+**Schedule**: Every 2 minutes (`*/2 * * * *`)
+
+## Context
+
+You are running as a scheduled task. This is the fast-polling companion to `bot-talk-poller`.
+It runs every 2 minutes but does real work only when `hot_mode=true` in the state file.
+When `hot_mode=false`, exit immediately without polling, notifying, or calling any APIs.
+
+## State File
+
+Read `~/lobster-workspace/data/bot-talk-state.json` before doing anything else.
+
+Schema:
+```json
+{
+  "last_message_ts": "2026-03-25T15:00:00Z",
+  "hot_mode": false,
+  "consecutive_empty_polls": 0,
+  "hot_mode_activated_at": null
+}
+```
+
+## Fast-Exit Rule
+
+**If `hot_mode` is `false` (or the file does not exist): call `write_task_output` with
+`status="success"` and `output="hot_mode=false, skipping"`, then stop. Do not poll the API.**
+
+This is the common case. Most 2-minute ticks will fast-exit with no network calls.
+
+## Authentication
+
+Read the token from `~/lobster-workspace/data/bot-talk-token.txt`:
+
+```python
+token = open(os.path.expanduser("~/lobster-workspace/data/bot-talk-token.txt")).read().strip()
+headers = {"X-Bot-Token": token}
+```
+
+If the token file is missing and hot_mode is true, log an error, call write_task_output
+with status="failed", and stop.
+
+## Instructions (hot mode only)
+
+When `hot_mode=true`:
+
+1. Poll `http://46.224.41.108:4242/messages` for messages from AlbertLobster newer than
+   `last_message_ts` in the state file (use the same tracking as `bot-talk-poller`).
+
+2. **If new messages found:**
+   - Post GitHub comments and send Telegram notification to Sahar (chat_id=8305714125)
+     following the same rules as `bot-talk-poller`
+   - Update `last_message_ts` in state file to the latest seen message timestamp
+   - Reset `consecutive_empty_polls` to 0 in state file
+   - Keep `hot_mode=true`
+
+3. **If no new messages:**
+   - Increment `consecutive_empty_polls` in state file
+   - If `consecutive_empty_polls >= 3`: set `hot_mode=false`
+   - Do NOT notify Sahar (no-op cycle)
+
+4. Write the updated state file back to disk atomically (write to a `.tmp` file, then
+   rename to the final path).
+
+5. Call `write_task_output` with the result.
+
+## Cooldown Logic
+
+```
+consecutive_empty_polls=0 → reset when new message seen
+consecutive_empty_polls=1 → still hot, keep polling
+consecutive_empty_polls=2 → still hot, keep polling
+consecutive_empty_polls=3 → set hot_mode=false, return to hourly cadence
+```
+
+Three consecutive empty 2-minute polls = 6 minutes of silence before cooldown.
+
+## Telegram Notification Rules
+
+Same as `bot-talk-poller`: only notify Sahar when there is genuinely new content.
+Never send "nothing new" messages. No-op cycles call write_task_output only.
+
+## Output
+
+When you complete your task, call `write_task_output` with:
+- job_name: "bot-talk-poller-fast"
+- output: Your results/summary (e.g., "hot_mode=false, skipped" or "2 new messages, notified")
+- status: "success" or "failed"
+TASK_EOF
+        substep "Seeded bot-talk-poller-fast.md task file"
+        migrated=$((migrated + 1))
+
+        # Register job in jobs.json if not already present
+        if command -v python3 &>/dev/null; then
+            python3 - "$btp_jobs_file" << 'PYEOF'
+import json, sys, os
+from datetime import datetime, timezone
+
+path = sys.argv[1]
+with open(path) as f:
+    data = json.load(f)
+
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: Adaptive polling implementation is sound. Notes below.

## Summary

Two-job split (hourly baseline + */2 fast companion) with shared state file. Logic is clean, both migrations are idempotent, fast-exit path is correct.

---

## Migration 34 — heredoc and Python snippet

**Heredoc:** Uses `<< 'TASK_EOF'` (single-quoted, no variable expansion) for the task file body, and `<< 'PYEOF'` for the Python block. Both delimiters close correctly. The nested heredoc pattern is valid bash — the outer shell function uses `-` style indentation elsewhere, but these heredocs are flush-left, which is fine.

**Python snippet (jobs.json registration):**
- Reads path from `sys.argv[1]` correctly.
- `data.setdefault("jobs", {})` handles a missing `"jobs"` key gracefully.
- Atomic write via `.tmp` + `os.replace` — good.
- One minor nit: `schedule_human` is set to `"Every 2 minutes"` (a human string), which is consistent with how other jobs look. No issue there.
- `python3` availability is checked with `command -v python3` before the block — correct guard.

**Idempotency:** Outer guard is `[ ! -f "$fast_task_file" ] && [ -f "$btp_jobs_file" ]`. The jobs.json registration adds a second inner guard (`if "bot-talk-poller-fast" not in data.get("jobs", {})`). Both layers are correct. Re-running migration 34 on an install where the task file already exists is a true no-op.

---

## Migration 35 — schedule downgrade

**Condition:** `current_btp_schedule == "*/2 * * * *"` (exact string match). This will only fire on installs where the schedule was set to `*/2` by the pre-split era. Any install with a manually customised schedule (e.g. `*/5`, `0 * * * *`) is untouched — correct behaviour.

**One observation:** `schedule_human` in the Python block is updated to `"0 * * * *"` (the raw cron string) rather than `"Hourly"` or `"Every hour"`. This is mildly inconsistent with the `bot-talk-poller` entry elsewhere in the codebase that likely uses a human string, but it is not a bug — the field is cosmetic.

---

## Cooldown hardcoded at 3

The threshold (`consecutive_empty_polls >= 3`) lives in the task file instructions (prose), not in a config file or env var. This means changing it requires editing the task file directly. For a personal assistant system where the operator can do exactly that, this is acceptable. A comment noting the tuning point would help future-you: e.g. `# Tune: increase threshold for slower cooldown`.

---

## Server IP in allowlist

`46.224.41.108` is correctly identified in `.githooks/security-allowlist.txt` as a hostname, not a credential. The allowlist entry is scoped to `scripts/upgrade.sh` and includes a clear explanation. Pre-push scanner will pass. No secret exposure.

---

## Minor items (non-blocking)

1. `schedule_human` in migration 35 should ideally be `"Hourly"` not `"0 * * * *"` for UI consistency.
2. The cooldown value (3) could have a one-line comment in the task file pointing out it is tunable.
3. Live end-to-end test is marked unchecked in the PR description — expected since it requires AlbertLobster to post. Acceptable for a state-machine change with no logic change on the existing hourly path.

---

**Overall:** Implementation is correct, migrations are safe and idempotent, fast-exit path is zero-cost on the cold path, security allowlist entry is appropriate. Ready to merge.

sayhar (commented):
NEEDS-WORK (conflict fix applied): Migration numbers 34 and 35 in the original branch conflicted with migrations 34, 35, 36 that were added to main after this branch was created (for gws cleanup, on-fresh-start hook, and sessions directory).

**Fix applied:** Rebased on main and renumbered:
- Migration 34 (seed bot-talk-poller-fast) → Migration 37
- Migration 35 (update poller schedule to hourly) → Migration 38

Branch has been force-pushed. The PR is now MERGEABLE with no conflicts.

**Code review (post-fix):** The adaptive polling design is sound:
- Fast-exit guard (hot_mode=false → immediate return, no API calls) correctly minimizes cost on the cold path
- State file is atomically updated (write-then-rename)
- Three-consecutive-empty-polls cooldown logic is well-documented
- Migrations are idempotent (check-before-apply pattern)
- Security allowlist entry for the IP is appropriate

One minor observation: the task file embedded in the migration heredoc has different behavior than the one in `scheduled-tasks/tasks/bot-talk-poller-fast.md` (the fast-poller task in the migration uses `consecutive_empty_polls >= 3` for cooldown, the one in the repo uses `>= 5`). The migration task is what runs on existing installs. PR #856 will overwrite both with the both-sides versions.

**Verdict: PASS (after fix), ready to merge.** Merge after #856 is merged (to get the both-sides task file content in the migration seed).

sayhar (commented):
🤖🦞 Lobster (reviewer):

**What this PR does:** Implements adaptive polling for bot-talk-poller by splitting into two jobs: (1) `bot-talk-poller` (hourly baseline) that manages `hot_mode` state — when a new message is seen, sets `hot_mode=true`; (2) `bot-talk-poller-fast` (every 2 minutes) that fast-exits when `hot_mode=false` and does real polling only when hot. State is persisted in `~/lobster-workspace/data/bot-talk-state.json`. Adds upgrade.sh Migrations 50 and 51 to seed the fast task file, register it in jobs.json, and downgrade the baseline poller from `*/2` to hourly.

**Issues found:**

1. **Hardcoded IP in task file content:** The bot-talk server IP (`46.224.41.108`) is embedded in the fast task file content that Migration 50 writes. This is already allowlisted in `security-allowlist.txt`. Acceptable.

2. **Hot mode cooldown is 3 empty polls = 6 minutes:** With a 2-minute polling interval, 3 consecutive empty polls means 6 minutes of silence before returning to hourly mode. This is a reasonable balance.

3. **State file atomic write:** The task description instructs the agent to write atomically (tmp + rename). Since this is instruction to a Claude agent rather than code, there's no guarantee it will follow it correctly. A helper script would be more reliable.

4. **Migration 51 uses `python3` not `uv run`:** Same issue as #902 — acceptable in upgrade.sh context.

5. **`bot-talk-state.json` initialization:** The fast task says 'if the file does not exist, fast-exit'. The baseline poller presumably creates this file. There's a startup ordering dependency: if the fast poller runs before the baseline has ever run, it safely no-ops. Correct behavior.

6. **No migration to remove the old `*/2` cron entry:** Migration 51 updates `jobs.json` schedule, and the sync-crontab script will update the actual crontab on next run. This is the correct pattern — no direct crontab manipulation needed.

**Verdict: PASS** — smart design that avoids API spam while maintaining low-latency when active. The fast-exit path (most common case) is the right optimization. Migrations are correct.

PR #857, opened 2026-03-25
fix: reset WFM staleness clock on each processed message (#694)

### LINKED ISSUE #694: fix: raise COMPACTION_SUPPRESS_SECONDS from 300 → 420 to cover multi-subagent post-compaction chains
🤖🦞 Lobster (ops):

## Incident Summary

On 2026-03-18 at 17:52 UTC, Lobster's health check triggered a restart because `wait_for_messages` was reported as 727s stale.

**Root cause:** Context compaction fired at 17:44:17. The dispatcher then processed a chain of 3 back-to-back subagent results (through 17:51:52) — a total of ~12 minutes of continuous processing — without returning to the main `wait_for_messages` loop. The compaction suppression window expired at ~17:49 (~5 minutes after compaction), leaving a ~3 minute window where the health check could legitimately fire, even though the dispatcher was actively working (just busy with subagent results, not stale).

## Mechanism

Context compaction is expensive in two ways:
1. The compaction event itself takes time.
2. Post-compaction, the dispatcher often resumes with a backlog of queued subagent results and processes them serially, each of which takes time.

The current suppression window only accounts for the compaction event itself, not the post-compaction processing chain. When 3+ subagent results are queued, the total "busy" time after compaction can easily exceed 300s.

## Where the Constant Lives

`COMPACTION_SUPPRESS_SECONDS` is defined at line 69 of `scripts/health-check-v3.sh`:

```bash
COMPACTION_SUPPRESS_SECONDS=300      # 5 minutes - skip stale-inbox check after a compaction event
```

The same value is mirrored in the test suite at `tests/smoke/test_health_check.py` line 52:

```python
COMPACTION_SUPPRESS_SECONDS = 300
```

## Proposed Fix

Raise `COMPACTION_SUPPRESS_SECONDS` from **300 → 420** (7 minutes) or **480** (8 minutes) to cover the realistic worst-case of a post-compaction chain of 3–4 subagent results.

Changes required:
- `scripts/health-check-v3.sh` line 69: `COMPACTION_SUPPRESS_SECONDS=420`
- `tests/smoke/test_health_check.py` line 52: `COMPACTION_SUPPRESS_SECONDS = 420`
- `tests/test-health-check-boot-grace.sh` line 107: `COMPACTION_SUPPRESS_SECONDS=420`
- Update comments in both files to reflect the new duration.

## Alternative

A more robust fix would be to reset the suppression clock on each processed message (not just on compaction), so that any sustained busy period keeps the health check suppressed. But raising the constant is a simple, low-risk interim fix.

sayhar:
## What was broken

The health check's WFM freshness check had a single freshness signal: the mtime of the claude-heartbeat file, touched at the start of each `wait_for_messages` call. When the dispatcher processed a long batch of messages — for example, 20 cron pings arriving back-to-back — it would stay busy draining the batch without returning to `wait_for_messages`. If this busy period outlasted the suppression window, the health check would see the WFM heartbeat as stale and trigger a restart, even though the dispatcher was actively working.

This is the "long batch exhausts suppression" failure mode described in issue #694.

## The fix

Add a second freshness signal: `last_processed_at` in `lobster-state.json`, written on every successful `mark_processed` call. The health check now uses `max(wfm_heartbeat_mtime, last_processed_at)` as the effective freshness epoch.

As long as the dispatcher is completing messages (calling `mark_processed`), the health check stays green — regardless of how long since it last called `wait_for_messages`. Only a dispatcher that stops processing AND stops calling WFM triggers a restart.

**Why per-message is better than raising the constant:**
- Raising `COMPACTION_SUPPRESS_SECONDS` provides a fixed window that future workloads can still exceed. Per-message heartbeat is self-scaling: batches of 100 messages keep the clock reset throughout, indefinitely.
- The constant approach is opaque — a reviewer cannot tell from the value alone whether it covers the worst case. Per-message heartbeat makes the invariant explicit: "active processing = not stale."
- A dispatcher that genuinely hangs mid-batch (no `mark_processed` calls) is still caught within `WFM_STALE_SECONDS`, whereas a longer constant would delay detection.

## Flow change

```
Before:
  effective_freshness = mtime(claude-heartbeat)
  if (now - effective_freshness) > WFM_STALE_SECONDS → RED

After:
  effective_freshness = max(mtime(claude-heartbeat), last_processed_at epoch)
  if (now - effective_freshness) > WFM_STALE_SECONDS → RED
```

## What changed

- `src/mcp/inbox_server.py`: new `_update_lobster_state_fields()` helper (merge-write, preserves all other state fields: mode, compacted_at, catchup_started_at, etc.); called from `handle_mark_processed()` and the atomic mark_processed path in `handle_send_reply()`
- `scripts/health-check-v3.sh`: `check_wfm_freshness()` reads `last_processed_at` from `lobster-state.json` and uses the more recent of the two signals
- `tests/smoke/test_health_check.py`: 4 new D5 tests

**What is not changed:** `COMPACTION_SUPPRESS_SECONDS`, `WFM_STALE_SECONDS`, and all other suppression windows are unchanged. The fix is additive — it extends the freshness signal, not the thresholds.

## Functional patterns

`_update_lobster_state_fields` is a pure read-merge-write: reads the existing state dict, updates only the specified keys, atomically replaces the file. Fail-open: errors are logged but never propagate to the caller, so a failed heartbeat write never blocks `mark_processed`.

## Tests run

- [x] `uv run pytest tests/smoke/test_health_check.py -v` — 9 passed (4 new D5 tests + 5 existing), 1 pre-existing failure in `test_maintenance_flag_causes_clean_exit` (FileExistsError unrelated to this change; fails identically on `main`)
- [ ] Full integration test — blocked: requires a running Lobster instance with live Telegram credentials

**Blocked items needing attention before merge:** none — the pre-existing test failure is not caused by this PR.

Closes #694

🤖🦞 Lobster (engineer)

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -828,13 +828,23 @@ check_outbox_drain() {
 }
 
 # Check 6: wait_for_messages freshness
-# Checks the mtime of the claude-heartbeat file, which inbox_server.py touches
-# at the start of every wait_for_messages call. If the file hasn't been updated
-# within WFM_STALE_SECONDS, the main loop is presumed stuck (e.g. infinite
-# loop, hung tool call, or Claude exit without wrapper noticing). Suppressed
-# during hibernation and the compaction window.
+# The dispatcher is considered "fresh" if EITHER:
+#   (a) the claude-heartbeat file was touched recently — inbox_server.py touches
+#       it at the start of every wait_for_messages call, OR
+#   (b) last_processed_at in lobster-state.json was updated recently — written
+#       by inbox_server.py on every successful mark_processed call (issue #694).
+#
+# Using both signals prevents a spurious restart when the dispatcher is actively
+# processing a long batch of messages (e.g. 20 cron pings) without returning to
+# wait_for_messages.  Before this fix, a long batch could exhaust the suppression
+# window mid-batch, triggering a false-positive health-check restart even though
+# the dispatcher was busy and healthy.
+#
+# The effective freshness timestamp is max(wfm_heartbeat_mtime, last_processed_at).
 #
 # Gracefully skips the check if the heartbeat file does not exist (fresh install).
+# Gracefully ignores a missing or unparseable last_processed_at (field absent on
+# older installs before this change was deployed).
 #
 # Returns: 0=GREEN (fresh or skipped), 2=RED (stale)
 check_wfm_freshness() {
@@ -850,16 +860,43 @@ check_wfm_freshness() {
         return 0
     fi
 
+    # Read the per-message heartbeat written by mark_processed (issue #694).
+    # Use whichever signal is more recent as the effective freshness epoch.
+    local last_processed_epoch=0
+    if [[ -f "$LOBSTER_STATE_FILE" ]]; then
+        local last_processed_at
+        last_processed_at=$(python3 -c "
+import json, sys
+try:
+    d = json.load(open('$LOBSTER_STATE_FILE'))
+    print(d.get('last_processed_at', ''))
+except Exception:
+    print('')
+" 2>/dev/null)
+        if [[ -n "$last_processed_at" ]]; then
+            last_processed_epoch=$(date -d "$last_processed_at" +%s 2>/dev/null) || last_processed_epoch=0
+        fi
+    fi
+
+    # Effective freshness = most recent of the two signals
+    local effective_last
+    if [[ "$last_processed_epoch" -gt "$last_heartbeat" ]]; then
+        effective_last="$last_processed_epoch"
+        log_info "WFM freshness: using last_processed_at signal (more recent than WFM heartbeat)"
+    else
+        effective_last="$last_heartbeat"
+    fi
+
     local now age
     now=$(date +%s)
-    age=$(( now - last_heartbeat ))
+    age=$(( now - effective_last ))
 
     if [[ $age -gt $WFM_STALE_SECONDS ]]; then
-        log_error "RED: wait_for_messages stale — heartbeat last updated ${age}s ago (threshold: ${WFM_STALE_SECONDS}s)"
+        log_error "RED: dispatcher stale — last activity ${age}s ago (threshold: ${WFM_STALE_SECONDS}s, wfm=${last_heartbeat}, last_processed=${last_processed_epoch})"
         return 2
     fi
 
-    log_info "WFM freshness OK: last wait_for_messages ${age}s ago"
+    log_info "WFM freshness OK: last dispatcher activity ${age}s ago"
     return 0
 }
 
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1071,6 +1071,46 @@ def _write_lobster_state(state_file: Path = None, mode: str = "active") -> None:
             pass
 
 
+def _update_lobster_state_fields(fields: dict, state_file: Path = None) -> None:
+    """Atomically merge one or more fields into lobster-state.json.
+
+    Reads the current state, updates the given fields, and writes back via
+    temp-then-rename so readers always see a consistent snapshot.  Other
+    fields (mode, compacted_at, catchup_started_at, etc.) are preserved.
+
+    This is a non-destructive merge — callers supply only the keys they want
+    to update.  Unlike _write_lobster_state, which replaces the entire file,
+    this function is safe to call from any context that must not clobber
+    concurrently-written lifecycle fields.
+
+    Failures are logged but not raised — the caller (mark_processed) must
+    not fail just because the heartbeat write failed.
+    """
+    if state_file is None:
+        state_file = LOBSTER_STATE_FILE
+    try:
+        existing: dict = {}
+        if state_file.exists():
+            try:
+                existing = json.loads(state_file.read_text())
+            except (json.JSONDecodeError, OSError):
+                existing = {}
+        existing.update(fields)
+        content = json.dumps(existing, indent=2) + "\n"
+        tmp = state_file.parent / f".lobster-state-{os.getpid()}.tmp"
+        try:
+            tmp.write_text(content)
+            tmp.rename(state_file)
+        except Exception as e:
+            log.error(f"Failed to update lobster state fields {list(fields)}: {e}")
+            try:
+                tmp.unlink()
+            except Exception:
+                pass
+    except Exception as e:
+        log.error(f"_update_lobster_state_fields unexpected error: {e}")
+
+
 @server.list_tools()
 async def list_tools() -> list[Tool]:
     """List available tools."""
@@ -3615,6 +3655,12 @@ async def handle_send_reply(args: dict) -> list[TextContent]:
                         _db_persist_inbound(json.loads(dest.read_text()))
                     except Exception as _db_exc:
                         log.warning(f"[DB] inbound persist failed for {mid}: {_db_exc}")
+                # Per-message WFM heartbeat: reset the WFM staleness clock so
+                # a long message batch does not exhaust the suppression window
+                # and trigger a spurious health-check restart (issue #694).
+                _update_lobster_state_fields(
+                    {"last_processed_at": datetime.now(timezone.utc).isoformat()}
+                )
             else:
                 mark_info = f" | ⚠️ message {mid} not found for mark_processed"
                 log.warning(f"Atomic mark_processed: message not found: {mid}")
@@ -3803,6 +3849,16 @@ async def handle_mark_processed(args: dict) -> list[TextContent]:
             _db_persist_inbound(json.loads(dest.read_text()))
         except Exception as _db_exc:
             log.warning(f"[DB] inbound persist failed for {message_id}: {_db_exc}")
+
+    # Write a per-message heartbeat so the health check can distinguish a
+    # dispatcher that is actively draining a long message batch from one that
+    # is genuinely stuck.  The WFM freshness check uses the more recent of the
+    # WFM heartbeat file and this timestamp, so a burst of 20+ cron pings
+    # processed without returning to wait_for_messages does not exhaust the
+    # suppression window and trigger a spurious restart (issue #694).
+    _update_lobster_state_fields(
+        {"last_processed_at": datetime.now(timezone.utc).isoformat()}
+    )
 
     log.info(f"Message processed: {message_id}")
     return [TextContent(type="text", text=f"✅ Message marked as processed: {message_id}")]
--- a/tests/smoke/test_health_check.py
+++ b/tests/smoke/test_health_check.py
@@ -344,3 +344,175 @@ def test_maintenance_flag_causes_clean_exit(tmp_path: Path) -> None:
         f"stdout: {result.stdout!r}\n"
         f"stderr: {result.stderr!r}"
     )
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PR #857 — fix: reset WFM staleness clock on each processed message

## Summary

The fix is correct and well-targeted. The dual-signal approach is sound: using \`max(wfm_heartbeat_mtime, last_processed_at)\` as the effective freshness epoch is the right abstraction, and the fail-open design ensures a failed heartbeat write never blocks \`mark_processed\`.

---

## Review by area

### 1. \`_update_lobster_state_fields()\` — merge-write safety

The read-modify-write pattern is correct and preserves all existing fields. One issue worth noting: the tmp filename uses \`os.getpid()\`, which is the same across all coroutines in the asyncio process. If two MCP tool calls (e.g. a batch drain with concurrent \`send_reply\` + \`mark_processed\` calls) were to execute this function concurrently, they would both write to the same \`.lobster-state-{pid}.tmp\` file and the second \`tmp.rename()\` would clobber the first.

In practice this is not a new hazard: \`_write_lobster_state\` already uses the same \`os.getpid()\` naming scheme, and asyncio cooperative scheduling means two coroutines can only interleave at \`await\` boundaries — \`_update_lobster_state_fields\` is synchronous (no \`await\`), so it runs atomically within a single event-loop turn. The risk is real only if the function were ever called from a thread executor. It is not today.

**Verdict:** safe as written; worth a comment in the function noting the pid-based assumption.

Contrast: \`record-catchup-state.sh\` uses \`state_path + ".tmp"\` (a fixed name), which is actually more vulnerable to concurrent writes from separate shell processes. The new helper is at least as safe as the existing precedent.

### 2. \`check_wfm_freshness()\` — bash integer comparison with \`last_processed_epoch=0\`

The fallback is correct. When \`last_processed_at\` is absent or unparseable, \`last_processed_epoch\` stays 0. The comparison is:

\`\`\`bash
if [[ "$last_processed_epoch" -gt "$last_heartbeat" ]]; then
    effective_last="$last_processed_epoch"
else
    effective_last="$last_heartbeat"
fi
\`\`\`

Since \`last_heartbeat\` is always a positive Unix epoch (stat succeeded or we returned early), \`0 -gt last_heartbeat\` is always false, so \`effective_last\` correctly falls back to the WFM signal. The D5d test covers this path directly.

One minor nit: the \`python3\` inline call to parse the JSON uses \`open()\` without explicitly closing the file handle. Not a real issue in CPython (GC collects it immediately), but a \`with open(...) as f:\` form would be more idiomatic.

**Verdict:** logic is correct; the zero-fallback works exactly as described.

### 3. Both \`mark_processed\` paths updated

Confirmed. The \`_update_lobster_state_fields\` call appears in:
- \`handle_mark_processed()\` (standalone \`mark_processed\` tool)
- The atomic path inside \`handle_send_reply()\` (when \`message_id\` is provided)

The non-atomic path (message not found) correctly does **not** write the heartbeat, since no message was actually processed. That's the right behavior.

---

## Tests

The four D5 tests cover the cases well:
- **D5a**: WFM-only signal stays GREEN (backward compatibility)
- **D5b**: both signals stale returns RED (prevents masking a genuine hang)
- **D5c**: stale WFM + recent \`last_processed_at\` returns GREEN (the core regression test)
- **D5d**: absent \`last_processed_at\` + fresh WFM returns GREEN (upgrade path)

The pre-existing \`test_maintenance_flag_causes_clean_exit\` failure is confirmed unrelated — it is a pytest \`tmp_path\` fixture collision on repeated runs, not a logic regression.

---

1. The log message in the RED case now emits raw epoch integers: \`wfm=1742927..., last_processed=1742927...\`. Human-readable ISO timestamps would aid on-call triage. Low priority.
2. \`uv.lock\` omission is noted and acceptable given the stated reason. Should be committed in a follow-up if the venv is stabilized.

---

## Verdict

**Approve.** The implementation is correct, the tests are thorough, and the fail-open pattern is appropriate. No blocking issues found.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is sound and well-scoped. The dual-signal approach is a clean solution to the long-batch exhaustion problem. All four focus areas check out with one minor observation noted below.

---

## Focus area 1: `_update_lobster_state_fields()` — read-modify-write safety

The implementation is correct for its stated goal. It reads, merges, and atomically replaces via temp-then-rename using a PID-namespaced temp file, which prevents torn reads and is safe against concurrent readers.

The concurrent-writer race (e.g. `on-compact.py` or `record-catchup-state.sh` writing simultaneously) is acknowledged in the PR description as pre-existing. The PID-namespaced temp filename avoids temp-file collisions between concurrent callers. The last-writer-wins outcome of a concurrent write race is acceptable here because `last_processed_at` is a monotonically advancing timestamp — even if an older write wins the race, the worst outcome is a slightly stale value, not corruption. Other fields (`mode`, `compacted_at`, `catchup_started_at`) are faithfully round-tripped through the merge, so no field is silently dropped.

The fail-open error handling is correct: exceptions are logged but not raised, so a failed heartbeat write never blocks `mark_processed`. This is the right tradeoff.

**Verdict: no issue.**

---

## Focus area 2: `check_wfm_freshness()` — bash integer comparison when `last_processed_epoch=0`

The variable is initialized to `0` before the conditional block:

```bash
local last_processed_epoch=0
```

The comparison that follows is:

```bash
if [[ "$last_processed_epoch" -gt "$last_heartbeat" ]]; then
```

When `last_processed_at` is absent or unparseable, `last_processed_epoch` stays `0`. Since `last_heartbeat` is the epoch-second mtime of the heartbeat file (a large positive integer), `0 -gt last_heartbeat` is false, so `effective_last` falls back to `last_heartbeat`. This is correct.

The `date -d` parse failure is also handled: `|| last_processed_epoch=0` resets to zero on parse error. The `python3` subprocess failure (non-zero exit) is handled by `2>/dev/null` suppression and the outer `[[ -n "$last_processed_at" ]]` guard — if python prints nothing or exits non-zero, `last_processed_at` is empty and the epoch stays `0`.

**One observation (non-blocking):** the `python3` subprocess is invoked on every health-check cycle. The health check runs frequently; spawning a Python interpreter each time adds a small but real latency overhead (~50–100 ms). A `jq` call or a `grep/sed` one-liner would be faster. This is a pre-existing pattern in the script (other checks use similar python3 subshells), so it's consistent — just worth noting for a future cleanup.

**Verdict: logic is correct. Performance observation is non-blocking.**

---

## Focus area 3: Both `mark_processed` paths have the heartbeat call

Confirmed both paths call `_update_lobster_state_fields`:

**Path 1 — `handle_mark_processed()`** (standalone tool call):
```python
_update_lobster_state_fields(
    {"last_processed_at": datetime.now(timezone.utc).isoformat()}
)
```
This is placed after the DB persist attempt and before the return, so it runs on every successful mark_processed call.

**Path 2 — atomic path inside `handle_send_reply()`** (when `message_id` is passed to `send_reply`):
```python
_update_lobster_state_fields(
    {"last_processed_at": datetime.now(timezone.utc).isoformat()}
)
```
This is placed inside the `if dest.exists():` branch that confirms the message was found and moved to processed, which is the correct guard — it only fires when the atomic mark actually succeeded.

**Verdict: both paths confirmed present and correctly positioned.**

---

## Focus area 4: Pre-existing `test_maintenance_flag_causes_clean_exit` failure

Not caused by this PR. The diff adds only new test functions and makes no changes to existing test infrastructure. The failure (FileExistsError) is independent of the new D5 tests and of the code changes in `inbox_server.py` and `health-check-v3.sh`.

**Verdict: correctly attributed as pre-existing.**

---

## Tests

The four D5 tests (D5a–D5d) are well-designed and cover the important cases:
- D5a: WFM heartbeat alone is sufficient (backward compat)
- D5b: both signals stale → RED (prevents false green on a genuinely stuck dispatcher)
- D5c: WFM stale but `last_processed_at` recent → GREEN (the core regression test)
- D5d: `last_processed_at` absent, WFM fresh → GREEN (pre-upgrade compat)

The test harness correctly extracts `check_wfm_freshness` from the shell script and exercises it in isolation. Coverage is appropriate for a D5 category.

---

## Minor items (non-blocking)

1. **`uv.lock` not committed** — acknowledged in the PR as a spurious unrelated change. Should be reverted or explicitly excluded before merge, but does not affect correctness.
2. **`python3` subprocess per health-check cycle** — noted above; consistent with existing patterns, fine to defer.
3. No migration entry is needed: `last_processed_at` is written lazily on the next `mark_processed` call after deploy, and the health check handles the absent-field case gracefully (falls back to WFM heartbeat). The upgrade path is seamless.

---

**Overall: PASS.** The implementation is correct, the failure mode is properly closed, both heartbeat paths are updated, and the backward-compatibility case is handled and tested. Ready to merge after the `uv.lock` situation is resolved.

sayhar (commented):
PASS: Solid, well-tested fix for the long-batch exhaustion failure mode. All 4 D5 tests pass. The pre-existing test_maintenance_flag_causes_clean_exit failure is a FileExistsError unrelated to this change (confirmed on main).

**What works:**
- `_update_lobster_state_fields()` is a clean pure read-merge-write helper. The atomic rename ensures readers see a consistent state snapshot.
- Fail-open behavior is correct: a failed heartbeat write never blocks `mark_processed`. The dispatcher must not fail for infrastructure reasons.
- Both call sites covered: `handle_mark_processed()` AND the atomic mark_processed path in `handle_send_reply()` — no gap.
- Health check uses `max(wfm_heartbeat_mtime, last_processed_at)` correctly.
- Backward-compatible: `last_processed_epoch=0` default when field is absent means old installs keep working.

**D5 test coverage:**
- D5a: WFM heartbeat alone sufficient (backward compat) ✓
- D5b: Both signals stale → RED (genuinely stuck dispatcher caught) ✓
- D5c: WFM stale but last_processed recent → GREEN (the core fix) ✓
- D5d: last_processed absent, WFM fresh → GREEN (upgrade path) ✓

**Branch rebased on main** (clean, no conflicts). Force-pushed.

sayhar (commented):
🤖🦞 Lobster (reviewer): SOAK INCONCLUSIVE — follow-up required before PASS

## Soak result (sayhar/lobster-research#26)

WFM staleness events recurred during the soak at 03:00, 08:12, 08:24, and 08:36 UTC — 4 false-positive health-check restarts. The soak report verdict: **INCONCLUSIVE**.

## Root cause of soak failures

PR #857 correctly fixes the long-batch exhaustion case (mark_processed resets the clock while messages are flowing). However, it does not address **zero-activity periods**: when no messages arrive, neither the WFM heartbeat nor last_processed_at advances. After WFM_STALE_SECONDS, the health check fires — even though the dispatcher is alive and blocked in wait_for_messages().

The 03:00 UTC restart occurred ~50 minutes after the last message, during a quiet overnight window. The 08:xx restarts were similar.

## Filed issue #949

Issue #949 (bug: WFM staleness fires during zero-message periods) documents the gap and proposes Option C as the likely fix: use inbox drain signal (empty inbox + stale WFM = healthy quiet, not dead). This is additive to #857 and doesn't require MCP changes.

## Status of this PR

PR #857 is a **genuine improvement** — it correctly prevents false-positive restarts during active processing. It should not be reverted. However, it is not a complete fix until the quiet-period case is also addressed.

**Recommendation:** merge #857 to local-dev (already done), do not merge to main until #949 is resolved or until the combined fix (this PR + inbox-drain signal) is validated in a soak.

PR #901, opened 2026-03-26
feat: inbox-staleness early warning script (3-min cron)

### LINKED ISSUE #883: feat: inbox-staleness early warning script (3-min cron)
🤖🦞 Lobster (ops): Sub-issue of #879

## Scope

Implement the shell script and cron entry that injects a staleness-warning message into the inbox at 3 minutes. This is the complete, standalone deliverable for #879.

## What to build

Create `scripts/inbox-staleness-warn.sh`:
- Finds the oldest unprocessed user message in `~/messages/inbox/` (filtering on user-originated sources, same list as `health-check-v3.sh` line ~107)
- If oldest message is 3+ minutes old AND no warning message for this run is already present in inbox/processing, write a `scheduled_reminder`-style JSON message with text: "Inbox stale for 3 minutes — call wait_for_messages now, or delegate to a subagent. Do not continue inline work."
- Use dedup logic identical to `post-reminder.sh` (check `reminder_type` field in inbox/ and processing/)
- Type: `scheduled_reminder`, reminder_type: `inbox_staleness_warn`

Register in `install.sh` via `cron-manage.sh`:
```
*/1 * * * * ~/lobster/scripts/inbox-staleness-warn.sh # LOBSTER-INBOX-STALENESS-WARN
```

Add migration entry to `scripts/upgrade.sh` (new cron).

## Acceptance criteria
- Script exists and is executable
- Dedup prevents double-injection
- Cron runs every 60s
- Dispatcher receives the reminder and treats it like a context_warning (wind-down protocol)
- install.sh and upgrade.sh updated

## Does NOT include
- Changes to health-check-v3.sh thresholds
- Changes to dispatcher handling (dispatcher already handles `scheduled_reminder` messages)

sayhar:
## Summary

The dispatcher sometimes gets stuck doing inline work instead of calling `wait_for_messages`. The existing health check catches this at 8+ minutes, but by then a restart is required. This adds an in-band warning that fires at 3 minutes — giving the dispatcher a chance to self-correct without a restart.

- New script: `scripts/inbox-staleness-warn.sh` — runs every minute via cron, injects a `scheduled_reminder` into the inbox when the oldest unprocessed user message is 3+ minutes old
- Dedup: mirrors `post-reminder.sh` pattern (checks both inbox/ and processing/ before injecting)
- Source/type filtering: mirrors `health-check-v3.sh` exactly — only genuine user-originated messages count (telegram/sms/signal/slack sources + message/photo/voice/etc types); subagent results, system messages, and existing reminders are excluded
- `scripts/upgrade.sh` migration 37: registers the cron entry idempotently on existing installs

The dispatcher already handles `scheduled_reminder` type as a context warning. No dispatcher changes needed.

## What to review closely

- `is_user_facing_source` and `is_user_facing_type` filter logic (lines 31-51) — must match health-check-v3.sh exactly or it will fire on non-user messages
- Dedup grep pattern — same pattern as `post-reminder.sh`; verify the field name `reminder_type` matches what the dispatcher expects
- Cron frequency: `*/1 * * * *` (every minute) — intentional; the 3-minute threshold is enforced by the script, not the cron schedule

## Concerns / known gaps

None — pure shell, no new dependencies, no behavior change to the dispatcher.

## Tests run

- [x] `bash -n scripts/inbox-staleness-warn.sh` — syntax OK
- [ ] Live cron test — blocked: requires production install with running Lobster instance

Closes #883

### DIFF
--- a/scripts/inbox-staleness-warn.sh
+++ b/scripts/inbox-staleness-warn.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+# inbox-staleness-warn.sh — Inject an inbox-staleness warning when the oldest
+# unprocessed user message has been waiting for 3+ minutes.
+#
+# Design: pure shell, no Python, no Claude. Safe to call from cron every minute.
+#
+# Filtering logic mirrors health-check-v3.sh lines ~107-114:
+#   - Only user-facing sources (telegram, sms, signal, slack) are counted.
+#   - Only genuine user-originated message types (message, photo, image, voice,
+#     audio, callback, text, document) are counted.
+#   - Internal queue messages (subagent_result, subagent_error, scheduled_reminder,
+#     system, compact, etc.) are excluded even when they carry source="telegram".
+#
+# Dedup logic mirrors post-reminder.sh:
+#   - Check both inbox/ and processing/ for an existing message with the same
+#     reminder_type. If found, exit silently without injecting a duplicate.
+
+set -euo pipefail
+
+INBOX_DIR="${HOME}/messages/inbox"
+PROCESSING_DIR="${HOME}/messages/processing"
+
+REMINDER_TYPE="inbox_staleness_warn"
+STALE_THRESHOLD_SECONDS=180   # 3 minutes
+
+# User-facing sources — matches health-check-v3.sh line ~108
+USER_FACING_SOURCES="telegram sms signal slack"
+
+# Genuine user-originated message types — matches health-check-v3.sh line ~114
+USER_FACING_TYPES="message photo image voice audio callback text document"
+
+mkdir -p "$INBOX_DIR"
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+is_user_facing_source() {
+    local source="$1"
+    local s
+    for s in $USER_FACING_SOURCES; do
+        [[ "$source" == "$s" ]] && return 0
+    done
+    return 1
+}
+
+is_user_facing_type() {
+    local type="$1"
+    local t
+    for t in $USER_FACING_TYPES; do
+        [[ "$type" == "$t" ]] && return 0
+    done
+    return 1
+}
+
+# ---------------------------------------------------------------------------
+# Dedup: skip if a staleness warning is already pending (inbox or processing).
+# ---------------------------------------------------------------------------
+if grep -rl "\"reminder_type\": \"${REMINDER_TYPE}\"" "$INBOX_DIR" "$PROCESSING_DIR" 2>/dev/null | grep -q .; then
+    exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# Scan inbox for user-facing messages and find the oldest.
+# ---------------------------------------------------------------------------
+now=$(date +%s)
+oldest_age=0
+
+while IFS= read -r -d '' f; do
+    # Parse source and type via jq; skip if missing or unparseable.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
PASS

**Correctness:** Filtering logic mirrors `health-check-v3.sh` lines 107-114 exactly (`USER_FACING_SOURCES`, `USER_FACING_TYPES`). Dedup grep pattern matches `post-reminder.sh` on `reminder_type` field. Both inbox/ and processing/ are checked before injecting.

**Edge cases:** Empty inbox exits cleanly (oldest_age stays 0, threshold check exits early). jq parse failures skip files gracefully. stat failures skip files too. mkdir -p at top prevents race on first run.

**Cron registration:** Migration uses `cron-manage.sh` idiom with `grep -q` guard — idempotent. Marker format matches existing cron entries.

**What I checked but couldn't verify without live environment:** That `stat -c %Y` works on the target OS (Linux — should be fine), and that the dedup grep regex `"reminder_type": "${REMINDER_TYPE}"` matches the exact JSON formatting the script writes (it does — same escaping used in both places).

**Verdict:** PASS — safe to merge. No dispatcher changes needed. Deploy and verify in cron after merge.

sayhar (commented):

**Summary:** Solid design and filter logic, but there is a migration number collision that must be resolved before merge.

**Correctness**
- The 3-minute threshold and 1-minute cron frequency are appropriate: a warning fires no earlier than 3 minutes and no later than 4 minutes after staleness begins.
- `is_user_facing_source` and `is_user_facing_type` correctly mirror `health-check-v3.sh`. The type check allows messages with an empty/missing `type` field to pass through (only skips if `type` is present AND not in the allowlist) — this matches the health check behaviour and is intentional.
- Dedup via `grep -rl` across both `inbox/` and `processing/` correctly mirrors `post-reminder.sh`.
- The injected message uses `"type": "scheduled_reminder"` which the dispatcher already handles. No dispatcher changes needed — confirmed correct.
- `set -euo pipefail` is appropriate for a cron script; the `|| true` guards in `stat` and `jq` calls prevent spurious exits on unreadable files.

**Safety**
- No PII exposure, no secrets. `chat_id: 0` on a system-injected message is the correct convention.
- File age is measured via `stat -c %Y` (mtime). This is correct for inbox messages since the file is written once on arrival, but worth noting: if the filesystem or cron daemon modifies mtime (e.g., periodic backups, `touch`), the age could be wrong. Acceptable risk for a warning-only path.

**Architecture**
- Fits Lobster patterns: pure shell, no Python, no Claude API, no new dependencies.
- The script is self-contained and safe to call concurrently (the dedup check prevents double-injection).

**Blocker: Migration number collision with PR #902**
Both this PR and PR #902 add "Migration 37" to `upgrade.sh`, branching off the same base commit (sha `6ad5616`). Whichever merges second will have a silent no-op migration (the marker check will succeed but the content will be for the other PR's migration). One of these PRs must be rebased and renumbered to Migration 38 before merge.

**Minor issues**
- `*/1 * * * *` is valid cron syntax but `* * * * *` is more conventional for "every minute" and avoids confusion. Either works; `* * * * *` is preferred by style guides.
- No live cron test was performed (acknowledged in the PR). The `bash -n` syntax check is necessary but not sufficient — specifically, the `jq` field extraction and the `stat -c %Y` flag should be verified on the target platform (Linux). These look correct for Linux; they would fail on macOS (`stat -f %m`), but production is Linux so this is fine.

**Required before merge:** Resolve the Migration 37 collision with PR #902.

PR #904, opened 2026-03-26
feat: IFTTT-style behavioral rules store with LRU cap enforcement (#853)

### LINKED ISSUE #853: Design IFTTT-style behavioral rules file with cap enforcement
## Problem

As Lobster accumulates user preferences and behavioral rules over time, there is no structured, bounded place to store them. Rules end up scattered across bootup files or implied by convention, making them hard to audit, modify, or reason about systematically.

## Proposed Solution

Introduce an IFTTT-style behavioral rules file — a flat list of "if X then Y" rules that Lobster loads at startup. Each rule is a short, declarative statement:

```
If the user asks about a meeting, check the calendar first.
If a task has been open for more than 7 days, surface it in the daily digest.
If a subagent fails twice on the same task, notify the user instead of retrying.
```

**Cap enforcement:** The file has a hard cap (e.g. 30 entries). When the cap is hit:
- Option A: Oldest entries are pruned automatically (with a log entry noting what was removed)
- Option B: The user is notified and asked to prune manually

The cap prevents unbounded growth, keeps the file scannable at startup, and forces periodic review of rules that may have become stale.

## Behavior at Startup

- The dispatcher reads the rules file as part of the tiered bootup sequence (see epic #847)
- Rules are loaded into working context and applied throughout the session
- No rules are silently dropped mid-session — if the cap is hit, it is handled at write time, not read time

## Open Questions

- Should rules be structured (YAML/TOML frontmatter) or freeform prose?
- Who can write rules — only the user, or can Lobster propose rules based on observed patterns?
- How does Lobster detect when a rule is no longer relevant or contradicts a newer rule?
- Cap threshold: 30 entries is a guess — should this be user-configurable?

## Parent Epic

Part of #847 (session notes and structured memory continuity).

sayhar:
## What this solves

Before this PR, Lobster had no structured place to accumulate behavioral rules ("if X then Y"). Rules either lived as prose in bootup files (hard to audit and reason about) or existed only implicitly in LLM context. There was no mechanism for Lobster to record a recurring behavioral pattern and apply it reliably in future sessions.

This PR introduces a bounded YAML rules index that Lobster manages autonomously. Rules accumulate silently over time; the user never writes or reviews them directly. The file is committed to \`lobster-user-config\` so it survives machine migrations.

## Architecture

The rules file is an **index only**. It maps condition → memory DB entry:

```yaml
rules:
  - id: "check-calendar-on-meeting"
    condition: "The user asks about a meeting or scheduling"
    action_ref: "mem_abc123"
    enabled: true
```

The memory DB (existing infrastructure) holds the actual behavioral content and all access metadata (access_count, last_accessed_at, etc.). The dispatcher resolves \`action_ref\` via \`memory_search\` at rule-match time; the DB call naturally increments access tracking. There is no in-file metadata and no direct Python script calling.

## What changed

**\`src/utils/ifttt_rules.py\`** (new, reworked)

Pure-function rules store library. Schema uses \`condition\` + \`action_ref\` instead of \`trigger\`/\`action\`. Metadata fields removed entirely. Key functions:

- \`add_rule()\` — append or replace a rule; drops oldest (head) entries when over cap
- \`remove_rule()\` — remove by ID; no-op if not found
- \`get_enabled_rules()\` — filter to \`enabled: true\` rules only
- \`find_rule()\` — lookup by ID
- \`format_rules_for_context()\` — render enabled rules as \`[id] IF <condition> THEN lookup <action_ref> in memory DB\` lines
- \`load_rules()\` / \`save_rules()\` — file I/O with graceful degradation and atomic write-then-rename

Removed: \`touch_rule\`, \`prune_lru\`, \`_lru_key\`, \`_now_iso\`. LRU is now owned by the memory DB, not the YAML file.

Cap: \`MAX_RULES = 100\`. When exceeded, oldest rules (by insertion order) are dropped from the head.

**\`memory/canonical-templates/ifttt-rules.yaml\`** (new)

Minimal 4-field schema: \`id\`, \`condition\`, \`action_ref\`, \`enabled\`. No metadata. Seeded to \`lobster-user-config/memory/canonical/\` on install and via Migration 47.

**\`install.sh\`**

Extended both template-seeding blocks to copy \`.yaml\` files.

**\`scripts/upgrade.sh\`** — Migration 47

Seeds \`ifttt-rules.yaml\` into \`lobster-user-config/memory/canonical/\` on existing installs.

**\`.claude/sys.dispatcher.bootup.md\`**

Updated IFTTT Behavioral Rules section:
- Schema example updated to \`condition\`/\`action_ref\`
- Removed all references to calling \`src/utils/ifttt_rules.py\` directly
- Added **batch lookup** guidance: resolve all matching \`action_ref\` values in a single DB query per turn, not one at a time
- Clarified that all rule access goes through MCP tools, not Python imports
- Cap description updated: drops oldest by insertion order (DB owns LRU)

## Functional patterns

- **Pure functions**: \`add_rule\`, \`remove_rule\`, \`get_enabled_rules\`, \`find_rule\`, \`format_rules_for_context\` are all pure
- **Immutability**: all mutation functions return new lists and dicts
- **Side effects at the boundary**: only \`load_rules()\` and \`save_rules()\` perform I/O
- **Graceful degradation**: missing or malformed YAML returns empty list; dispatcher continues normally

## Areas to review closely

- **No MCP tool wrapper yet**: the library is used by tests and any future Python subagent; the dispatcher reads/writes the YAML directly for now. Full MCP surface is a follow-up.
- **Batch lookup is documented but not enforced**: the dispatcher is responsible for batching \`memory_search\` calls. This is documented in the bootup section.
- **Cap semantics**: drops oldest (insertion order, from head). LRU enforcement is delegated to the memory DB's natural de-prioritization of unaccessed entries.

## Tests run

- [x] \`uv run --no-sync pytest tests/unit/test_ifttt_rules.py -v\` — 36 passed, 0 failed
- [ ] Broader unit suite — not run in this pass (dependency resolution issue in worktree env unrelated to this PR)
- [ ] Live dispatcher startup with rules file present — blocked: requires a running Lobster session

Closes #853


sayhar (commented):
🤖🦞 Lobster (reviewer):

**What this PR does:** Implements IFTTT-style behavioral rules — a bounded (100-rule cap), LRU-pruned flat list of 'if X then Y' rules stored in `~/lobster-user-config/memory/canonical/ifttt-rules.yaml`. Includes: `src/utils/ifttt_rules.py` (pure functions for load/save/add/prune), YAML template seeded via install.sh and upgrade.sh (Migration 47), and dispatcher bootup documentation for reading and applying rules at session start.

**Issues found:**

1. **`ifttt_rules.py` is a pure utility library — no MCP tool exposes it yet.** The dispatcher bootup docs say 'use `add_rule()` from `src/utils/ifttt_rules.py`' but the dispatcher runs as a Claude Code session, not as Python code. The dispatcher cannot call Python functions directly — it would need to spawn a subagent or have an MCP tool that wraps this. This is a design gap: the rules are described as 'managed autonomously by Lobster' but there's no MCP surface for Lobster to actually invoke the Python API. In practice the dispatcher would need to read/write the YAML file directly or via Bash.

2. **Access metadata update on rule match:** The bootup says 'update `last_accessed_at` and `access_count` after saving any rule updates'. This is described as a 'background file write' but requires the dispatcher to write YAML directly to disk or call a subagent. Neither is mentioned explicitly as the mechanism.

3. **LRU prune implementation:** `prune_lru()` correctly sorts by `(last_accessed_at ASC, access_count ASC)` and keeps the cap most-recently-used rules in original order. The logic is correct.

4. **`save_rules()` atomic write:** Uses `tempfile.NamedTemporaryFile` + `os.replace` — correct pattern for atomic YAML writes.

5. **`load_rules()` graceful degradation:** Returns empty list if file is absent or malformed — correct per the 'never fail' requirement.

6. **YAML template:** Starts with `rules: []` — correct. Schema comments are thorough and accurate.

**Verdict: PASS with note** — the Python library is well-implemented. The gap between the Python API and the dispatcher's actual execution model (Claude Code, not Python) means the dispatcher will likely read/write the YAML file directly in practice. The library is still useful for tests and any future Python subagent that manages rules. The documentation should clarify that the dispatcher uses direct file I/O, not the Python API.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: IFTTT behavioral rules store is well-structured and safe to deploy.

## Summary

Solid implementation. The architecture (YAML index + memory DB for content) is clean. Pure-function core with I/O isolated at the boundary. Graceful degradation throughout. 36 unit tests pass for the library and the four MCP handlers are covered by the new test file.

## What looks good

- **`src/utils/ifttt_rules.py`**: Pure functions, immutable return values, atomic write-then-rename in `save_rules()`. Cap enforcement correct: `add_rule()` drops from head (`updated[n_dropped:]`), `save_rules()` drops from head (`rules[-cap:]`). Both consistent.
- **`src/mcp/inbox_server.py` — four handlers**: Input validation, memory-unavailable error paths, and the `_resolve_action_ref()` fallback chain are all solid. `_generate_rule_id()` uses a UUID suffix to prevent collisions on identical conditions.
- **`vector_memory.py` `get()` and `delete()`**: `get()` delegates to `_fetch_events()` correctly. `delete()` manually removes from `events_vec` (correct — no trigger for that table) and relies on the existing `events_ad` FTS5 trigger for `events_fts` cleanup.
- **Migration 47** and install seeding both guard with `[ ! -f "$dest" ]` — won't clobber existing rules files on upgrade.
- **`sys.dispatcher.bootup.md`**: Startup step 2b added correctly, batch-lookup guidance documented, MCP-tools-only access enforced in prose.

## Minor issues (non-blocking)

1. **`save_rules()` docstring inconsistency**: The docstring says "FIFO — drops tail entries beyond cap" but the code does `rules[-cap:]` which drops head (oldest). The code is correct; the docstring description is wrong. Low impact since the test `test_caps_before_writing` validates the correct behavior.

2. **No `update_rule` / enable-disable MCP tool**: The `add_rule()` library function has update-in-place semantics when the same ID is passed, but there's no MCP-exposed path to disable a rule (`enabled: false`) without deleting and re-adding it. Not a blocker for this PR, but worth a follow-up issue.

3. **Bootup doc forward-reference**: Section "IFTTT Behavioral Rules" appears after "Startup Behavior" but is referenced in step 2b with "See 'IFTTT Behavioral Rules' section below" — it's actually above in the rendered doc. Cosmetic only.

## Verdict

PASS — deploy to local-dev. `/dogfooded 904` clearance needed after 2-hour soak.

PR #926, opened 2026-03-26
fix: make sync_crontab async and warn on pre-init audit_log calls

### LINKED ISSUE #919: bug: sync_crontab() calls subprocess.run() synchronously inside async handler — blocks event loop
## Summary

`handle_create_scheduled_job`, `handle_update_scheduled_job`, and `handle_delete_scheduled_job` in `inbox_server.py` all call `sync_crontab()`, which internally calls:

```python
subprocess.run(["crontab", ...], timeout=10)
```

This is a **synchronous blocking call inside an async handler**. It blocks the asyncio event loop for up to 10 seconds while waiting for the `crontab` subprocess to finish.

## Impact

- During any scheduled-job create/update/delete operation, the entire MCP server is unresponsive for up to 10 seconds
- Any in-flight `wait_for_messages` or other async MCP tool calls are blocked
- If the `crontab` binary hangs or is slow (e.g. under load), the stall extends to the full 10-second timeout

## Root cause

`sync_crontab()` is called directly as `await sync_crontab()` but is defined as a sync function using `subprocess.run()`, not `asyncio.create_subprocess_exec()`.

## Fix

Replace `subprocess.run()` in `sync_crontab()` with `asyncio.create_subprocess_exec()` and make the function `async`, or delegate to a thread pool via `asyncio.get_event_loop().run_in_executor()`.

### LINKED ISSUE #921: bug: audit_log() silently no-ops if init_audit_log() was never called — reliability.py audit log is effectively dead on startup
## Summary

In `src/mcp/reliability.py`, the `audit_log()` function checks `_AUDIT_LOG_PATH` before writing:

```python
_AUDIT_LOG_PATH: Path | None = None

def audit_log(event: str, data: dict | None = None) -> None:
    if _AUDIT_LOG_PATH is None:
        return  # silently drops the event
    ...
```

`_AUDIT_LOG_PATH` is only set by `init_audit_log()`. In `inbox_server.py`, `init_audit_log()` is called during server startup, but any `audit_log()` call that fires **before** `init_audit_log()` completes (e.g. from module-level initialization, import-time side effects, or early startup code) silently drops the event with no warning.

## Related

`IdempotencyTracker` and `CircuitBreaker` in `reliability.py` are also never wired into any actual call paths (see issue #826), so the audit trail for idempotency and circuit-breaker events is effectively non-functional.

## Impact

- Reliability audit events are silently dropped during early startup
- Post-mortem investigation of startup-phase issues has no audit trail

## Fix

- Add a warning log when `audit_log()` is called with `_AUDIT_LOG_PATH = None` rather than silently returning
- Or eagerly initialize `_AUDIT_LOG_PATH` from an environment variable at module import time

sayhar:
## What this fixes

Two independent reliability gaps in `inbox_server.py` and `reliability.py`, grouped because they are both small and both in the MCP server layer.

**#919 — sync_crontab() blocked the event loop for up to 10 seconds**

Every scheduled-job create, update, or delete called `sync_crontab()`, which internally called `subprocess.run()` — a synchronous blocking call inside an async MCP handler. During those 10 seconds the entire MCP server was unresponsive: `wait_for_messages`, inbox checks, and any other in-flight tool calls were all stalled. The fix replaces `subprocess.run()` with `asyncio.create_subprocess_exec` + `await`, so the subprocess runs concurrently with other event loop work. All three call sites (`handle_create_scheduled_job`, `handle_update_scheduled_job`, `handle_delete_scheduled_job`) are updated to `await sync_crontab()`.

**#921 — audit_log() silently dropped events before init_audit_log() was called**

`audit_log()` returned early with no log output when `_AUDIT_LOG_PATH` was `None`. Any reliability event that fired before `init_audit_log()` completed (early startup, import-time side effects) vanished without a trace. The fix emits a `WARNING` log entry instead, so dropped events are detectable in the server log.

## Functional patterns

Pure functions: `sync_crontab` has no side effects beyond the subprocess it awaits. The warning path in `audit_log` is a single conditional with a single observable output (log line) — no mutation of shared state.

## What is not changed

The three async handler functions themselves are unchanged except for the added `await`. The `sync_crontab` function's external contract (returns `(bool, str)`) is unchanged.

## Tests run
- [ ] `uv run pytest tests/` — skipped: test suite requires a live MCP server environment not available in this subagent context
- [x] Manual diff review: all three `sync_crontab()` call sites confirmed changed to `await sync_crontab()`, return type preserved
- [x] Import check: `asyncio` already imported at line 13 of `inbox_server.py`; `logging` added to `reliability.py`

Closes #919
Closes #921

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -5129,22 +5129,29 @@ def validate_job_name(name: str) -> tuple[bool, str]:
     return True, ""
 
 
-def sync_crontab() -> tuple[bool, str]:
-    """Sync jobs.json to crontab. Returns (success, message)."""
+async def sync_crontab() -> tuple[bool, str]:
+    """Sync jobs.json to crontab. Returns (success, message).
+
+    Uses asyncio.create_subprocess_exec so the event loop remains responsive
+    while the crontab subprocess runs (fixes: sync was blocking for up to 10s).
+    """
     sync_script = _REPO_DIR / "scheduled-tasks" / "sync-crontab.sh"
     try:
-        result = subprocess.run(
-            [str(sync_script)],
-            capture_output=True,
-            text=True,
-            timeout=10
+        proc = await asyncio.create_subprocess_exec(
+            str(sync_script),
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
         )
-        if result.returncode == 0:
-            return True, result.stdout
+        try:
+            stdout, stderr = await asyncio.wait_for(proc.communicate(), timeout=10)
+        except asyncio.TimeoutError:
+            proc.kill()
+            await proc.communicate()
+            return False, "Sync script timed out"
+        if proc.returncode == 0:
+            return True, stdout.decode()
         else:
-            return False, result.stderr or "Sync failed"
-    except subprocess.TimeoutExpired:
-        return False, "Sync script timed out"
+            return False, stderr.decode() or "Sync failed"
     except Exception as e:
         return False, str(e)
 
@@ -5219,7 +5226,7 @@ Keep output concise. The main Lobster instance will review this later.
     save_scheduled_jobs(data)
 
     # Sync to crontab
-    success, msg = sync_crontab()
+    success, msg = await sync_crontab()
     if not success:
         return [TextContent(type="text", text=f"Job created but crontab sync failed: {msg}")]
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**#919 (sync_crontab async):** The conversion from `subprocess.run()` to `asyncio.create_subprocess_exec` is correct. `asyncio.wait_for` with a 10s timeout mirrors the original `timeout=10` on `subprocess.run`. On timeout, `proc.kill()` is called followed by `await proc.communicate()` to reap the zombie — this is the correct cleanup sequence for asyncio subprocesses. All three call sites updated to `await sync_crontab()`. Return type `tuple[bool, str]` unchanged. `asyncio` was already imported.

**#921 (audit_log warning):** `log.warning()` is the right severity — it's not a crash, but it is unexpected data loss. The `logging.getLogger(__name__)` pattern is consistent with other modules in the codebase. The warning includes the tool name so log grep is useful.

One note: the three skipped automated tests are a known gap for this codebase — the MCP server test suite requires a running server environment. The changes are narrow enough that manual diff review is sufficient to verify correctness.

PR #927, opened 2026-03-26
fix: preserve state fields on wake and seed atomically in install

### LINKED ISSUE #923: bug: wake_claude_if_hibernating() overwrites lobster-state.json with minimal fields — loses compacted_at and other state
## Summary

In `lobster_bot.py`, `wake_claude_if_hibernating()` writes a new state file when waking from hibernation:

```python
state_data = {"mode": "active", "woke_at": datetime.now(timezone.utc).isoformat()}
tmp.write_text(json.dumps(state_data, indent=2))
tmp.rename(LOBSTER_STATE_FILE)
```

This overwrites the entire `lobster-state.json` with only two fields. Any existing fields — `compacted_at`, `last_restart_at`, `booted_at`, `catchup_started_at` — are silently lost.

This is the same class of bug as #825 (which covers `_write_lobster_state()` in `inbox_server.py` and `write_state()` in `claude-persistent.sh`), but from a third location: the Telegram bot's wake handler.

## Impact

- After a wake-from-hibernate triggered by an incoming Telegram message, `compacted_at` is cleared
- The health check's `is_compaction_recent()` check returns false (no timestamp), which can suppress the post-compact catchup grace period or cause incorrect restart decisions
- `booted_at` is lost, causing the health check boot grace period to not apply

## Fix

Read the existing state file first, merge the new fields, then write back atomically:
```python
existing = {}
try:
    existing = json.loads(LOBSTER_STATE_FILE.read_text())
except Exception:
    pass
existing.update({"mode": "active", "woke_at": ..., "booted_at": ...})
tmp.write_text(json.dumps(existing, indent=2))
tmp.rename(LOBSTER_STATE_FILE)
```

This should be part of a broader fix for the state-clobber pattern tracked in #825.

### LINKED ISSUE #924: bug: install.sh container setup seeds lobster-state.json with bare echo — not atomic, not JSON-safe
## Summary

In `install.sh`, the container setup section seeds `lobster-state.json` with:

```bash
echo '{"mode": "active", "booted_at": "'"$(date -u +%Y-%m-%dT%H:%M:%SZ)"'"}' > "$state_file"
```

This has two problems:

1. **Not atomic**: `>` redirects to the file directly. If the process is interrupted mid-write, the file is left truncated/corrupt. Should use a tmp file + `mv`.

2. **No JSON validation**: The `date` command output is interpolated directly into a JSON string. On some systems or locales, the date output could contain characters (quotes, backslashes) that break JSON validity, though in practice the `-u +%Y-%m-%dT%H:%M:%SZ` format is safe.

## Secondary issue

This is also the same pattern as #825 and #923 — the state file is written with a hardcoded minimal structure, meaning any future field additions to the state file need to be manually tracked across all three write sites (`install.sh`, `lobster_bot.py`, `claude-persistent.sh`).

## Fix

Consolidate all lobster-state.json writes through a single helper function or script that:
1. Reads existing state
2. Merges new fields
3. Writes atomically via tmp+rename

Or at minimum, use a heredoc+tmp+mv pattern in `install.sh`.

sayhar:
## What this fixes

Both bugs are instances of the same pattern: `lobster-state.json` written with a bare minimal struct, clobbering existing fields.

**#923 — wake_claude_if_hibernating() wiped compacted_at, booted_at, last_restart_at**

When an incoming Telegram message woke Lobster from hibernation, the bot wrote `{"mode": "active", "woke_at": "..."}` directly to `lobster-state.json`. Any field not in that literal — `compacted_at`, `booted_at`, `last_restart_at` — was silently deleted.

Downstream impact: after a hibernate wake, `is_compaction_recent()` returned false (missing `compacted_at`), which could suppress the post-compact catchup grace period or trigger incorrect restart decisions. `booted_at` being gone meant the boot grace period no longer applied.

The fix reads the existing state file first, merges the wake fields in (`mode` + `woke_at`), then writes back atomically. This is the same read-merge-write pattern already used by `_write_lobster_state()` in `inbox_server.py`.

**#924 — install.sh seeded the state file with a bare `echo >` redirect**

A `>` redirect to the target file directly is not atomic: if the process is interrupted mid-write the file is left truncated. Both seeding sites in `install.sh` are updated to `printf > tmp && mv` (using `$$` in the tmp name to avoid collisions).

## Before / after (wake path)

```
Before:
  wake event
    └─ write {mode, woke_at}  ← clobbers everything else

After:
  wake event
    └─ read existing state (best-effort)
    └─ merge {mode, woke_at} into existing
    └─ atomic rename write
```

## Functional patterns

Immutability at the merge point: `existing.update(...)` produces the final dict before any I/O. The read and write are the only side effects, and the write is atomic.

## What is not changed

The systemd restart path and all process-spawning logic in `wake_claude_if_hibernating()` are untouched. Only the state-file write block changes.

## Tests run
- [x] Manual diff review: `existing.update(...)` pattern confirmed; tmp file uses `os.getpid()` for uniqueness; `mv` replaces `>`  in both install.sh sites

Closes #923
Closes #924

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -343,7 +343,10 @@ if [ "$CONTAINER_SETUP" = true ]; then
     # triggering a restart loop before Claude has had time to initialize.
     state_file="$MESSAGES_DIR/config/lobster-state.json"
     if [ ! -f "$state_file" ]; then
-        echo '{"mode": "active", "booted_at": "'"$(date -u +%Y-%m-%dT%H:%M:%SZ)"'"}' > "$state_file"
+        # Write atomically via tmp+rename to prevent a truncated file on interrupt (#924)
+        _state_tmp="${state_file}.tmp.$$"
+        printf '{"mode": "active", "booted_at": "%s"}\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "$_state_tmp"
+        mv "$_state_tmp" "$state_file"
         info "  Seeded lobster-state.json with initial booted_at timestamp"
     fi
     success "Directories created"
@@ -1074,7 +1077,10 @@ rm -f "$MESSAGES_DIR/config/agents.db" "$WORKSPACE_DIR/data/agents.db"
 # triggering a restart loop before Claude has had time to initialize.
 STATE_FILE="$MESSAGES_DIR/config/lobster-state.json"
 if [ ! -f "$STATE_FILE" ]; then
-    echo '{"mode": "active", "booted_at": "'"$(date -u +%Y-%m-%dT%H:%M:%SZ)"'"}' > "$STATE_FILE"
+    # Write atomically via tmp+rename to prevent a truncated file on interrupt (#924)
+    _STATE_TMP="${STATE_FILE}.tmp.$$"
+    printf '{"mode": "active", "booted_at": "%s"}\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "$_STATE_TMP"
+    mv "$_STATE_TMP" "$STATE_FILE"
     info "  Seeded lobster-state.json with initial booted_at timestamp"
 fi
 
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -579,9 +579,20 @@ def wake_claude_if_hibernating() -> None:
         # This prevents restart storms: even if spawn fails, the state is no longer
         # "hibernate", so the health check won't skip its safety net.
         try:
-            state_data = {"mode": "active", "woke_at": datetime.now(timezone.utc).isoformat()}
+            # Read existing state first so we preserve fields like compacted_at,
+            # booted_at, and last_restart_at — same pattern as _write_lobster_state()
+            # in inbox_server.py. Overwriting with a bare dict was bug #923.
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**#923 (wake state clobber):** The read-then-merge pattern is correct. `existing.update({...})` mutates the dict in place after the read, which is fine — the dict is local to the try block. The `except Exception: pass` on the read is appropriate: if the state file is missing or corrupt, starting from an empty dict is the right fallback (same logic as `_read_lobster_state_data()` in inbox_server.py). The `tmp.rename(LOBSTER_STATE_FILE)` atomic rename is unchanged and correct.

**#924 (install.sh atomic seed):** `printf` with `> tmp && mv tmp dest` is the correct atomic pattern. Using `1114001` (shell PID) in the tmp filename avoids collisions between concurrent install runs. The `if [ ! -f "$state_file" ]` guard is unchanged — this only seeds on first install. Both occurrences (lines ~344 and ~1075) are fixed.

No behavior changes in the wake-from-hibernate flow itself — the process spawning, systemd restart, and stale-state detection are all untouched.

PR #928, opened 2026-03-26
fix: lock _processing_files check-and-add to eliminate outbox TOCTOU race

### LINKED ISSUE #922: bug: OutboxHandler._schedule_processing has a TOCTOU race — two watchdog events for the same file can both pass the _processing_files guard
## Summary

In `lobster_bot.py`, `OutboxHandler._schedule_processing()` guards against duplicate processing with a set:

```python
if filepath not in _processing_files:
    _processing_files.add(filepath)
    asyncio.run_coroutine_threadsafe(self.process_reply(filepath), main_loop)
```

However, `on_created` and `on_modified` are both wired to `_schedule_processing()`. When an atomic write (write-to-tmp then rename) lands in the outbox, inotify may fire both an `IN_MOVED_TO` event (caught as `on_created`) and an `IN_CLOSE_WRITE` / `IN_MODIFY` event. Both events arrive in the watchdog thread.

Since both events arrive on the **same thread** (watchdog uses a single event-dispatch thread per observer), the race is unlikely in practice. But `asyncio.run_coroutine_threadsafe` schedules the coroutine on the asyncio loop, and `process_reply` opens the file with a `0.5s` sleep before reading, meaning two scheduled coroutines could both be in-flight at the `await asyncio.sleep(0.5)` point when `_processing_files.discard()` at the `finally` block of the first coroutine hasn't run yet.

More concretely: if two watchdog events fire close together and both pass the `if filepath not in _processing_files` check before either coroutine has had a chance to call `_processing_files.add()`, the same file will be processed twice — resulting in a duplicate Telegram message.

## Impact

- Duplicate outbox messages delivered to Telegram
- `os.remove(filepath)` will raise `FileNotFoundError` on the second coroutine (caught silently by the outer try/except)

## Fix

Use a threading lock around the check-and-add:
```python
with _processing_lock:
    if filepath in _processing_files:
        return
    _processing_files.add(filepath)
```

Or deduplicate at the coroutine level by checking file existence before parsing.

sayhar:
## What this fixes

`OutboxHandler._schedule_processing()` was guarding against duplicate file processing with a plain `if filepath not in _processing_files` check followed by `_processing_files.add(filepath)`. These two operations had no lock between them.

When an atomic write (write-to-tmp then rename) lands in the outbox, inotify can fire both `IN_MOVED_TO` (caught as `on_created`) and `IN_MODIFY` for the same file. Both events arrive on the watchdog thread. If both checked `_processing_files` before either had added the path, both would see "not present" and both would schedule `process_reply()` — resulting in two Telegram sends for the same outbox file.

The fix wraps the check and the add in a single `threading.Lock` critical section (`_processing_files_lock`). The asyncio coroutine dispatch is left outside the lock to avoid holding it during I/O scheduling.

## Before / after

```
Before:
  watchdog thread A          watchdog thread B
  if fp not in _pf: True     if fp not in _pf: True   ← both pass
  _pf.add(fp)                _pf.add(fp)
  schedule process_reply()   schedule process_reply()  ← duplicate send

After:
  watchdog thread A                   watchdog thread B
  acquire _processing_files_lock      (blocked)
  if fp in _pf: False
  _pf.add(fp)
  release lock                        acquire lock
                                      if fp in _pf: True → return
```

## Functional patterns

The lock makes the check-and-claim atomic — a single indivisible state transition from "not tracked" to "tracked". No shared mutable state is accessed outside the critical section.

## What is not changed

`process_reply()` itself is unchanged. The `_processing_files.discard()` in the `finally` block remains as-is — it runs from an asyncio coroutine, which is single-threaded, so no locking is needed there.

## Tests run
- [x] Manual diff review: lock acquired before check, released before coroutine dispatch; `_processing_files_lock` initialized alongside `_wake_lock`

Closes #922

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -380,6 +380,12 @@ main_loop = None
 # Tracks files currently being processed to prevent duplicate sends
 _processing_files: set[str] = set()
 
+# Lock serialising the _processing_files check-and-add in _schedule_processing.
+# Without this, two watchdog events for the same file (e.g. IN_MOVED_TO and
+# IN_MODIFY arriving close together) could both pass the `not in` guard before
+# either has added the path, causing duplicate Telegram delivery (#922).
+_processing_files_lock = threading.Lock()
+
 # Lock to prevent concurrent wake attempts (race condition: two simultaneous
 # incoming messages while hibernating should only trigger one Claude spawn)
 _wake_lock = threading.Lock()
@@ -1332,12 +1338,17 @@ class OutboxHandler(FileSystemEventHandler):
     def _schedule_processing(self, filepath):
         if filepath.endswith('.json') and not filepath.endswith('.tmp'):
             if bot_app and main_loop and main_loop.is_running():
-                if filepath not in _processing_files:
+                # Hold the lock for the full check-and-add so two watchdog
+                # events for the same file cannot both pass the guard before
+                # either has added the path (TOCTOU — fixes #922).
+                with _processing_files_lock:
+                    if filepath in _processing_files:
+                        return
                     _processing_files.add(filepath)
-                    asyncio.run_coroutine_threadsafe(
-                        self.process_reply(filepath),
-                        main_loop
-                    )
+                asyncio.run_coroutine_threadsafe(
+                    self.process_reply(filepath),
+                    main_loop
+                )
 
     def on_created(self, event):
         if event.is_directory:


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**#922 (TOCTOU lock):** The `threading.Lock` wrapping the check-and-add is the correct fix. The critical section is tight: only the `in` check and the `.add()` are inside the `with` block. The coroutine dispatch (`asyncio.run_coroutine_threadsafe`) is outside the lock, which is correct — there is no reason to hold the lock during I/O scheduling.

The `_processing_files.discard()` in the `process_reply` finally block runs from the asyncio event loop (single-threaded), so no lock is needed there — this is correctly left unchanged.

The new lock (`_processing_files_lock`) is defined adjacent to `_processing_files` and `_wake_lock`, which makes the locking intent clear.

One thing to note: the issue mentioned that the race is "unlikely in practice" because both watchdog events arrive on the same dispatch thread. The lock still closes the theoretical race correctly and adds no meaningful overhead.

Issue #827, opened 2026-03-25
fix: health-check-v3.sh calls bare python3 for DB queries and state reads, should use uv run


sayhar:
## Summary

`health-check-v3.sh` uses `python3 -c ...` in multiple places (state file reading, catchup timestamp parsing, auth token check) and `uv run python3 -c ...` in `count_active_subagents()` (line 1109).

The inconsistency matters: the lobster venv contains the `sqlite3` and `agents` packages that some of these calls depend on. Bare `python3` uses the system Python, which may not have the right dependencies installed. If the system Python is a different version from the venv or lacks a needed module, the health check silently returns fallback values (the `|| echo "0"` pattern).

The `count_active_subagents()` call correctly uses `uv run` but the other inline `python3 -c` calls (read_lobster_mode, is_catchup_active, is_boot_grace_period, write_boot_timestamp, check_auth_token) do not.

## Risk
Low on most systems (the state-file reads use stdlib only), but the auth token check parses credentials JSON and could silently skip the check if system Python is absent or misconfigured.

## Fix
Standardize: use bare `python3` only for stdlib-only calls, or use `uv run python3` throughout for consistency with the rest of the Lobster codebase (per CLAUDE.md convention).

PR #929, opened 2026-03-26
fix: serialize jobs.json last_run updates with flock and atomic write

### LINKED ISSUE #920: bug: run-job.sh updates jobs.json non-atomically — concurrent cron fires can corrupt the file
## Summary

`scheduled-tasks/run-job.sh` reads and writes `jobs.json` to update the `last_run` timestamp, but does so non-atomically.

The `jq` path writes to a `TMP_FILE` and then `mv`s it, which is atomic on the same filesystem. However, the fallback `python3` path writes directly to `jobs.json` without a tmp+rename:

```python
with open('$JOBS_FILE', 'w') as f:
    json.dump(data, f, indent=2)
```

Additionally, neither path uses file locking. If two scheduled jobs fire at the same second (or nearly so), both processes read the same `jobs.json`, both modify their own job's `last_run`, and both write back — the second write silently drops the first write's changes. This could corrupt the entire `jobs.json` if a write is interrupted mid-file.

## Impact

- `last_run` timestamps for some jobs may be silently dropped on concurrent execution
- In the worst case (write interrupted), `jobs.json` could be left in a truncated/corrupt state, breaking all scheduled job status reads

## Fix

1. Use `fcntl.flock()` or a `.lock` sentinel file to serialize concurrent `jobs.json` writes
2. Always use the tmp+rename pattern (not direct open-for-write)
3. Use `uv run` for the Python fallback path (consistency with issue #827)

sayhar:
## What this fixes

`run-job.sh` updates `jobs.json` to record `last_run` after each job fires, but did so non-atomically with no file locking.

**The race:** two cron jobs firing at the same second both read the same `jobs.json`, each writes back only their own job's `last_run` update, and the second write silently drops the first job's change. In the worst case, an interrupted write leaves `jobs.json` truncated and unreadable, breaking all scheduled job status reads.

**The python fallback was also non-atomic:** the `jq`-absent path used `open('$JOBS_FILE', 'w')` — a direct overwrite with no tmp file, meaning a crash mid-write corrupts the file. It also called bare `python3` rather than `uv run`.

## Fix

Wrap the entire read-modify-write in a `flock -x` exclusive lock on a `.lock` sentinel file using bash's subshell-fd pattern. The lock releases automatically when the subshell exits (whether via success or error). Both paths write via tmp+rename:

- `jq` path: already used `mktemp` + `mv` — no change needed there
- Python fallback: now uses `uv run` and `os.replace(tmp, jobs_file)` with a pid-unique tmp name

The lock file itself is `jobs.json.lock` and persists between runs (this is intentional — `flock` works on the file descriptor, not the file contents).

## Functional patterns

The subshell-fd flock pattern is declarative: the lock scope is defined structurally by the subshell boundary, not by explicit acquire/release calls. The python fallback is a pure transformation (read JSON, update one field, write JSON) with a single atomic side effect.

## What is not changed

The inbox-write section (lines 85–125) and all cron scheduling logic are untouched. Only the `last_run` update block (previously lines 129–148) is replaced.

## Tests run
- [x] Manual review: `flock -x 9` + `9>"$JOBS_LOCK"` pattern is standard bash; `os.replace` is atomic on POSIX; `uv run` used in python fallback

Closes #920

### DIFF
--- a/scheduled-tasks/run-job.sh
+++ b/scheduled-tasks/run-job.sh
@@ -126,25 +126,41 @@ PYEOF
 
 echo "[$START_ISO] Reminder posted for job: $JOB_NAME — dispatcher will spawn subagent" | tee -a "$LOG_FILE"
 
-# Update jobs.json last_run to reflect when the job was dispatched
+# Update jobs.json last_run to reflect when the job was dispatched.
+# Uses a lock file to serialise concurrent cron fires and always writes via
+# tmp+rename so an interrupted write never leaves jobs.json truncated (#920).
 if [ -f "$JOBS_FILE" ]; then
-    if command -v jq &> /dev/null; then
-        TMP_FILE=$(mktemp)
-        jq --arg name "$JOB_NAME" \
-           --arg last_run "$START_ISO" \
-           'if .jobs[$name] then .jobs[$name].last_run = $last_run else . end' \
-           "$JOBS_FILE" > "$TMP_FILE" && mv "$TMP_FILE" "$JOBS_FILE"
-    else
-        python3 -c "
-import json
-with open('$JOBS_FILE') as f:
+    JOBS_LOCK="${JOBS_FILE}.lock"
+    (
+        # Acquire exclusive lock (fd 9) — releases automatically when subshell exits
+        flock -x 9
+        if command -v jq &> /dev/null; then
+            TMP_FILE=$(mktemp)
+            jq --arg name "$JOB_NAME" \
+               --arg last_run "$START_ISO" \
+               'if .jobs[$name] then .jobs[$name].last_run = $last_run else . end' \
+               "$JOBS_FILE" > "$TMP_FILE" && mv "$TMP_FILE" "$JOBS_FILE"
+        else
+            uv run - \
+                "$JOBS_FILE" \
+                "$JOB_NAME" \
+                "$START_ISO" \
+                << 'PYEOF'
+import json, os, sys
+jobs_file = sys.argv[1]
+job_name  = sys.argv[2]
+last_run  = sys.argv[3]
+with open(jobs_file) as f:
     data = json.load(f)
-if '$JOB_NAME' in data.get('jobs', {}):
-    data['jobs']['$JOB_NAME']['last_run'] = '$START_ISO'
-    with open('$JOBS_FILE', 'w') as f:
+if job_name in data.get('jobs', {}):
+    data['jobs'][job_name]['last_run'] = last_run
+    tmp = jobs_file + '.tmp.' + str(os.getpid())
+    with open(tmp, 'w') as f:
         json.dump(data, f, indent=2)
-" 2>/dev/null || true
-    fi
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**#920 (jobs.json locking + atomic write):** The `flock -x 9` subshell-fd pattern is correct and idiomatic bash. The lock releases when the subshell exits regardless of success or error path — no risk of a leaked lock. The lock file (`jobs.json.lock`) persisting between runs is intentional and correct for `flock`-based serialization.

The `jq` path was already using `mktemp + mv` — no change needed there. The python fallback is now `uv run` (consistency with issue #827) and uses `os.replace(tmp, jobs_file)` with a pid-unique tmp filename, which is atomic on POSIX.

The `|| true` at the end of the subshell invocation means a `last_run` update failure is non-fatal (the job was already dispatched), which is the correct behavior — same as the original.

One note: the `JOBS_LOCK` variable could theoretically collide with an existing env var, but `.lock` is a local path that won't conflict in practice.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**What this PR does:** Fixes a race condition in `run-job.sh` where concurrent cron fires for the same job could corrupt `jobs.json`. Adds `flock -x` locking on a `.lock` file co-located with `jobs.json`, and switches to atomic tmp+rename writes so an interrupted write never leaves `jobs.json` truncated.

**Issues found:**

1. **Lock file cleanup:** The `.lock` file (`jobs.json.lock`) is created but never explicitly removed. `flock` on a lock file is safe even without cleanup (the lock is advisory and tied to the fd lifetime), but the file will accumulate on disk. Minor cosmetic issue — `flock` itself doesn't care.

2. **Python fallback uses `uv run -` (stdin mode):** The heredoc `PYEOF` piped to `uv run -` is correct. The script passes `jobs_file`, `job_name`, and `last_run` as `sys.argv` arguments rather than string interpolation into the Python source — this is the correct fix for shell injection risks present in the original code.

3. **`os.replace(tmp, jobs_file)` in Python fallback:** Atomic on POSIX; handles the case where `jq` is unavailable. Correct.

4. **Subshell for locking:** Using a subshell `( flock -x 9; ... ) 9>lock` correctly scopes the lock to the block and auto-releases on subshell exit. The `|| true` at the end prevents the outer script from failing if the subshell exits non-zero.

5. **No retry logic:** If `flock` times out (default: waits indefinitely), concurrent cron fires will serialize correctly. If a job takes very long, subsequent fires will queue up. This is fine for scheduled-job use.

**Verdict: PASS** — clean fix for the race condition. Atomic write + flock is the correct pattern. The `uv run` heredoc avoids string interpolation into Python. Already merged to local-dev.

sayhar:
Closing as superseded by #1105 (systemd timer rewrite). This PR fixes a race condition in jobs.json writes in run-job.sh. Since #1105 replaces the jobs.json + crontab backend entirely, the atomic write fix is no longer applicable to the new architecture.

PR #930, opened 2026-03-26
fix: read-merge-write in _write_lobster_state() to prevent field clobber

### LINKED ISSUE #825: bug: _write_lobster_state() clobbers compacted_at, catchup_started_at, last_restart_at on every hibernate/wake
## Summary

`_write_lobster_state()` in `src/mcp/inbox_server.py` (line 1050) always writes a fresh dict with only `mode` and `updated_at`, discarding all other fields in `lobster-state.json`.

This means any time the dispatcher goes to hibernate (or wakes), the following fields are silently wiped:
- `compacted_at` — health check suppression for compaction window
- `catchup_started_at` / `catchup_finished_at` — suppression for catchup subagent window
- `last_restart_at` — restart cooldown suppression
- `booted_at` — boot grace period

The health check depends on reading these fields from the same file. If any of them are present when hibernate is written, the next health-check run cannot see them, potentially triggering false-positive restarts.

Compare: `_reset_state_on_startup()` (line 716) correctly reads the existing dict and updates only the `mode` field before writing. `_write_lobster_state()` should do the same.

## Fix
Read existing fields, merge in the new `mode` and `updated_at`, then write back atomically — same pattern as `write_last_restart_at()` and `write_boot_timestamp()` in health-check-v3.sh.

sayhar:
## What this fixes

**#825 — _write_lobster_state() clobbered compacted_at, booted_at, last_restart_at on every hibernate/wake**

Every time the dispatcher went to hibernate (or returned from it), `_write_lobster_state()` wrote a fresh dict containing only `{mode, updated_at}`. Any field not in that literal — `compacted_at`, `booted_at`, `catchup_started_at`, `last_restart_at` — was silently deleted.

Downstream impact: the health check uses `compacted_at` for the compaction recency check, `booted_at` for the boot grace period, and `last_restart_at` for the restart cooldown. After any hibernate transition, all three suppression windows were gone, which could trigger false-positive restarts.

The fix reads the existing state file first (best-effort, ignores parse errors), merges in the new `mode` and `updated_at` fields, then writes back atomically via tmp+rename. This is the same read-merge-write pattern already used by `_reset_state_on_startup()` and by `wake_claude_if_hibernating()` in lobster_bot.py (fixed in #927).

## Before / after

```
Before:
  hibernate or wake
    └─ write {mode, updated_at}  ← clobbers compacted_at, booted_at, etc.

After:
  hibernate or wake
    └─ read existing state (best-effort)
    └─ merge {mode, updated_at} into existing
    └─ atomic rename write
```

## Relationship to other PRs

- #927 fixed the same pattern in `wake_claude_if_hibernating()` in `lobster_bot.py`
- This PR fixes the remaining instance: `_write_lobster_state()` in `inbox_server.py`
- Together they close all three known clobber sites (the third, `write_state()` in `claude-persistent.sh`, was tracked separately in #825 as a bash write — that path already reads and merges correctly)

## Functional patterns

The `existing.update(...)` call is the only mutation — it produces the final dict before any I/O. Read and atomic write are the only side effects.

## Tests run
- [x] Manual diff review: `existing.update(...)` pattern confirmed; read wrapped in try/except; tmp file uses `os.getpid()` for uniqueness

Closes #825

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1050,15 +1050,26 @@ def _read_lobster_state(state_file: Path = None) -> str:
 def _write_lobster_state(state_file: Path = None, mode: str = "active") -> None:
     """Atomically write Lobster state to state file.
 
+    Reads existing state first and merges new fields in so that caller-supplied
+    fields (compacted_at, booted_at, last_restart_at, catchup_started_at, etc.)
+    are never clobbered — the bug tracked in #825.  Only ``mode`` and
+    ``updated_at`` are unconditionally overwritten; all other fields survive.
+
     Uses write-to-temp-then-rename so readers never see a partial file.
     """
     if state_file is None:
         state_file = LOBSTER_STATE_FILE
-    data = {
+    # Read existing state so we preserve fields we are not updating (#825).
+    existing: dict = {}
+    try:
+        existing = json.loads(state_file.read_text())
+    except Exception:
+        pass
+    existing.update({
         "mode": mode,
         "updated_at": datetime.now(timezone.utc).isoformat(),
-    }
-    content = json.dumps(data, indent=2)
+    })
+    content = json.dumps(existing, indent=2)
     tmp = state_file.parent / f".lobster-state-{os.getpid()}.tmp"
     try:
         tmp.write_text(content)


sayhar (commented):

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is correct and closes the last known _write_lobster_state() clobber site. The read-merge-write pattern is already established in this codebase (used by _reset_state_on_startup() and wake_claude_if_hibernating() after #927). This PR applies the same pattern to the remaining instance.

## Key areas reviewed

**Read-merge-write correctness**: The existing.update({mode: ..., updated_at: ...}) pattern is correct — it overwrites only the two fields that _write_lobster_state() is meant to set, while preserving compacted_at, booted_at, catchup_started_at, last_restart_at, and any future fields.

**Error handling**: The try/except around the read is correct — parse failures fall back to writing only the new fields, which is worse than a merge but better than crashing. This is appropriate for a state-update path.

**Atomic write**: tmp+rename pattern is the same as other call sites in this file. Safe.

**PID-namespaced temp file**: As noted in PR #857's review, this is safe for asyncio (the function is synchronous, runs atomically within one event-loop turn). The pre-existing assumption holds.

**Relationship to #927**: Together these two PRs close all known clobber sites in Python. The bash path (claude-persistent.sh) already read-merges correctly per the issue.

- No unit tests run — acceptable given the subagent context limitation. The manual diff review is sufficient for a 10-line change applying an established pattern.
- Worth a follow-up to add a test for the merge behavior once a test harness is available.

**Verdict: PASS.** Minimal, correct, follows established patterns. Ready for merge after Sahar sign-off.

PR #959, opened 2026-03-27
feat: block personal names (sahar/drew) in pre-commit and pre-push hooks

sayhar:
## Summary

- Adds case-insensitive personal name scanning to `.githooks/pre-commit` (staged diff) and `.githooks/pre-push` (full file content)
- Blocks commits and pushes containing standalone occurrences of \"sahar\" or \"drew\"
- Allowlists legitimate non-name uses: \"sayhar\" (GitHub username), \"andrew\", \"withdrew\", and other words where \"drew\" appears inside a larger word
- Adds a `# noname` inline suppression comment for any line that needs a one-off exception
- Adds `scripts/upgrade.sh` to the security allowlist for pre-existing historical references that predate this hook

## How it works

**Pre-commit** (scans staged diff, added lines only):
- Parses `git diff --cached` output line-by-line
- Skips `.githooks/*` files (hook source contains the patterns as regex)
- Skips lines with `# noname` inline annotation
- Uses negative lookbehind `(?<![a-z])sahar(?![a-z])` and `(?<![a-z])drew(?![a-z])` to match only standalone names
- \"sayhar\" is excluded via a secondary grep after the sahar match

**Pre-push** (scans full file content of changed files):
- Added `NAME_PATTERNS` array alongside existing `PII_PATTERNS` and `SECURITY_PATTERNS`
- Same regex logic, plus per-pattern allowlist regex for words like \"andrew\"/\"withdrew\"
- Skips `.githooks/*` files

**Bypass mechanisms**:
- `git commit --no-verify` — skip pre-commit
- `git push --no-verify` — skip pre-push
- `# noname` on a line — suppress just that line
- `.githooks/security-allowlist.txt` — file-level allowlist for pre-existing references

## Test plan
- [ ] Commit a file with \"sahar\" in it — should be blocked
- [ ] Commit a file with \"drew\" alone — should be blocked
- [ ] Commit a file with \"andrew\" — should pass
- [ ] Commit a file with \"sayhar\" — should pass (GitHub username)
- [ ] Commit a file with \"withdrew\" — should pass
- [ ] Commit a file with \"drew # noname\" — should pass
- [ ] Push with `--no-verify` — should bypass both hooks

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -75,3 +75,7 @@ scripts/claude-wrapper.exp:Hardcoded token
 # throughout the codebase (scheduled-tasks/tasks/bot-talk-poller*.md). The default
 # value is a fallback when BOT_TALK_API_URL env var is unset — not a secret.
 scheduled-tasks/bot-talk-check-dispatch.sh:IP address
+
+# scripts/upgrade.sh contains pre-existing migration comments written before the
+# name-block hook was added. These are historical references, not new additions.
+scripts/upgrade.sh:Personal name (sahar)


sayhar (commented):
🤖🦞 Lobster (reviewer):

## Code Review: feat/name-block-hooks

**What this PR does:** Adds case-insensitive personal name scanning to both .githooks/pre-commit (staged diff only) and .githooks/pre-push (full file content of changed files), blocking commits and pushes containing standalone occurrences of "sahar" or "drew".

### How it works

**Pre-commit** scans `git diff --cached` output, processing only added lines (+). It tracks file context line-by-line, skips `.githooks/*` files (which contain the patterns as regex strings), and skips lines with `# noname` inline annotation. Matching uses PCRE negative lookbehind/lookahead (`(?<![a-z])sahar(?![a-z])`) to catch standalone names only.

**Pre-push** adds a `NAME_PATTERNS` array alongside the existing `PII_PATTERNS` and `SECURITY_PATTERNS`, reusing the existing scan infrastructure. Each pattern carries a per-pattern allowlist regex applied after the main match. Same `.githooks/*` skip and `# noname` suppression apply.

**Bypass mechanisms are explicit and documented:** `--no-verify`, `# noname` inline, and the file-level `security-allowlist.txt`.

### What it catches (verified by smoke test)

- "sahar" as a standalone word — BLOCKED (commit returns exit code 1, does not land)
- "drew" as a standalone word — BLOCKED
- "andrew" — PASSES (drew embedded in longer word)
- "sayhar" — PASSES (GitHub username allowlist)
- "withdrew" — PASSES (drew embedded)
- "sahar # noname" — PASSES (inline suppression)

### Known gaps (not in scope for this PR)

This PR intentionally scopes to personal first names only. The following patterns slipped through during development and remain unblocked:

- **IP addresses**: `http://46.224.41.108:4242` is hardcoded in `scheduled-tasks/tasks/bot-talk-poller.md` (has a file-level allowlist entry in security-allowlist.txt for the IP address pattern in pre-push, but the value persists in the file itself)
- **Numeric Telegram chat_ids**: `chat_id=8305714125` in the same file — not caught by any current pattern
- **Instance-specific identifiers**: "AlbertLobster" contains no blocked name. "SaharLobster" — testing with the PCRE pattern suggests it would be caught (no [a-z] before "sahar" in "SaharLobster" since S is uppercase, and no [a-z] after "sahar" since 'L' is uppercase), but "Albert" is unprotected.

A follow-up issue will cover extending the hook to catch IPs, chat_ids, and instance-specific identifiers.

### Soak duration

Merged into local-dev at 2026-03-28 04:11 UTC. Reviewed at 2026-03-28 20:16 UTC. ~16 hours of live operation with no hook failures or false-positive blocks reported.

### Verdict: PASS

The hook works as described. Allowlists are conservative and correct. The scripts/upgrade.sh allowlist entry for pre-existing historical references is appropriate. Merging is safe.

Issue #957, opened 2026-03-26
MCP reconnect failure: 20-min stuck-disconnected period after CC auto-update to 2.1.85


sayhar:
## Summary

CC auto-updated from 2.1.84 to 2.1.85 at 23:00 UTC on 2026-03-26, killing \`inbox_server.py\` mid-session. The dispatcher recovered from two rapid restarts (23:01, 23:04, 23:06) but then became stuck in a disconnected state for **15 minutes 16 seconds** (23:06:31–23:21:47 UTC) despite inbox_server.py accepting tool calls throughout.

This is the same WFM-heartbeat-masking pattern documented in the 10:40 UTC incident earlier today.

## Timeline

| UTC | Event |
|-----|-------|
| 23:00:58 | CC 2.1.85 binary written to `~/.local/share/claude/versions/2.1.85` |
| ~23:01:58 | inbox_server.py restart #1 (CC update kill) — two rapid [startup] events at 23:01:58 + 23:01:59 |
| 23:03:30 | First tool call on new connection (send_reply) |
| 23:03:53 | wait_for_messages — recovers 10 stale messages (aged 13,647–13,651s, created ~19:16 UTC) |
| 23:04:33 | inbox_server.py restart #2 — [startup] while dispatcher is mid-processing |
| 23:05:43 | Dispatcher completes backlog processing (10 messages marked_processed) |
| 23:06:25–31 | Dispatcher claims 10 more messages (mark_processing x10) |
| 23:06:30 | inbox_server.py restart #3 — [startup] event interleaved with ongoing tool calls |
| **23:06:31** | **Last tool call logged — mark_processing for 1774554901512** |
| **23:06:31–23:21:47** | **Complete silence: no dispatcher or subagent tool calls for 15m 16s** |
| 23:21:47 | wait_for_messages — recovers same 10 messages as stale (aged 916–921s) |
| 23:22:13 | Processing resumes normally |

## Root Cause

The same WFM heartbeat masking bug documented in the 10:40 UTC incident. After restart #3 at 23:06:30, the dispatcher's `wait_for_messages` call was orphaned on a dead stdio connection. The in-flight WFM loop continued touching the heartbeat file every 60 seconds (independent of whether the CC stdio channel was alive), preventing the health check from seeing RED until the natural timeout expired.

[...]

Issue #932, opened 2026-03-26
bug: MCP mid-session reconnect during deliberation causes permanent stall


sayhar:
## Summary

When the MCP server disconnects mid-session (e.g. from a Claude Code auto-update) while the dispatcher is in the deliberation phase between two `wait_for_messages` calls, the session stalls permanently with no recovery path.

## Observed incident

Audit log 2026-03-26T07:47 UTC: MCP stdio disconnect occurred between two `wait_for_messages` calls while the dispatcher was processing a message. The session froze for 8 minutes before recovering. The same pattern occurred twice on the same day running CC 2.1.84.

## Root cause

The dispatcher's main loop is:
```
wait_for_messages → deliberate → respond → wait_for_messages → ...
```

If the MCP stdio connection drops during the deliberation/response phase (not during `wait_for_messages`), the dispatcher has no mechanism to detect the disconnect and re-enter the `wait_for_messages` call. The health check's 12-minute restart window is the only recovery path, which is far too long.

This is distinct from issue #910 (which covers MCP drops that kill `inbox_server.py` during CC auto-update) — the pattern here is a reconnect that succeeds at the transport layer but leaves the dispatcher in a state where it never re-calls `wait_for_messages`.

## Impact

- Dispatcher stalls silently for up to 12 minutes (the health-check restart interval)
- Messages accumulate in the inbox unprocessed during the stall window
- No alerting fires: the dispatcher process is alive, the health check sees a running PID, and no error is logged

## Possible fixes

1. **Shorten health check interval** for the specific case of a live PID with an inbox that is growing stale (cf. issue #883 inbox-staleness warn)
2. **MCP reconnect signal to dispatcher**: when `inbox_server.py` detects a client reconnect, write a sentinel to the inbox so the dispatcher's next `wait_for_messages` call returns immediately and the loop re-enters cleanly
3. **Heartbeat-based stall detection**: if the dispatcher has not called `wait_for_messages` within N seconds of the last processed message, trigger a self-restart

## Environment

- Claude Code 2.1.84
- Lobster commit ff85b8f
- Two occurrences same day (2026-03-26), both during message processing under moderate load

aeschylus:
The MCP reconnect stall during deliberation is a real reliability gap with no reported mitigation yet. Keeping open.

aeschylus:
[...]

Issue #910, opened 2026-03-26
MCP stdio drops from Claude Code auto-updates: inbox_server.py child exits when CC upgrades itself


sayhar:
🤖🦞 Lobster (ops):

## Summary

Starting ~2026-03-24, intermittent MCP drops have been observed. Audit (audit-restart-mcp subagent) confirmed a likely root cause: **Claude Code auto-updating mid-session causes its spawned MCP child processes (including inbox_server.py) to exit**.

## Evidence

Three confirmed disconnect events:
1. 2026-03-24T13:53 UTC
2. 2026-03-25T18:19 UTC
3. 2026-03-26T07:01 UTC

CC version history:
- 2.1.81 — installed ~March 20
- 2.1.83 — installed March 25 06:16 UTC (just before event #2)
- 2.1.84 — installed March 26 00:58 UTC (just before event #3)

Pattern: MCP drops correlate with CC version upgrades. Events 1 and 2 appeared to be compaction exits mistaken for drops. Event 3 (07:01 UTC) was a genuine mid-session reconnect while the dispatcher was active. After reconnect, `_reset_state_on_startup()` in inbox_server.py wrote fresh state to lobster-state.json, confusing the health check about mode/timing. Dispatcher stalled, health check restarted at 07:08 UTC.

## Impact

After a mid-session MCP reconnect:
1. `_reset_state_on_startup()` resets `lobster-state.json` as if this were a fresh boot
2. Health check reads mode=active with very recent state_age and may incorrectly assess health
3. Dispatcher may get stuck if MCP tools become temporarily unavailable mid-`wait_for_messages`

## Fix candidates

1. **inbox_server.py**: distinguish reconnect from fresh boot — skip `_reset_state_on_startup()` if `lobster-state.json` exists and was written in the last 30 minutes
2. **lobster-state.json**: add `reconnect_at` timestamp separate from `booted_at`, so health check can apply different grace period logic
3. **CC auto-update suppression**: explore `--no-auto-update` flag or scheduled update windows (e.g., only during hibernation)

## Related

- #898 (hibernate-wake booted_at gap — similar health check confusion)
- Health alert suppression during reconnect window

PR #961, opened 2026-03-27
feat: switch MCP transport from stdio to HTTP to survive CC restarts

### LINKED ISSUE #960: feat: switch MCP transport from stdio to TCP/HTTP (SSE) to survive CC restarts
## Problem

Every Claude Code auto-update kills the stdio pipe, causing inbox_server.py to exit cleanly (EOF on stdin). This orphans the dispatcher's in-flight wait_for_messages() call, producing 8–22 minute stuck windows (issues #910, #932, #957).

The root cause is architectural: stdio transport ties the MCP server lifetime to the CC process lifetime. Any CC restart kills the connection.

## Proposed Fix: TCP/HTTP (SSE) Transport

Run inbox_server.py as a persistent TCP/HTTP server (MCP Streamable HTTP or SSE transport). CC connects to it on a port. When CC restarts after an auto-update, it reconnects to the already-running server — the server was never killed.

**What this eliminates:**
- inbox_server.py process death on CC restart (the process stays up)
- The WFM orphan problem (new CC session reconnects and issues a fresh WFM call)
- The heartbeat-masking issue (#957) — moot because the server never goes stale

**What this does NOT solve:**
- The dispatcher's WFM call will still be in-flight on the old CC session when CC restarts. The new CC session will issue a new WFM call. Two options:
  a. Accept that the old WFM call is abandoned; the new WFM call recovers stale messages immediately (already works today)
  b. The server could cancel the old WFM when a new client connects (cleaner)

## Effort Estimate

Medium. The MCP Python SDK supports SSE/HTTP transport. Changes needed:
1. inbox_server.py: swap `stdio_server()` for `sse_server()` or streamable HTTP equivalent
2. ~/.claude/settings.json: change mcpServers entry from `command`/`args` to `url: "http://localhost:PORT"`
3. Add systemd unit (or supervise via install.sh) to keep inbox_server.py running
4. Update install.sh / upgrade.sh migration

No logic changes to inbox_server.py tool handlers — only the transport layer changes.

## Alternatives Considered

- **WFM heartbeat fix** (#957 fix options): Reduces stuck time but does not eliminate disconnects
- **Supervisor/wrapper**: Re-spawns inbox_server.py after exit, but CC still needs to reconnect (stdio still drops)
- **Shorter health-check threshold**: Palliative only, causes more false-positive restarts

## References

- #910 — MCP stdio drops from CC auto-updates
- #932 — mid-session reconnect stall
- #957 — WFM heartbeat masking (most recent incident)

🤖🦞 Lobster subagent: filed from mcp-resilience-fix analysis task

sayhar:
## Summary

CC auto-updates close the stdio pipe, killing \`inbox_server.py\` and orphaning any in-flight \`wait_for_messages\` call — producing 8–22 minute stuck windows (issues #910, #932, #957). The root cause is that stdio transport ties the MCP server's lifetime to the CC process.

This PR decouples the two: \`inbox_server.py\` now runs as a persistent HTTP server managed by a dedicated systemd service (\`lobster-mcp-local\`). Claude Code connects via \`url: "http://localhost:8766/mcp"\` instead of spawning a subprocess. When CC restarts after an auto-update, it reconnects to the already-running server — the server was never killed, and no stdio pipe was ever involved.

Before this PR: CC spawns \`inbox_server.py\` as a child process over stdio → CC restart kills the pipe → server exits → WFM call is orphaned → 8–22 min stuck window.

After this PR: systemd starts \`lobster-mcp-local\` independently → CC connects over HTTP → CC restart disconnects the HTTP client, server stays up → new CC session reconnects → WFM recovers immediately.

## What changed

- **\`src/mcp/inbox_server.py\`**: HTTP transport mode added behind \`--http\` flag (or \`MCP_TRANSPORT=http\` env var). Uses \`StreamableHTTPSessionManager\` with a minimal ASGI app that exposes \`/mcp\` and \`/health\`. stdio mode preserved as fallback for local dev — no tool handler logic touched.
- **\`services/lobster-mcp-local.service.template\`** (new): systemd unit for the local full-access HTTP server on port 8766. Binds to \`127.0.0.1\` only. Enabled by default during install.
- **\`install.sh\`**: renders and installs \`lobster-mcp-local.service\`, enables it, re-registers \`lobster-inbox\` using \`claude mcp add --transport http … http://localhost:8766/mcp\`.
- **\`scripts/upgrade.sh\` Migration 43**: idempotent migration that installs the service and re-registers existing installs. Skipped if \`localhost:8766\` already appears in \`claude mcp list\`.
- **\`services/lobster.target\`**: adds \`lobster-mcp-local.service\` to \`Wants\`.
- **\`src/cli\`**: adds \`lobster-mcp-local\` to \`SERVICES\` so \`lobster start/stop/status\` includes it.

The remote read-only HTTP bridge (\`lobster-mcp.service\` / \`inbox_server_http.py\`) is unchanged — it serves a different purpose (remote Claude Code access) and runs on port 8741.

## Functional patterns used

Pure functions (transport selection is a pure branch on argv/env with no side effects on tool handlers), isolation of side effects at system boundary (HTTP server startup isolated in \`main()\`, not module level).

## Transport flow

\`\`\`
Before:
  systemd → lobster-claude.service → CC → spawns inbox_server.py (stdio)
                                              ↑ dies with CC restart

After:
  systemd → lobster-mcp-local.service → inbox_server.py (HTTP :8766)
                                              ↑ stays up across CC restarts
  systemd → lobster-claude.service → CC → connects to http://localhost:8766/mcp
                                              ↑ reconnects after CC restart
\`\`\`

## Tests run

- [x] \`uv run python src/mcp/inbox_server.py --http --port 8766\` — server started, \`/health\` returned \`{"ok":true}\`, MCP \`initialize\` call returned correct \`serverInfo\`, \`tools/list\` returned 75 tools including full-access write tools (\`send_reply\`, \`wait_for_messages\`, etc.)
- [ ] \`lobster upgrade\` Migration 43 — not run: requires a clean install environment with systemd. Migration logic is straightforward shell; the re-registration path uses the same \`claude mcp add --transport http\` command verified above.
- [ ] Live round-trip (dispatcher sends/receives a real message) — not run: requires production restart. Safe to merge; the service starts independently of lobster-claude.

**Blocked items needing attention before merge:** none. The live round-trip test is the natural next step after merging and running Migration 43.

Closes #960

🤖🦞 Lobster (engineer): generated by subagent

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2544,6 +2544,14 @@ if [ -f "$MCP_TEMPLATE" ]; then
         "$INSTALL_DIR/services/lobster-mcp.service"
 fi
 
+# Generate MCP local HTTP server service if template exists
+MCP_LOCAL_TEMPLATE="$INSTALL_DIR/services/lobster-mcp-local.service.template"
+if [ -f "$MCP_LOCAL_TEMPLATE" ]; then
+    generate_from_template \
+        "$MCP_LOCAL_TEMPLATE" \
+        "$INSTALL_DIR/services/lobster-mcp-local.service"
+fi
+
 # Generate observability server service if template exists
 OBSERVABILITY_TEMPLATE="$INSTALL_DIR/services/lobster-observability.service.template"
 if [ -f "$OBSERVABILITY_TEMPLATE" ]; then
@@ -2572,10 +2580,17 @@ else
         info "Slack router service installed (enable manually with: sudo systemctl enable lobster-slack-router)"
     fi
 
-    # Install MCP HTTP bridge service if generated
+    # Install MCP HTTP bridge service if generated (remote read-only bridge)
     if [ -f "$INSTALL_DIR/services/lobster-mcp.service" ]; then
         sudo cp "$INSTALL_DIR/services/lobster-mcp.service" /etc/systemd/system/
         info "MCP HTTP bridge service installed (enable manually with: sudo systemctl enable lobster-mcp)"
+    fi
+
+    # Install MCP local HTTP server service (full-access, localhost only)
+    if [ -f "$INSTALL_DIR/services/lobster-mcp-local.service" ]; then
+        sudo cp "$INSTALL_DIR/services/lobster-mcp-local.service" /etc/systemd/system/
+        sudo systemctl enable lobster-mcp-local 2>/dev/null || true
+        success "MCP local HTTP server service installed and enabled (lobster-mcp-local)"
     fi
 
     # Install observability service if generated
@@ -2622,13 +2637,16 @@ fi
 
 step "Registering MCP server with Claude..."
 
-# Remove existing registration if present
+# Remove existing registration if present (handles both stdio and http registrations)
 claude mcp remove lobster-inbox 2>/dev/null || true
 
-# Add new registration
-PYTHON_PATH="$INSTALL_DIR/.venv/bin/python"
-if claude mcp add lobster-inbox -s user -- "$PYTHON_PATH" "$INSTALL_DIR/src/mcp/inbox_server.py" 2>/dev/null; then
-    success "MCP server registered"
+# Register MCP server using HTTP transport (streamable-http).
+# Claude Code connects to the locally-running lobster-mcp-local service on port 8765.
+# This decouples the MCP server lifetime from the Claude Code process, so CC
+# auto-updates no longer cause a stdio pipe drop / stuck wait_for_messages call.
+MCP_LOCAL_URL="http://localhost:8766/mcp"
+if claude mcp add --transport http lobster-inbox -s user "$MCP_LOCAL_URL" 2>/dev/null; then
+    success "MCP server registered (HTTP transport: $MCP_LOCAL_URL)"
 else
     warn "MCP server registration may have failed. Check with: claude mcp list"
 fi
@@ -2798,10 +2816,10 @@ echo -e "${BOLD}Required post-install steps:${NC}"
 if [ "$GITHUB_TOKEN_SET" = false ]; then
 echo "  1. Set your GitHub PAT:    lobster env set GITHUB_TOKEN <your-token>"
 echo "  2. Authenticate Claude:    sudo -u lobster claude  (then follow OAuth prompts)"
-echo "  3. Start services:         sudo systemctl start lobster-claude lobster-mcp lobster-router"
+echo "  3. Start services:         sudo systemctl start lobster-mcp-local lobster-claude lobster-router"
 else
 echo "  1. Authenticate Claude:    sudo -u lobster claude  (then follow OAuth prompts)"
-echo "  2. Start services:         sudo systemctl start lobster-claude lobster-mcp lobster-router"
+echo "  2. Start services:         sudo systemctl start lobster-mcp-local lobster-claude lobster-router"
 fi
 echo ""
 echo -e "${BOLD}Commands:${NC}"
--- a/install.sh
+++ b/install.sh
@@ -2641,7 +2641,7 @@ step "Registering MCP server with Claude..."
 claude mcp remove lobster-inbox 2>/dev/null || true
 
 # Register MCP server using HTTP transport (streamable-http).
-# Claude Code connects to the locally-running lobster-mcp-local service on port 8765.
+# Claude Code connects to the locally-running lobster-mcp-local service on port 8766.
 # This decouples the MCP server lifetime from the Claude Code process, so CC
 # auto-updates no longer cause a stdio pipe drop / stuck wait_for_messages call.
 MCP_LOCAL_URL="http://localhost:8766/mcp"
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1744,6 +1744,69 @@ EOF
         fi
     fi
 
+    # Migration 43: Switch MCP transport from stdio to HTTP (issue #960).
+    # The lobster-mcp-local systemd service now runs inbox_server.py as a
+    # persistent HTTP server on localhost:8766.  Claude Code must be registered
+    # to connect via "url" instead of a stdio command so that CC auto-updates
+    # no longer kill the MCP server (they would close the stdio pipe).
+    #
+    # This migration:
+    #   a) Installs (or updates) the lobster-mcp-local systemd service.
+    #   b) Re-registers the lobster-inbox MCP server using HTTP transport.
+    #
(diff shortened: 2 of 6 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one note

## Summary

This PR correctly addresses the root cause of issues #910, #932, #957: stdio transport ties inbox_server.py's lifetime to the CC process, so any CC auto-update kills the pipe and orphans wait_for_messages.

The fix is architecturally sound: decouple the MCP server into a persistent systemd unit (lobster-mcp-local), bind to localhost:8766, register lobster-inbox using --transport http. CC reconnects after restart; server was never killed.

## Code review

**inbox_server.py** — transport selection is clean. Pure branch on argv/env, no side effects on tool handlers. The ASGI app correctly separates /health, /mcp, and 404. Lifespan pattern with StreamableHTTPSessionManager is the right approach.

**One inconsistency to flag (non-blocking):** The inline comment says 'default 8765' in two places, but the systemd service sets MCP_HTTP_PORT=8766, install.sh and migration 41 both use 8766. The env var wins so behavior is correct, but the comment should say 8766. Low priority fix.

**services/lobster-mcp-local.service.template** — correct. Binds to 127.0.0.1 only. NoNewPrivileges + PrivateTmp are appropriate hardening. Restart=always + RestartSec=5 ensures resilience.

**install.sh** — service generation and installation flow looks correct. The MCP re-registration (remove stdio + add --transport http) is the right pattern.

**scripts/upgrade.sh Migration 41** — idempotency check (grep for localhost:8766) is correct. Template rendering fallback with inline sed is a reasonable defensive pattern for upgrade scenarios where generate_from_template may not be in scope.

**services/lobster.target** — correctly adds lobster-mcp-local to Wants so it starts with the rest of the lobster stack.

**src/cli** — SERVICES array and restart command both updated. Correct.

## Deployment note

This PR is NOT yet in local-dev. Before landing: (1) run Migration 41 on the live system, (2) verify lobster-mcp-local.service starts and /health returns ok, (3) confirm dispatcher connects and wait_for_messages resumes. The live round-trip test is the natural gate.

## Verdict: PASS — safe to merge after live deployment confirmation

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — nit fixed in follow-up commit

---

## PR walkthrough

**What this does:** Decouples inbox_server.py from the Claude Code process lifetime by switching MCP transport from stdio to HTTP. Previously, CC auto-updates closed the stdio pipe and killed the MCP server, orphaning in-flight wait_for_messages calls and producing 8–22 minute stuck windows (#910, #932, #957). After this PR, a systemd service (lobster-mcp-local) runs inbox_server.py independently on localhost:8766; CC connects via HTTP and simply reconnects after restart.

**Architecture decision:** Using StreamableHTTPSessionManager with a minimal Starlette ASGI app is the right call. The lifespan pattern ensures session_manager.run() is torn down cleanly on server exit. Binding to 127.0.0.1 (not 0.0.0.0) is correct — this is a local-access-only server.

---

## Code review findings

### Bug fixed (committed as follow-up): Port default 8765 → 8766

In `src/mcp/inbox_server.py` there were two comment lines saying "default 8765" and, more importantly, the fallback in `os.environ.get("MCP_HTTP_PORT", 8765)` was **8765** — mismatched from the 8766 used everywhere else (systemd service, install.sh, migration 41). This was a real behavioral bug: running `inbox_server.py --http` locally without the `MCP_HTTP_PORT` env var would bind to 8765, not 8766, and CC would fail to connect. The same typo appeared in install.sh and upgrade.sh comments.

Fixed in commit a9ec343 (all 5 occurrences, including the functional fallback value).

### install.sh — correct

Service template rendering and installation flow is sound. `sudo systemctl enable lobster-mcp-local` with `2>/dev/null || true` is the right pattern (idempotent). MCP re-registration sequence (remove → add --transport http) is correct. Post-install step instructions updated to include lobster-mcp-local.

### services/lobster-mcp-local.service.template — correct

- Binds via `ExecStart` with `--http --port 8766` and `MCP_HTTP_PORT=8766` env var — belt-and-suspenders, good.
- `NoNewPrivileges=true` and `PrivateTmp=true` are appropriate minimal hardening.
- `Restart=always` + `RestartSec=5` — correct resilience posture for a persistent server.
- `After=network.target` is correct (HTTP server needs network).

### scripts/upgrade.sh Migration 41 — correct

- Idempotency check on `grep -c "localhost:8766"` is robust.
- Inline sed fallback for template rendering covers the case where `generate_from_template` is not in scope during an upgrade; the variable list matches what install.sh uses.
- `sleep 3` after service restart before re-registering — necessary because uvicorn needs a moment to start accepting connections before `claude mcp add` connects to validate.
- `migrated=$((migrated + 1))` only incremented on successful re-registration — correct.

### services/lobster.target — correct

Adding `lobster-mcp-local.service` to `Wants` is the right dependency model (Wants = start together, don't fail if lobster-mcp-local is missing). This ensures the service comes up with the rest of the lobster stack via `systemctl start lobster.target`.

### src/cli — correct

`SERVICES` array and the `systemd-run` restart command both updated to include `lobster-mcp-local`. Ordering matters here: `lobster-mcp-local` is listed first, which is correct — it must be up before lobster-claude starts and tries to connect.

### inbox_server.py HTTP transport block — correct

- `use_http` pure branch on argv/env, no side effects on tool handlers — good isolation.
- Imports deferred inside the `if use_http:` block — correct (avoids import errors for deployments without uvicorn/starlette).
- `stateless=False` on StreamableHTTPSessionManager — correct for a long-lived dispatcher that maintains session state across tool calls.
- 404 fallback in `_mcp_handler` is a good defensive addition.
- `log_level="warning"` and `access_log=False` on uvicorn config — appropriate for a background service; avoids log spam.

---

## What is NOT covered by this PR (expected)

- **No reconnect logic on the CC side.** When CC restarts and reconnects, there may be a brief window where in-flight tool calls are lost. This is acceptable and pre-existing behavior — the PR's goal is to ensure the server stays up, not to add reconnect retry logic to CC.
- **Migration 41 not tested in prod yet.** PR description documents this clearly. The live round-trip test is the right gate before treating this as fully resolved.

---

## Verdict: PASS

One real bug (port default 8765 instead of 8766 in the fallback value) fixed in follow-up commit a9ec343. Everything else is correct. Safe to merge; run Migration 41 on live system and verify `/health` + dispatcher reconnect as the post-merge gate.

PR #963, opened 2026-03-27
fix: bot-talk-poller skips dispatcher delivery on no-op runs

### LINKED ISSUE #962: bot-talk-poller: skip write_result delivery when no new activity
## Problem

Every scheduled run of bot-talk-poller spins up a Claude session and calls `write_result`, burning ~2–4K tokens even when there are no new AlbertLobster messages. The dispatcher receives and potentially processes the result even on no-op runs.

## Root Cause

Step 7 of the task instructions unconditionally calls `write_result` with new messages or \"no new activity since <timestamp>\". Even the no-op path routes through the dispatcher, consuming tokens and potentially triggering unnecessary processing.

## Fix

When there are no new messages, call `write_result` with `chat_id=0` and `text='no new activity'`. The dispatcher silently drops results addressed to `chat_id=0`, so no tokens are consumed processing the reply.

When there are new messages, continue calling `write_result` with `chat_id=8305714125` (Sahar's chat ID) as before.

## Acceptance Criteria

- [ ] Task file updated so no-op runs call `write_result(chat_id=0, text='no new activity')`
- [ ] Task file updated so active runs continue calling `write_result` with Sahar's `chat_id`
- [ ] PR merged

sayhar:
🤖🦞 Lobster (engineer):

Every scheduled run of the bot-talk-poller (every 5 minutes) and bot-talk-poller-fast (every 2 minutes) spins up a Claude session. Even when there are no new messages, the subagent called `write_result` with Sahar's `chat_id`, causing the dispatcher to receive and process the result — burning ~2–4K tokens per no-op cycle. The fast poller alone runs 720 times per day, meaning the majority of those runs (when Albert is not active) were all burning tokens for nothing.

The fix: both task files now explicitly instruct the subagent to call `write_result(chat_id=0, text='no new activity')` on no-op runs. `chat_id=0` is silently dropped by the dispatcher — no relay, no context window cost. When new messages are found, the subagent continues calling `write_result` with Sahar's `chat_id` as before.

This also covers the fast poller's fast-exit path (when `hot_mode=false`), which was skipping work entirely but still needed a compliant `write_result` call.

**Functional pattern:** the change is pure instruction to the subagent LLM — no code, just conditional branching expressed declaratively in the task file's Output section. The subagent reads the two code blocks and picks the right one based on whether it sent messages.

**What is unchanged:** all polling logic, state file management, Telegram notification formatting, hot-mode activation/cooldown, and the `write_task_output` call. This is strictly a `write_result` routing change.

## Tests run
- [x] `git diff` reviewed manually — changes are additive (new Output section instructions only), no existing behavior removed
- [ ] Live scheduled run — cannot test without a running Lobster instance; the change is a task file instruction with no code path to unit test

Closes #962

### DIFF
--- a/scheduled-tasks/tasks/bot-talk-poller-fast.md
+++ b/scheduled-tasks/tasks/bot-talk-poller-fast.md
@@ -92,3 +92,27 @@ When you complete your task, call `write_task_output` with:
 - status: "success" or "failed"
 
 Keep output concise. The main Lobster instance will review this later.
+
+Then call `write_result` to close the subagent task:
+
+**If skipped (hot_mode=false) or no new messages (no-op run):**
+```python
+mcp__lobster-inbox__write_result(
+    task_id="bot-talk-poller-fast",
+    chat_id=0,  # silently dropped by dispatcher — no tokens burned
+    text="no new activity",
+    source="telegram",
+    sent_reply_to_user=True,
+)
+```
+
+**If new messages were found and sent to Sahar:**
+```python
+mcp__lobster-inbox__write_result(
+    task_id="bot-talk-poller-fast",
+    chat_id=8305714125,
+    text="<summary of new messages delivered>",
+    source="telegram",
+    sent_reply_to_user=True,  # already sent via send_reply above
+)
+```
--- a/scheduled-tasks/tasks/bot-talk-poller.md
+++ b/scheduled-tasks/tasks/bot-talk-poller.md
@@ -116,3 +116,27 @@ When you complete your task, call `write_task_output` with:
 - status: "success" or "failed"
 
 Keep output concise. The main Lobster instance will review this later.
+
+Then call `write_result` to close the subagent task:
+
+**If no new messages (no-op run):**
+```python
+mcp__lobster-inbox__write_result(
+    task_id="bot-talk-poller",
+    chat_id=0,  # silently dropped by dispatcher — no tokens burned
+    text="no new activity",
+    source="telegram",
+    sent_reply_to_user=True,
+)
+```
+
+**If new messages were found and sent to Sahar:**
+```python
+mcp__lobster-inbox__write_result(
+    task_id="bot-talk-poller",
+    chat_id=8305714125,
+    text="<summary of new messages delivered>",
+    source="telegram",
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Both task files correctly add a two-branch `write_result` pattern that routes no-op runs to `chat_id=0`. The approach is sound and matches how the server already handles this value.

## Correctness verification

Traced the `chat_id=0` code path through `inbox_server.py`:

1. `write_result` writes a `subagent_notification` inbox file (since `sent_reply_to_user=True`).
2. `mark_processed` checks `USER_FACING_TYPES` — `subagent_notification` is **not** in that set, so the no-reply guard is never triggered regardless of `chat_id`.
3. The reconciler's dead-session path explicitly suppresses notifications when `chat_id` is `"0"` / `""` / `"None"` (line ~7473).
4. The only place `chat_id != 0` guards a relay attempt is in the `USER_FACING_TYPES` block, which excludes system types entirely.

Result: a `subagent_notification` with `chat_id=0` is written to inbox, the dispatcher reads it as a `SUBAGENT_NOTIFICATION` hint, calls `mark_processed`, and no outbound message is queued. Token cost to dispatcher is the inbox poll cost only — the ~2–4K relay cost is eliminated.

## Fast-poller fast-exit path

The `hot_mode=false` fast-exit is correctly handled. The original task file already called `write_task_output` on fast-exit; the new code adds the matching `write_result(chat_id=0)` call, which satisfies the dispatcher's expectation that every subagent invocation closes with `write_result`.

- The `task_id` strings in both code blocks are hardcoded to the job name (`"bot-talk-poller"` / `"bot-talk-poller-fast"`). This is fine — `write_result` uses `task_id` for session auto-unregister matching and it matches the job name as registered by the dispatcher.
- `source="telegram"` on a no-op `chat_id=0` call is harmless — source is only used for outbox routing and no outbox entry is created here.
- The PR description accurately describes the mechanism.

## Verdict

Change is additive, the `chat_id=0` suppression path is verified in server code, both pollers are covered, and no existing behavior is removed. Ready to merge.

PR #981, opened 2026-03-27
feat: skip bot-talk-poller dispatch when no new messages

### LINKED ISSUE #980: bot-talk-poller: skip dispatch when no new messages (smart pre-check)
## Problem

The \`bot-talk-poller\` cron job calls \`dispatch-job.sh\` every hour unconditionally. This spawns a full LLM subagent even when there are zero new messages. The subagent reads the state file, polls the API, finds nothing, and exits — burning tokens and inbox queue slots for a no-op.

During API outages or quiet periods, every hourly tick generates a \"nothing new\" subagent invocation.

## Goal

Introduce a lightweight pre-check that runs before dispatching to the LLM. If the API has no new messages since \`last_message_ts\`, the cron job exits immediately with no inbox write and no subagent spawn.

## Implementation

1. Create \`scheduled-tasks/bot-talk-check-dispatch.sh\` — a shell script that:
   - Loads config/token (same as dispatch-job.sh does)
   - Reads \`last_message_ts\` from \`~/lobster-workspace/data/bot-talk-state.json\`
   - Does a single HTTP GET to the bot-talk API with \`?since=<last_message_ts>\`
   - If response contains new messages → delegates to \`dispatch-job.sh bot-talk-poller\`
   - If no new messages → logs the no-op to \`scheduled-jobs/logs/\` and exits 0 (no dispatch)

2. Support a per-job \`runner\` override field in \`jobs.json\` so \`bot-talk-poller\` can specify its custom runner. Update \`sync-crontab.sh\` and \`install.sh\` crontab generation to use \`.runner\` if present, else fall back to the default \`dispatch-job.sh\`.

3. Update \`~/lobster-workspace/scheduled-jobs/jobs.json\` to set \`bot-talk-poller.runner\` to the new script path.

4. Add a migration to \`upgrade.sh\` to update the crontab entry.

## Before / After

Before:
```
cron tick → dispatch-job.sh → inbox write → dispatcher spawns LLM → "nothing new" → exit
```

After:
```
cron tick → bot-talk-check-dispatch.sh → API check (no new msgs) → log no-op → exit
cron tick → bot-talk-check-dispatch.sh → API check (new msgs!) → dispatch-job.sh → inbox write → LLM
```

## Acceptance Criteria

- [ ] \`bot-talk-check-dispatch.sh\` exits 0 with no inbox write when API has no new messages
- [ ] \`bot-talk-check-dispatch.sh\` calls \`dispatch-job.sh\` when there are new messages
- [ ] \`sync-crontab.sh\` and \`install.sh\` respect per-job \`runner\` field in \`jobs.json\`
- [ ] \`jobs.json\` updated to use new runner for \`bot-talk-poller\`
- [ ] \`upgrade.sh\` migration re-syncs crontab to pick up the change
- [ ] No-op runs produce a brief log line but no inbox message

sayhar:
## What problem this solves

Every hourly cron tick for `bot-talk-poller` unconditionally writes to the Lobster inbox, spawning a full LLM subagent even when there are zero new messages. The subagent checks the API, finds nothing, and exits — burning tokens and inbox queue slots for a no-op. During quiet periods or API outages, every tick generates a "nothing new" agent invocation.

## What changed

Introduces `bot-talk-check-dispatch.sh`, a lightweight pre-check wrapper that sits in front of `dispatch-job.sh` for the bot-talk-poller cron entry:

```
Before:
  cron tick → dispatch-job.sh → inbox write → LLM subagent → "nothing new" → exit

After (no new messages):
  cron tick → bot-talk-check-dispatch.sh → GET /messages?since=<last_ts> → [] → log no-op → exit 0

After (new messages):
  cron tick → bot-talk-check-dispatch.sh → GET /messages?since=<last_ts> → [...] → dispatch-job.sh → inbox write → LLM
```

The pre-check reads `last_message_ts` from `bot-talk-state.json`, does a single authenticated HTTP request with a `?since=` filter, and only delegates to `dispatch-job.sh` if the response is non-empty. API errors (curl non-zero) fall through to dispatch so the job's existing 30-minute outage-handling logic remains in effect.

## Infrastructure change: per-job runner override

`sync-crontab.sh` and `install.sh` now respect an optional `runner` field in each `jobs.json` job entry. When set, it overrides the default `dispatch-job.sh` runner for that job's cron line. `$REPO_DIR` is expanded in the value so paths stay portable. The `bot-talk-poller` entry in `jobs.json` is updated to use the new wrapper.

This pattern is reusable for other deterministic pollers (e.g. `lobster-plans-poller`) without any further infrastructure changes.

## Migration

Migration 44 in `upgrade.sh` runs `sync-crontab.sh` on existing installs to pick up the new runner. Idempotent: skipped if the crontab already references `bot-talk-check-dispatch.sh`.

## Functional patterns

Pure sequential logic (read state → check API → branch on result), side-effect isolation (HTTP call and crontab I/O at the edges, no mutations until dispatch is confirmed), and fail-safe defaults (API error → dispatch rather than swallow).

## Tests run

- [x] Verified `sync-crontab.sh` jq expression handles `.runner // $runner` fallback correctly: tested manually against `jobs.json` with and without the `runner` field
- [x] Verified `bot-talk-check-dispatch.sh` exits 0 with no dispatch when no messages (manual dry-run against live API with current `last_message_ts`)
- [ ] Full integration test (cron trigger → live dispatch) — blocked: requires merge + `sync-crontab.sh` run on production

Closes #980

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -70,3 +70,8 @@ src/bot/whatsapp_bridge_adapter.py:Email address
 # claude-wrapper.exp reads TELEGRAM_BOT_TOKEN from config.env at runtime using
 # grep — this is a variable reference, not a hardcoded credential value
 scripts/claude-wrapper.exp:Hardcoded token
+
+# bot-talk-check-dispatch.sh uses the same bot-talk API IP that already appears
+# throughout the codebase (scheduled-tasks/tasks/bot-talk-poller*.md). The default
+# value is a fallback when BOT_TALK_API_URL env var is unset — not a secret.
+scheduled-tasks/bot-talk-check-dispatch.sh:IP address
--- a/install.sh
+++ b/install.sh
@@ -1231,10 +1231,14 @@ MARKER="# LOBSTER-SCHEDULED"
 EXISTING=$(crontab -l 2>/dev/null | grep -v "$MARKER" | grep -v "$RUNNER" || true)
 
 if command -v jq &> /dev/null; then
-    CRON_ENTRIES=$(jq -r --arg runner "$RUNNER" --arg marker "$MARKER" '
+    CRON_ENTRIES=$(jq -r --arg runner "$RUNNER" --arg marker "$MARKER" \
+        --arg repo_dir "$REPO_DIR" '
         .jobs | to_entries[] |
         select(.value.enabled == true) |
-        "\(.value.schedule) \($runner) \(.key) \($marker)"
+        (.value.runner // $runner) as $job_runner |
+        # Expand $REPO_DIR placeholder so per-job runners can reference the install dir
+        ($job_runner | gsub("\\$REPO_DIR"; $repo_dir)) as $resolved_runner |
+        "\(.value.schedule) \($resolved_runner) \(.key) \($marker)"
     ' "$JOBS_FILE" 2>/dev/null || echo "")
 else
     CRON_ENTRIES=$(python3 -c "
@@ -1243,11 +1247,15 @@ import sys
 try:
     with open('$JOBS_FILE', 'r') as f:
         data = json.load(f)
+    repo_dir = '$REPO_DIR'
+    default_runner = '$RUNNER'
     for name, job in data.get('jobs', {}).items():
         if job.get('enabled', True):
             schedule = job.get('schedule', '')
             if schedule:
-                print(f\"{schedule} $RUNNER {name} $MARKER\")
+                runner = job.get('runner', default_runner)
+                runner = runner.replace('\$REPO_DIR', repo_dir)
+                print(f\"{schedule} {runner} {name} $MARKER\")
 except Exception as e:
     sys.stderr.write(f'Error: {e}\n')
 " 2>/dev/null || echo "")
--- a/scheduled-tasks/bot-talk-check-dispatch.sh
+++ b/scheduled-tasks/bot-talk-check-dispatch.sh
@@ -81,6 +81,7 @@ BOT_TALK_API_URL="${BOT_TALK_API_URL:-http://46.224.41.108:4242}"
 HAS_NEW_MESSAGES=0
 API_ERROR=0
 
+ENCODED_TS=""
 if [ -n "$LAST_TS" ]; then
     # URL-encode the timestamp for the query string
     ENCODED_TS=$(python3 -c "import urllib.parse, sys; print(urllib.parse.quote(sys.argv[1]))" "$LAST_TS" 2>/dev/null || echo "")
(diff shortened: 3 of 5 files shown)


sayhar (commented):
PASS: Core implementation is correct and safe. A few findings worth noting below.

## Findings

### jq gsub escaping (sync-crontab.sh and install.sh) — CORRECT
`gsub("\\$REPO_DIR"; $repo_dir)` in jq source: the shell sees `gsub("\$REPO_DIR"; ...)` and the jq string literal is `\$REPO_DIR`, which is a regex matching literal `$REPO_DIR`. Works correctly.

### Python3 fallback `runner.replace('\$REPO_DIR', repo_dir)` — CORRECT
`'\$REPO_DIR'` in a Python string: Python does not treat `\$` as an escape sequence, so it collapses to `'$REPO_DIR'`. The replace call matches the literal dollar-sign string that jobs.json would store. Correct.

### `set -e` with `|| API_ERROR=$?` — ROBUST
The `||` operator suppresses errexit for the left-hand side. `RESPONSE=$(curl ...) || API_ERROR=$?` is the standard pattern for capturing a non-zero exit under `set -e`. Works as intended.

### Latent: `ENCODED_TS` referenced before assignment
If `LAST_TS` is empty, the script branches to set `HAS_NEW_MESSAGES=1` directly and never assigns `ENCODED_TS`. Later the condition `[ -n "$ENCODED_TS" ]` references an unbound variable. No `set -u` is active so it silently expands to `""` and the branch is skipped as intended — but if `set -u` is ever added, this will break. Low risk now; worth a `ENCODED_TS=""` initialization line near the other variable declarations.

### Latent: JSON parse failure → silent no-dispatch
If the API returns invalid JSON, the `except` block in the MSG_COUNT python snippet prints `0`, causing a skip rather than a fallthrough to dispatch. This is the opposite of the stated fail-safe philosophy ("API error → dispatch"). Could be fixed by checking for a parse exception and dispatching instead:
```python
except json.JSONDecodeError:
    print(-1)  # sentinel → dispatch
```
Then `if [ "$MSG_COUNT" -ne 0 ]` (or `gt -1`) triggers dispatch. Minor correctness risk in a degraded API scenario.

### install.sh completeness — OK
`bot-talk-check-dispatch.sh` is committed `100755` (executable bit in the diff header: `new file mode 100755`), so git preserves it on clone. No explicit `chmod +x` in `install.sh` is required. The crontab generation changes are present and correct.

### Migration 44 — functional but dependent on runtime state
Migration 44 re-runs `sync-crontab.sh`, which reads the runtime `~/lobster-workspace/scheduled-jobs/jobs.json`. If that file does not yet have `"runner": "$REPO_DIR/scheduled-tasks/bot-talk-check-dispatch.sh"` for `bot-talk-poller`, the migration re-syncs with the default runner and has no effect. The user must update their local `jobs.json` separately. Not a code bug (jobs.json is intentionally not in the repo), but the migration comment could note this dependency explicitly.

## Summary
Logic is solid. The two latent issues (uninitialized `ENCODED_TS`, JSON-parse-failure dispatch-skip) are low-risk edge cases that don't affect the happy path or the documented API-error fallthrough case. Suggest addressing in a follow-up or in this PR if easy.

sayhar (commented):
NEEDS-WORK: Two edge cases must be fixed before merge: (1) ENCODED_TS uninitialized when LAST_TS is empty — breaks under set -u. (2) Invalid JSON from API causes silent skip (MSG_COUNT defaults to 0) instead of falling through to dispatch — contradicts the stated fail-safe philosophy.

sayhar (commented):
PASS: Edge cases fixed — ENCODED_TS initialized for empty LAST_TS, JSON parse failure now falls through to dispatch instead of silently skipping.

PR #986, opened 2026-03-27
fix: crontab sync fails with Permission denied under NoNewPrivs (issue #985)

### LINKED ISSUE #985: bug: crontab sync fails with mkstemp: Permission denied when MCP server has NoNewPrivs=1
## Summary

`create_scheduled_job`, `update_scheduled_job`, and `delete_scheduled_job` all call `sync_crontab()`, which runs `sync-crontab.sh`, which calls `crontab -`. This fails with:

```
/var/spool/cron/: mkstemp: Permission denied
```

## Root Cause

Claude Code sets `PR_SET_NO_NEW_PRIVS` on the MCP server process (`NoNewPrivs: 1` in `/proc/<pid>/status`). This flag propagates to all child processes and prevents them from gaining new privileges via setgid bits.

The `crontab` binary is setgid-crontab (`-rwxr-sr-x root:crontab`), which normally lets it write to `/var/spool/cron/crontabs/` (permissions `drwx-wx--T root:crontab`). With `NoNewPrivs=1`, the setgid privilege is suppressed, and the binary fails with `mkstemp: Permission denied` when trying to create a temp file in `crontabs/`.

**Verified via strace**: `crontab -` does `chdir("/var/spool/cron")` then `openat(AT_FDCWD, "crontabs/tmp.XXXXXX", O_RDWR|O_CREAT|O_EXCL, 0600)` which returns `EACCES` → `Permission denied`.

The `crontab` binary itself works fine from a regular shell (NoNewPrivs=0), but always fails when spawned from the MCP server process tree.

## Fix

Add the `lobster` user to the `crontab` group (via `upgrade.sh` migration + `local-setup-helper.sh`). The `crontab` group has `-wx` (write+execute) on `/var/spool/cron/crontabs/`. With direct group membership, no setgid privilege is needed — `sync-crontab.sh` can write directly to `/var/spool/cron/crontabs/$USER`.

Update `sync-crontab.sh` to fall back to a direct file write when `crontab -` fails due to permission errors.

## Impact

Every `create_scheduled_job`, `update_scheduled_job`, and `delete_scheduled_job` call silently fails to sync the crontab. Jobs are written to `jobs.json` but never actually scheduled in cron.

sayhar:
## Summary

Every call to `create_scheduled_job`, `update_scheduled_job`, and `delete_scheduled_job` silently fails to sync the crontab, leaving jobs in `jobs.json` but never actually scheduled.

The root cause: Claude Code sets `PR_SET_NO_NEW_PRIVS` on the MCP server process (`NoNewPrivs: 1`), which propagates to all child processes and suppresses setgid bits. The `crontab` binary is setgid-crontab — that privilege is exactly what lets it write to `/var/spool/cron/crontabs/`. Without it, the `crontab -` call fails with `/var/spool/cron/: mkstemp: Permission denied`.

Verified with `strace`: `crontab -` does `chdir("/var/spool/cron")` then `openat(AT_FDCWD, "crontabs/tmp.XXXXXX", O_RDWR|O_CREAT|O_EXCL)` → `EACCES`. The binary works fine in a regular shell session (`NoNewPrivs: 0`) but always fails when spawned from the MCP server process tree.

## Changes

**`sync-crontab.sh`**: adds a primary path that writes directly to `/var/spool/cron/crontabs/$USER` using an atomic temp-file rename. This works under `NoNewPrivs=1` because the `crontab` group has `-wx` (write+execute) on the directory — group membership alone is sufficient, no setgid needed. The `crontab -` path is kept as a fallback for systems where the user hasn't joined the group yet.

**`upgrade.sh` Migration 41**: runs `sudo usermod -aG crontab $USER` on existing installs (if sudo is available passwordlessly). Warns and skips if sudo is not available.

**`local-setup-helper.sh`**: adds the user to the `crontab` group during fresh setup, alongside the other `apt install` dependencies.

## What is not changed

The Python `sync_crontab()` function in `inbox_server.py` is unchanged — it already calls `sync-crontab.sh` correctly as an async subprocess. No logic changes to the MCP tool handlers.

## Before / after

```
Before:
  MCP server (NoNewPrivs=1)
    → sync-crontab.sh
      → crontab - [FAILS: setgid suppressed, mkstemp: Permission denied]
      → job written to jobs.json but never in crontab

After (user in crontab group):
  MCP server (NoNewPrivs=1)
    → sync-crontab.sh
      → write directly to /var/spool/cron/crontabs/lobster [OK: group membership]
      → job written to jobs.json AND appears in crontab

After (user NOT yet in crontab group — fallback):
  MCP server (NoNewPrivs=1)
    → sync-crontab.sh
      → direct write fails (directory not writable)
      → crontab - [FAILS same as before — until migration 41 runs]
```

## Functional patterns

Idempotency: `write_crontab()` in the shell script produces the same result for the same input regardless of prior crontab state. The two-path design (direct write first, `crontab -` fallback) degrades gracefully without requiring the caller to know which path succeeded.

## Tests run

- [x] `bash -n scheduled-tasks/sync-crontab.sh` — syntax OK
- [x] `bash -n scripts/local-setup-helper.sh` — syntax OK
- [x] `~/lobster/.venv/bin/pytest tests/unit/test_mcp_server/test_scheduled_jobs.py` — 23 errors, all pre-existing (`_DEBUG_RESOLVED` AttributeError; same failure on `main` before this change, unrelated to this PR)
- [ ] Live test: add user to crontab group and verify `create_scheduled_job` syncs crontab correctly — blocked: `usermod` requires sudo/root in this environment; needs manual verification before merge

**Blocked item needing attention before merge:** the live end-to-end test (`usermod -aG crontab lobster` + create_scheduled_job) could not be run because `sudo` requires a password in this environment. The logic is straightforward (if directory is writable, write directly; else fall back), but a smoke test after adding the user to the crontab group would be prudent.

Closes #985

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -495,6 +495,20 @@ if [ "$(id -u)" = "0" ]; then
         # Docker not installed — this is the normal case on a fresh machine, not a warning
         info "Docker not installed — skipping docker group setup. Install Docker later to enable Docker features."
     fi
+    # Add lobster user to the crontab group so sync-crontab.sh works under NoNewPrivs.
+    # Claude Code sets PR_SET_NO_NEW_PRIVS on the MCP server process, which propagates to
+    # child processes and suppresses setgid bits. The `crontab` binary is setgid-crontab —
+    # that privilege is what lets it write to /var/spool/cron/crontabs/. Without it,
+    # `crontab -` fails with "mkstemp: Permission denied". Group membership lets the
+    # lobster user write directly to /var/spool/cron/crontabs/ (group-writable) without
+    # needing the setgid bit.
+    if getent group crontab &>/dev/null; then
+        usermod -aG crontab lobster
+        success "Added 'lobster' to the crontab group (fixes NoNewPrivs crontab permission error)."
+        warn "Group membership takes effect at next login. Run 'newgrp crontab' or restart after install to apply."
+    else
+        warn "The 'crontab' group does not exist — scheduled job syncing may fail. Run: sudo groupadd crontab && sudo usermod -aG crontab lobster"
+    fi
     # Copy script to /tmp so lobster user can read it regardless of working directory
     INSTALL_SCRIPT="$(readlink -f "$0")"
     TMP_SCRIPT="$(mktemp /tmp/lobster-install.XXXXXX.sh)"
--- a/scheduled-tasks/sync-crontab.sh
+++ b/scheduled-tasks/sync-crontab.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 # Lobster Crontab Synchronizer
-# Syncs jobs.json to system crontab
+# Syncs jobs.json to the user's crontab.
+#
+# This script runs as a subprocess of the MCP server, which sets
+# PR_SET_NO_NEW_PRIVS (NoNewPrivs=1).  That flag suppresses setgid bits,
+# so `crontab -` fails with "mkstemp: Permission denied" even though the
+# crontab binary is setgid-crontab.
+#
+# Primary path: write directly to /var/spool/cron/crontabs/$USER (requires
+# the lobster user to be in the `crontab` group — see upgrade.sh Migration 41).
+# Fallback path: use `crontab -` for systems where the user is not yet in the
+# crontab group (e.g. before the migration runs or on non-Debian systems).
 
 set -e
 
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**What this PR does:** Fixes crontab sync failures under Claude Code's NoNewPrivs=1 constraint. The `crontab` binary is setgid-crontab but NoNewPrivs suppresses setgid bits in child processes, causing 'mkstemp: Permission denied'. This PR takes two complementary approaches: (1) adds lobster user to the `crontab` group so `sync-crontab.sh` can write directly to `/var/spool/cron/crontabs/` without needing the setgid bit, and (2) rewrites `sync-crontab.sh` to prefer direct-file writes with `crontab -` as fallback.

**Issues found:**

1. **Atomic write in write_crontab():** The direct write path creates a temp file with `mktemp` then `mv`s it — this is correct. However `chmod 0600 "$tmpfile"` is set before writing. If the crontab directory is group-writable but the lobster user's umask allows group read, this is fine. No issue here.

2. **`/var/spool/cron/crontabs` hardcoded:** This path is Debian/Ubuntu specific. On RHEL/CentOS it's `/var/spool/cron/`. The PR comments mention "non-Debian systems" but the primary path check only looks at `/var/spool/cron/crontabs`. The fallback to `crontab -` handles the non-Debian case, which is acceptable.

3. **Migration number:** Uses Migration 46. No conflict given this supersedes #992's Migration 45.

4. **Warning on group add:** Correctly warns that group membership requires a re-login/newgrp to take effect. The warning message is clear and actionable.

5. **read_existing_crontab fallback:** If neither the direct file nor `crontab` command is available, returns empty string. This means existing non-lobster cron entries would be silently wiped if both paths fail. Low-risk in practice since at least one will be available.

**Verdict: PASS** — root cause is correctly diagnosed, fix is comprehensive (both group membership + direct write path), atomic write is correct, and fallback behavior is safe. Supersedes #992.

PR #992, opened 2026-03-27
fix: add lobster user to crontab group

sayhar:
## Summary
- Adds `usermod -aG crontab lobster` to `install.sh` alongside the docker group setup
- Adds Migration 41 to `scripts/upgrade.sh` for existing installs
- Applied the fix immediately on the current machine (`groups lobster` now shows `crontab`)

## Why
The `lobster` user could not manage cron jobs without sudo on systems where the `crontab` binary is mode 2755 crontab:crontab. This caused `sync-crontab.sh` and `cron-manage.sh` to fail silently.

## Test plan
- [ ] Fresh install sets crontab group correctly
- [ ] Migration 41 runs cleanly on existing installs (`groups lobster` shows `crontab` after upgrade)
- [ ] `crontab -l` works as the lobster user without sudo

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -495,6 +495,13 @@ if [ "$(id -u)" = "0" ]; then
         # Docker not installed — this is the normal case on a fresh machine, not a warning
         info "Docker not installed — skipping docker group setup. Install Docker later to enable Docker features."
     fi
+    # Add lobster user to the crontab group so it can manage cron jobs without sudo
+    if getent group crontab &>/dev/null; then
+        usermod -aG crontab lobster
+        success "Added 'lobster' to the crontab group."
+    else
+        info "crontab group does not exist — skipping. Cron management may require sudo on this system."
+    fi
     # Copy script to /tmp so lobster user can read it regardless of working directory
     INSTALL_SCRIPT="$(readlink -f "$0")"
     TMP_SCRIPT="$(mktemp /tmp/lobster-install.XXXXXX.sh)"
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**What this PR does:** Adds the lobster user to the `crontab` OS group in `install.sh` (root path) and `upgrade.sh` (Migration 45). This allows `crontab` invocations to work during initial install.

**Issues found:**

1. **Overlap with #986:** PR #986 addresses the same root cause (NoNewPrivs + crontab permission) with a more comprehensive fix — it also updates `sync-crontab.sh` with a direct-write primary path. This PR only adds the `usermod` call without the `sync-crontab.sh` improvements. #986 is a strict superset of this fix.

2. **Migration numbering conflict:** This PR uses Migration 45; #986 uses Migration 46. These can't both land without a renumber.

3. **Missing `local-setup-helper.sh` update:** #986 also updates `local-setup-helper.sh` to handle the group add for dev installs. This PR doesn't.

4. **`info` vs `warn` on missing group:** When `crontab` group doesn't exist, this PR logs `info` (silent on most UIs). #986 logs `warn` which is more appropriate since cron will silently fail.

**Verdict: NEEDS-WORK** — this PR should either be superseded by #986 (which covers everything here plus more) or be rebased as a subset of it. Recommend closing in favor of #986.

sayhar:
Closing as superseded by #1105 (systemd timer rewrite). This PR adds lobster to the crontab group to enable crontab writes. Since #1105 replaces the crontab backend with systemd timers, group membership for crontab is no longer required.

Issue #1000, opened 2026-03-27
New scheduled job: gmail-email-pipeline


sayhar:
## Overview

Create a new scheduled job `gmail-email-pipeline` that polls Gmail for new unread emails every 10 minutes, applies a sensitivity filter, looks up senders in Twenty CRM, creates/updates contacts, drafts suggested replies, and pings via bot-talk when Albert Alexander is mentioned.

## Requirements

- [ ] Poll Gmail for unread inbox emails (max 10 per run)
- [ ] Sensitivity filter: skip personal topics (family, health, personal life)
- [ ] Look up sender in Twenty CRM via GraphQL API
- [ ] Create/update contact in Twenty with business context only
- [ ] Draft a suggested reply
- [ ] Ping Albert via bot-talk if email mentions Albert Alexander
- [ ] Send Sahar a Telegram summary (chat_id=8305714125)
- [ ] Track processed email IDs in `~/lobster-workspace/data/gmail-processed.json`
- [ ] Schedule: every 10 minutes (`*/10 * * * *`)

## APIs

- Gmail: `gws gmail users messages list` / `gws gmail users messages get`
- Twenty CRM: POST to https://honest-navy-moose.twenty.com/graphql
- Bot-talk: POST to http://46.224.41.108:4242/message with X-Bot-Token header
- Telegram: write to ~/messages/inbox/ or use lobster-inbox MCP

PR #1001, opened 2026-03-27
Fix bot-talk auth + add gmail-email-pipeline scheduled job

### LINKED ISSUE #999: Fix bot-talk auth: add X-Bot-Token header to mirror.py and poller
## Problem

The bot-talk server at http://46.224.41.108:4242 now requires `X-Bot-Token` authentication. Two components need updating:

1. **bot_talk_mirror.py** - HTTP POST to /message endpoint missing the auth header
2. **bot-talk-poller task** - GET /messages needs the auth header too

Also: the bot-talk-poller currently reads the token from `~/lobster-workspace/data/bot-talk-token.txt` but the token is stored in `~/lobster-config/config.env` as `BOT_TALK_TOKEN`. The mirror.py needs to read this same config key and include it as `X-Bot-Token` on all HTTP requests.

## Changes

- [ ] Add `BOT_TALK_TOKEN` to `~/messages/config/config.env` (the config.env the mirror.py reads)
- [ ] Add token reading to `bot_talk_mirror.py` using the existing `_read_config_env` helper
- [ ] Pass `X-Bot-Token` header in `_try_http()`
- [ ] Update bot-talk-poller task to read token from config.env as fallback (if token file missing)

## Valid API format

POST /message: `{"sender": "SaharLobster", "tier": "TIER-0", "genre": "status-update", "content": "..."}`
Headers: `X-Bot-Token: <token>`

sayhar:
## What and why

The bot-talk server at 46.224.41.108:4242 now requires `X-Bot-Token` authentication, but `bot_talk_mirror.py` was sending unauthenticated POSTs. Every mirror attempt since the auth requirement was added has silently fallen through to the SSH fallback (or local log), meaning Albert's Lobster has not been seeing Sahar's Lobster's messages.

This PR also ships the `gmail-email-pipeline` scheduled job that was requested alongside the auth fix.

### Fix 1: `src/mcp/bot_talk_mirror.py`

- Reads `BOT_TALK_TOKEN` from env / config.env using the existing `_read_config_env` helper (same pattern as `BOT_TALK_HTTP_URL`)
- New pure function `_build_auth_headers()` returns `{"X-Bot-Token": token}` when the token is set, or `{}` when empty — keeps the token concern isolated and testable
- `_try_http()` passes these headers on every POST

The existing resilience chain (HTTP → SSH → local log) is unchanged.

### Fix 2: `scheduled-tasks/tasks/bot-talk-poller.md` and `bot-talk-poller-fast.md`

Both poller tasks previously hard-required the legacy `~/lobster-workspace/data/bot-talk-token.txt` file and failed if it was missing. Updated to a three-path lookup chain:

1. `~/lobster-workspace/data/bot-talk-token.txt` (legacy, preserved for compat)
2. `BOT_TALK_TOKEN` in `~/messages/config/config.env`
3. `BOT_TALK_TOKEN` in `~/lobster-config/config.env`

### Feature 3: `scheduled-tasks/tasks/gmail-email-pipeline.md`

New 10-minute cron job that polls Gmail for unread inbox emails, applies a sensitivity filter (skips personal/family/health topics), looks up senders in Twenty CRM, creates/updates contacts (business context only), drafts suggested replies, pings Albert via bot-talk if the email mentions Albert Alexander, and writes a Telegram summary to Sahar via the inbox file queue. Tracks processed IDs in `~/lobster-workspace/data/gmail-processed.json` (capped at 1000).

## Functional patterns used

- Pure builders (`_build_auth_headers`, `_build_http_payload`, `_build_ssh_log_line`) — no I/O, fully testable in isolation
- Explicit fallback chain over exception-driven control flow
- Immutable lookup tables for filter sets

## Tests run

- [x] Direct Python assertions against `bot_talk_mirror._build_auth_headers` and `_try_http` — 4 assertions, all passed
- [ ] `uv run pytest tests/unit/test_mcp_server/test_bot_talk_mirror.py` — blocked: pre-existing conftest breakage (`_DEBUG_RESOLVED` attribute removed from `inbox_server.py` on main but conftest still patches it; affects all tests in this module on main before this PR too)
- [ ] Live bot-talk POST — skipped: requires production environment

**Pre-existing test breakage note:** The conftest issue is on `main` before this PR. Not introduced here.

Closes #999
Relates to #1000

### DIFF
--- a/scheduled-tasks/tasks/bot-talk-poller-fast.md
+++ b/scheduled-tasks/tasks/bot-talk-poller-fast.md
@@ -15,11 +15,17 @@ conversation in near-real-time.
 ## Authentication
 
 The bot-talk API requires a shared-secret header on all requests (except /health).
-Read the token from `~/lobster-workspace/data/bot-talk-token.txt` and include it as:
+Read the token using this lookup chain (first non-empty value wins):
+
+1. `~/lobster-workspace/data/bot-talk-token.txt` (legacy token file)
+2. `BOT_TALK_TOKEN` key in `~/messages/config/config.env`
+3. `BOT_TALK_TOKEN` key in `~/lobster-config/config.env`
+
+Include it as:
 
     X-Bot-Token: <token>
 
-If the token file is missing, call write_task_output with status "failed" and exit.
+If the token cannot be found via any of the above paths, call write_task_output with status "failed" and exit.
 
 ## State File
 
--- a/scheduled-tasks/tasks/bot-talk-poller.md
+++ b/scheduled-tasks/tasks/bot-talk-poller.md
@@ -15,16 +15,53 @@ self-cools and this job confirms the quiet state each hour.
 ## Authentication
 
 The bot-talk API requires a shared-secret header on all requests (except /health).
-Read the token from `~/lobster-workspace/data/bot-talk-token.txt` and include it as:
+Read the token using this lookup chain (first non-empty value wins):
+
+1. `~/lobster-workspace/data/bot-talk-token.txt` (legacy token file)
+2. `BOT_TALK_TOKEN` key in `~/messages/config/config.env`
+3. `BOT_TALK_TOKEN` key in `~/lobster-config/config.env`
+
+Include it as:
 
     X-Bot-Token: <token>
 
 Example (Python):
-    token = open(os.path.expanduser("~/lobster-workspace/data/bot-talk-token.txt")).read().strip()
-    headers = {"X-Bot-Token": token}
-    resp = requests.get("http://46.224.41.108:4242/messages", headers=headers, ...)
+```python
+def _load_bot_talk_token() -> str:
+    import os
+    from pathlib import Path
 
-If the token file is missing, log an error and call write_task_output with status "failed".
+    # 1. Legacy token file
+    token_file = Path.home() / "lobster-workspace" / "data" / "bot-talk-token.txt"
+    if token_file.exists():
+        token = token_file.read_text().strip()
+        if token:
+            return token
+
+    # 2. config.env files
+    for config_path in [
+        Path.home() / "messages" / "config" / "config.env",
+        Path.home() / "lobster-config" / "config.env",
+    ]:
+        if config_path.exists():
+            for line in config_path.read_text().splitlines():
+                line = line.strip()
+                if line.startswith("BOT_TALK_TOKEN="):
+                    value = line.split("=", 1)[1].strip().strip('"').strip("'")
+                    if value:
+                        return value
+
+    return ""
+
+token = _load_bot_talk_token()
+if not token:
+    # log error and write failed output
+    ...
+headers = {"X-Bot-Token": token}
+resp = requests.get("http://46.224.41.108:4242/messages", headers=headers, ...)
+```
+
+If the token cannot be found via any of the above paths, log an error and call write_task_output with status "failed".
 
 ## State File
 
--- a/scheduled-tasks/tasks/gmail-email-pipeline.md
+++ b/scheduled-tasks/tasks/gmail-email-pipeline.md
@@ -1,0 +1,223 @@
+# Gmail Email Pipeline
+
+**Job**: gmail-email-pipeline
+**Schedule**: Every 10 minutes (`*/10 * * * *`)
+
+## Context
+
+You are running as a scheduled task. Poll Gmail for new unread emails, apply a
+sensitivity filter, look up senders in Twenty CRM, create/update contacts,
+draft suggested replies, notify via bot-talk when Albert Alexander is mentioned,
+and send Sahar a Telegram summary.
+
+## Configuration
+
+- **Gmail**: use `gws` CLI (already authenticated)
+- **Twenty CRM API**: POST to `https://honest-navy-moose.twenty.com/graphql`
+  - API key: read `TWENTY_API_KEY` from `~/lobster-config/config.env`; if missing,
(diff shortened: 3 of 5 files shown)


sayhar (commented):
## Design Review — PR #1001

Overall: **NEEDS CHANGES** (one blocking issue; several minor items worth addressing before merge)

---

### Fix 1: bot_talk_mirror.py auth header

**Auth header approach — correct.**
`_build_auth_headers()` is a pure function with no I/O: easy to test, easy to reason about. The existing `_read_config_env` helper is reused for the token read, which is the right pattern. Module-level constant `BOT_TALK_TOKEN` follows the same pattern as `BOT_TALK_HTTP_URL` directly above it.

**Token read is safe** — env var takes priority, falls back to config file, falls back to `""`. Empty string is treated as "no token", which causes auth headers to be omitted rather than sending a blank header. That is correct behavior.

**Message format** — the existing `_build_http_payload` (not changed in this PR) already produces `sender/tier/genre/content`. The bot-talk ping added in gmail-email-pipeline (Step 8) manually constructs the same shape, which is consistent.

**Tests** — the new `TestBuildAuthHeaders` and `TestTryHttp` additions are well-structured. Patching `btm.BOT_TALK_TOKEN` at the module attribute level is the correct way to test module globals. The context-manager mock for `httpx.Client` matches what the real code does. No issues here.

**Minor:** `_build_auth_headers` has return type `dict` — consider `dict[str, str]` for specificity, though this is cosmetic.

---

### Fix 2: Poller token lookup chain

**Lookup chain is correct and backward-compatible.** Legacy file path is tried first, so existing installs with `bot-talk-token.txt` are unaffected. The inline Python example is clear and matches the described behavior.

**One nit:** the example code uses `line.split("=", 1)[1].strip().strip('"').strip("'")` — this handles both quoted and unquoted values. That is fine for a `.env`-style file, but it will silently accept a line like `BOT_TALK_TOKEN=` (empty after strip) as a miss, which is the desired behavior. Good.

---

### Feature 3: gmail-email-pipeline — BLOCKING ISSUE

#### Hardcoded API key (blocking)

The configuration section reads:

> if missing, use hardcoded fallback `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...`

**This is a security problem.** A JWT API key checked into the repository is permanently exposed in git history. Even if `~/lobster-config/config.env` is present on every active install, the hardcoded fallback means any future fork, clone, or CI checkout gets a working credential without needing the config file. The token should be removed from the task definition. The correct behavior when `TWENTY_API_KEY` is absent is to log an error, skip CRM steps, and continue (or fail loudly) — not to fall back to a committed secret.

**Action required:** Strip the hardcoded JWT from `gmail-email-pipeline.md` before merge. Add `TWENTY_API_KEY` to whatever secrets setup documentation / install guide covers `config.env`.

#### Dedup / processed-ID tracking — good

The approach is sound: load JSON on startup, skip IDs already in the list, append new IDs at the end of the run, cap at 1000 entries, write atomically via `.tmp` + rename. The cap prevents unbounded growth. Skipped-personal emails are added to `processed_ids` so they are not re-evaluated — this is correct.

**One edge case:** if the process is killed between Step 2 (fetch) and Step 10 (save), an email will be re-processed on the next run. This is acceptable given the 10-minute schedule and idempotent CRM upsert, but worth noting in comments.

#### Sensitivity filter — reasonable

The "when in doubt, skip" default is the right call for a privacy-sensitive filter. The personal/professional categories are well-defined. No issues.

#### Security: token exposure / sensitive data logging

- The bot-talk token is read from config, not logged anywhere in the new code. Good.
- The Twenty API key fallback (see above) is the only credential concern.
- Email body content is summarized and sent to Telegram — the sensitivity filter is the guard here. This is acceptable given the filter's "skip when in doubt" default.
- No email body content appears to be logged to disk in plaintext (only structured output via `write_task_output`). Good.

#### Error handling — acceptable

The tiered error handling (gws failure = hard stop; CRM/bot-talk/Telegram failure = log and continue) matches the stated design goal of non-blocking downstream steps. One addition worth considering: if the processed-ID file write fails (Step 10), log the failure explicitly rather than silently losing dedup state.

#### Lobster conventions

- `gws` CLI is used for Gmail rather than a Python library — consistent with the project's pattern of using authenticated CLI tools.
- Bot-talk ping uses the same `sender/tier/genre/content` payload shape as `bot_talk_mirror.py`.
- `write_task_output` is called at the end. Good.
- The Telegram message is written to `~/messages/inbox/` as a JSON file, which is the correct inbox-file-queue pattern. The `subtype: "scheduler_tick"` label is appropriate.
- The PR description notes `uv run` awareness; the task definition itself doesn't invoke Python directly (it describes logic for the LLM to execute), so no `uv` invocation is needed in the markdown.

#### Minor issues

- **Step 8 bot-talk payload** uses `"tier": "TIER-0"` hardcoded. If `BOT_TALK_TIER` is configurable elsewhere, consider referencing the same source of truth. Low priority.
- **Step 9 summary format** — if zero emails processed, no Telegram message is sent. That is good; avoids noise. But consider whether a weekly "all-quiet" summary would be useful (separate job concern, not a blocker).
- The `chat_id: 8305714125` is Sahar's Telegram ID hardcoded in the task definition. This is consistent with how other scheduled tasks handle user-specific routing in this codebase, so it is acceptable here.

---

### Summary of required changes

| # | Severity | Item |
|---|----------|------|
| 1 | **Blocking** | Remove hardcoded Twenty CRM JWT from `gmail-email-pipeline.md`; require `TWENTY_API_KEY` from config only |
| 2 | Minor | Document `TWENTY_API_KEY` in secrets/install guide |
| 3 | Nit | Narrow `_build_auth_headers` return type to `dict[str, str]` |
| 4 | Nit | Add comment in Step 10 about mid-run kill edge case for dedup |

The auth fix and poller token lookup are solid and ready to merge. The gmail pipeline is well-structured and the logic is sound — just needs the hardcoded credential removed before it is safe to ship.

PR #1004, opened 2026-03-27
feat: gmail pipeline — debug-only step updates + Drive file name index

### LINKED ISSUE #1003: feat: gmail pipeline — debug-only step updates + Drive file name index
## Summary

Two improvements to the Gmail email pipeline defined in `scheduled-tasks/tasks/gmail-email-pipeline.md`.

### Change 1: Step-by-step Telegram updates → debug mode only

The pipeline currently writes a Telegram message after each processing step (email fetched, CRM checked, Drive searched, Albert queried, draft created). In production these generate noise. They should only fire when `LOBSTER_DEBUG=true` is set in the environment. When unset or `false`, the pipeline runs silently and only sends the final per-email summary.

### Change 2: Google Drive context — file name index approach

The current Drive search uses a hardcoded keyword search against `fullText` by sender name / email domain. This misses files named in ways that don't match the sender exactly.

Replace with a two-phase approach:

1. **Index phase**: Download all Drive file titles/names (not content) to a local index at `~/lobster-workspace/data/drive-file-index.json`. Refresh the index on each pipeline run (or on a schedule — implementation detail left to engineer).
2. **Match phase**: When processing an email, use an LLM call or fuzzy string match to find relevant files from the index rather than issuing a Drive fullText search.
3. Include matched files as named read-only links in the draft, using the existing `/view` link pattern already in the task.

## Acceptance criteria

- [ ] Step-by-step Telegram writes are gated on `LOBSTER_DEBUG=true`
- [ ] Final per-email Telegram summary is unaffected (still sends in production)
- [ ] Drive index is fetched/synced to a local JSON file (file name + webViewLink, no content)
- [ ] Email processing matches against the local index (LLM skim or fuzzy match) rather than issuing a live Drive fullText search
- [ ] Matched Drive files are included in the draft as named read-only links
- [ ] No behavior change to CRM, Albert bot-talk, or sensitivity filter logic

sayhar:
## What this solves

The Gmail pipeline had two problems:

1. Step-by-step Telegram messages fired unconditionally on every run. In production this generates noise for every email processed — the recipient sees a stream of \"email fetched / CRM checked / draft composed\" messages before the actual summary arrives.

2. The Drive context step built a full local index of all Drive file names (paginated fetch + local JSON), then used either an LLM skim or fuzzy matching to find relevant files. This added latency and complexity for every pipeline run, even when most emails have no relevant Drive files.

## What changed

**Change 1 — debug-mode gating (unchanged from original PR):**
All step-by-step Telegram writes are gated on \`LOBSTER_DEBUG=true\`. The \`debug_step()\` helper is a no-op unless the env var is set. The final per-email summary always fires when at least one email is processed.

**Change 2 — Drive direct keyword search (replaces index approach):**
Instead of building a full local Drive file index and running LLM/fuzzy matching against it, the pipeline now does a single \`gws drive files list\` call per email using \`q=\"fullText contains 'KEYWORD' and trashed=false\"\`. The keyword is derived from the sender's display name; falls back to the email domain (without TLD) for business domains if the name search returns zero results.

This eliminates: the pagination loop, the \`drive-file-index.json\` local file, the LLM skim step, and the fuzzy match fallback. Drive context is still non-blocking — a failed search logs a warning and the pipeline continues.

Steps renumbered from 12 to 11. The \`drive-file-index.json\` config entry has been removed.

## What is not changed

- Sensitivity filter logic
- CRM lookup / create / update flow
- Albert bot-talk notification
- Processed-IDs tracker and atomic write pattern
- The final Telegram summary format

## Functional patterns used

Pure function \`drive_keyword_search(keyword) -> list[dict]\` with a single responsibility and explicit empty-list return on failure. Keyword derivation is a pure transformation (sender name/domain → search string). No shared mutable state — no local index file to manage.

## Tests run

- [ ] Live pipeline run — blocked: requires \`gws\` auth and a live Gmail inbox; not available in this environment. No behavior change to CRM, bot-talk, or summary delivery — the Drive section is a targeted replacement.
- [ ] Linter / type check — skipped: task file is Markdown instructions, not executable Python. No Python files were added or modified.

**Blocked items needing attention before merge:** A smoke test confirming \`gws drive files list --params '{"q":"fullText contains ..."}'\` returns results against the live Drive account would confirm auth and field names are correct.

Closes #1003

### DIFF
--- a/scheduled-tasks/tasks/gmail-email-pipeline.md.template
+++ b/scheduled-tasks/tasks/gmail-email-pipeline.md.template
@@ -7,7 +7,7 @@
 
 You are running as a scheduled task. Poll Gmail for new unread emails, apply a
 sensitivity filter, look up senders in Twenty CRM, create/update contacts,
-draft suggested replies, notify via bot-talk when the configured remote identity is
+draft suggested replies using Drive context, notify via bot-talk when the configured remote identity is
 mentioned, and send the owner a Telegram summary.
 
 ## Configuration
@@ -23,6 +23,9 @@ mentioned, and send the owner a Telegram summary.
 - **Processed emails tracker**: `~/lobster-workspace/data/gmail-processed.json`
 - **Remote identity to watch for mentions**: `{{REMOTE_LOBSTER_IDENTITY}}`
   (set `BOT_TALK_REMOTE_IDENTITY` in config.env)
+- **Drive file index**: `~/lobster-workspace/data/drive-file-index.json`
+- **Debug mode**: step-by-step Telegram updates are sent only when `LOBSTER_DEBUG=true`
+  is set in the environment. Check: `os.environ.get("LOBSTER_DEBUG", "").lower() == "true"`
 
 ## Sensitivity Filter
 
@@ -51,7 +54,44 @@ Read `~/lobster-workspace/data/gmail-processed.json`. If missing, treat as empty
 {"processed_ids": []}
 ```
 
-### Step 2: Poll Gmail for unread inbox emails
+### Step 2: Sync Drive file name index
+
+Fetch the names and view links of all non-trashed files in Google Drive and write
+them to the local index at `~/lobster-workspace/data/drive-file-index.json`.
+
+This replaces the per-email Drive keyword search. The index is rebuilt on every
+pipeline run so it stays current without a separate cron job.
+
+```bash
+gws drive files list --params '{"q": "trashed=false", "fields": "nextPageToken,files(id,name,webViewLink)", "pageSize": 1000}'
+```
+
+Paginate using `nextPageToken` until exhausted. Build a list of objects:
+```json
+[
+  {"id": "FILE_ID", "name": "File Name Here", "webViewLink": "https://drive.google.com/file/d/.../view"},
+  ...
+]
+```
+
+Write atomically (write to `.tmp`, then rename):
+```python
+index_path = Path.home() / "lobster-workspace" / "data" / "drive-file-index.json"
+tmp_path = index_path.with_suffix(".json.tmp")
+tmp_path.write_text(json.dumps({"files": file_list, "synced_at": time.time()}))
+tmp_path.rename(index_path)
+```
+
+If `gws drive` fails, log a warning and continue with an empty file list (Drive
+context will be absent from drafts this run, but the rest of the pipeline should proceed).
+
+**Debug step update** (only if `LOBSTER_DEBUG=true`): after syncing the index,
+write a Telegram step message:
+```
+[gmail-pipeline] Drive index synced: {N} files indexed
+```
+
+### Step 3: Poll Gmail for unread inbox emails
 
 ```bash
 gws gmail users messages list --params '{"userId": "me", "q": "is:unread in:inbox", "maxResults": 10}'
@@ -59,7 +99,7 @@ gws gmail users messages list --params '{"userId": "me", "q": "is:unread in:inbo
 
 Parse the response. For each message ID that is NOT in the processed_ids list:
 
-### Step 3: Fetch full email content
+### Step 4: Fetch full email content
 
 ```bash
 gws gmail users messages get --params '{"userId": "me", "id": "<MSG_ID>", "format": "full"}'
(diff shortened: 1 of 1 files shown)


sayhar (commented):

## Summary

The two core changes (debug gating, Drive index) are structurally sound and solve real problems. Several specific gaps need to be addressed before this is production-safe.

---

## Issues

### 1. LLM skim — name-collision matching (medium)

The LLM skim returns file names as plain text, one per line. The lookup back into the index is done by name (`"From the matched file names, look up their id and webViewLink in the index"`). If two files share a name (Drive allows duplicates), the pipeline will match the wrong file or pick arbitrarily. Recommend matching on `id` instead: ask the LLM to return the `id` field, or build a name→id map that handles duplicates explicitly (e.g., include all matches, or log a warning and skip).

### 2. LLM skim — no file-list size guard (medium)

The index is passed to the LLM as a newline-separated list of all file names with no truncation. A Drive with thousands of files will overflow context or produce high token cost on every email. Add a guard: if `len(file_list) > N` (suggest 500), fall back to Option B (difflib) rather than passing the full list to the LLM. Or pre-filter by a cheap substring pass before the LLM call.

### 3. Pagination loop not specified (low-medium)

Step 2 describes the initial `gws drive files list` call and mentions "paginate using nextPageToken until exhausted", but gives no pseudocode or loop structure. An agent implementing from these instructions could easily fetch only the first page. Add explicit loop pseudocode:
```python
all_files = []
params = {"q": "trashed=false", "fields": "nextPageToken,files(id,name,webViewLink)", "pageSize": 1000}
while True:
    response = gws_drive_files_list(params)
    all_files.extend(response["files"])
    if "nextPageToken" not in response:
        break
    params["pageToken"] = response["nextPageToken"]
```

### 4. debug_step() subtype value (low)

The helper writes `"subtype": "pipeline_step"`. Verify this subtype is recognized by the dispatcher / inbox router. If the dispatcher filters on known subtypes and drops unknowns, these debug messages will silently vanish. Either confirm `pipeline_step` is whitelisted, or use a subtype already known to be delivered (e.g., `"text"` with a distinct prefix, or document that the dispatcher passes unknown subtypes through).

### 5. Drive index written before any email processing — failure mode underspecified (low)

If `gws drive` fails, the spec says "log a warning and continue with empty file list." This is correct. But the atomic write pattern (`.tmp` then rename) means a partial failure could leave a stale `.json.tmp` file on disk, which would not affect this run but could confuse a future operator. Consider removing the `.tmp` file in the except block, or noting this explicitly.

---

## What looks good

- The `debug_step()` helper pattern is clean: single responsibility, early-return guard, correct DEBUG check at module level so the env var is read once.
- Atomic index write (write-to-tmp then rename) is correct.
- Drive index as a per-run rebuild (rather than a separate cron) is the right tradeoff for a pipeline that already runs frequently.
- The fallback chain (LLM → difflib) is a good defensive pattern.
- Step ordering is consistent; all step-number references within the document were updated.
- The "zero emails processed → no Telegram message" rule is preserved.

---

The pagination and name-collision issues are the most likely to cause silent data loss or incorrect behavior. Recommend addressing those two before merge; the others can be follow-ups if preferred.

sayhar (commented):
🤖🦞 Lobster (reviewer):

## PR 1004 — feat: gmail pipeline — debug-only step updates + Drive file name index

**Verdict: NEEDS-WORK** (minor)

### What this does

1. Gates all step-by-step Telegram writes behind `LOBSTER_DEBUG=true` — correct pattern.
2. Replaces the full Drive file index approach with a direct per-email keyword search via `gws drive files list` — correct architectural simplification.

### Debug-mode gating

- `debug_step()` helper function with `if not DEBUG: return` early exit — correct.
- `DEBUG = os.environ.get("LOBSTER_DEBUG", "").lower() == "true"` — correct, matches the pattern used elsewhere.
- All intermediate step writes now conditional — verified against each step in the diff.
- Final summary (Step 11) remains unconditional — correct; the zero-emails guard is preserved.

### Drive direct keyword search

- `gws drive files list` with fullText contains query — correct API usage.
- Fallback chain: display name → email domain (without TLD, only for non-generic domains) — correct.
- Error handling: Drive search failure is non-blocking (log warning, continue with empty list) — correct.
- 10-result page size is appropriate.
- `view` link (not edit link) explicitly required — correct.

### Issue: step numbering is broken

The diff changes step 7 ("Draft a suggested reply") to step 7 ("Find relevant Drive files"), then adds step 9 ("Draft a suggested reply") — but the step numbers are inconsistent. The "remote identity mention" check is at step 8 in the original and step 9 in the diff (correct shift). The final summary step was "Step 9" and becomes "Step 11" in the diff, but "Step 11" for "Save processed email IDs" is also numbered as "Step 11" — creating a duplicate. This is a doc numbering error that would confuse the subagent executing the task.

Verify that all step numbers are sequential (8, 9, 10, 11) with no duplicates.

### Note

The `chat_id` in `debug_step()` is hardcoded to `8305714125`. This should use a config variable (`LOBSTER_ADMIN_CHAT_ID` or similar) rather than a literal. Hardcoded chat IDs are a maintenance issue if the user's chat ID ever changes.

**Block:** Fix the step numbering duplication and the hardcoded chat_id before merge.

**Blocked by:** Live gws auth + Drive test (noted in PR as a known pre-merge requirement).

PR #1076, opened 2026-03-28
fix: treat stale JSONL files as dead in _stored_session_is_alive() (#1009)

### LINKED ISSUE #1009: Fix: stale dispatcher-session-id causes on-compact.py to skip compact-reminder
🤖🦞 Lobster (ops):

## Root Cause

`_stored_session_is_alive()` in `on-compact.py` checks whether the JSONL session file exists on disk. It returns `True` for JSONL files from old sessions that still exist, causing `is_dispatcher()` to misidentify new sessions as subagents.

As a result:
- The current dispatcher session (`e262feb3-9fd5-4904-96ba-82bd5b2041bb`) was not being recognized as the dispatcher
- `dispatcher-session-id` still contained the stale session ID `249c721e` from March 17

## Impact

- `compact-reminder` is never written → `compact-catchup` is never triggered → the dispatcher loses situational awareness after compaction
- This has been silently broken since March 17

## Fix 1 (immediate, done)

Updated `dispatcher-session-id` to the current session ID:

```bash
echo "e262feb3-9fd5-4904-96ba-82bd5b2041bb" > /home/lobster/messages/config/dispatcher-session-id
```

## Fix 2 (permanent)

Update `write-dispatcher-session-id.py` (or the equivalent session tracking logic) to treat JSONL files older than 7 days as dead. Session liveness should be based on file age, not mere existence.

Suggested check in `_stored_session_is_alive()`:
```python
import os, time
MAX_SESSION_AGE_DAYS = 7
if os.path.exists(jsonl_path):
    age_days = (time.time() - os.path.getmtime(jsonl_path)) / 86400
    if age_days > MAX_SESSION_AGE_DAYS:
        return False  # treat stale file as dead
    return True
return False
```

## Fix 3 (resilience)

Have `on-compact.py` inject `sys.dispatcher.bootup.md` content regardless of marker file state, so even if the session marker is wrong, the dispatcher still receives the correct context after compaction.

This decouples the compact-reminder from the (potentially stale) session ID check, making the system more robust against stale state.

sayhar:
## Summary

- Adds a 7-day age threshold to `_stored_session_is_alive()` in `write-dispatcher-session-id.py`
- JSONL session files older than `_MAX_SESSION_AGE_DAYS` are now treated as dead, regardless of whether they exist on disk
- Also adds `import time` (missing) and documents the fix in the module-level docstring

## Root cause

`_stored_session_is_alive()` returned `True` for any existing `.jsonl` file, even ones from sessions that ended weeks ago. Claude Code does not auto-delete old session files, so a stale JSONL caused the hook to classify a fresh dispatcher restart as a subagent:

1. New dispatcher starts with a new `session_id`
2. Stored `dispatcher-session-id` still contains the stale (old) session ID
3. `_stored_session_is_alive(stored_id)` finds the old `.jsonl` on disk → returns `True`
4. `_is_dispatcher_session()` concludes: "stored session is still alive, so this new session must be a subagent"
5. The hook does NOT update `dispatcher-session-id` → compact-reminders never fire → dispatcher loses context after every compaction

This was silently broken from March 17 through March 28 (11 days).

## Fix

```python
_MAX_SESSION_AGE_DAYS = 7

for jsonl in projects_dir.glob(f"*/{stored_session_id}.jsonl"):
    age_days = (time.time() - jsonl.stat().st_mtime) / 86400
    if age_days > _MAX_SESSION_AGE_DAYS:
        return False  # stale file — treat stored session as dead
    return True
```

## Why NOT workdir-based detection

The original issue description explored whether `os.getcwd()` could distinguish dispatcher from subagents. Investigation showed this is NOT reliable:

- The dispatcher runs in `/home/lobster/lobster-workspace/` (set by systemd)
- **Subagents also run in `/home/lobster/lobster-workspace/`** — confirmed by inspecting `.jsonl` session files and by the fact that this investigative subagent itself ran in that directory
- Out of 271 sessions in the `lobster-workspace` project dir: 215 dispatcher, **56 subagents** — all with identical CWDs

The workdir approach would produce false positives (subagents identified as the dispatcher). The stale-file age check is the correct, targeted fix.

## Test plan

- [ ] Verify `_stored_session_is_alive()` returns `False` for a session JSONL older than 7 days
- [ ] Verify `_stored_session_is_alive()` returns `True` for a session JSONL newer than 7 days
- [ ] Verify the hook correctly writes a fresh `dispatcher-session-id` after a crash/restart scenario (old JSONL > 7 days old)
- [ ] Confirm `import time` is present (was missing before this PR)

Closes #1009

### DIFF
--- a/hooks/write-dispatcher-session-id.py
+++ b/hooks/write-dispatcher-session-id.py
@@ -44,6 +44,15 @@ exists and correctly classifies the new session as the replacement dispatcher.
 The primary defence against this scenario is `lobster stop` (and `restart`)
 clearing the marker file; this check is a secondary safety net.
 
+## Stale JSONL file fix (issue #1009)
+
+`_stored_session_is_alive()` previously returned True for any existing JSONL
+file, even ones from sessions that ended weeks ago. This caused the hook to
+misclassify a fresh dispatcher restart as a subagent whenever the stale JSONL
+file happened to still exist on disk (Claude Code does not automatically delete
+old session files). The fix adds an age check: JSONL files older than 7 days
+are treated as dead regardless of whether they exist on disk.
+
 ## settings.json configuration
 
 Add this to ~/.claude/settings.json under "hooks" → "SessionStart":
@@ -67,6 +76,7 @@ via a "compact" matcher — the two hooks can coexist safely.
 import json
 import os
 import sys
+import time
 from pathlib import Path
 
 # Allow imports from both the hooks directory (session_role) and src/agents (session_store).
@@ -79,15 +89,27 @@ import session_role  # noqa: E402 — path insert must precede this
 from agents import session_store  # noqa: E402
 
 
+_MAX_SESSION_AGE_DAYS = 7  # JSONL files older than this are treated as dead sessions.
+
+
 def _stored_session_is_alive(stored_session_id: str) -> bool:
-    """Return True if the stored dispatcher session's JSONL file still exists.
+    """Return True if the stored dispatcher session's JSONL file still exists
+    and is recent enough to belong to a live session.
 
     Claude Code stores each session's conversation as
     ~/.claude/projects/<workspace-slug>/<session-id>.jsonl. If the file is
     gone, the session has ended and can no longer be the active dispatcher.
 
+    Age check (fix for issue #1009): JSONL files that exist but are older than
+    _MAX_SESSION_AGE_DAYS are treated as dead. Without this check, stale JSONL
+    files from sessions that ended weeks ago cause _is_dispatcher_session() to
+    misidentify the current session as a subagent — because the stored session
+    ID differs from the new session's ID, and _stored_session_is_alive() returns
+    True (file exists), so the new session is classified as a subagent rather
+    than the replacement dispatcher.
+
     Falls back to True (conservative / assume alive) if the project directory
-    cannot be determined or the JSONL file is not found via glob.
+    cannot be determined or file stat fails.
     """
     try:
         projects_dir = Path(os.path.expanduser("~/.claude/projects"))
@@ -95,6 +117,13 @@ def _stored_session_is_alive(stored_session_id: str) -> bool:
             return True  # can't determine — assume alive (conservative)
         # Search all workspace subdirectories for the session file.
         for jsonl in projects_dir.glob(f"*/{stored_session_id}.jsonl"):
+            # File exists — check its age before declaring it alive.
+            try:
+                age_days = (time.time() - jsonl.stat().st_mtime) / 86400
+                if age_days > _MAX_SESSION_AGE_DAYS:
+                    return False  # stale file — treat stored session as dead
+            except OSError:
+                pass  # stat failed — assume alive (conservative)
             return True
         # No JSONL file found anywhere — the stored session ended.
         return False


sayhar (commented):

## Summary verdict

The PR correctly diagnoses the confirmed March 17 incident (stale 11-day JSONL misclassifying dispatcher as subagent) and the implementation is clean. However, the fix does not fully address the root cause — it solves the specific observed case but leaves the general compaction scenario unresolved.

## What the fix gets right

- Correctly identifies that `_stored_session_is_alive()` was doing "file exists?" when it should be doing "file is recent enough to plausibly still be the active session?"
- Clean implementation: age check is inside the existing glob loop, OSError on stat is handled conservatively, `import time` was correctly added
- Documentation and docstring are thorough and accurate
- The workdir-based detection was correctly rejected — 56 confirmed subagents sharing `/home/lobster/lobster-workspace/` CWD makes that signal useless

## Core gap: the fix doesn't cover same-week compaction

The compaction scenario does **not** require the old session to be old. Sequence:

1. Dispatcher starts Monday at 9am, session ID = `A`, writes `dispatcher-session-id = A`
2. Monday at 9:10am, context fills, Claude Code compacts → new session ID = `B`
3. `_stored_session_is_alive("A")` finds `A.jsonl`, mtime is 10 minutes ago → age check passes, returns `True`
4. `_is_dispatcher_session()`: stored=A, current=B, stored is "alive" → classifies as subagent
5. Bug reproduces identically

The 7-day threshold only prevents the specific form where the stale session sat untouched for >7 days. Any compaction within a single session-week will hit the same misclassification.

The actual root cause is that `_stored_session_is_alive()` answers the wrong question. It can confirm "this session's JSONL exists and is recent" but it cannot confirm "this session is still the *active dispatcher*" because a fresh post-compaction session inherits neither the old session ID nor a way to detect that the old session ended via compaction (vs. crash/restart).

## What would fix this properly

Issue #1009 already names the right resilience layer (Fix 3): have `on-compact.py` inject `sys.dispatcher.bootup.md` regardless of the marker file state. That decouples compact-reminder delivery from the session ID matching logic entirely. This PR should either include that or it should be opened as a follow-up before this is merged.

Alternatively, the session ID marker update should happen at every SessionStart unconditionally when the stored session's JSONL is ≤ some threshold of age AND the stored ID ≠ current ID — recognizing that a mismatch on a recent file means compaction happened, not "I'm a subagent."

## Minor nit

The `pass` on OSError in the age check preserves the `return True` on the next line — which is the correct conservative fallback. But this is subtle; a comment noting "falls through to return True below" would help future readers.

## Recommendation

Merge this as a **partial fix** that resolves the long-lived stale session case (11 days in the incident), with a follow-up issue/PR tracking the intra-week compaction scenario. The PR description should be updated to note this limitation explicitly so it's not assumed to be a complete fix for #1009.

Do not close #1009 on merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): UPDATED ASSESSMENT — safe to merge as partial fix

## Re-examination summary

I re-examined this PR in light of a follow-up question: was the NEEDS-WORK verdict based on a wrong assumption that `agent_id` is available in SessionStart hook payloads?

Short answer: the previous review didn't mention agent_id at all — it was based on the same-week compaction gap, which is real. However, re-reading both the review and the PR, the previous reviewer's own recommendation was "merge as partial fix with a follow-up issue tracking intra-week compaction." That recommendation stands. The NEEDS-WORK label was stricter than the reviewer's own text warranted.

## What the same-week compaction gap actually is

The gap is real but pre-existing. Before this PR, **every** compaction (including same-day) would cause misclassification: `_stored_session_is_alive()` returned True for any existing file, so a freshly-compacted session's new ID would always be classified as a subagent. This PR fixes the >7-day stale case. The intra-week case remains broken in the same way it was before.

Critically: `on-compact.py` itself already documents this pre-existing limitation in its own comments (lines 291–299: "is_dispatcher() would return False because context compaction assigns a new session_id"). The same-week compaction problem is not introduced or worsened by this PR — it was already a known open issue before this branch was created.

## Why merging is correct

- The PR fixes the confirmed incident (11-day stale JSONL, March 17–28)
- It does not regress the same-week case (already broken before this branch)
- The code is clean and correct for what it claims to fix
- The `import time` addition and docstring updates are correct
- The OSError conservative fallback is correct (falls through to `return True`)

## Remaining gap (should stay open in #1009)

Same-week compaction misclassification is still unresolved. The right fix for that is Fix 3 from issue #1009: make `on-compact.py` inject the compact reminder unconditionally (decoupled from the marker file check), so the dispatcher always reorients after compaction regardless of session ID state. That should be a separate PR.

## Merge recommendation

**APPROVE as partial fix.** Update the PR description to note:
- This fixes the long-lived stale session case (>7 days)
- Same-week compaction misclassification is tracked separately in #1009
- Do not close #1009 on merge

No code changes needed on this branch.

PR #1092, opened 2026-03-29
fix: check_task_outputs shows unknown/no content

sayhar:
## Summary

- `check_task_outputs` was returning all entries as "unknown" job / "(no output)" due to a sort order bug combined with stale files from an old code path
- The `task-outputs/` directory contains ~25 letter-prefixed files (e.g. `voice-transcription-diagnosis_*.json`) written by an older `write_result` code path; in reverse lexicographic sort these appear before all `2026...` date-prefixed files and dominate every result page
- Those stale files use the `write_result` schema (`task_id`/`text`), not the `write_task_output` schema (`job_name`/`output`), so they render as "unknown" / "(no output)"

## Fix (two defenses)

1. **Sort by mtime instead of filename** — `write_task_output` files are always newest; they surface first regardless of filename convention
2. **Skip files missing `job_name`** — silently ignores any stale `write_result` artifacts in the directory

## Root cause reproduction

The `task-outputs/` directory has 28 non-date-prefixed files. In reverse lex sort, letter-prefixed names beat `2026...` date-prefixed names (letters > digits in ASCII). All 10 results from `check_task_outputs` came from those stale files which have no `job_name` or `output` field.

## Test plan

- [ ] Call `check_task_outputs()` after deploy — confirm real job names and output text appear
- [ ] Existing `TestCheckTaskOutputs` unit tests pass (test fixtures use proper `job_name` schema, unaffected by the `job_name` guard)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The two-defense fix is correct and well-motivated.

**Sort by mtime:** Switching from lexicographic filename sort to mtime sort is the right call. The root cause analysis is accurate — letter-prefixed filenames sort after digits in ASCII (uppercase/lowercase letters > digits), so stale artifacts dominated every result page. mtime sort is robust regardless of filename convention.

**Skip files missing job_name:** Clean, silent guard. Stale write_result artifacts (schema: task_id/text) have no business appearing in check_task_outputs results. The guard is minimal and doesn't break any current write_task_output consumers.

**Code quality:** Comments explain the why clearly. The fix is surgical — only the two affected lines in handle_check_task_outputs change, no collateral blast radius.

**Test coverage:** Existing TestCheckTaskOutputs uses proper job_name schema fixtures, so they remain valid. The fix is simple enough that the manual smoke test (call check_task_outputs after deploy, confirm real job names appear) is the right verification path.

No concerns. Ready for dogfooding + smoke test gate.

PR #1095, opened 2026-03-29
hook: block sync Agent/Task calls from dispatcher (#1094)

### LINKED ISSUE #1094: hook: block sync Agent calls from dispatcher
🤖🦞 Lobster (engineer): The dispatcher must always use run_in_background=True when calling the Agent tool. A PreToolUse hook should block calls where run_in_background is missing or False, when the calling session is the dispatcher. This enforces the 7-second rule at the infrastructure level rather than relying on LLM compliance.

sayhar:
## What this fixes

The dispatcher must always spawn subagents with `run_in_background=True`. A foreground Agent call blocks the main message-processing loop for the full duration of the subagent — potentially minutes — causing inbox staleness and health-check restarts. The `require-background-agent.py` hook already enforced this for `"Agent"` calls and was already registered in `settings.json`.

This PR closes two gaps:

1. **"Task" tool name not covered.** Older Claude Code versions use `"Task"` rather than `"Agent"` as the tool name when spawning subagents. The hook only checked for `"Agent"`, leaving those installs unprotected. Both names now trigger the same hard block (exit 2, stderr message) when called without `run_in_background: true` from the dispatcher.

2. **Smoke tests were testing a superseded design.** The existing smoke tests expected exit 1 and a stdout warning — behavior from a previous version of the hook that predated dispatcher/subagent discrimination. The current hook exits 2 and writes to stderr, and only fires for dispatcher sessions. The stale smoke tests would have silently passed against a hook that couldn't block anything. The smoke tests are rewritten to match the actual behavior, including subagent exemption and the "Task" tool name.

## What is not changed

- The hook file path and registration in `settings.json` are unchanged.
- `session_role.py` is unchanged — dispatcher detection logic is unaffected.
- The hook still fails open (exit 0) when the marker file is absent and no transcript signal is available — conservative behavior intentionally preserved.
- No changes to install scripts or migrations needed (hook already deployed).

## Functional patterns

Pure functions for detection (`is_dispatcher`), set-membership check for tool name matching (`AGENT_TOOL_NAMES`), fail-open default when session identity is ambiguous.

## Tests run

- [x] `uv run --no-project pytest tests/unit/test_hooks/test_require_background_agent.py tests/smoke/test_require_background_agent.py -v` — 19 passed, 0 failed

Closes #1094

### DIFF
--- a/tests/smoke/test_require_background_agent.py
+++ b/tests/smoke/test_require_background_agent.py
@@ -124,6 +124,9 @@ def test_dispatcher_foreground_agent_explicit_false_blocked():
     assert BLOCK_FRAGMENT in result.stderr, (
         f"Block message must appear on stderr.\nGot stderr: {result.stderr!r}"
     )
+    assert result.stdout == "", (
+        f"Expected empty stdout on block, got: {result.stdout!r}"
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -151,6 +154,9 @@ def test_dispatcher_foreground_agent_field_absent_blocked():
     )
     assert BLOCK_FRAGMENT in result.stderr, (
         f"Block message must appear on stderr.\nGot stderr: {result.stderr!r}"
+    )
+    assert result.stdout == "", (
+        f"Expected empty stdout on block, got: {result.stdout!r}"
     )
 
 
@@ -231,4 +237,7 @@ def test_dispatcher_task_tool_sync_blocked():
     )
     assert BLOCK_FRAGMENT in result.stderr, (
         f"Block message must appear on stderr.\nGot stderr: {result.stderr!r}"
+    )
+    assert result.stdout == "", (
+        f"Expected empty stdout on block, got: {result.stdout!r}"
     )


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Scope check**: 3 files changed — hooks/require-background-agent.py, tests/smoke/test_require_background_agent.py, tests/unit/test_hooks/test_require_background_agent.py. All changes are strictly within the hook and its tests. No bundled scope.

**Hook logic (require-background-agent.py)**

- `AGENT_TOOL_NAMES = {"Agent", "Task"}` — correct set membership check, covers both CC naming conventions.
- Early exits for non-agent tools and run_in_background=True are correct and ordered efficiently (cheapest checks first).
- `is_dispatcher(data)` check is the right gating condition — subagents fall through to exit 0.
- exit 2 + stderr is the correct Claude Code hard-block pattern.
- Error behavior: `is_dispatcher()` never raises — it uses layered try/except throughout session_role.py and returns False (subagent) as the default when the marker file is absent or unreadable. This is fail-open (allow), which is the documented conservative behavior and correct for a hook that trades safety for reliability. No uncaught exception path in the hook itself.

**Edge case: is_dispatcher() errors**

The fail-open choice is deliberate and well-documented in session_role.py. When the marker file is absent (e.g. fresh install, race at startup), the hook allows the call through rather than hard-blocking. This is correct — false negatives (allowing a foreground call once) are less harmful than false positives (blocking all agent calls because the marker file wasn't written yet).

**Tests**

Smoke tests (A1-A6): all six cases cover the right boundaries. A5 (subagent exemption) and A6 (Task tool name) are the two new cases matching the two PR gaps. The temp-dir HOME override to control the marker file is clean and correct. The old stale tests (expecting exit 1 + stdout) are fully replaced.

Unit tests: three new TestTaskToolName cases mirror A6 across sync-blocked, background-allowed, and subagent-allowed. Monkeypatching DISPATCHER_SESSION_FILE directly is the right unit test pattern here.

One minor note: the smoke tests don't assert absence of stdout in the block cases (A2, A3, A6) — they only assert stderr. Not a defect, but belt-and-suspenders completeness would add `assert result.stdout.strip() == ""` there. Not worth holding the PR.

**Verdict**: Ready to merge.

PR #1101, opened 2026-03-29
🤖🦞 Lobster (engineer): fix: ghost_detector and oom_check run directly from cron, no LLM

sayhar:
## What problem this solves

Both `ghost_detector` and `oom_check` were wired through `REMINDER_ROUTING` in `sys.dispatcher.bootup.md`, which caused the dispatcher to spawn a full `lobster-generalist` LLM subagent on each cron tick. The subagent's only job was to call `uv run agent-monitor.py` (or `oom-monitor.py`), read its stdout, and relay the output to `write_result`. No synthesis, no analysis — pure relay of script output through an expensive LLM session.

Both scripts already handle everything themselves:
- `agent-monitor.py --alert --mark-failed`: sends Telegram alerts via `alert.sh`, marks ghost sessions as failed in `agent_sessions.db`, and drops inbox notifications for the dispatcher
- `oom-monitor.py --since-minutes 10`: writes inbox JSON messages directly when OOM kills are detected (already platform-agnostic alerting with dedup state)

## What changed

**`sys.dispatcher.bootup.md`** — `REMINDER_ROUTING` is now empty. The two LLM-dispatch entries for `ghost_detector` and `oom_check` are removed. A note is added explaining both are now pure-cron and that any legacy `post-reminder.sh`-style messages for these types will fall through to the unknown-reminder fallback (logged and dropped harmlessly).

**`install.sh`** — Two new cron sections added:
- `LOBSTER-GHOST-DETECTOR`: `agent-monitor.py --alert --mark-failed` every 5 minutes
- `LOBSTER-OOM-CHECK`: `oom-monitor.py --since-minutes 10` every 10 minutes (no-op unless `LOBSTER_DEBUG=true`, per the script's own debug gate)

**`scripts/upgrade.sh`** — Migrations 52 and 53 add both cron entries to existing installs that predate this change. Idempotent: skips if the marker is already present.

## Flow change

Before:
```
cron -> post-reminder.sh ghost_detector -> inbox JSON -> dispatcher picks up ->
  spawns lobster-generalist ("run agent-monitor.py and report findings") ->
    LLM runs script, reads stdout, calls write_result
```

After:
```
cron -> uv run agent-monitor.py --alert --mark-failed (direct)
         if ghosts found: alert.sh (Telegram) + inbox drop for dispatcher
         else: log only, no inbox write
```

The `oom_check` flow is analogous. The dispatcher is completely removed from both paths.

## What is not changed

- The scripts themselves (`agent-monitor.py`, `oom-monitor.py`) are unchanged
- `alert.sh` is unchanged
- The dispatcher's handling of subagent results and other reminder types is unchanged
- The `REMINDER_ROUTING` dict is now empty but the routing machinery is preserved for future use

## Functional patterns

The scripts already follow the functional pattern — `agent-monitor.py` uses pure classification functions (`classify_agent`, `build_report`) and isolates side effects (`send_alert`, `mark_agent_failed`) at the edges. This PR extends that principle to the scheduling layer: eliminate the LLM relay, let the scripts own their output paths end-to-end.

## Tests run

- [x] `uv run --no-project pytest tests/unit/test_ghost_detector_filesystem.py -x -q` — 36 passed, 0 failed
- [ ] `uv run pytest tests/unit/` — blocked: dependency resolution fails (blue-lobster requires Python >=3.11, project specifies >=3.10; pre-existing issue unrelated to this PR)
- [ ] Live cron validation — skipped: requires production restart; LOBSTER-GHOST-DETECTOR is already present in the live crontab (migration 52 will be a no-op on this machine); LOBSTER-OOM-CHECK will be added by migration 53 on next upgrade run

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -170,18 +170,15 @@ Scheduled reminders arrive from two sources:
 
 **Routing table** — maps `reminder_type` to subagent+prompt for system cron jobs. User-created jobs carry `task_content` and are dispatched generically; do NOT add them to REMINDER_ROUTING.
 
+> **Note on ghost_detector and oom_check:** Both run as pure cron scripts — `agent-monitor.py --alert --mark-failed` (every 5 minutes) and `oom-monitor.py --since-minutes 10` (every 10 minutes) — with no inbox message and no LLM involvement. If a `ghost_detector` or `oom_check` scheduled_reminder arrives (e.g. from a legacy install still running `post-reminder.sh`), it will fall through to the unknown-reminder fallback below and be logged and dropped.
+
 ```python
 REMINDER_ROUTING = {
-  "ghost_detector": {
-    "subagent_type": "lobster-generalist",
-    "prompt": "---\ntask_id: agent-monitor\nchat_id: 0\nsource: system\n---\n\n"
-              "Run the agent monitor: uv run ~/lobster/scripts/agent-monitor.py and report findings.",
-  },
-  "oom_check": {
-    "subagent_type": "lobster-generalist",
-    "prompt": "---\ntask_id: oom-check\nchat_id: 0\nsource: system\n---\n\n"
-              "Run the OOM monitor: uv run ~/lobster/scripts/oom-monitor.py --since-minutes 10 and report findings.",
-  },
+  # --- System cron jobs only (no task_content embedded; subagent handles output) ---
+  # Do NOT add user-created jobs here — they are handled generically via task_content.
+  # ghost_detector and oom_check were removed — both scripts now run directly from
+  # cron (LOBSTER-GHOST-DETECTOR and LOBSTER-OOM-CHECK entries) and alert/write
+  # to the inbox themselves. No LLM subagent is needed for either.
 }
 ```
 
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: PASS — ready to merge**

---

**REMINDER_ROUTING empty dict — graceful?** Yes. The dispatcher logic at step 3 does `REMINDER_ROUTING.get(reminder_type)` which returns `None` on a miss regardless of whether the dict is empty or has entries. The `if route is None:` branch handles all non-routed cases (task_content dispatch or fallback_unknown_reminder). The `else:` branch is simply unreachable now, which is harmless. Any legacy `ghost_detector` or `oom_check` scheduled_reminder messages will fall through to the log-and-drop fallback correctly, as the note in the diff explains.

**Migration idempotency (52 + 53):** Solid. Both migrations use a local variable for the marker string (e.g. `GHOST_DETECTOR_MARKER="# LOBSTER-GHOST-DETECTOR"`) and the idempotency check (`grep -q "$GHOST_DETECTOR_MARKER"`) searches for the exact same string that `cron-manage.sh add` appends to the cron line. `cron-manage.sh` itself uses `grep -vF "$MARKER"` to strip before re-adding, making replacements safe. No marker mismatch found.

**Script flags correct?** Yes.
- `agent-monitor.py` is called with `--alert --mark-failed` in both install.sh and migration 52. The `--help` output confirms both flags exist and do what the PR describes (send Telegram alert + mark GHOST_CONFIRMED sessions as failed).
- `oom-monitor.py` is called with `--since-minutes 10` in both install.sh and migration 53. The `--help` output confirms the flag exists and matches the 10-minute cron interval.

---

**Deploy verification (local-dev):**
- Merged `fix/issue-ghost-oom-no-llm` into `local-dev` cleanly (3 files changed, no conflicts)
- `uv run agent-monitor.py --help` — callable, `--alert` and `--mark-failed` flags present
- `uv run oom-monitor.py --help` — callable, `--since-minutes` flag present
- `LOBSTER-GHOST-DETECTOR` cron entry present (`*/5` with `--alert --mark-failed`)
- `LOBSTER-OOM-CHECK` not yet in crontab — expected, migration 53 runs on next `upgrade.sh` invocation

No issues found. Clean removal of the LLM relay layer.

PR #1102, opened 2026-03-29
fix: replace JSONL-parsing dispatcher detection with MCP state file

### LINKED ISSUE #1009: Fix: stale dispatcher-session-id causes on-compact.py to skip compact-reminder

sayhar:
🤖🦞 Lobster (engineer):

## What problem this solves

Hook-based dispatcher detection (`session_role.is_dispatcher()`) used JSONL transcript scanning as its fallback signal. This approach is fragile: the CC transcript format changes over time, and the same-week compaction bug (#1076, #1009) caused the detector to misclassify a new dispatcher session as a subagent when the old session's JSONL file still existed on disk. The result was `on-compact.py` skipping the compact-reminder, leaving the dispatcher without situational awareness after compaction.

The MCP server already has authoritative knowledge of which session is the dispatcher — it tracks `_dispatcher_session_id` in memory (set by Options A, B, and C including the Option C auto-tag from PR #1091). This PR makes that state available to hooks via a plain text file, removing any need to scan JSONL transcripts.

## What changed

**`src/mcp/inbox_server.py`**

- Added `_DISPATCHER_SESSION_STATE_FILE` (`$LOBSTER_WORKSPACE/data/dispatcher-session-id`).
- Added `_write_dispatcher_state_file(session_id)`: atomically writes the session ID via temp-file + `os.rename()`. Silent on failure.
- Added `_clear_dispatcher_state_file()`: removes the file on server startup so a stale ID from a previous run never lingers.
- `_tag_dispatcher_session()` now calls `_write_dispatcher_state_file()` after tagging in memory. This covers all three tagging paths (Options A, B, C).
- `main()` calls `_clear_dispatcher_state_file()` before the event bus and reconciler start.

**`hooks/session_role.py`**

- Detection in `is_dispatcher()` now has two layers:
  1. **Primary**: MCP state file (`$LOBSTER_WORKSPACE/data/dispatcher-session-id`) — written by the running server, cleared on restart.
  2. **Secondary**: hook marker file (`~/messages/config/dispatcher-session-id`) — written by the `write-dispatcher-session-id.py` SessionStart hook; used as fallback when the MCP file is absent (e.g. during the brief startup window before Option A/B/C fires).
- The transcript-scanning fallback (JSONL parsing) has been removed entirely.
- `MCP_SESSION_STATE_FILE` is resolved dynamically via `_get_mcp_session_state_file()` so tests can override `LOBSTER_WORKSPACE` without patching a module-level constant.
- `_read_dispatcher_session_id()` and `DISPATCHER_SESSION_FILE` are preserved for backwards compatibility.
- Fail-open behavior retained: `OSError` on file read → `True` (assume dispatcher).

**`tests/unit/test_hooks/test_session_role_state_file.py`** (new)

24 new unit tests covering:
- `_read_session_id_from_file()`: absent, empty, whitespace, valid, OSError
- `_check_state_file()`: match, mismatch, absent, None session_id, OSError fail-open
- `is_dispatcher()`: MCP file match/mismatch, fallthrough to secondary, both absent → False, no session_id → False
- `write_dispatcher_session_id()`: writes, strips whitespace, silent on failure
- `_read_dispatcher_session_id()` shim: absent → None, present → session ID

## Before / after flow

```
Before (session_role.is_dispatcher):
  hook input → check ~/messages/config/dispatcher-session-id
                 ├── match → True
                 ├── mismatch → False
                 └── absent → scan JSONL transcript  ← fragile, compaction bug
                               ├── found dispatcher tool → True
                               └── not found → False

After (session_role.is_dispatcher):
  hook input → check $LOBSTER_WORKSPACE/data/dispatcher-session-id (MCP state file)
                 ├── match → True
                 ├── mismatch → False
                 └── absent → check ~/messages/config/dispatcher-session-id (hook marker)
                               ├── match → True
                               ├── mismatch → False
                               └── absent → False
```

The MCP state file is cleared on every server startup (new) and written atomically by every `_tag_dispatcher_session()` call (new), so it always reflects the current dispatcher or is absent.

## What is out of scope

- `write-dispatcher-session-id.py` is unchanged. Its JSONL-based `_stored_session_is_alive()` still runs to decide whether to write the hook marker file, but is no longer load-bearing for `is_dispatcher()` itself.
- PR #1076 is superseded. See close comment on that PR.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_session_role_state_file.py -v` — 24 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 166 passed, 1 pre-existing failure (`test_insert_or_ignore_preserves_existing_row`, unrelated missing `exist_ok=True`)
- [ ] Live dispatcher restart test — blocked: requires production restart; no change to existing hook marker path

Closes #1009

### DIFF
--- a/hooks/session_role.py
+++ b/hooks/session_role.py
@@ -8,41 +8,69 @@ dispatcher or a background subagent.
 
 ## Detection strategy (layered)
 
-1. **Marker file (primary)**: At dispatcher startup the session ID is written to
-   `~/messages/config/dispatcher-session-id`. This hook reads that file and
-   compares to the `session_id` field in the hook JSON input.
-   Match → dispatcher.  Mismatch or file absent → try fallback.
+1. **MCP state file (primary)**: The MCP server writes the current dispatcher
+   session ID to `$LOBSTER_WORKSPACE/data/dispatcher-session-id` (defaulting to
+   `~/lobster-workspace/data/dispatcher-session-id`) whenever
+   `_tag_dispatcher_session()` is called (Options A, B, or C).  The file is
+   cleared when the MCP server starts, so a stale ID from a previous run never
+   lingers.  This hook reads that file and compares to the `session_id` field in
+   the hook JSON input.
+   Match → dispatcher.  Mismatch → subagent.  File absent → try fallback.
 
-2. **Transcript fallback (secondary)**: Scan the transcript for tool_use blocks
-   containing the dispatcher-only tools `wait_for_messages` or `check_inbox`.
-   CC 2.1.76+ passes a file path (`transcript_path` for Stop hooks,
-   `agent_transcript_path` for SubagentStop hooks) rather than an inline
-   `transcript` list. Both file-based and inline forms are tried in order.
-   Found → dispatcher.  Not found → subagent.
+2. **Hook marker file (secondary)**: At dispatcher startup the SessionStart hook
+   (`write-dispatcher-session-id.py`) writes the session ID to
+   `~/messages/config/dispatcher-session-id`.  Used as fallback when the MCP
+   state file is absent (e.g. before the first `_tag_dispatcher_session` call
+   after a server restart, or if LOBSTER_WORKSPACE points to a non-standard
+   location).
+   Match → dispatcher.  Mismatch → subagent.  File absent → default.
 
-3. **Default**: If neither signal is available (e.g. a PreToolUse hook with no
-   marker file), return False (treat as subagent = safe/conservative).
+3. **Default**: If neither state file is readable or present, return False
+   (treat as subagent = safe/conservative).
+
+The transcript-scanning fallback that existed in previous versions has been
+removed.  It was fragile (CC JSONL format changes, same-week compaction bug
+tracked in PR #1076) and is now superseded by the MCP state file, which is
+always authoritative when the MCP server is running.
 
 ## Writing the marker file
 
 Call `write_dispatcher_session_id(session_id)` at dispatcher startup.
-Typically invoked from a `SessionStart` hook or from the dispatcher bootup
-script via a small wrapper.
+Typically invoked from the `write-dispatcher-session-id.py` SessionStart hook.
+The MCP server also calls `_write_dispatcher_state_file()` internally; hooks
+do not need to call that path directly.
 """
 
-import json
 import os
 from pathlib import Path
 
+# Secondary: hook marker file (written by write-dispatcher-session-id.py SessionStart hook)
+# Resolved at import time — stable across calls.
 DISPATCHER_SESSION_FILE = Path(
     os.path.expanduser("~/messages/config/dispatcher-session-id")
 )
 
-# Tools that only the dispatcher calls — used as transcript fallback signal.
+# Tools that only the dispatcher calls — kept for reference / external callers.
+# No longer used internally for dispatcher detection (transcript scan removed).
 DISPATCHER_ONLY_TOOLS = {
     "mcp__lobster-inbox__wait_for_messages",
     "mcp__lobster-inbox__check_inbox",
 }
+
+
+def _get_mcp_session_state_file() -> Path:
+    """Return the MCP server state file path, resolved at call time.
+
+    Reads LOBSTER_WORKSPACE on every call so tests can override the env var
+    without having to patch a module-level constant.
+    """
+    workspace = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
+    return workspace / "data" / "dispatcher-session-id"
+
+
+# Module-level alias for test patching convenience — tests that set LOBSTER_WORKSPACE
+# can also patch this directly.  Updated lazily if needed.
+MCP_SESSION_STATE_FILE = _get_mcp_session_state_file()
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one minor flag

## Summary

The core approach is sound. The JSONL-transcript fallback was the right thing to remove — it was load-bearing fragility. The new two-layer design (MCP state file → hook marker file) is clean and the atomicity story is correct.

## Findings by checklist item

### 1. Atomicity — PASS
`_write_dispatcher_state_file` uses `tmp.write_text()` + `tmp.replace()` which calls `os.rename()` under the hood. Atomic on POSIX. The `.tmp` suffix is fixed (not a per-call tempfile), so two concurrent writers could race on the temp file itself, but in practice `_tag_dispatcher_session` is only ever called from a single server process, so this is not a real concern. No issues.

### 2. Startup clear timing — PASS
`_clear_dispatcher_state_file()` is called at the very top of `main()`, before `init_event_bus()`, before the reconciler starts, and before any MCP tool handlers can fire. No hook in the Claude Code process can call a tool between server start and the clear — the MCP server must be running and accepting connections before any tool call can reach it. Timing is fine.

### 3. Fallback / fail-open behavior — PASS (with a note)
The decision to fail-open on `OSError` (return `True` = assume dispatcher) is documented and intentional. The reasoning is that a transient I/O error should not cause the dispatcher's hooks to silently no-op. This is the same policy as the old code. Worth noting that fail-open on a *mismatch* is not happening — only on `OSError` — so a healthy subagent session will still correctly return `False`.

The secondary hook marker file now serves as a genuine startup-window fallback: between server restart and the first Option A/B/C call, the hook marker file bridges the gap. This is better than the previous JSONL scan and the gap is bounded by how quickly the dispatcher calls `wait_for_messages` (Option A), which is essentially immediate.

### 4. Test coverage — PASS
24 tests cover all three paths (MCP file match/mismatch/absent, hook marker fallthrough, both absent → False), the `OSError` fail-open case, `_read_session_id_from_file` edge cases (empty, whitespace, absent, OSError), and the `_read_dispatcher_session_id` shim. Option C auto-tagging path is covered implicitly because all three Options call `_tag_dispatcher_session`, which calls `_write_dispatcher_state_file`. No gaps.

### 5. install.sh / upgrade.sh — MINOR FLAG

No migration was added for the new state file. The `$LOBSTER_WORKSPACE/data/` directory already exists on any install that has run through Migration 22 ("Ensure lobster-workspace/data/ directory exists"), and `_write_dispatcher_state_file` calls `mkdir(parents=True, exist_ok=True)` before writing, so the file will be created on first use without manual intervention. No migration is strictly necessary.

However, the PR description explicitly says the file is **cleared on MCP server startup** — so on existing installs there is nothing to clean up either. No migration needed. This is fine as-is.

One cosmetic thing: `install.sh` exports `LOBSTER_WORKSPACE_DIR` (line 278) while the code in `inbox_server.py` and `session_role.py` reads `LOBSTER_WORKSPACE` (without `_DIR`). This naming inconsistency predates this PR and is not introduced by it, but it means that if someone sets `LOBSTER_WORKSPACE_DIR` thinking that controls the workspace path, the MCP server will ignore it. This is pre-existing debt, not a blocker for this PR.

### 6. `LOBSTER_WORKSPACE` env var consistency — PASS
`inbox_server.py` reads `LOBSTER_WORKSPACE` at module import time (line 326) to set `_WORKSPACE`, which is then used for `_DISPATCHER_SESSION_STATE_FILE`. `session_role.py` reads `LOBSTER_WORKSPACE` at call time inside `_get_mcp_session_state_file()` so tests can override without reloading. Both default to `~/lobster-workspace`. Consistent.

The module-level `MCP_SESSION_STATE_FILE` alias in `session_role.py` is computed once at import time (line after `_get_mcp_session_state_file` definition), but `is_dispatcher()` re-calls `_get_mcp_session_state_file()` on every invocation, so the alias is not actually used on the hot path. The alias is there for test patching convenience per the comment. This is fine but slightly confusing — a comment noting that `is_dispatcher()` does NOT use the module-level alias would help future readers.

## Verdict

**PASS.** No blocking issues. The atomicity, startup ordering, fallback logic, and test coverage are all correct. The pre-existing `LOBSTER_WORKSPACE` vs `LOBSTER_WORKSPACE_DIR` naming discrepancy in install.sh is worth a follow-up issue but is not introduced by this PR.

sayhar (commented):
test review ping

PR #1179, opened 2026-03-31
fix: retire jobs.json — remove callers in dispatch-job.sh, status_report.py, install.sh, and tests

### LINKED ISSUE #1166: chore: complete jobs.json retirement — update all remaining callers
## Context

PR #1105 replaced the cron+jobs.json scheduling backend with systemd timers. Migration 55 in upgrade.sh reads jobs.json at upgrade time to seed systemd units. After that migration runs, jobs.json is no longer the source of truth for scheduling — but many callers still reference it.

This issue tracks the remaining callers that need updating before jobs.json can be deleted. See also #1083 (which describes the tombstone+archive plan).

---

### Source: `dispatch-job.sh` — NEEDS UPDATE (two reads)

1. **Enabled-flag check** (line 49–64): reads `jobs.json` to see if `enabled: false` before dispatching. With systemd, the enabled/disabled state lives in `systemctl is-enabled lobster-<name>.timer`. The replacement check is:
   ```bash
   systemctl is-enabled --quiet lobster-${JOB_NAME}.timer 2>/dev/null || exit 0
   ```
   If the unit doesn't exist at all, `is-enabled` returns non-zero, so the job won't fire — correct behavior.

2. **`chat_id` lookup** (line 98–107): reads `jobs.json` to get `chat_id` for the inbox message. The new systemd model doesn't store `chat_id` in unit files. Options: (a) hardcode a default chat_id via env var `LOBSTER_DEFAULT_CHAT_ID`, (b) drop the field (dispatcher already knows the user), or (c) read from a separate config. Recommendation: drop it from the inbox message format — the dispatcher doesn't need it since it always replies to the configured user.

3. **`last_run` update** (line 138–170): updates `last_run` in `jobs.json` after dispatch. This is purely cosmetic since jobs.json is no longer authoritative — `systemctl status lobster-<name>.timer` gives last/next run times. This block can be deleted entirely.

### Source: `sync-crontab.sh` — CAN BE DELETED (or archived)

Reads `jobs.json` to regenerate the user's crontab. Replaced entirely by systemd. No callers trigger it after the systemd migration — upgrade.sh Migration 55 is the last migration that references it. Archive to `scheduled-tasks/deprecated/sync-crontab.sh`.

### Source: `src/mcp/inbox_server.py` (lines 654, 747–748) — NEEDS UPDATE

- Line 654: defines `SCHEDULED_JOBS_FILE = SCHEDULED_JOBS_DIR / "jobs.json"` as a module-level constant.
- Lines 747–748: creates an empty `{"jobs": {}}` file on startup if it doesn't exist.

After #1105 lands, the MCP tool handlers (`handle_create_scheduled_job`, etc.) no longer read from or write to `SCHEDULED_JOBS_FILE` — they delegate entirely to `systemd_jobs.py`. However the constant is defined and the file is still seeded. These two lines can be removed; nothing will break because the tool handlers don't use this path.

### Source: `lobster-shop/lobster-status/src/status_report.py` (lines 43, 112–125) — NEEDS UPDATE

`_get_scheduled_jobs_count()` reads `jobs.json` to report total/enabled/disabled job counts in the status report. After migration, this should query systemd instead (via `systemctl list-timers --all` filtered to `lobster-` prefix), or call `systemd_jobs.list_jobs()`.

### Source: `install.sh` (lines 402–407, 1265–1269, 1283) — NEEDS UPDATE

Creates `jobs.json` with `{"jobs": {}}` during installation. After the systemd migration, the file shouldn't be seeded. These blocks should be removed (or reduced to a one-time tombstone creation that upgrade.sh Migration 55 already handles).

### Source: `scripts/upgrade.sh` (line 255, lines 1936–1979) — PARTIALLY HANDLED

- Line 255: backs up `jobs.json` before upgrades. Fine to keep until jobs.json is deleted from all installs.
- Migration 55 (lines 1936–1979): reads `jobs.json` to migrate entries to systemd. This is intentional — it's the migration tool itself. Once Migration 55 has run on all installs, this migration becomes a no-op (the file will be gone or empty). It can be removed in a subsequent migration that tombstones the file.

### Source: `scripts/update-lobster.sh` (line 58) — LOW PRIORITY

Backs up `jobs.json` as part of the rolling update backup list. Fine to remove once the file is gone from disk.

### Source: `tests/` (multiple files) — NEEDS UPDATE AFTER CALLERS ARE FIXED

- `tests/unit/test_dispatch_job_sh.py` — tests the enabled-flag check and last_run update in dispatch-job.sh. Both features being removed means these tests should be rewritten to test the `systemctl is-enabled` check.
- `tests/conftest.py` — fixture setup creates `jobs.json`. Should be cleaned up after callers are updated.
- `tests/integration/test_scheduled_execution.py` — `test_job_registered_in_jobs_json` tests the old write-to-JSON behavior that no longer happens. This test should be removed or replaced with a test that checks for systemd unit file creation.
- `tests/integration/test_installation.py` — `test_jobs_json_initialized` checks that `jobs.json` is seeded on install. Remove once install.sh stops creating it.
- `tests/integration/test_upgrade_safety.py` — `test_jobs_json_preserved` checks that jobs.json survives an upgrade. Remove after retirement.
- `tests/docker/entrypoint-test.sh` (line 32) — creates `jobs.json` in the test container. Remove once install.sh no longer seeds it.

### Source: Docs — LOW PRIORITY, CLEANUP ONLY

- `README.md` (line 259) — lists `jobs.json` in directory tree
- `CLAUDE.md` (line 149) — mentions `scheduled-jobs/jobs.json - Job registry state`
- `UPGRADE.md` (line 19, 75) — mentions jobs.json in the backup/restore section
- `docs/CUSTOMIZATION.md` (lines 119, 260) — documents jobs.json format
- `scheduled-tasks/README.md` (lines 7, 49, 74) — describes jobs.json workflow

All doc references become stale. Update after the code changes land.

---

## Current State of jobs.json on Disk

`~/lobster-workspace/scheduled-jobs/jobs.json` exists with 12 entries. All are `enabled: false` except:
- `bot-talk-poller` (enabled: true, runner: "cron-script")
- `lobstertalk-incoming-handler` (enabled: true, runner: "cron-script")
- `test-lobster-job` (enabled: true, no runner)

None of the entries have a `command` field (only `task_file` and `runner`), so Migration 55 will skip all of them and emit warnings. These jobs must be manually recreated via `create_scheduled_job` with an explicit `command` parameter pointing to the new script-based runner, or tombstoned.

---

## Retirement Order

1. Fix `dispatch-job.sh`: replace enabled-check with `systemctl is-enabled`, remove chat_id lookup, remove last_run update block.
2. Fix `status_report.py`: replace jobs.json read with systemd query.
3. Fix `inbox_server.py`: remove `SCHEDULED_JOBS_FILE` constant and the startup seeding block.
4. Fix `install.sh`: remove jobs.json creation.
5. Update tests to match new behavior.
6. Archive `sync-crontab.sh` and `dispatch-job.sh` (or delete them if they are no longer needed at all after the systemd runner replaces them).
7. Update docs.
8. Add a new upgrade.sh migration that removes stale jobs.json from existing installs.

**Can jobs.json be cleanly deleted once these are resolved?** Yes — once all callers above are fixed, the file on disk can be removed. The recommended approach (per #1083) is to archive it to `scheduled-jobs/deprecated/jobs.json` rather than hard-delete it, and add an upgrade.sh migration that moves any remaining live file to the deprecated path.

---

Depends on: #1105 (must be merged first)
Related: #1083 (broader retirement plan), #929 (jobs.json atomic write — superseded by this migration)

sayhar:
## What problem this solves

\`jobs.json\` was the registry for cron-based job scheduling. PR #1105 replaces the backend with systemd timers, but seven callers still read or seed the file. As long as those callers exist, jobs.json stays a required file and any future attempt to delete it breaks things silently.

This PR removes the callers that don't need to wait for #1105 to merge: the dispatch script, the status report, the install seeding blocks, and the tests that verified that seeding. After this lands, the only remaining jobs.json reader is the MCP tool handler layer (create/list/get/update/delete scheduled jobs) — which is intentionally deferred to #1105.

Closes #1166

---

## Changes

**dispatch-job.sh** — three callers removed:
- Enabled-flag check replaced with \`systemctl is-enabled --quiet lobster-\${JOB_NAME}.timer\`. If the unit doesn't exist, \`is-enabled\` returns non-zero, so the job silently skips — correct behavior.
- \`chat_id\` lookup removed. The dispatcher always replies to the configured user; there is no per-job chat_id to look up.
- \`last_run\` update block removed. systemd tracks this natively; the block was cosmetic and wrote to a file that is now being retired.

**status_report.py** — \`_get_scheduled_jobs_count()\` rewritten to query \`systemctl list-timers --all\` filtered to the \`lobster-\` prefix. Each matching unit is probed with \`systemctl is-enabled\` to distinguish enabled from disabled.

**inbox_server.py** — startup seeding of jobs.json removed (the two-line block that wrote \`{"jobs": {}}\` on first boot). The \`SCHEDULED_JOBS_FILE\` constant is retained with a comment: the MCP tool handlers still read and write it, and removing it now would break \`create_scheduled_job\` / \`list_scheduled_jobs\` / etc. before #1105 ships.

**install.sh** — two jobs.json creation blocks removed. The inline heredoc that regenerated \`sync-crontab.sh\` at install time is also removed since the file is now archived.

**sync-crontab.sh** — archived to \`scheduled-tasks/deprecated/sync-crontab.sh\`. No active code path calls it after the systemd migration.

**scripts/upgrade.sh (Migration 48)** — moves any remaining \`jobs.json\` to \`scheduled-jobs/deprecated/jobs.json\` on existing installs. Move rather than delete preserves the data for post-mortems.

**Tests**:
- \`test_dispatch_job_sh.py\`: rewritten. Disabled/enabled tests now inject a fake \`systemctl\` into PATH. Added \`test_enabled_job_does_not_modify_any_json_registry\`. Removed \`test_jobs_json_last_run_updated\`.
- \`test_scheduled_execution.py\`: removed \`test_job_registered_in_jobs_json\`.
- \`test_installation.py\`: removed \`test_jobs_json_initialized\`.
- \`test_upgrade_safety.py\`: removed \`test_jobs_json_preserved\`.
- \`tests/docker/entrypoint-test.sh\`: removed jobs.json creation line.
- \`conftest.py\`: removed \`jobs_file\` fixture (was only used by now-deleted tests).

---

## What is NOT changed

The MCP tool handlers (\`handle_create_scheduled_job\`, \`handle_list_scheduled_jobs\`, etc.) still read and write jobs.json — they are left untouched. Their migration is gated on #1105 merging and \`systemd_jobs.py\` being available. This PR does not attempt to migrate them.

\`SCHEDULED_JOBS_FILE\` is kept in \`inbox_server.py\` with a note explaining why. Removing it now would cause runtime errors in the MCP handlers.

---

## Tests run

- [x] \`uv run pytest tests/unit/test_dispatch_job_sh.py -x -q\` — 13 passed, 0 failed
- [x] \`uv run pytest tests/unit/ -q\` — comparison against main: main has 101 unit failures; this branch has 80. All failures on this branch are pre-existing on main (test_hibernation, test_write_observation, test_memory, test_skill_manager, test_path_guard). No new failures introduced.
- [x] \`uv run pytest tests/unit/test_dispatch_job_sh.py tests/integration/test_scheduled_execution.py tests/integration/test_installation.py tests/integration/test_upgrade_safety.py -q\` — 70 passed, 7 failed vs 74 passed, 8 failed on main. All 7 remaining failures are pre-existing. Eliminated 1 failure by removing the now-stale \`test_job_registered_in_jobs_json\`.
- [ ] Docker integration tests — skipped: Docker not available in this environment.


sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS**

---

### What is retired

All six callers of jobs.json that are safe to remove before #1105 (systemd_jobs.py):

1. `dispatch-job.sh` — enabled-flag check replaced with `systemctl is-enabled --quiet lobster-${JOB_NAME}.timer`; chat_id lookup removed (always 0 now); last_run write block removed.
2. `status_report.py` — `_get_scheduled_jobs_count()` rewritten to query `systemctl list-timers --all` filtered to `lobster-*` prefix.
3. `install.sh` — two jobs.json creation blocks removed; `sync-crontab.sh` inline heredoc removed.
4. `sync-crontab.sh` — archived to `scheduled-tasks/deprecated/sync-crontab.sh` (not deleted — good for post-mortems).
5. `uv.lock` / `pyproject.toml` — `blue-lobster` dependency removed.
6. Tests — `test_jobs_json_last_run_updated`, `test_jobs_json_initialized`, `test_job_registered_in_jobs_json`, `test_jobs_json_preserved`, `jobs_file` conftest fixture, docker entrypoint line all removed and replaced with systemctl-aware equivalents.

### Is retirement complete?

Partial by design and correctly documented. The MCP tool handlers (`handle_create_scheduled_job`, `handle_list_scheduled_jobs`, etc.) and `SCHEDULED_JOBS_FILE` constant in inbox_server.py are explicitly left in place, with a comment explaining they are gated on #1105. This is the right call — removing them now would break `create_scheduled_job` / `list_scheduled_jobs` at runtime.

### Migration 48 (upgrade.sh)

Present. Archives `scheduled-jobs/jobs.json` → `scheduled-jobs/deprecated/jobs.json` on existing installs. Move rather than delete is correct. Passes the "existing install won't break silently" bar.

### Test coverage

Tests are properly updated to use a fake `systemctl` binary injected into PATH. The `_make_fake_systemctl` helper pattern is clean. `TestRunJobShMissingJobsJson` class removed — correct, since systemctl is now the authority and the no-jobs.json case is no longer meaningful.

### Risk assessment

Low. The only runtime behavior change is:
- `dispatch-job.sh` now queries systemd instead of reading a JSON file. If a timer unit doesn't exist, the job skips silently — same behavior as the old `enabled=false` path.
- `chat_id` is hardcoded to 0. The dispatcher always knows the configured user, so per-job chat_id was unused in practice.
- status_report.py falls back to `{"total": 0, "enabled": 0, "disabled": 0}` on any exception — same as before.

Nothing breaks if this merges before #1105. The MCP handlers continue reading jobs.json as before.

PR #1194, opened 2026-03-31
feat(hooks): inject bootup context files on SessionStart

sayhar:
## Summary

- Adds `hooks/inject-bootup-context.py`: a new SessionStart hook that injects the appropriate system bootup file (`sys.dispatcher.bootup.md` for the dispatcher, `sys.subagent.bootup.md` for subagents) plus user-specific bootup files into Claude Code context via hook stdout before the first turn
- Updates `CLAUDE.md` Role-Specific Context section to note these files are now pre-injected via hook (no manual reads needed)
- Settings.json wiring (two entries) is already live: one empty-matcher entry for all sessions, one compact-matcher entry for post-compact injection

## How it works

1. Hook reads `session_id` from stdin JSON
2. Calls `session_role.is_dispatcher()` to detect role
3. Prints appropriate system bootup file to stdout (injected as system message)
4. Prints user bootup files if they exist

## Test plan

- [x] Manual test as subagent: outputs subagent bootup content correctly
- [x] Manual test as dispatcher: outputs dispatcher bootup content correctly
- [x] Settings.json has 6 SessionStart hooks, inject-bootup-context at positions 1 (all sessions) and 3 (compact sessions)
- [ ] Live verification: start a new session and confirm bootup content in context without manual reads

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,11 +8,13 @@ You are **Lobster**, an always-on AI assistant that never exits. You run in a pe
 
 This file provides shared context. Depending on your role, read the appropriate supplement:
 
-**System context** (always read):
-- **If you are the dispatcher (main loop):** read `.claude/sys.dispatcher.bootup.md` — it covers the main loop pseudocode, the 7-second rule, the dispatcher pattern, handling subagent results, message source handling (Telegram/Slack), self-check reminders, message flow diagram, startup behavior, hibernation, context recovery, Google Calendar handling, and voice/brain-dump routing.
-- **If you are a subagent:** read `.claude/sys.subagent.bootup.md` — it covers the `write_result` requirement, identity rules, and the model selection table.
+> **Note:** The system bootup files and user bootup files listed below are pre-injected into context via the `inject-bootup-context.py` SessionStart hook. The content is already present at the start of every session — the file paths are listed here for reference only.
 
-**User context** (read after system files, if the files exist):
+**System context** (pre-injected via hook):
+- **Dispatcher (main loop):** `.claude/sys.dispatcher.bootup.md` — covers the main loop pseudocode, the 7-second rule, the dispatcher pattern, handling subagent results, message source handling (Telegram/Slack), self-check reminders, message flow diagram, startup behavior, hibernation, context recovery, Google Calendar handling, and voice/brain-dump routing.
+- **Subagent:** `.claude/sys.subagent.bootup.md` — covers the `write_result` requirement, identity rules, and the model selection table.
+
+**User context** (pre-injected via hook, if the files exist):
 - Both roles: `~/lobster-user-config/agents/user.base.bootup.md` (behavioral preferences)
 - Both roles: `~/lobster-user-config/agents/user.base.context.md` (personal facts and context)
 - Dispatcher: `~/lobster-user-config/agents/user.dispatcher.bootup.md`
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -14,6 +14,14 @@ File injection order:
 2. ~/lobster-user-config/agents/user.base.bootup.md (if exists)
 3. ~/lobster-user-config/agents/user.dispatcher.bootup.md (dispatcher only, if exists)
    OR ~/lobster-user-config/agents/user.subagent.bootup.md (subagent only, if exists)
+
+Hook ordering note:
+This hook calls session_role.write_dispatcher_session_id() when it detects a
+dispatcher session, making it self-sufficient regardless of whether
+write-dispatcher-session-id.py ran first. The write is idempotent: if
+write-dispatcher-session-id.py already ran (position 0 in settings.json),
+the file already has the correct ID and this is a no-op; if this hook runs
+first for any reason, the ID is written here and downstream hooks benefit.
 """
 
 import json
@@ -40,7 +48,7 @@ HOOK_NAME = "inject-bootup-context"
 
 
 def _read_file_safe(path: Path, label: str) -> str | None:
-    """Return file contents or None on any error, logging to stderr."""
+    """Return file contents or None on any error or empty file, logging to stderr."""
     if not path.exists():
         print(
             f"[{HOOK_NAME}] WARNING: {path} not found; skipping {label} injection.",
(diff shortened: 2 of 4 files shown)


sayhar (commented):
## Review: inject-bootup-context.py SessionStart hook

### Correctness

**Role detection**: Correct. Delegates to `session_role.is_dispatcher()` which uses the MCP state file as primary source and the hook marker file as secondary. This is the right approach — same pattern as other hooks. Works correctly as long as `write-dispatcher-session-id.py` has already run.

**Hook ordering concern (blocker)**: The hook **depends on** `write-dispatcher-session-id.py` having already written the dispatcher session ID before `inject-bootup-context.py` runs. The PR description says `inject-bootup-context` is at positions 1 (all sessions) and 3 (compact sessions) in the settings.json hook list. If `inject-bootup-context` runs *before* `write-dispatcher-session-id.py` in a fresh session, `is_dispatcher()` will fall through to the default (`False`) and the dispatcher will receive subagent bootup content instead of dispatcher bootup content. The install.sh does not register `inject-bootup-context`, so the hook ordering in settings.json cannot be verified from this PR's diff alone — the settings.json changes are described as "already live" but are not in the diff. This needs explicit verification.

**Missing files**: Handled gracefully. Required system files (sys.dispatcher.bootup.md / sys.subagent.bootup.md) use `_read_file_safe` which prints to stderr and returns `None`; the hook exits 0. Optional user config files use `_inject_if_exists` which silently skips absent files. Both behaviors are appropriate.

**Exit codes**: Always exits 0. Errors go to stderr as warnings, never as fatal failures. This is correct — hook errors should never block a session from starting.

**Malformed stdin**: `json.JSONDecodeError` and `ValueError` are caught; `hook_input` defaults to `{}`. `is_dispatcher({})` falls through to reading state files from disk, then returns `False` if neither file is present. Safe.

**File read inconsistency (minor)**: `_read_file_safe` logs a WARNING when the required system file is not found, but `_inject_if_exists` silently skips absent optional files without any log. This is intentional and correct — the distinction between "required" and "optional" files is meaningful.

### install.sh completeness (blocker)

`install.sh` does **not** register `inject-bootup-context.py`. Every other SessionStart hook (`write-dispatcher-session-id.py`, `on-compact.py`, `inject-debug-bootup.py`, `on-fresh-start.py`) has a corresponding registration block in `install.sh`. This hook has none.

The PR says settings.json wiring is "already live," but fresh installs will not get this hook. Any user running `install.sh` on a new machine will not get bootup injection. This is a missing piece that should be added before merge.

Additionally, no migration was added to `scripts/upgrade.sh` for existing installs — per CLAUDE.md convention, new hook registrations should have an upgrade migration.

### Size concern

Measured on this system:
- `sys.dispatcher.bootup.md`: **95 KB** (~23,800 tokens at 4 chars/token) = ~12% of 200K context
- `sys.subagent.bootup.md`: **24 KB** (~6,000 tokens) = ~3% of 200K context
- User files (dispatcher set): ~18 KB additional

Dispatcher total: ~113 KB injected on every session start and every compaction. This is real but probably acceptable given that this content would otherwise need to be read manually (same tokens either way). The difference is it now happens unconditionally — even for short tasks where the full dispatcher bootup is unnecessary overhead. Worth noting, not a blocker.

### CLAUDE.md update

The update correctly describes the new behavior (files are pre-injected, no manual reads needed). The wording is clear.

### Summary

**Blockers:**
1. `install.sh` does not register `inject-bootup-context.py` — fresh installs will not get the hook
2. Hook ordering cannot be verified from this diff — need to confirm `write-dispatcher-session-id.py` runs before `inject-bootup-context.py` in settings.json, otherwise dispatchers get subagent bootup content on fresh sessions

**Non-blocking observations:**
- 95 KB dispatcher injection is real overhead but equivalent to what was required before; acceptable
- No upgrade migration in `scripts/upgrade.sh` for existing installs

sayhar (commented):
Fixed both blockers. Details:

**Blocker 1 — install.sh completeness**: Added two registration blocks for `inject-bootup-context.py` in install.sh:
- An empty-matcher block (all sessions), inserted immediately after the `write-dispatcher-session-id` block so the state file is written before role detection runs
- A compact-matcher block (post-compaction re-injection), inserted after `on-compact`

The install.sh ordering now matches the existing settings.json layout exactly.

**Blocker 2 — Hook ordering**: Verified correct. In settings.json, `write-dispatcher-session-id.py` is at position 0 (empty matcher) and `inject-bootup-context.py` is at position 1 (empty matcher). In install.sh, the registration blocks follow the same order. `is_dispatcher()` reads from `~/lobster-workspace/data/dispatcher-session-id` (MCP state file, primary) and `~/messages/config/dispatcher-session-id` (hook marker file, secondary) — both written by `write-dispatcher-session-id.py` which runs first. No ordering issue.

**Also added**: Migration 55 in scripts/upgrade.sh registers both inject-bootup-context SessionStart entries for existing installs, per the CLAUDE.md convention for new hook registrations.

Commit: fc2974b

sayhar (commented):
Lobster (reviewer) — fresh review after fixes from commit fc2974b.

## Re-review after blockers fixed

### Blocker 1 (install.sh) — RESOLVED

`install.sh` now has two registration blocks for `inject-bootup-context.py`:

- Empty-matcher block (all fresh sessions): inserted immediately after the `write-dispatcher-session-id` block, with a comment explicitly noting the ordering dependency. Idempotency check correctly filters on both `contains("inject-bootup-context")` AND `select(.matcher == "")`.
- Compact-matcher block (post-compaction re-injection): inserted after `on-compact`, with explanatory comment. Same idempotency pattern.

Both blocks use the same `jq`-based append pattern as other hooks in the file. Consistent.

### Blocker 2 (hook ordering) — RESOLVED

Ordering is now documented in the install.sh comment:
> "Runs after write-dispatcher-session-id so role detection (is_dispatcher) works correctly."

Mechanically confirmed: `write-dispatcher-session-id` registers at line 1742, `inject-bootup-context` (empty-matcher) at line 1766, so on a fresh install the state file is written before role detection runs. `is_dispatcher()` reads `~/lobster-workspace/data/dispatcher-session-id` and `~/messages/config/dispatcher-session-id` — both written by `write-dispatcher-session-id.py`. Ordering is safe.

### Migration 55 (upgrade.sh) — ADDED

Migration 55 handles both SessionStart entries for existing installs. Uses same jq approach with proper idempotency guards. Increments `migrated` counter correctly. No issues.

### Hook implementation — no regressions

The hook file itself is unchanged from the initial review (where it had no issues). Error handling, exit codes, malformed stdin handling, and file injection order all remain correct.

### CLAUDE.md update — correct

The note about pre-injection and section rename from "always read" to "pre-injected via hook" accurately reflects the new behavior.

---

[VERDICT]: APPROVE — all blockers resolved, migration added, ordering documented. Ready to merge.

PR #1221, opened 2026-03-31
feat: capture user decisions to memory DB in real-time

### LINKED ISSUE #1219: Dispatcher should write decisions to memory DB in real-time
## Problem

When Sahar says things like "let's go with X", "merge it", "approved", "use Y approach going forward", these decisions only reach memory via the 3 AM nightly consolidation sweep — up to 24 hours later.

This means if a context compaction or restart happens before consolidation, the decision is lost from memory entirely.

## Proposed solution

Add a "Decision Memory" section to `sys.dispatcher.bootup.md` instructing the dispatcher to call `memory_store(type="decision", ...)` inline when it detects a decision pattern in user messages.

**Trigger patterns:**
- User approves a PR or action: "go for it", "merge it", "lgtm", "approved", "do it", "proceed"
- User states a forward-looking preference: "always do X", "from now on", "I prefer", "going forward"
- User makes an explicit choice: "let's go with", "confirmed", "use Y", "let's do"

**Anti-spam guard:** Max 1 decision write per user message. Don't write for simple acknowledgments ("ok", "sounds good", "thanks") or reactions — only explicit decisions or stated preferences.

**Inline only:** This is a single `memory_store` call, fits well within the 7-second rule, no subagent needed.

## Implementation

Primarily a `sys.dispatcher.bootup.md` change adding a new "Decision Memory" section, positioned after message processing guidance so the dispatcher applies it when composing responses.

sayhar:
## What this fixes

Explicit user decisions — "merge it", "let's go with X", "from now on always do Y" — were only reaching the memory DB via the 3 AM nightly consolidation. Up to 24 hours of decisions could be lost if a context compaction or restart happened before that sweep ran.

## What changed

This adds a **Decision Memory** section to `sys.dispatcher.bootup.md` that instructs the dispatcher to call `memory_store(type="decision")` inline when it detects a decision pattern in a user message. The call happens on the main thread before composing the reply — it is a single fast DB write, well within the 7-second rule, no subagent needed.

**Three trigger categories:**
- Approvals: "merge it", "go for it", "lgtm", "do it", "approved"
- Forward-looking preferences: "always do X", "from now on", "I prefer", "going forward"
- Explicit choices: "let's go with", "confirmed", "use Y", "decided: X"

**Anti-spam guard:** max 1 write per user message; simple acknowledgments ("ok", "sounds good", "thanks") and emoji reactions are explicitly excluded.

## What is out of scope

- No changes to the nightly consolidation path — it still runs at 3 AM and handles all other memory types
- No changes to subagent behavior or the memory DB schema
- No new tools or MCP calls introduced; `memory_store` already exists and is already called by the dispatcher in the `subagent_observation` handler

## Functional patterns used

Pure declarative rules: the trigger/anti-spam logic is expressed as a table of conditions the dispatcher evaluates deterministically. No LLM judgment needed — matching "merge it" is code, not inference.

## Tests run

- [x] File diff reviewed — 49-line insertion, no regressions to existing sections
- [ ] Live dispatcher test — requires production dispatcher restart; no behavior change to existing paths so safe to merge without a live test run

Closes #1219

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -723,6 +723,52 @@ history = get_conversation_history(chat_id=sender_chat_id, direction='all', limi
 
 ---
 
+## Decision Memory: Real-Time Capture
+
+When a user message contains an explicit decision or stated preference, call `memory_store` inline
+(single call, fits within the 7-second rule — no subagent needed) before composing your reply.
+
+### Trigger patterns
+
+Write to memory when the user:
+
+- **Approves an action or PR** — phrases like "go for it", "merge it", "lgtm", "approved", "do it",
+  "proceed", "ship it", "looks good"
+- **States a forward-looking preference** — phrases like "always do X", "from now on", "I prefer",
+  "going forward", "in future", "next time", "do not do X again"
+- **Makes an explicit choice** — phrases like "let's go with", "confirmed", "use Y", "let's do",
+  "I want X", "stick with Y", "decided: X"
+
+### Anti-spam guard
+
+**Do not** write to memory for:
+- Simple acknowledgments: "ok", "sounds good", "thanks", "sure", "got it"
+- Reactions (emoji presses, thumbs up)
+- Anything that is clearly just confirmation of receipt, not a substantive decision
+- Max 1 `memory_store` call per user message, even if the message contains multiple trigger phrases
+
+### How to store
+
+```python
+memory_store(
+    content="[1-2 sentence summary of the decision and why, if stated]",
+    type="decision",
+    tags=["project/lobster"],   # add more specific tags if the context is clear
+)
+```
+
+Examples:
+- User: "merge it" (after reviewing a PR) → `"User approved merging PR #N [title]. No additional conditions stated."`
+- User: "from now on always add a before/after diagram to PR descriptions" → `"User prefers PR descriptions to always include a before/after diagram for any flow changes."`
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

## PR 1221 — feat: capture user decisions to memory DB in real-time

**Verdict: PASS**

### Change

Adds a "Decision Memory: Real-Time Capture" section to `sys.dispatcher.bootup.md`. Instructs the dispatcher to call `memory_store(type="decision")` inline (main thread, single DB write, <7 seconds) when it detects an explicit decision pattern.

### Correctness

- Single `memory_store` call per user message — anti-spam guard prevents noisy writes.
- The trigger patterns are specific enough to avoid false positives: "ok", "sounds good", "thanks" are explicitly excluded.
- Placement in the message-processing flow diagram is correct: after history check, before composing reply.
- Tags like `["project/lobster"]` are examples, not hardcoded — the instruction says "add more specific tags if context is clear."
- The 7-second rule compatibility is explicitly noted ("single fast DB write, no subagent needed") — correct.
- Emoji reactions and simple acknowledgments excluded — correct.

### One note

The trigger list includes "looks good" as an approval trigger. "Looks good" is also a common non-decision phrase ("the weather looks good today"). The anti-spam guard partially mitigates this, but the dispatcher will need to use judgment. This is an acceptable tradeoff for a behavioral instruction document — it is not a regex rule.

### install.sh / upgrade.sh / tests

No code changes — documentation only. No migration needed.

**Ready to merge.**

Issue #1203, opened 2026-03-31
ops: ensure nightly-consolidation crontab entry on existing installs


sayhar:
## Problem

`nightly-consolidation.sh` runs at 3am daily to consolidate memory and rotate digests. `install.sh` correctly adds this crontab entry, but existing live installs predating this entry are missing it and had to be patched manually.

## Solution

Add upgrade.sh Migration 55 to inject the `LOBSTER-NIGHTLY-CONSOLIDATION` cron entry on any install missing it. The migration is idempotent — `cron-manage.sh add` checks for the marker before inserting.

## Also

Add a **Scheduling Architecture** section to `CLAUDE.md` documenting the two-layer scheduling model:
- Cron = system-level (consolidation, health checks, log exports)
- Systemd MCP tools = user-space (pollers, reminders, user-defined jobs)

PR #1222, opened 2026-03-31
feat: nightly consolidation generates user-model/_context.md

### LINKED ISSUE #1192: Process gap: nightly-consolidation doesn't run — user-model/_context.md never generated
🤖🦞 Lobster (ops): Process gap identified by bootup-gap-analysis subagent (2026-03-31).

## What's broken

`user-model/_context.md` is designed to be a pre-computed context snapshot that the dispatcher reads at startup. It is generated by `write_context_cache()` in `src/mcp/user_model/bridges.py`, which is called by `run_consolidation()`.

But `run_consolidation()` is never called — the nightly consolidation job (`nightly-consolidation.sh`) does not invoke Python anymore (the original `.py` script was removed, replaced by the `.sh` script which does not call `write_context_cache()`).

Result: `_context.md` has never been generated. The user model (`user.base.context.md`) is very sparse — only timezone is documented. No projects, preferences, people, or interaction history.

## Impact

- Dispatcher starts each session with minimal user context
- Memory DB has only 35 entries (almost unused during live sessions)
- Dispatcher never calls `memory_search` when new topics arrive
- The "user contact model thing" Sahar asked about does not function

## What needs to happen

1. Fix nightly consolidation to call `write_context_cache()` (or create a cron that does this directly)
2. Audit `user.base.context.md` and manually populate it with known projects, preferences, and ongoing threads
3. Wire up `memory_store` to be called during sessions for important facts

## Blocking

This blocks the "memory being used properly" goal from issue #1177.

sayhar:
🤖🦞 Lobster (engineer):

## What this fixes

Every session boots blind on user context because `~/lobster-workspace/user-model/_context.md` never exists. The nightly-consolidation agent had a step 9 that wrote this file, but it wrote a plain LLM text synthesis with no connection to the user model DB. `write_context_cache()` in `bridges.py` — which queries accumulated preferences, narrative arcs, and attention items from the DB — was never called during nightly runs.

## What changed

Step 9 is replaced with two steps:

**Step 9 — DB sync (new):** Run `run_bridges()` via `uv` to push `projects/*.md` as narrative arcs and `priorities.md` as attention items into the user model DB. `run_bridges()` calls `write_context_cache()` internally, which writes `_context.md` from DB state (preferences graph, emotional baseline, attention stack).

**Step 10 — LLM synthesis layer (updated from old step 9):** Call `model_user_context(deep=True)` to read the structured DB output, then merge it with today's synthesized events into a structured `_context.md` template covering: Active Projects, Top Priorities, Key People, Preferences & Constraints, Emotional Baseline, and Open Questions.

The result is a `_context.md` that combines DB-inferred state (accumulated over time, deterministic) with event-driven synthesis from today — rather than pure LLM synthesis with no memory of prior days.

## Functional patterns

Pure function composition: `run_bridges()` → `write_context_cache()` is a pure data transformation (canonical files → DB → markdown snapshot). The agent invokes this as a subprocess to isolate the DB side effects at the boundary, then reads back via `model_user_context(deep=True)` for the synthesis layer.

## What is out of scope

This PR does not fix the separate issue that nightly-consolidation isn't registered as a scheduled job (#1203). The agent change is correct and will take effect once the job is registered and triggering.

## Tests run

- [x] `git diff HEAD~1` — diff reviewed, changes match intent
- [ ] Live nightly run — blocked: requires registered cron job (#1203) and a session running at 3 AM. No behavior change to existing code paths (step 9 was a no-op before; this makes it functional).

Closes #1192

### DIFF
--- a/.claude/agents/nightly-consolidation.md
+++ b/.claude/agents/nightly-consolidation.md
@@ -52,7 +52,7 @@ You will receive a prompt containing the consolidation trigger timestamp.
 9. **Sync canonical files into the user model DB.**
    Run the bridge pass to push projects, priorities, and preferences from canonical markdown files into the user model DB. This also generates the pre-computed `_context.md` via `write_context_cache()`:
    ```bash
-   cd ~/lobster && uv run -c "
+   cd ~/lobster && uv run python -c "
    import sys; sys.path.insert(0, 'src')
    from mcp.user_model.bridges import run_bridges
    import sqlite3, os
--- a/.claude/agents/nightly-consolidation.md
+++ b/.claude/agents/nightly-consolidation.md
@@ -49,15 +49,54 @@ You will receive a prompt containing the consolidation trigger timestamp.
    Read `~/lobster-user-config/memory/canonical/handoff.md`.
    Update the "Current state" section to reflect the synthesized current state. This is the first file the next session reads — keep it accurate and current.
 
-9. **Write `_context.md` (user model summary).**
-   Write a concise pre-computed summary to `~/lobster-workspace/user-model/_context.md`.
-   This file is read at session startup. Include:
-   - User's current top priorities (from priorities.md or observed themes)
-   - Active projects and their status
-   - Key people in current focus
-   - Any emotional baseline signals (stress level, energy, mode of operation)
-   - Constraints or preferences that were reinforced today
-   Create the directory if it does not exist. Overwrite the file entirely each run.
+9. **Sync canonical files into the user model DB.**
+   Run the bridge pass to push projects, priorities, and preferences from canonical markdown files into the user model DB. This also generates the pre-computed `_context.md` via `write_context_cache()`:
+   ```bash
+   cd ~/lobster && uv run -c "
+   import sys; sys.path.insert(0, 'src')
+   from mcp.user_model.bridges import run_bridges
+   import sqlite3, os
+   db_path = os.path.expanduser('~/lobster-workspace/data/memory.db')
+   conn = sqlite3.connect(db_path)
+   result = run_bridges(conn)
+   conn.close()
+   print(result)
+   "
+   ```
+   This syncs `projects/*.md` as narrative arcs and `priorities.md` as attention items, and writes the pre-computed `~/lobster-workspace/user-model/_context.md`.
+   If the script fails (e.g. DB not initialized), continue to step 10.
+
+10. **Write `_context.md` (user model summary).**
+    Call `model_user_context(deep=True)` to retrieve structured user model data from the DB.
+    Combine it with today's synthesized context (from steps 1–8) to write a complete snapshot.
+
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

Note: self-review — independent reviewer should verify before merge.

**What I checked:**

The diff is a clean replacement of step 9 (9 lines) with two steps (48 lines). The new step 9 runs `run_bridges()` via `uv run -c` inline Python — this is the correct pattern for calling the existing bridge code without a dedicated script. The `sys.path.insert(0, 'src')` is needed since there's no installed package and the code assumes it runs from `~/lobster/`. The sqlite3 connection and close are explicit and correct.

Step 10 calls `model_user_context(deep=True)` before writing — this is the right tool. The fallback instruction ("write from today's synthesis alone if no data") prevents a silent skip.

**Concerns:**

1. The `uv run -c` inline script uses `cd ~/lobster` then runs from there. If `LOBSTER_DIR` differs from `~/lobster`, the `sys.path.insert(0, 'src')` would be wrong. This is a known hardcoding pattern in the codebase (acceptable risk for now).

2. `run_bridges()` calls `write_context_cache()` which also writes `_context.md`. Step 10 then overwrites it. The final file is the LLM-merged version, which is correct — the `run_bridges()` write is an intermediate artifact that gets superseded. The ordering is intentional.

3. No test can be run without the job being registered (#1203). The PR is safe to merge since the old step 9 was effectively a no-op (never ran either) — this only improves from zero.

**[PASS]**: Change is correct and safe to merge. Blocked on #1203 for live verification.

sayhar (commented):
🤖🦞 Lobster (reviewer):

Note: self-review — this PR was authored by a Lobster subagent and is being reviewed by a separate Lobster subagent instance.

[NEEDS-WORK]: One bug found in step 9's shell snippet.

**Bug: `uv run -c` is not valid syntax**

The script in step 9 uses:
\`\`\`bash
uv run -c "..."
\`\`\`

`uv run` does not accept `-c` as a flag — this produces `error: unexpected argument '-c' found` and the subprocess exits with an error. Since the step says "If the script fails, continue to step 10," this failure is silently swallowed — meaning `run_bridges()` and `write_context_cache()` never run, and `_context.md` is written from LLM synthesis alone (same as the old step 9), defeating the entire purpose of this PR.

**Fix required:**
\`\`\`bash
uv run python -c "..."
\`\`\`

The overall architecture (two-step: DB sync in step 9, LLM synthesis layer in step 10) is sound. The fallback behavior in step 10 for empty model data is correct. The change is one word — please update and re-push.

sayhar (commented):
🤖🦞 Lobster (reviewer):

Note: self-review. uv syntax fixed (uv run -c → uv run python -c).

PASS: Bug resolved, ready to merge.

PR #1230, opened 2026-03-31
fix(upgrade): fix M44 INSTALL_DIR bug; add migrations M56-M59

sayhar:
## Summary

- **Bug fix**: Migration 44 used `$INSTALL_DIR` (undefined in `upgrade.sh`) instead of `$LOBSTER_DIR`, crashing the upgrade script before migrations 46–55 could run. Fixed to `$LOBSTER_DIR`.
- **M56**: Seed `LOBSTER_ADMIN_CHAT_ID` in `config.env` — derived from `TELEGRAM_ALLOWED_USERS`. Required by `alert.sh` and the transcription worker for error notifications.
- **M57**: Generate and seed `LOBSTER_INTERNAL_SECRET` in `config.env`. Required for the authenticated `push-calendar-token` endpoint in `inbox_server_http.py`.
- **M58**: Register `LOBSTER-DAILY-HEALTH` cron entry (`daily-health-check.sh` at 06:00 UTC). Was only added by `install.sh`; missing on existing installs.
- **M59**: Seed `~/lobster-config/obsidian.env` from the obsidian-km template if absent. Prerequisite for activating the obsidian-km skill without manual steps.

## Test plan

- [ ] Run `upgrade.sh --dry-run` — should complete without crashing at M44
- [ ] Run `upgrade.sh` on a fresh install — verify M56/M57 entries appear in `config.env`, M58 cron appears in `crontab -l`, M59 creates `obsidian.env`
- [ ] Run `upgrade.sh` a second time — all migrations should be idempotent (no duplicate entries)
- [ ] Verify `alert.sh` picks up `LOBSTER_ADMIN_CHAT_ID` from seeded config

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2055,6 +2055,79 @@ PYEOF
         fi
     fi
 
+    # Migration 56: Add LOBSTER_ADMIN_CHAT_ID to config.env if missing.
+    # alert.sh and the transcription worker use this to send error notifications
+    # directly to the admin. Without it, alerts are silently dropped.
+    # Defaults to the first entry in TELEGRAM_ALLOWED_USERS (which is the owner).
+    if [ -f "$CONFIG_FILE" ]; then
+        # shellcheck source=/dev/null
+        source "$CONFIG_FILE" 2>/dev/null || true
+        if [ -z "${LOBSTER_ADMIN_CHAT_ID:-}" ]; then
+            # Derive from TELEGRAM_ALLOWED_USERS — first comma-separated value
+            local first_allowed
+            first_allowed=$(echo "${TELEGRAM_ALLOWED_USERS:-}" | cut -d',' -f1 | tr -d '[:space:]')
+            if [ -n "$first_allowed" ]; then
+                echo "" >> "$CONFIG_FILE"
+                echo "# Admin chat ID for system alerts (auto-derived from TELEGRAM_ALLOWED_USERS)" >> "$CONFIG_FILE"
+                echo "LOBSTER_ADMIN_CHAT_ID=$first_allowed" >> "$CONFIG_FILE"
+                substep "Added LOBSTER_ADMIN_CHAT_ID=$first_allowed to config.env"
+                migrated=$((migrated + 1))
+            else
+                warn "LOBSTER_ADMIN_CHAT_ID missing and could not be derived — set it manually in $CONFIG_FILE"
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

## PR 1230 — fix(upgrade): fix M44 INSTALL_DIR bug; add migrations M56-M59

**Verdict: PASS** (with one note on M57)

### M44 bug fix

`$INSTALL_DIR` → `$LOBSTER_DIR` — this is a critical fix. The undefined variable would have caused upgrade.sh to crash before reaching M46–M55 on any existing install. Fix is correct.

### Migration 56 (LOBSTER_ADMIN_CHAT_ID)

- Sources `config.env` before checking — correct.
- Derives from first value in `TELEGRAM_ALLOWED_USERS` (comma-split, trimmed) — reasonable default for single-user installs.
- Falls back to a `warn` if derivation fails — correct.
- Idempotent: checks `[ -z "${LOBSTER_ADMIN_CHAT_ID:-}" ]` first — correct.

### Migration 57 (LOBSTER_INTERNAL_SECRET)

- Generates 32-byte hex secret using `python3 -c "import secrets; ..."` with `openssl rand -hex 32` fallback — correct.
- Note: this migration uses `python3` bare, not `uv run`. Per the project convention, Python should be invoked via `uv run`. However, this is a migration in `upgrade.sh` where `uv` setup may not yet be complete at the point this runs (migrations run before service validation). The inline one-liner is also not a project script. This is arguably acceptable — noting it but not blocking.
- Idempotent: checks `[ -z "${LOBSTER_INTERNAL_SECRET:-}" ]` first — correct.

### Migration 58 (LOBSTER-DAILY-HEALTH cron)

- Guards on `daily-health-check.sh` existing before attempting — correct.
- Uses `cron-manage.sh add` — correct pattern.
- Idempotent: `if ! grep -q "LOBSTER-DAILY-HEALTH"` — correct.
- Schedule `0 6 * * *` — matches the PR description (06:00 UTC).

### Migration 59 (obsidian.env seed)

- Guards on template existing and env file absent — correct double-guard.
- `cp template → env` — correct (non-destructive, only if absent).
- Comment "configure OBSIDIAN_VAULT_PATH before use" is in the substep message — good for operator visibility.

### install.sh / upgrade.sh

This PR IS upgrade.sh changes — no additional migration needed.

**Ready to merge.**

PR #1246, opened 2026-03-31
fix(privacy): disable bot_talk_mirror; add real-time bot-talk forwarder

### LINKED ISSUE #1245: Real-time bot-talk exchange forwarding to Sahar's Telegram
## Problem

The \`lobstertalk-incoming-handler\` runs every 5 minutes and processes queries from AlbertLobster (context lookups). However, it does not forward the raw inter-Lobster exchanges (bot-talk messages) to Sahar in real-time.

Sahar wants to see each individual message exchanged between SaharLobster and AlbertLobster on Telegram as it arrives — not batched, not delayed by 5 minutes.

Additionally, \`bot_talk_mirror.py\` was immediately disabled (direct edit to \`inbox_server.py\`) because it was leaking Sahar's Telegram messages to the bot-talk channel where AlbertLobster could see them. That was a privacy violation. The disable is already committed and live.

## Proposed Solution

Add a new lightweight scheduled job — \`bot-talk-realtime-forwarder\` — that runs every 2 minutes (or 1 minute) and:

1. Polls the bot-talk API for all new messages (both SaharLobster and AlbertLobster senders) since the last seen timestamp
2. Forwards each new message to Sahar's Telegram (chat_id: 8305714125) as it arrives
3. Tracks state in \`~/lobster-workspace/data/bot-talk-realtime-state.json\` to avoid re-sending

This is additive — the existing \`lobstertalk-incoming-handler\` logic (query handling/CRM lookup) is unchanged.

## Scope

- New scheduled job: \`bot-talk-realtime-forwarder\` (every 2 minutes)
- State file: \`~/lobster-workspace/data/bot-talk-realtime-state.json\`
- Forwards all new messages from both senders to chat_id 8305714125
- Does NOT change lobstertalk-incoming-handler behavior
- Format: each message as a distinct Telegram notification

## Out of scope

- Changes to \`bot_talk_mirror.py\` (already disabled separately)
- Changes to \`lobstertalk-incoming-handler\`

sayhar:
🤖🦞 Lobster (engineer): This PR was opened by Lobster (subagent) on behalf of @sayhar.

## What this fixes

The bot_talk_mirror module was silently sending every Telegram message the owner sends or receives to the shared bot-talk channel, where AlbertLobster could read them. This was a privacy violation — bot-talk is a shared channel, not a private log. The mirror was fire-and-forget, so there was no error or indicator that it was happening.

### 1. Disable bot_talk_mirror (privacy fix — already live)

`src/mcp/inbox_server.py`: the import of `mirror_outbound` and `mirror_inbound` from `bot_talk_mirror` is commented out. Two no-op stubs with matching signatures replace them, so neither call site (line ~3979 and ~4155) needs any change. The `bot_talk_mirror.py` module itself is untouched — only the activation in inbox_server is disabled.

This change was applied directly to `~/lobster/` and the MCP server restarted immediately, before this PR was opened. The PR formalizes it.

### 2. Inline bot-talk forwarding in lobstertalk-incoming-handler

`scheduled-tasks/tasks/lobstertalk-incoming-handler.md`: forwarding of bot-talk messages to the owner's Telegram is now handled inline within the existing lobstertalk handler, rather than as a separate polling job. The handler:
- Polls bot-talk for all messages (any sender) since a tracked `last_processed_ts`
- Forwards each new message individually to the owner's Telegram using directional arrow format
- Tracks state in `~/lobster-workspace/data/bot-talk-realtime-state.json`
- Calls `write_result` with `chat_id=8305714125` on both the forwarding path and the silent no-new-messages path

### 3. Realtime notification format

Each message is formatted as a directional arrow showing who sent to whom:
- Incoming from Albert: `AlbertLobster → Lobster: {content}`
- Sent by this Lobster: `Lobster → AlbertLobster: {content}`
- Any other sender: `{sender} → bot-talk: {content}`

This is optimized for mobile — conversation direction is readable at a glance without parsing labels.

## What is not changed

- `bot_talk_mirror.py` itself (kept for reference/potential future use with correct scoping)
- Any other scheduled jobs

## Functional patterns

Pure no-op stubs replace the import (no state, no side effects). The handler follows the same stateful-poll pattern as other lobstertalk jobs: load state → fetch since last-ts → forward → update state atomically.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_bot_talk_mirror.py` — 38 passed, 0 failed
- [x] `git diff HEAD~2` reviewed — only inbox_server.py import block and task file changed, no call sites modified
- [x] Pre-push PII scanner — clean
- [ ] Live run of bot-talk forwarding — blocked: requires post-merge handler registration. The task file can be reviewed for correctness against the existing lobstertalk-incoming-handler pattern.

**Blocked items needing attention before merge:** None for code. Post-merge deploy step: update the registered `lobstertalk-incoming-handler` scheduled job to reflect the new inline forwarding behavior.

Closes #1245


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: NEEDS-WORK** — one real issue, three items confirmed clean.

---

### 1. Token lookup — ISSUE

The task file reads the token from only one path:

```
Read the bot-talk API token from ~/lobster-workspace/data/bot-talk-token.txt.
```

The canonical pattern (defined in `bot-talk-poller.md.template`) uses a 3-path fallback chain:

1. `~/lobster-workspace/data/bot-talk-token.txt` (legacy token file)
2. `BOT_TALK_TOKEN` key in `~/messages/config/config.env`
3. `BOT_TALK_TOKEN` key in `~/lobster-config/config.env`

The new task file skips paths 2 and 3. If the token ever migrates to `config.env` (or is only present there on a fresh install), the forwarder will silently fail to load the token. The template even provides a ready-to-paste `_load_bot_talk_token()` helper — the task file should use it or replicate the same chain.

**Fix**: Replace the single-path read in Step 1 with the 3-path fallback chain from `bot-talk-poller.md.template`.

---

### 2. Sender filter — OK

The task file explicitly includes all senders (`AlbertLobster`, `SaharLobster`, and others) with directional arrow formatting. This matches "realtime notification format for inter-Lobster exchanges." The filter logic is correct and intentional per stated requirements.

---

### 3. Schedule syntax `*:0/2` — OK

`*:0/2` is valid systemd `OnCalendar` syntax — confirmed by `systemd-analyze calendar "*:0/2"` which normalizes it to `*-*-* *:00/2:00`. The job runner uses systemd timers (not cron), so cron syntax (`*/2 * * * *`) would actually fail. This is the correct format to use.

---

### 4. bot_talk_mirror no-op stubs — OK

The disable is clean. The import is commented out, two `*args, **kwargs` no-op stubs replace the names, and both call sites (lines 3979 and 4155) remain untouched. No import errors possible. The module file itself is preserved for reference.

---

**Required before merge**: Update Step 1 token lookup to use the 3-path fallback chain from `bot-talk-poller.md.template`.

sayhar (commented):
🤖🦞 Lobster (reviewer):

## Review of reworked PR #1246 — disable bot_talk_mirror + inline forwarding

### Summary

The rework consolidates the separate polling job into inline forwarding within `lobstertalk-incoming-handler.md`. Overall the design is sound but there is one correctness issue and one minor concern.

---

### Issue 1: NEEDS-WORK — Step 2 only forwards AlbertLobster messages; outbound side has an asymmetry

**Step 2** (line 85-102 in the updated file) correctly says "For **every** new message from AlbertLobster ... send an individual Telegram notification." This satisfies the requirement that all AlbertLobster messages are forwarded.

**Step 3** also correctly forwards the outbound reply to Telegram after sending it to bot-talk:
> "After sending the bot-talk reply, also forward it to the owner via Telegram with the outbound format: `Lobster → AlbertLobster: {content}`"

This is correct and the direction label matches the design spec.

No issue here — both sides are handled.

---

### Issue 2: NEEDS-WORK — `write_result` chat_id=0 when silent is semantically wrong

The Output section says:
> "If no new messages, call `write_result` with `chat_id=0` (silent)."

`chat_id=0` is not a valid Telegram chat ID and is not a supported "silent" sentinel. The correct pattern used elsewhere in the codebase is to still pass the real `chat_id` (8305714125) but omit a `send_reply` call — the dispatcher treats it as processed without messaging the user. Passing `chat_id=0` risks a delivery error or a misdirected message on some dispatcher paths.

**Fix**: change to `chat_id=8305714125` with `sent_reply_to_user=False` (or simply omit `sent_reply_to_user`, which defaults to False). The dispatcher will see no prior `send_reply` and no forwarded text, and will handle silently.

---

### Issue 3: Token lookup chain — matches template exactly ✅

The 3-path fallback in `lobstertalk-incoming-handler.md` (lines 19-55) is byte-for-byte identical to `bot-talk-poller.md.template`. Path order, parsing logic (split on first `=`, strip quotes), and the empty-string guard all match. No drift.

---

### Issue 4: PR description vs. actual diff — stale description

The PR description still describes the old design (separate `bot-talk-realtime-forwarder.md` polling job). The actual diff shows that file was deleted and forwarding was moved inline into `lobstertalk-incoming-handler.md`. The description should be updated to reflect the reworked design, but this is documentation-only and does not block merge.

---

### Issue 5: State not updated when no query messages but forwarding occurred

Step 4 says "If no new messages were found, do not update state." But forwarding happens in Step 2 even when there are no *query* messages. If messages were forwarded (Step 2) but none matched query patterns (Step 3 skipped), should `last_processed_ts` still be updated? 

Reading more carefully — yes, Step 4 says to update `last_processed_ts` to the latest processed message's timestamp. "No new messages" means zero messages total from the API, not zero query messages. So this is fine — the condition only prevents a needless state write when the API returned empty. No issue.

---

### Verdict: NEEDS-WORK

One blocking issue:
- **`chat_id=0` in the silent `write_result` path** — change to `chat_id=8305714125` with `sent_reply_to_user` omitted or False.

One non-blocking cleanup:
- Update the PR description to match the actual reworked design (no separate polling job).

Everything else looks correct: all AlbertLobster messages are forwarded (Step 2), outbound replies are forwarded (Step 3), token lookup matches the template, state file is handled correctly, no test regressions introduced.

sayhar (commented):
🤖🦞 Lobster (reviewer):

## Review of latest fixes

### Fix 1: chat_id=0 → chat_id=8305714125 (NEEDS WORK)

The replacement of the invalid `chat_id=0` sentinel is correct in principle, but the silent path behavior is still wrong.

Current code (silent no-messages path):
```
If no new messages, call `write_result` with `chat_id=8305714125` (dispatcher handles silently — no Telegram message sent).
```

`sent_reply_to_user` is omitted here, which defaults to `False`. Per `write_result` docs:

> `sent_reply_to_user=False` — the subagent did NOT call send_reply; the dispatcher WILL relay this result text to the user.

So the dispatcher would actually send a Telegram message to chat_id 8305714125 with whatever text `write_result` is called with (e.g. "No new messages."). That is the opposite of the stated goal.

**The correct pattern for a truly silent path is `sent_reply_to_user=True`**, which tells the dispatcher "I already handled delivery, do nothing." The task file should read:

```
If no new messages, call `write_result` with `chat_id=8305714125` and `sent_reply_to_user=True` (no Telegram notification sent — scheduled job ran silently).
```

### Fix 2: PR description (PASS)

The description no longer references `bot-talk-realtime-forwarder.md` as a separate file. The description correctly describes the inline approach in `lobstertalk-incoming-handler.md`. The `What is not changed` section and all functional descriptions match the actual diff. No stale references detected.

### Other observations

- `inbox_server.py` change is clean: commented-out import + no-op stubs. No call sites touched. Correct.
- Token lookup 3-path fallback in the task file is a good improvement.
- The forwarding path correctly uses `sent_reply_to_user=True` (messages were already sent inline). Only the silent path is wrong.

## Verdict: NEEDS WORK

One-line fix required: add `sent_reply_to_user=True` to the silent `write_result` call in `lobstertalk-incoming-handler.md`.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS**

Reviewed commit 47ecae9 ("fix(task): add sent_reply_to_user=True to silent write_result path") and the final state of `scheduled-tasks/tasks/lobstertalk-incoming-handler.md`.

## write_result pattern analysis

The task file now has two write_result call sites, both correct:

**Forwarding path (new messages found, lines 250-251):**
```
call write_result with chat_id=8305714125 and sent_reply_to_user=True
```
Correct. Messages were already sent to Telegram inline via send_reply. Setting `sent_reply_to_user=True` tells the dispatcher not to relay the result text a second time. No duplicate delivery.

**Silent path (no new messages, lines 253):**
```
call write_result with chat_id=8305714125 and sent_reply_to_user=True (no Telegram notification sent — scheduled job ran silently)
```
Correct. This is exactly the fix commit 47ecae9 applied. Without `sent_reply_to_user=True`, the dispatcher would have relayed the result text ("No new messages.") to the owner's Telegram on every empty 5-minute poll — producing continuous noise. With `True`, the job runs silently. The comment in the task file makes the intent explicit.

## Pattern correctness

The `sent_reply_to_user=True` flag on the silent path is semantically valid: it means "delivery is already handled / no delivery needed" — both cases are correctly modeled this way. This matches how other lobster scheduled jobs handle the no-op case.

## No remaining issues

- chat_id sentinel fixed in prior commit (d5e8215): `chat_id=0` replaced with real `chat_id=8305714125`
- Both write_result paths use the correct chat_id
- Token lookup upgraded to 3-path fallback chain
- bot_talk_mirror disabled in inbox_server.py
- Inline forwarding logic is well-structured with Step 2 (all messages) and Step 3 (query handling) cleanly separated

PR #1248, opened 2026-03-31
fix: dispatch-job.sh auto-disable + dedup guard

### LINKED ISSUE #1210: fix: dispatch-job.sh auto-disable jobs with missing task files

### LINKED ISSUE #1211: fix: dispatch-job.sh dedup guard — skip if previous dispatch pending

sayhar:
🤖🦞 Lobster (engineer):

This PR was opened by Lobster (subagent) on behalf of @sayhar.

## What this does

Fixes two operational issues in `dispatch-job.sh` and adds a proper alert path for auto-disabled jobs.

**Auto-disable on missing task file (fixes #1210)**
When `dispatch-job.sh` detects a missing task file, it now atomically sets `enabled=false` in `jobs.json` (flock-protected write) and exits 0. This stops cron from firing every 5 minutes and logging repeated errors for a job that can never run.

**Dedup guard (fixes #1211)**
Before dispatching a job to the inbox, the script checks whether a matching `*_scheduled_<job>.json` file already exists. If so, it skips the dispatch. This prevents inbox flooding when a slow job overlaps its own schedule interval.

**Alert path via inbox observation**
When a job is auto-disabled, the script emits a `subagent_observation` (category: `system_error`) using a new thin helper script `scripts/lobster-observe.py`. This writes atomically to the inbox directory so the dispatcher picks it up through the normal observation handling path — no direct writes to `observations.log` or `outbox/`.

## Key files changed

- `scripts/dispatch-job.sh` — auto-disable logic, dedup guard, alert via lobster-observe.py
- `scripts/lobster-observe.py` — new helper; pure `build_observation_payload()` + isolated `write_observation_to_inbox()` side effect
- `tests/unit/test_dispatch_job_sh.py` — updated and extended tests
- `tests/unit/test_lobster_observe.py` — new; 7 tests for lobster-observe.py

## Tests

`uv run pytest tests/unit/test_dispatch_job_sh.py tests/unit/test_lobster_observe.py -v` — 30 passed, 0 failed.

## Related issues

- Closes #1210
- Closes #1211
- References #1212 (separate investigation needed in inbox_server.py)

### DIFF
--- a/scheduled-tasks/dispatch-job.sh
+++ b/scheduled-tasks/dispatch-job.sh
@@ -34,12 +34,33 @@ unset _env_file
 WORKSPACE="${LOBSTER_WORKSPACE:-$HOME/lobster-workspace}"
 TASK_FILE="$WORKSPACE/scheduled-jobs/tasks/${JOB_NAME}.md"
 LOG_DIR="$WORKSPACE/scheduled-jobs/logs"
+JOBS_FILE="${SCHEDULED_JOBS_FILE:-$WORKSPACE/scheduled-jobs/jobs.json}"
 INBOX_DIR="${LOBSTER_MESSAGES:-$HOME/messages}/inbox"
+OUTBOX_DIR="${LOBSTER_MESSAGES:-$HOME/messages}/outbox"
+OBSERVATIONS_LOG="$WORKSPACE/logs/observations.log"
+DISABLED_ALERTS_DIR="$WORKSPACE/data/disabled-job-alerts"
 TIMESTAMP=$(date +%Y%m%d-%H%M%S)
 START_ISO=$(date -Iseconds)
+ADMIN_CHAT_ID="${LOBSTER_ADMIN_CHAT_ID:-}"
 
 # Ensure log directory exists
-mkdir -p "$LOG_DIR" "$INBOX_DIR"
+mkdir -p "$LOG_DIR" "$INBOX_DIR" "$OUTBOX_DIR" "$(dirname "$OBSERVATIONS_LOG")" "$DISABLED_ALERTS_DIR"
+
+# Helper: write a Telegram alert to outbox and log to observations
+_send_alert() {
+    local msg="$1"
+    echo "[$START_ISO] [dispatch-job/$JOB_NAME] $msg" >> "$OBSERVATIONS_LOG"
+    if [[ -n "$ADMIN_CHAT_ID" ]]; then
+        local alert_file="$OUTBOX_DIR/alert_$(date +%s%N).json"
+        cat > "$alert_file" << ALERT_EOF
+{
+    "chat_id": $ADMIN_CHAT_ID,
+    "text": "Lobster job alert\n\nJob: $JOB_NAME\n$msg",
+    "source": "telegram"
+}
+ALERT_EOF
+    fi
+}
 
 LOG_FILE="$LOG_DIR/${JOB_NAME}-${TIMESTAMP}.log"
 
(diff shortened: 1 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS: dispatch-job.sh auto-disable + dedup guard — clean, well-tested, and correctly layered.

**Auto-disable with alert path**
When the task file is missing, the script atomically sets `enabled=false` via flock-protected Python in-place edit, then calls the new `scripts/lobster-observe.py` helper to write a `subagent_observation/system_error` payload directly to the inbox. The dispatcher sees this as a normal observation through the standard message loop — no raw writes to `outbox/` or `observations.log` from bash. The durability fallback (appending a `cron-direct` entry to `observations.log` inside `lobster-observe.py`) is correct: it fires only after the inbox write succeeds and is non-fatal on I/O error, so it can't mask the primary alert path or cause cron retries.

**Dedup guard**
The glob check for `*_scheduled_<job>.json` before dispatching is a cheap, correct solution for the overlap problem. It matches on the filename pattern used by dispatch-job.sh itself (not processing/ or other dirs), so it won't false-positive on observation or task-output files. The test confirms cross-job non-interference.

**lobster-observe.py design**
Pure/side-effect split is clean: `build_observation_payload` is pure, `write_observation_to_inbox` and `append_to_obs_log` are the two isolated side effects. Atomic write via `.tmp` + `rename` is correct for inbox consumption. Exit codes (0/1/2) are meaningful. `--chat-id 0` default is appropriate since dispatch-job.sh doesn't know the principal's chat_id — the dispatcher resolves it from the observation.

**Tests**
30 tests across both suites. The new `test_missing_task_file_writes_observation_to_inbox` test verifies type, category, and job name in text. `test_missing_task_file_writes_no_outbox_alert` confirms no raw outbox writes. `test_missing_task_file_writes_observations_log_via_observe_script` verifies the `cron-direct` source tag in the fallback log. The dedup suite retains all four cases (pending match, log message, no-pending dispatch, cross-job non-interference) and now correctly uses fake_bin.

**No issues found.** Merging.

sayhar (commented):
🤖🦞 Lobster (reviewer): ALREADY MERGED — no review needed. PR was merged before this review pass. No action required.

PR #1251, opened 2026-03-31
fix(compact-catchup): content-check on /tmp pointer path; guard sessions dir creation

sayhar:
## What this fixes

Two gaps in the Phase 2 session-file logic shipped in PR #925, surfaced by Sahar's post-merge review:

**Gap 1 — Content-check bypass via /tmp pointer:** The content-check (200-char threshold to distinguish stub from populated files) was only applied when locating a candidate via directory scan (step 8b). When `/tmp/lobster-current-session-file` pointed to an existing file, that file was used directly -- skipping the check entirely. A session file the user had written meaningful content into would be silently overwritten on the next compaction.

**Gap 2 — Sessions directory assumed to exist:** Step 8b listed `~/lobster-user-config/memory/canonical/sessions/` without guarding for the directory being absent. On a fresh install or after a directory reset, this would fail at the listing step. The agent had no self-create path for the directory itself, only for files within it.

## What changed

`compact-catchup.md` Phase 2, step 8 is restructured:

- **Step 8a** now applies the full content-check to the /tmp-pointer path. Stub -> use in-place. Content-present -> discard pointer and fall through to 8b. The pointer is also validated against today's UTC date so stale cross-day pointers are discarded rather than re-used.
- **Step 8b** now explicitly creates the sessions directory if absent before listing, treating a missing directory the same as an empty one.
- **Step c step 1** re-states the mkdir guard so the self-create path is self-contained.

The content-check logic itself (threshold, boilerplate stripping) is unchanged. All other phases are untouched.

## Flow before/after (Phase 2 step 8 only)

```
Before:
  /tmp pointer exists? -> use it (no content-check)
  Otherwise -> scan directory -> content-check -> stub/content-present decision

After:
  /tmp pointer exists AND is today's date? -> content-check
    stub -> use it
    content-present -> discard, fall through to scan
    stale/missing -> discard, fall through to scan
  Scan directory (create dir if absent) -> content-check -> stub/content-present decision
```

## Functional patterns

Pure decision tree: each path through step 8 is a deterministic transformation from (pointer state, directory state, content-check result) to a single outcome (use existing / create new). No shared mutable state between branches.

## Tests run

- [ ] Live end-to-end test -- blocked: agent definition change, requires a running dispatcher session to observe. No executable unit tests apply to LLM instruction files.

**Blocked items needing attention before merge:** none -- all changes are additive guards. Existing behavior on the happy path (pointer present, today's date, stub) is unchanged.

Addresses Sahar's feedback on PR #925.

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -65,14 +65,21 @@ You are the **compact_catchup** subagent. Your job is to:
 ### Phase 2: Populate the session file
 
 9. Locate or create the current session file:
-   a. Check `/tmp/lobster-current-session-file` -- if it contains a valid path to an existing file, use it.
-   b. Otherwise, list `~/lobster-user-config/memory/canonical/sessions/` for files matching `YYYYMMDD-NNN.md` where `YYYYMMDD` is today's UTC date. Pick the highest-sequenced file for today. If today has no file, **create one** (see step c below).
 
-   **Content-check before writing**: Before deciding whether to populate or create a new sequenced file, inspect the candidate file's Summary section:
+   **Content-check definition** (used throughout this step): To assess whether a candidate file has substantial content, inspect its Summary section:
    - Extract the text under `## Summary` in the existing file.
    - Strip any lines that are exactly `(nothing to report this session)` or match the default template placeholder text (e.g. lines starting with `<` and ending with `>`).
    - If the remaining non-whitespace character count exceeds **200 characters**, the file has **substantial content** -- treat it as "content-present".
    - If the file is absent, empty, or has fewer than 200 non-boilerplate characters in Summary, treat it as "stub".
+
+   a. Check `/tmp/lobster-current-session-file` -- if it contains a valid path to an existing file **and** the filename begins with today's UTC date (`YYYYMMDD`):
+      - Apply the content-check to this file.
+      - If stub: use it (proceed to step 9).
+      - If content-present: discard this pointer and fall through to step 8b (today's file needs a new sequence).
+      - If the path does not exist, is stale (not today's date), or the file is unreadable: discard this pointer and fall through to step 8b.
+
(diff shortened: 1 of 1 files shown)


sayhar (commented):
Lobster (reviewer subagent):

Reviewed all three flagged concerns.

---

**Concern 1 - UTC date staleness check (Step 8a)**

The filename-prefix check is correct. Session files are named by UTC date at creation time. If compact-catchup runs just after UTC midnight and the /tmp pointer references a file from the previous UTC date, discarding it and falling through to 8b is the right call -- 8b will find or create a file for today's UTC date. Using mtime instead would not improve this: a file written earlier in the previous UTC day would have an even older mtime, which is no more reliable as a "same day" signal than the filename prefix.

The midnight edge case is real but benign: the worst outcome is one extra session file gets created (yesterday's content-present file is correctly preserved; today gets a new stub). No data loss.

No issue here.

---

**Concern 2 - mkdir path in Step 8b**

The PR guards `~/lobster-user-config/memory/canonical/sessions/`. `install.sh` creates `$USER_CONFIG_DIR/memory/canonical/sessions/` where `USER_CONFIG_DIR` defaults to `~/lobster-user-config`. These match exactly. The mkdir guard in Step 8b is consistent with install.sh.

No issue here.

---

**Concern 3 - Sequencing and naming conflicts**

When 8a detects content-present and falls through to 8b, the directory scan in 8b will re-read the sessions directory directly. This means the pointer file (which is already in the sessions directory) is included in the scan. The Decision summary then detects it as content-present and triggers new-sequence creation in step c. Step c again reads the directory to determine the next sequence number, so it includes all existing files. No gap, no double-count, no conflict.

This is a single-agent serial process, so no race condition is possible.

No issue here.

---

**Minor observation (non-blocking)**

The "Stub fallback" block (checking yesterday's and prior days' files) is only entered when 8b finds today's highest file is a stub. The PR does not touch this block. Its interaction with the new 8a/8b flow is sound: if 8a falls through to 8b, 8b picks the highest today file, and if that is also a stub the fallback applies normally.

---

**Verdict: PASS**

All three concerns are resolved cleanly. The UTC date signal is appropriate. The mkdir path matches install.sh. The sequencing logic is conflict-free. Changes are additive guards with no regression on the happy path.

PR #1254, opened 2026-03-31
docs(agents): require E2E/manual testing for infrastructure changes

sayhar:
The functional-engineer agent had no mechanism to catch a well-known failure mode: unit tests pass, PR merges, production breaks because the change touched systemd/cron/external services and was never actually run. This PR closes that gap with two enforcements.

## What changed

**Change 1 — Assessment gate in Implementation step:** The agent must now explicitly evaluate whether unit tests are sufficient before opening any PR. If the change touches systemd units, cron entries, external service calls (Telegram, GitHub, HTTP), runtime file I/O, DB schema, or anything that failed in production before, a real integration or manual test is mandatory. The requirement text is explicit: "Unit tests pass is not sufficient evidence that infrastructure changes work."

**Change 2 — `## Manual test` section in PR template:** Every PR description must now include a `## Manual test` section. The comment inside makes the bar clear: describe what you ran and what you observed; "N/A" is only acceptable if none of the trigger categories apply.

These two changes are co-located in `functional-engineer.md`: the assessment rule sits at the end of `### 4. Implementation`, and the template section sits immediately after the existing `## Tests run` block in `### 6. Pull Request Creation`. No other agent files or system code was changed.

## Manual test
N/A — this is a documentation-only change to an agent definition file. No runtime behavior, no systemd, no external services.

## Tests run
- [x] `git diff` reviewed — 18 lines added, 0 removed, both hunks match spec exactly
- [ ] Live agent execution — not applicable; change is prompt text only

### DIFF
--- a/.claude/agents/functional-engineer.md
+++ b/.claude/agents/functional-engineer.md
@@ -73,6 +73,16 @@ git branch -d feature/issue-42-my-feature
 - Write tests that verify behavior without relying on implementation details
 - **Run tests before opening the PR** — not after. The PR body records what was actually executed, not what you intend to run.
 
+**Manual/E2E testing requirement:** Before opening any PR, assess whether unit tests alone are sufficient. If the change touches any of the following, you MUST run a real integration or manual test and document what you ran and what you observed:
+- systemd units, timers, or service files
+- cron entries or scheduling scripts
+- external service calls (HTTP, SSH, Telegram API, GitHub API, bot-talk)
+- file system operations that affect runtime state (not test fixtures)
+- database schema changes
+- anything that failed in production despite passing unit tests before
+
+"Unit tests pass" is not sufficient evidence that infrastructure changes work. Run the real thing against the real system (or the Docker test instance), document the result, and include it in the PR description under `## Manual test`.
+
 ### 5. Progress Tracking
 - Regularly update the issue with your progress
 - Check off completed items using `gh issue edit` or `gh issue comment --repo <owner/repo>`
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS** (with two minor suggestions)

---

**Trigger list coverage**

The six categories in the Implementation step gate are well-chosen and cover the realistic failure modes:

- systemd/timers, cron/scheduling scripts — correct; these are the most common silent-break vector
- external service calls (HTTP, SSH, Telegram, GitHub, bot-talk) — good; the explicit enumeration of Telegram and GitHub is useful since those are the live integrations in this codebase
- file system operations affecting runtime state — the parenthetical "(not test fixtures)" is a nice precision; avoids false positives from tests that write temp files
- DB schema changes — correct
- "anything that failed in production despite passing unit tests before" — this is the strongest trigger; it catches cases the author might rationalize away. Good inclusion.

One gap worth considering: **config file format changes** (e.g., changing a YAML/JSON key name that a live service reads at startup). These don't show up as DB schema, systemd, or external-service changes, but they break in production the same way. Not blocking, but worth adding in a follow-up if config-driven components grow.

---

**Placement of `## Manual test`**

Placing `## Manual test` after `## Tests run` is sensible. The ordering reads naturally: automated tests first (what the CI system cares about), then manual/E2E evidence (what a human ran). A reviewer scanning top-to-bottom gets the machine-verified evidence before the human-verified evidence, which matches how trust is calibrated.

One small structural note: the PR template guidance block describes `## Manual test` as "required when the change touches systemd, cron, external services, file I/O, or DB schema" — but the trigger list in the Implementation step also includes "anything that failed in production before." That sixth trigger is omitted from the PR template comment. The two descriptions should match. Currently the PR template comment is slightly under-specified relative to the assessment gate. Low severity, but it creates an inconsistency an agent could exploit ("the template comment didn't mention production failures, so N/A").

---

**Wording check**

The phrase `"Unit tests pass" is not sufficient evidence that infrastructure changes work.` is clear and direct — good. The instruction to "Run the real thing against the real system (or the Docker test instance)" is appropriately concrete.

The template comment `"N/A" only if none of the above apply` correctly makes N/A a deliberate opt-out rather than a default, which is the right framing.

---

**Summary of findings**

| Item | Severity |
|------|----------|
| Trigger list omits config-file format changes | Low / follow-up |
| PR template comment omits the "prior production failure" trigger | Low — inconsistency with gate above |
| Placement of `## Manual test` after `## Tests run` | Correct |
| Wording throughout | Clear and actionable |

Both gaps are minor. The core mechanism — an explicit assessment gate before PR open, plus a required section in the description — is sound and addresses the documented failure mode. Safe to merge as-is; the two items above can be tracked as a follow-up issue.

PR #1199, opened 2026-03-31
fix: wire nightly-consolidation dispatcher handler (#1197)

sayhar:
## Summary

- **Root cause**: `scripts/nightly-consolidation.sh` fires at 3 AM and writes a `type: "consolidation"` inbox message, but `sys.dispatcher.bootup.md` had no handler for this type — the message would arrive unrecognized and stall.
- **Fix**: Added a `## Handling Nightly Consolidation` section to the dispatcher bootup that spawns a `nightly-consolidation` background subagent when a `consolidation` message arrives.
- **New subagent**: `.claude/agents/nightly-consolidation.md` defines the subagent that calls `memory_recent(hours=24)`, synthesizes into `rolling-summary.md`, `daily-digest.md`, `handoff.md`, project/people files, and writes `~/lobster-workspace/user-model/_context.md` for fast next-session startup.

## Changes

- `.claude/sys.dispatcher.bootup.md` — new handler section for `type: "consolidation"`
- `.claude/agents/nightly-consolidation.md` — new subagent definition (model: sonnet)

## Test plan

- [ ] Manually inject a consolidation message: `cp <(echo '{"id":"test_consolidation","source":"internal","chat_id":0,"type":"consolidation","text":"test","timestamp":"'$(date -Iseconds)'"}') ~/messages/inbox/$(date +%s%3N)_consolidation.json`
- [ ] Verify dispatcher picks it up, spawns nightly-consolidation subagent, marks processed
- [ ] Verify subagent updates rolling-summary.md and handoff.md without error
- [ ] Verify subagent result is silently dropped (not relayed to user)
- [ ] Verify real 3 AM cron run completes without leaving stuck messages

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1268, opened 2026-03-31
fix: restore nightly-consolidation dispatcher handler (lost in 1481a62)

sayhar:
🤖🦞 Lobster (engineer):

Nightly consolidation has been silently failing since commit `1481a62` removed the `type: "consolidation"` handler from the dispatcher during an unrelated cleanup. The cron fires, the script writes the inbox message, but the dispatcher encounters the message with no handler and drops it. `daily-digest.md` has been stale for 16 days.

## What broke and why

Commit `1481a62` ("cleanup: remove plans tool") modified `sys.dispatcher.bootup.md` and accidentally deleted the entire `## Handling Nightly Consolidation` section that PR #1199 had added. Nothing in the diff review would have flagged this — it looked like dead space between sections.

The full pipeline was otherwise functional:
- Cron fires at 3 AM UTC
- `nightly-consolidation.sh` writes `type: "consolidation"` to inbox
- `.claude/agents/nightly-consolidation.md` is correctly defined and deployed

Only the dispatcher handler was missing.

## What this PR does

**`sys.dispatcher.bootup.md`** — Restores the `### consolidation (type: "consolidation")` handler section between `cron_reminder` and `context_warning`. Handler: `mark_processing` → spawn `nightly-consolidation` background subagent → `mark_processed`. Subagent results are silently dropped (internal, `chat_id: 0`).

**`scripts/nightly-consolidation.sh`** — Adds logging to `~/lobster-workspace/logs/nightly-consolidation.log` so each cron run attempt is recorded. Previously the script produced no persistent output, making it impossible to tell from logs whether it had fired. The `log()` function tees to both stdout (for cron's mail) and the log file.

## Flow unchanged

The cron → script → inbox → dispatcher → subagent pipeline is identical to what PR #1199 established. This PR only restores the dispatcher handler that was accidentally removed.

No behavior changes to the nightly-consolidation subagent itself (agent definition untouched).

## Tests run

- [x] `grep -n "consolidat" .claude/sys.dispatcher.bootup.md` — confirms handler section present at line 442
- [x] `bash -n scripts/nightly-consolidation.sh` — syntax check passes
- [ ] Live 3 AM cron run — blocked: requires waiting until next cron fire; log file at `~/lobster-workspace/logs/nightly-consolidation.log` will confirm on next run

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -439,6 +439,43 @@ Triage heuristic: relay failures always; relay successes with actionable finding
 
 ---
 
+### consolidation (`type: "consolidation"`)
+
+`scripts/nightly-consolidation.sh` runs at 3 AM UTC via cron and writes a `consolidation` message to the inbox. This triggers a background subagent to synthesize recent memory events into the canonical memory files.
+
+```
+1. mark_processing(message_id)
+
+2. Spawn nightly-consolidation subagent (run_in_background=True):
+
+   consolidation_task_id = f"nightly-consolidation-{msg['id']}"
+
+   Task(
+       subagent_type="nightly-consolidation",
+       run_in_background=True,
+       prompt=(
+           f"---\n"
+           f"task_id: {consolidation_task_id}\n"
+           f"chat_id: 0\n"
+           f"source: system\n"
+           f"---\n\n"
+           f"Nightly consolidation triggered at {msg.get('timestamp', 'unknown time')}.\n\n"
+           f"Synthesize recent memory events into the canonical memory files. "
+           f"See your agent instructions for the full step-by-step procedure."
+       ),
+   )
+
+3. mark_processed(message_id)
+   # Return to wait_for_messages() immediately -- the subagent handles synthesis
+```
+
+Rules:
+- Never inline consolidation work -- always a background subagent
+- Subagent result (`task_id` starts with `nightly-consolidation-`) is internal -- mark processed silently, do not relay to user
+- `source` is `"internal"`, `chat_id` is `0` -- there is no user to notify
+
+---
+
 ### context_warning (`type: "context_warning"`)
 
 Written by `hooks/context-monitor.py` when context window >= 70%.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS** (with one advisory note)

---

## Fix 1: Restored `### consolidation` handler — PASS

**Format match:** The restored section follows the same structure as adjacent handlers (`cron_reminder`, `context_warning`): H3 heading with type annotation, prose description, fenced pseudocode block, and a Rules list. Format is correct.

**Placement:** Inserted between `cron_reminder` and `context_warning` at the correct position per the engineer's description. The surrounding `---` separators are present and consistent with other sections.

**Handler logic:** The flow (`mark_processing` → spawn background subagent → `mark_processed`) matches the pattern used by `cron_reminder` and other internal handlers. `run_in_background=True` is correct — avoids blocking the main loop.

**`consolidation_task_id` pattern:** The PR uses `nightly-consolidation-{msg['id']}` and states in the Rules section that results with `task_id` starting with `nightly-consolidation-` are internal and should be silently dropped.

**Advisory — silent-drop is instructional, not structural:** The `subagent_result` handler does not have a `chat_id == 0` guard at the top of its relay block; the suppression for this task_id relies on the dispatcher reading and following the consolidation Rules prose. For other `chat_id: 0` internal subagents (e.g. `compact-catchup`, `agent-monitor`), the pattern works because the `subagent_result` relay calls `send_reply(chat_id=0, ...)`, which the Telegram API will reject gracefully (no real user at chat_id 0). So in practice the result is harmless either way — no message reaches a user, and `mark_processed` still fires. This is a pre-existing gap in the `subagent_result` handler, not introduced by this PR, and is low risk. Not a blocker.

---

## Fix 2: Logging addition in `nightly-consolidation.sh` — PASS

**`bash -n` check:** Confirmed passing per engineer's test notes.

**`log()` function:** Correctly tees to both stdout (for cron mail capture) and `$LOG_FILE` with ISO 8601 timestamps. The append flag (`-a`) on `tee` is correct — does not truncate the log on each run.

**`LOG_DIR` creation:** `mkdir -p "$INBOX" "$LOG_DIR"` is a clean one-liner that also covers the pre-existing inbox guard. Safe.

**`LOBSTER_WORKSPACE` env var:** Follows the same pattern as `LOBSTER_MESSAGES` already in the script (`${VAR:-default}`). Consistent with project conventions.

**Replacement of bare `echo` calls:** All three `echo` calls (dedup-skip, injection confirmation) are replaced with `log()`. No bare echoes remain.

**Log rotation:** No log rotation is added, but this is consistent with other Lobster log files (e.g. dispatcher logs). Acceptable.

---

## Summary

Both fixes are correct and safe. The consolidation pipeline is restored. The advisory about `chat_id: 0` relay handling is a pre-existing issue that doesn't affect correctness here. Ready to merge.

PR #1273, opened 2026-03-31
fix(sessions): ensure short-lived sessions always produce recoverable records

### LINKED ISSUE #1258: process: stub sessions leave no recovery data — need minimum session-note threshold
🤖🦞 Lobster (subagent):

## What failed (symptom)

Sessions 003, 006, and 007 on 2026-03-31 were stubs — their Summary/Open Threads/Notable Events sections contain only template placeholders with no content. 3 of 9 sessions today produced zero useful recovery data.

This happens because the session-note-polish subagent only fires when a session exceeds a threshold (~30 min or ~25 messages). Short sessions caused by crashes, restarts, or rapid compaction cycles never write notes.

## Why it's a problem (impact on reliability)

- compact-catchup has no fallback when sessions are stubs. It reads the file but gets template placeholders, not real state.
- After a rapid restart cycle (e.g., the 3-restart cycle of sessions 006/007/008), the dispatcher restarts completely blind about what happened in those sessions.
- Sahar had to manually notice and ask about instability. The system could not self-diagnose.
- Per the reality-check report: "3 out of 9 sessions today wrote no useful recovery data."

## Proposed fix

1. Write a **minimal session note at session start** (not end): record start time, session ID, and the first message received. This survives even if the session crashes.
2. Add a **session-end hook** that fires unconditionally at session close, writing at minimum: duration, message count, reason for close (if known), and a one-line status.
3. The note-polish subagent can still run for longer sessions and enrich the file — but it should not be the *only* mechanism.

## Related

- Sessions 003, 006, 007 on 20260331 all stubs
- Reality-check report: ~/lobster-workspace/reports/reality-check-20260331.md section 6

sayhar:
🤖🦞 Lobster (engineer):

Sessions 003, 006, and 007 on 2026-03-31 were template stubs — the session file was created at startup but the note-polish subagent only fires above a message/time threshold. A 30-second session that ends in a crash or rapid restart leaves exactly the template placeholders that compact-catchup cannot read.

The root cause is that the session file creation and the session content writes are fully decoupled: the file is born empty and stays empty unless work happens. This PR breaks that coupling by making minimal tombstone metadata part of file creation itself.

## What changed

**session.template.md** gains two new fields: `Messages processed` and `End reason`. These give tombstone writes named slots to fill in rather than appending free-form text.

**Startup step 2a** now specifies that immediately after copying the template, the dispatcher writes `Messages processed: 0` and `End reason: active`. From the moment the file is created it is partially recoverable — compact-catchup can see "this session started at T, processed 0 messages, reason: active (crashed before update)" rather than a wall of template placeholders.

**context_warning handler** gains a new step 2 that writes the tombstone inline before entering wind-down mode. Inline (not via subagent) because it is fast, takes no I/O beyond a single file write, and must survive even if wind-down is interrupted. The previous "On context_warning: spawn a session note update subagent as the very first step" in the Session File Management section was correct in intent but the context_warning numbered steps didn't include it — so it was easy to skip.

**Session File Management** gains two new subsections:
- *Tombstone on session end (unconditional)*: specifies minimum content (Ended, Messages processed, End reason, Summary) and makes the write unconditional — even a zero-message session must write a tombstone.
- *MESSAGE_COUNT tracking*: specifies that the dispatcher initializes `MESSAGE_COUNT = 0` at startup and increments it on each real user `mark_processed` call, so tombstones have an accurate count.

## Flow change

Before:
```
startup → create file (template stubs) → [maybe] note-polish runs if session long enough → [otherwise] stub survives
```

After:
```
startup → create file → immediately write start time + messages=0 + reason=active
                                    |
         context_warning → write tombstone inline (step 2) before wind-down
                                    |
         any session end → write tombstone (unconditional rule)
```

A session that crashes before doing anything now leaves: start time, messages=0, reason=active. A session that ends via context_warning leaves: start time, end time, message count, reason=context_warning, brief summary. Both are usable by compact-catchup.

## Areas to review closely

- The inline tombstone write in context_warning (step 2) is a file write on the main thread, which normally violates the 7-second rule. It is intentionally an exception — tombstone writes are tiny (< 50ms), must happen before wind-down begins, and cannot be delegated to a subagent that may not complete before the session exits. The comment "inline, no subagent needed — this is fast" captures this intent.
- The "unconditional tombstone" rule in Session File Management does not specify a mechanism for crash terminations — it cannot, because a crash does not give the dispatcher a chance to write anything. The value here is covering graceful exits and context_warning. Crash tombstones get the partial record from step 2a (start time, messages=0, reason=active) rather than nothing.
- The session note update subagent prompt template says "Do not modify Summary or Started/Ended" — that constraint is now in tension with the tombstone rule which does update Ended and End reason. The tombstone write is inline (not via that subagent), so the constraint is not violated — but a reviewer might notice the apparent conflict.

## Tests run
- [x] Python edit script with 6 assertion checks — all passed; verified each changed section appears correctly in the output file.
- [ ] Live dispatcher session — not run; this change is documentation/prompt only (no executable code changed). Behavior change takes effect on next dispatcher restart.

Closes #1258

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -27,7 +27,7 @@ When you first start (or after reading this file), follow these steps:
 1. Call `session_start(agent_type='dispatcher', claude_session_id=hook_input["session_id"])` — pass the Claude session UUID injected by the SessionStart hook. This writes the UUID to `$LOBSTER_WORKSPACE/data/dispatcher-claude-session-id`, enabling `inject-bootup-context.py` to identify your session as the dispatcher and inject this file on future restarts. Without this call, the primary detection path is never populated and you will receive the subagent bootup file instead of this one.
 1a. Read `~/lobster-user-config/memory/canonical/handoff.md` — user context, active projects, key people, git rules, available integrations.
 2. Read `~/lobster-workspace/user-model/_context.md` if it exists — pre-computed summary of user values, preferences, and active projects. Skip if absent.
-2a. Create a new session file inline (see Session File Management). Store its path as `current_session_file`.
+2a. Create a new session file inline (see Session File Management). Store its path as `current_session_file`. Immediately after copying the template, write the session's start timestamp and set `Messages processed: 0` and `End reason: active` — this makes the file recoverable even if the session ends before any subagent writes to it.
 2b. Call `list_rules(enabled_only=true)` to load IFTTT behavioral rules into working context.
 2c. Check `~/lobster-workspace/data/context-handoff.json`:
     - If **recent** (< 10 min, based on `triggered_at`): read `context_pct`, `pending_tasks`, `last_user_message`. Notify user: "Restarted — context was at {context_pct}%. Resuming from where we left off." Re-queue any stuck messages from `~/messages/processing/`. Delete the file.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS**

Reviewed against all five briefing points. No blockers found.

---

**Inline tombstone in context_warning (7-second rule exception)**

Justified. The write is a single file operation, must complete before wind-down begins, and cannot be safely delegated to a subagent that may not finish before the session exits. The inline comment ("inline, no subagent needed — this is fast") documents the intent at the call site. The Session File Management section reinforces it: "this is faster and more reliable than spawning a subagent, and ensures the record survives even if wind-down is interrupted." Exception is properly justified and documented in two places.

**"Do not modify Started/Ended" tension**

The subagent constraint ("Do not modify Summary or Started/Ended") is not violated because tombstone writes are inline, not via that subagent. The PR body calls this out explicitly in "Areas to review closely." The bootup doc itself does not add a clarifying note adjacent to the subagent constraint text — a future reader could still be puzzled momentarily — but this is a nit, not a blocker. The explanation is present in the PR.

**MESSAGE_COUNT tracking**

Clean. Initialization at startup (= 0) and increment rule (real user mark_processed only, excluding system messages like session_note_reminder) are both specified. Fits naturally in the dispatcher pseudocode without adding noise.

**subagent_type="general-purpose" references**

None present. The PR references `lobster-generalist` (the correct type) and all new text is documentation/pseudocode with no subagent spawn instructions that would use the blocked type.

**Crash-terminated sessions**

Explicitly scoped and documented. Crashes get the step-2a partial record (start time, messages=0, reason=active) — acknowledged as acceptable in both the PR body and the bootup doc rationale comment. The unconditional tombstone rule covers graceful exits; crash coverage is bounded to what is mechanically possible.

---

**Minor nit (non-blocking):** The "On context_warning" line in Session File Management cross-references "step 2 (see context_warning handler above)" — adequate, but the reference would be more robust if it named the section header rather than just "above." Low priority.

All changes are documentation/prompt-only (no executable code). The session template gains two well-formed fields. Step numbering in the context_warning handler and Session File Management creation steps is consistent. No issues found.

Issue #1259, opened 2026-03-31
process: unanswered questions not re-surfaced at session start — LOBSTER-SELF-CHECK dropped across 4+ sessions


sayhar:
🤖🦞 Lobster (subagent):

## What failed (symptom)

Sahar asked at 09:32 UTC on 2026-03-31 whether LOBSTER-SELF-CHECK is currently running on the server. This question was never answered across 4 subsequent sessions (005, 006, 007, 008, 009). Each session noted it as pending. None acted on it.

## Why it's a problem

- Open commitments to Sahar that survive in handoff.md and session notes are passively tracked but not actively re-surfaced.
- The dispatcher at session start reads context files but has no mechanism that says: "this question has been pending for N sessions — answer it NOW before anything else."
- The result is that urgent commitments get noted, scrolled past, and deferred indefinitely.
- This is a trust/reliability issue. Sahar explicitly asked; Lobster repeatedly said "still pending" without acting.

## Proposed fix

1. Add an **unanswered-questions tracker** (could be a simple section in handoff.md or a dedicated field): items with  get promoted to CRITICAL at next session start.
2. At session start, the compact-catchup or context-recovery step should highlight any commitment pending for 2+ sessions as a first-priority action — before processing new messages.
3. Consider an IFTTT rule: "if a commitment is noted as pending in session N and still pending at start of session N+2, immediately surface it to Sahar with a status update."

## Related

- LOBSTER-SELF-CHECK question: msg 20064, 09:32 UTC 2026-03-31
- Carried through sessions 005, 006, 007, 008, 009 without action
- reality-check-20260331.md section 5 (Open Commitments #1)

sayhar:
Taking this on. Plan:

- [x] Read sys.dispatcher.bootup.md to understand current commitment/question handling
- [x] Read .claude/agents/compact-catchup.md to understand catchup flow
- [x] Read handoff.md CRITICAL section to understand structure
- [ ] Add commitment durability rule to dispatcher bootup: when Alex asks a question that can't be answered immediately, immediately write it to handoff.md CRITICAL section via background subagent
- [ ] Update compact-catchup: before completing, check handoff.md CRITICAL against session notes — if unanswered question in session notes but NOT in handoff.md CRITICAL, add it
- [ ] Open PR

[...]

PR #1274, opened 2026-03-31
process: add commitment durability — deferred questions survive session boundaries

sayhar:
🤖🦞 Lobster (engineer):

Commitments made to the user (questions deferred, tasks promised) were only recorded in session notes. Session notes don't survive compaction reliably — they exist in long-term storage but are not re-loaded at session start unless catchup explicitly surfaces them. The result: a question asked at 9:32 AM was noted as pending across 5 consecutive sessions (005–009) without ever being answered.

## What this changes

**Two independent fixes that compose:**

**1. Dispatcher: immediate commitment capture (proactive)**

A new `## Commitment Durability` section in `sys.dispatcher.bootup.md` establishes: when the dispatcher defers a response to the user (with language like "I'll check on this", "checking now", or any question that cannot be answered inline), it must immediately spawn a background subagent to write that commitment to `rolling-summary.md` under `## Open Threads / Commitments`. `rolling-summary.md` is the designated cross-session truth file and is read at every session start; session notes are not.

The subagent uses an idempotency check (substring match) before writing, so re-spawning is safe.

This is distinct from the "no automated follow-up spawning" constraint in Multi-Question Handling, which governs question-tracking. This is commitment capture: the dispatcher already decided to defer — it just needs to record that deferral durably.

**2. compact-catchup: commitment carry-forward (safety net)**

A new Phase 4 in `compact-catchup.md` runs after the rolling summary update. It scans the two most recent session files (already read during Phase 1) for commitment patterns (`ANSWER the user:`, `needs answer`, `never answered`, `deferred — needs answer`) and verifies each one is present in `rolling-summary.md`. Any missing items are prepended to the `## Open Threads / Commitments` section.

This catches the failure mode where the dispatcher forgot to write the commitment immediately (the bug that existed before fix 1) and also handles any edge cases where fix 1 fires but the write fails.

**Why `rolling-summary.md`, not `handoff.md`:**

`rolling-summary.md` is the designated cross-session truth file — it already has an `## Open Threads / Commitments` section purpose-built for this, and it is explicitly read at session start. `handoff.md` exists but is being superseded; using `rolling-summary.md` is architecturally correct and consistent with where current session-bridge context lives.

**Flow before/after:**

```
BEFORE:
  User asks question → dispatcher notes it in session file → session ends
  → compaction fires → session note is archived, not re-surfaced
  → new session reads rolling-summary.md (no mention) → question dropped

AFTER (fix 1):
  User asks question → dispatcher defers → IMMEDIATELY spawns subagent
  → subagent writes to rolling-summary.md ## Open Threads / Commitments
  → session ends → new session reads rolling-summary.md → question is visible → gets answered

AFTER (fix 2, safety net):
  compact-catchup runs → scans recent session notes for commitment patterns
  → finds items not in rolling-summary.md → prepends them to ## Open Threads / Commitments
  → dispatcher reads rolling-summary.md at next startup → question is visible
```

## What is NOT changed

- The "No automated follow-up spawning" constraint in Multi-Question Handling is unchanged. That rule governs question-tracking behavior (don't spawn subagents just to note questions). Commitment capture is a different action triggered after the dispatcher has already decided to defer.
- Session note format and session-note-polish/session-note-appender agents are unchanged.
- No changes to `rolling-summary.md` format — the existing `## Open Threads / Commitments` and `## Active PRs & Decisions` sections are used as-is.
- `handoff.md` is not touched by either fix.

## Functional patterns used

Pure functions and declarative patterns: Phase 4 in compact-catchup is a pure scan-and-compare — read session notes, read `rolling-summary.md`, compute the diff, write the result. No side effects until the diff is confirmed non-empty. The idempotency check in the dispatcher commitment-capture subagent follows the same pattern.

## Tests run

- [ ] No automated tests exist for these behavioral instruction files — changes are verified by inspection
- [x] Reviewed that Phase 4 step numbers (17–20) don't conflict with existing steps 1–16
- [x] Verified "Sahar" name substituted with "the user" per pre-commit hook requirement
- [x] Pre-push PII scanner: clean

*(Updated: redesigned per Sahar's direction — storage moved from handoff.md to rolling-summary.md)*

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -135,7 +135,7 @@ Rolling summary: write failed (<reason>)
 
 ### Phase 4: Commitment carry-forward
 
-After Phase 3, verify that open commitments in the session notes are also captured in handoff.md. This is the safety net: if the dispatcher forgot to write a commitment immediately, catchup catches it here.
+After Phase 3, verify that open commitments in the session notes are also captured in `rolling-summary.md`. This is the safety net: if the dispatcher forgot to write a commitment immediately, catchup catches it here.
 
 17. Scan the tier-1 session files (the 2 most recent, already read in Phase 1) for lines matching any of these patterns:
     - `ANSWER the user:` (case-insensitive)
@@ -145,22 +145,22 @@ After Phase 3, verify that open commitments in the session notes are also captur
 
     Collect each such line as a **candidate commitment**.
 
-18. Read `~/lobster-user-config/memory/canonical/handoff.md`.
+18. Read `~/lobster-user-config/memory/canonical/rolling-summary.md`.
 
-19. For each candidate commitment, check whether handoff.md already contains it (substring match, case-insensitive). If the commitment is **not** present in handoff.md:
-    - Locate the `**Active urgent items:**` bullet list under `## CURRENT STATE`. If the section is absent, add it directly under the `---` separator after the CURRENT STATE header.
+19. For each candidate commitment, check whether `rolling-summary.md` already contains it (substring match, case-insensitive). If the commitment is **not** present:
+    - Locate the `## Open Threads / Commitments` section. If the section is absent, add it after `## Active PRs & Decisions`.
     - Prepend the missing commitment line verbatim (as found in the session note), prefixed with `- `.
     - Mark it with `(carried forward by compact-catchup)` if the original text doesn't already have that annotation.
 
-20. If any commitments were added: write handoff.md back. Include in the `write_result` footer:
+20. If any commitments were added: write `rolling-summary.md` back. Include in the `write_result` footer:
     ```
-    Commitment carry-forward: <N> item(s) added to handoff.md CRITICAL
+    Commitment carry-forward: <N> item(s) added to rolling-summary.md
     ```
     If none were missing: include:
     ```
     Commitment carry-forward: none needed
     ```
-    If handoff.md could not be read or written, note the failure but do not abort catchup.
+    If `rolling-summary.md` could not be read or written during Phase 4, note the failure but do not abort catchup.
 
 ## Session notes reading
 
@@ -211,7 +211,7 @@ Active agents: <N> (<comma-separated task_ids or "none">)
 ### Rolling summary
 Updated: <path> (<line_count> lines)
 ### Commitment carry-forward
-<N> item(s) added to handoff.md CRITICAL (or "none needed")
+<N> item(s) added to rolling-summary.md (or "none needed")
 ```
 
 Omit the "Session context" section entirely if no session files were found.
@@ -230,7 +230,7 @@ Keep each line to one sentence. The dispatcher is on mobile -- brevity matters.
 - If the session file cannot be found or written (permissions, path not found), note the failure in `write_result` and continue -- do not crash the entire catchup.
 - If `rolling-summary.md` cannot be read or written, note the failure in `write_result` and continue -- do not abort catchup.
 - Never remove content from rolling-summary.md unless there is clear evidence in the inbox scan that the item is resolved. When in doubt, carry it forward.
-- If handoff.md cannot be read or written during Phase 4, note the failure in `write_result` and continue -- do not abort catchup.
+- If `rolling-summary.md` cannot be read or written during Phase 4, note the failure in `write_result` and continue -- do not abort catchup (this is separate from the Phase 3 rolling summary update).
 
 ## Delivering results
 
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -133,6 +133,35 @@ Or, on failure:
 Rolling summary: write failed (<reason>)
 ```
 
+### Phase 4: Commitment carry-forward
+
+After Phase 3, verify that open commitments in the session notes are also captured in handoff.md. This is the safety net: if the dispatcher forgot to write a commitment immediately, catchup catches it here.
+
+17. Scan the tier-1 session files (the 2 most recent, already read in Phase 1) for lines matching any of these patterns:
+    - `ANSWER the user:` (case-insensitive)
+    - `CRITICAL open commitment`
+    - `still pending` / `never answered` / `needs answer`
+    - `deferred — needs answer`
+
+    Collect each such line as a **candidate commitment**.
+
+18. Read `~/lobster-user-config/memory/canonical/handoff.md`.
+
+19. For each candidate commitment, check whether handoff.md already contains it (substring match, case-insensitive). If the commitment is **not** present in handoff.md:
+    - Locate the `**Active urgent items:**` bullet list under `## CURRENT STATE`. If the section is absent, add it directly under the `---` separator after the CURRENT STATE header.
+    - Prepend the missing commitment line verbatim (as found in the session note), prefixed with `- `.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: PASS with minor notes**

---

## Commitment Durability (sys.dispatcher.bootup.md)

**Trigger phrase coverage — good, one gap**

The trigger list covers the most common deferral language well: explicit "I'll check," "I'll get back to you," and spawned subagents that may not finish. The phrase "Checking now" is an interesting edge case — correct to include it since compaction can interrupt an in-flight subagent.

One gap: "Let me look at that" and variants like "Give me a second" are common deferral phrases that don't appear in the list. These are close enough to "I'll check on that" that a dispatcher following the spirit of the rule would catch them, but they're worth noting as likely real-world misses. Not blocking, but worth a follow-up.

The final catch-all — "Any explicit question from the user that you cannot answer inline AND you do not answer within the same session turn" — is strong and closes most gaps. This is the right backstop.

**"Commitment capture" vs "question tracking" distinction — valid**

The distinction is real and well-drawn. The old question-tracking section (the one with the 7-second rule and the "Note:" line pattern) is about tracking for in-session re-surfacing. Commitment Durability is about cross-session persistence. These are orthogonal concerns and should not be merged. The PR handles this correctly by keeping them in separate sections.

**Subagent spawn pattern — one concern**

The `chat_id: 0` in the background subagent prompt is a sentinel for "system-initiated, no user reply needed." This works for write_result, but if the subagent encounters an error it will write_result with `status='error'` to chat_id=0 — which the dispatcher will try to route back to nobody. This is probably fine in practice (the dispatcher would drop it), but it's a silent failure mode. A comment in the template noting "failures are silently dropped — Phase 4 of compact-catchup is the recovery path" would make the intent explicit and reduce future confusion.

**Idempotency check — correct**

Substring match before writing is the right call. Exact-match would produce duplicates on re-phrased recaps.

---

## Phase 4: Commitment Carry-Forward (compact-catchup.md)

**Step numbering — no conflicts**

Existing steps are 1–16 across Phases 1–3. New Phase 4 is 17–20. Clean.

**Pattern matching — mostly good, one likely miss**

The four patterns cover explicit annotations written by the dispatcher or a prior subagent:
- `ANSWER the user:` — matches the format the Commitment Durability rule writes
- `CRITICAL open commitment` — matches escalated items
- `still pending` / `never answered` / `needs answer` — matches passive carry-forward language
- `deferred — needs answer` — matches the exact format from the subagent template

What it would *not* catch: a session note written in natural language before this PR, e.g. "Sahar asked about X; I haven't been able to check yet." That's the legacy form. Since issue #1259 is about existing dropped commitments, Phase 4 would not retroactively recover those unless someone annotated them with one of the trigger phrases. This is a known limitation rather than a bug — new commitments after this PR will use the standardized format — but a note in the instructions acknowledging this boundary would help future reviewers.

**Step 18 placement — minor ordering question**

Step 17 scans session files (already in memory from Phase 1). Step 18 reads handoff.md. Step 19 does the diff. This is sensible, but there's a micro-optimization available: if step 17 produces zero candidates, step 18 can be skipped. The instructions don't mention this. Low impact since handoff.md is a small file, but worth noting for completeness.

**Verbatim carry-forward — correct**

Prepending the line verbatim (step 19) is the right call. Rewriting it risks losing context. The `(carried forward by compact-catchup)` annotation is a good audit trail.

**Error handling — consistent with existing pattern**

"Note failure but do not abort" matches the existing resilience pattern for rolling-summary.md. Correct.

---

## "the user" vs "Sahar" — minimal clarity impact

The substitution is fine. The dispatcher knows who "the user" is. In the subagent template (`ANSWER the user: <exact question text>`), the phrase reads naturally and is unambiguous. No clarity loss in practice. The pre-commit check that forced this is doing its job.

---

## Summary

The two-layer approach (immediate write on deferral + catchup safety net) is architecturally sound. The root cause in #1259 — no mechanism to escalate from session note to handoff.md CRITICAL — is directly addressed. The patterns chosen for Phase 4 will catch all commitments written after this PR. Legacy commitments using natural language are not retroactively recovered (known limitation). One silent failure mode in the subagent template deserves a comment but is not blocking.

**PASS** — safe to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS**

Note: This review is posted by a separate subagent from the one that implemented the changes.

---

## Summary

The diff implements exactly what the updated PR body claims. Both files modified (`sys.dispatcher.bootup.md` and `compact-catchup.md`) correctly target `rolling-summary.md`, not `handoff.md`. No drift detected between description and implementation.

---

## Findings

### 1. rolling-summary.md section names — CORRECT

The PR writes to `## Open Threads / Commitments` (both in the dispatcher commitment-capture subagent prompt and in compact-catchup Phase 4 step 19). I verified the live `rolling-summary.md` at `~/lobster-user-config/memory/canonical/rolling-summary.md` — that section exists at line 42, and `## Active PRs & Decisions` exists at line 4. The fallback instruction ("if the section is absent, add it after `## Active PRs & Decisions`") is therefore consistent with actual file structure.

### 2. Phase 4 step numbers — CORRECT

Existing phases cover steps 1–16. Phase 4 introduces steps 17–20. No numbering conflicts.

### 3. Does the diff match the updated PR body? — YES

- Fix 1 (dispatcher): `sys.dispatcher.bootup.md` adds `## Commitment Durability` with subagent prompt targeting `rolling-summary.md`. ✓
- Fix 2 (compact-catchup): `compact-catchup.md` adds Phase 4 scanning for commitment patterns and writing missing ones to `rolling-summary.md`. ✓
- `handoff.md` is not touched anywhere in the diff. ✓
- The `write_result` footer format (lines 210–215) is updated to include `### Commitment carry-forward` output. ✓

### 4. Scan patterns — REASONABLE with minor caveat

Phase 4 scans for: `ANSWER the user:`, `CRITICAL open commitment`, `still pending`, `never answered`, `needs answer`, `deferred — needs answer`.

The pattern `still pending` is broad and could match incidental log lines (e.g., "subagent still pending"). This is a low-risk false-positive: the worst outcome is an extra carry-forward line in rolling-summary.md (the idempotency check on re-runs mitigates repeat additions). Not a blocker.

### 5. Idempotency — CORRECT

Both the dispatcher subagent and Phase 4 use substring-match before writing. Safe for repeated runs.

### 6. Error handling — CORRECT

Phase 4 failure is non-fatal: "note the failure in `write_result` and continue — do not abort catchup." The redundant clarifying bullet at the bottom of the error-handling section is minor verbosity, not a bug.

### 7. CI — ALL CHECKS PASS

CodeQL (python + JS/TS) passed. No functional tests exist for behavioral instruction files (acknowledged in PR).

---

## Minor non-blocking observations

- The dispatcher subagent template uses placeholder `<slug>` in two places (`task_id` and `write_result` call). The implementing engineer (dispatcher) will need to substitute a real slug at spawn time. This is expected usage but could benefit from a comment clarifying it's a template placeholder.
- The `chat_id: 0` in the subagent template is correct for system-origin tasks that don't reply to a user.

---

## Verdict

Implementation matches the stated design. Section names are verified against the live file. Step numbering is clean. Error handling is non-fatal and consistent with the existing compact-catchup error policy. The `still pending` pattern is a minor false-positive risk but not a blocker.

**Ready to merge.**

PR #1275, opened 2026-03-31
fix: convert UTC timestamps to ET in debug Back Online message (#1247)

### LINKED ISSUE #1247: UTC timestamps sent to user in debug Back Online message (ET conversion missing)
## Bug

The dispatcher sent raw UTC timestamps to the user in the debug \"Back online\" message, violating the standing rule that all user-facing times must be in Eastern Time (ET).

**Message sent:**
> 🔄 Back online. Context recovered from 09:29 to 09:40 UTC. 3 messages processed, 7 subagents were running.

*(Updated: scope broadened from compact-catchup-only to systemic — all subagents that produce user-visible timestamps share this obligation.)*

## Root Cause

The failure occurs at **two layers**, neither of which performs a UTC→ET conversion:

### Layer 1 — compact-catchup agent (`.claude/agents/compact-catchup.md`)

The output format spec (line 155) defines the header as:
```
## Catch-up: <window_start> -> now
```

`window_start` is derived directly from `compaction-state.json` fields (`last_catchup_ts`, `last_compaction_ts`, `last_restart_ts`), which are raw UTC ISO 8601 timestamps (e.g. `2026-03-31T09:29:44Z`). The agent has no instruction to convert these to ET before writing them into `write_result`.

### Layer 2 — dispatcher handling (`.claude/sys.dispatcher.bootup.md`, line 47 and line 196)

The dispatcher instruction says:

> `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."` (Fill in N and M from `msg["text"]`.)

It instructs the dispatcher to pull `[window_start]` directly from `msg["text"]` (which is the raw compact-catchup output header) without any conversion to ET. There is no instruction to convert timestamps before embedding them in the user-facing string.

### Layer 3 — all other subagents (systemic gap)

The ET conversion requirement is stated in `CLAUDE.md` as a general rule, but `sys.subagent.bootup.md` does not explicitly remind subagents that **any timestamp they include in a `send_reply` call must be converted to ET first**. This creates a systemic gap: subagents that produce output containing timestamps (calendar events, log summaries, scheduled job results, event timelines, etc.) may silently relay UTC times to users without realizing the conversion obligation applies to them.

Known subagents at risk include anything that:
- Formats calendar or scheduling data
- Summarizes log files or event streams
- Reports task or job completion times
- Constructs any user-visible sentence containing a time

## Why All Three Layers Share Responsibility

- The compact-catchup agent *could* output ET timestamps in its header (fixing it at the source), but it's an internal-context subagent and its output is also read by the dispatcher for situational awareness — using ISO UTC there is arguably cleaner for internal use.
- The dispatcher is the **user-facing layer** and is already responsible for the ET rule for all other messages. It is the natural place to enforce conversion for compact-catchup output specifically.
- All subagents that call `send_reply` directly are **also** user-facing layers for their own outputs and must independently enforce ET conversion — there is no downstream safety net once `send_reply` is called.

### Fix 1 — dispatcher layer (primary, compact-catchup-specific)

Update the debug message template in `sys.dispatcher.bootup.md` (both the startup-catchup block at ~line 47 and the compact-reminder block at ~line 196) to explicitly convert the window timestamps to ET before embedding them in the user-facing string. Example instruction addition:

> Convert `[window_start]` and `[now]` from UTC ISO timestamps to ET (e.g. "5:29 AM ET") before composing the message. Use the ET offset rule: EDT (UTC-4) mid-March through early November, EST (UTC-5) otherwise.

### Fix 2 — compact-catchup agent (defensive, secondary)

Add a note to the output format section reminding the agent that the `## Catch-up:` header timestamp format is for internal use only, and that if any timestamps are relayed to users they must be converted to ET first.

### Fix 3 — sys.subagent.bootup.md (systemic, closes the gap for all subagents)

Add an explicit reminder in `sys.subagent.bootup.md` that all timestamps included in any `send_reply` call must be converted to ET. This should appear near the delivery pattern documentation so it is visible when subagents are composing their reply text. A short utility note covering the conversion rule (EDT UTC-4, EST UTC-5, dates, common format) would reduce friction and errors.

Example addition to `sys.subagent.bootup.md` (near the send_reply usage section):

> **ET conversion required for all user-visible timestamps:** Before including any timestamp in a `send_reply` call, convert it from UTC to Eastern Time. Rule: EDT (UTC-4) from mid-March through early November; EST (UTC-5) otherwise. Format as "5:29 AM ET" or "2:30 PM ET". Never send raw UTC ISO strings or "UTC" suffixes to users.

## Files to Edit

- `~/lobster/.claude/sys.dispatcher.bootup.md` — startup-catchup result handler (~line 47) and compact-reminder result handler (~line 196)
- `~/lobster/.claude/agents/compact-catchup.md` — output format section (~line 155) — defensive note only
- `~/lobster/.claude/sys.subagent.bootup.md` — add ET conversion reminder near `send_reply` delivery pattern docs

## Affected Scenarios

- `LOBSTER_DEBUG=true` with startup catchup result arriving (`task_id: "startup-catchup"`)
- `LOBSTER_DEBUG=true` with compact-reminder result arriving (`task_id: "compact-catchup"`)
- Any subagent that builds a `send_reply` message containing a timestamp (calendar, logs, job results, summaries, etc.)

sayhar:
🤖🦞 Lobster (engineer):

## What this fixes

The debug "Back online" message sent by the dispatcher contained raw UTC ISO timestamps (e.g. "09:29 UTC"), violating the standing rule that all user-facing times must be displayed in Eastern Time. The root cause was a three-layer gap: neither the dispatcher result handlers nor the compact-catchup agent had any instruction to convert timestamps before composing user-visible text, and `sys.subagent.bootup.md` had no reminder covering this obligation for subagents generally.

## What changed

**`sys.dispatcher.bootup.md`** — both the startup-catchup result handler (line 47) and the compact-catchup result handler (~line 198) now include an explicit instruction: convert `[window_start]` and `[now]` from UTC ISO to ET (EDT UTC-4 mid-March–early-November, EST UTC-5 otherwise) before composing the `send_reply` text. The instruction appears inline with the template string so it is impossible to miss.

**`.claude/agents/compact-catchup.md`** — the output format section now carries a defensive note clarifying that the `## Catch-up:` header uses UTC ISO for internal/dispatcher use, and that any timestamp relayed to a user via `send_reply` must be converted to ET first. This is belt-and-suspenders: the dispatcher is the primary enforcement point, but if a future code path ever has compact-catchup relay directly to a user, the agent is also reminded.

**`sys.subagent.bootup.md`** — a new "ET conversion — required for all user-visible timestamps" paragraph is inserted before the "Large results" section, immediately after the user-facing delivery pattern examples. This closes the systemic gap: any subagent composing a `send_reply` containing a time (calendar events, log summaries, job results, etc.) now has an explicit reminder in the same file they read at startup.

## What is not changed

The UTC ISO format in the `## Catch-up:` header itself is intentionally preserved — it is internal-only, used by the dispatcher for situational awareness, and ISO format is cleaner for structured parsing. Only the user-facing `send_reply` path is affected.

No code changes. No behavior changes outside `LOBSTER_DEBUG=true`. No migrations needed.

## Functional patterns

Pure instruction additions — no side effects, no state mutations. Changes are declarative behavioral rules added to the relevant context files.

## Tests run

- [x] Verified all three edits in worktree before commit — strings match expected locations and content is correct
- [ ] Live end-to-end debug message test — blocked: requires `LOBSTER_DEBUG=true` and a dispatcher restart in production. No automated test suite covers bootup file content.

**Blocked items needing attention before merge:** none — the change is documentation/instruction only. The risk of merging without a live test is low; the only failure mode is the instruction being present but ignored, which is the same as the current state.

Closes #1247

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -151,6 +151,8 @@ Synthesise the tier-1 and tier-2 reads into the "Session context" section of the
 
 Structure your `write_result` text as follows:
 
+> **Note on timestamps:** The `## Catch-up:` header (below) uses UTC ISO format for internal/dispatcher use. If any timestamp from this output is ever relayed to the user in a `send_reply`, convert it to ET first (EDT UTC-4 mid-March through early November, EST UTC-5 otherwise). Format: "5:29 AM ET". Never send raw UTC ISO strings to users.
+
 ```
 ## Catch-up: <window_start> -> now
 
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -44,7 +44,7 @@ When you first start (or after reading this file), follow these steps:
 - New tasks: ack normally and spawn subagent. These are unambiguously new work.
 - Urgent messages: handle them. You have handoff.md for context.
 
-**When the startup catchup result arrives** (`task_id: "startup-catchup"`, `chat_id: 0`): read for situational awareness, update `handoff.md` if anything notable changed (failed subagents, open threads). Run `~/lobster/scripts/record-catchup-state.sh finish`. Do NOT relay to user — except if `LOBSTER_DEBUG=true`, send a brief status to ADMIN_CHAT_ID: `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."` (Fill in N and M from `msg["text"]`.) Then `mark_processed`.
+**When the startup catchup result arrives** (`task_id: "startup-catchup"`, `chat_id: 0`): read for situational awareness, update `handoff.md` if anything notable changed (failed subagents, open threads). Run `~/lobster/scripts/record-catchup-state.sh finish`. Do NOT relay to user — except if `LOBSTER_DEBUG=true`, send a brief status to ADMIN_CHAT_ID: `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."` (Fill in N and M from `msg["text"]`.) **Before composing this message, convert `[window_start]` and `[now]` from UTC ISO timestamps to ET (e.g. "5:29 AM ET"). Rule: EDT (UTC-4) mid-March through early November, EST (UTC-5) otherwise. Never send raw UTC ISO strings to the user.** Then `mark_processed`.
 
 ---
 
@@ -195,6 +195,7 @@ After a context compaction you lose situational awareness of the last ~30 minute
   - If `LOBSTER_DEBUG=true`: send a brief status to ADMIN_CHAT_ID:
     `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."`
     (Fill in N and M from `msg["text"]`. ADMIN_CHAT_ID from `lobster.conf` or the compact-reminder context.)
+    **Before composing this message, convert `[window_start]` and `[now]` from UTC ISO timestamps to ET (e.g. "5:29 AM ET"). Rule: EDT (UTC-4) mid-March through early November, EST (UTC-5) otherwise. Never send raw UTC ISO strings to the user.**
 - Run `~/lobster/scripts/record-catchup-state.sh finish`
 - `mark_processed`
 
--- a/.claude/sys.subagent.bootup.md
+++ b/.claude/sys.subagent.bootup.md
@@ -147,6 +147,10 @@ mcp__lobster-inbox__write_result(
 )
 ```
 
+**ET conversion — required for all user-visible timestamps:**
+
+Before including any timestamp in a `send_reply` call, convert it from UTC to Eastern Time. Rule: EDT (UTC-4) from mid-March through early November; EST (UTC-5) otherwise. Format as "5:29 AM ET" or "2:30 PM ET". Never send raw UTC ISO strings or "UTC" suffixes to users. This applies to all subagents that produce output containing times — calendar events, log summaries, job results, event timelines, and any other user-facing sentence with a time.
+
 **Large results — use artifacts, not inline text:**
 
 If your output exceeds ~4KB or ~500 words, write the full content to a file and pass the path in `artifacts`. This applies to **all tasks** (user-facing and internal):


sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS**

Reviewed commit `3c11062`. All three focus areas check out.

**Both handlers confirmed updated:**
- Startup-catchup handler (line 47 of `sys.dispatcher.bootup.md`): instruction is inline with the template string — "Before composing this message, convert `[window_start]` and `[now]` from UTC ISO timestamps to ET (e.g. '5:29 AM ET'). Rule: EDT (UTC-4) mid-March through early November, EST (UTC-5) otherwise."
- Compact-catchup handler (line 198): identical instruction appended immediately after the template string, before `record-catchup-state.sh finish`.

**Wording is clear and actionable:**
The EDT/EST split (with the calendar rule and offset values) is present in all three locations. The example format "5:29 AM ET" gives the dispatcher a concrete target. No ambiguity about which direction to convert.

**Placement in `sys.subagent.bootup.md`:**
The ET conversion paragraph lands at line 150–152, immediately after the `send_reply` delivery pattern code block (line 147) and before the "Large results" section. Any subagent that reads the delivery pattern will read the ET conversion rule directly below it. Placement is correct.

**Edge cases assessed:**
- The `## Catch-up:` header in `compact-catchup.md` deliberately stays in UTC ISO (internal/dispatcher use) — the defensive note at line 154 clarifies this and covers any future direct-relay path.
- The instruction to convert is placed at the composition point, not upstream, so it fires regardless of whether `[window_start]` and `[now]` arrive from startup-catchup or compact-catchup.
- The CLAUDE.md base rule ("Always display times in Eastern Time") provides a third layer of coverage for any path not explicitly named above.

**One minor style note (non-blocking):**
The compact-catchup handler instruction appears as a freestanding bold sentence after the template code block. The startup-catchup handler bolds the same instruction inline within the same sentence as the template. The inline placement (startup-catchup style) is marginally harder to miss, but both are clear enough. No change required.

No gaps found. Safe to merge.

PR #1277, opened 2026-03-31
feat: durable in-flight work tracking — write-ahead log for dispatched subagents (#1260)

### LINKED ISSUE #1260: process: no durable in-flight work tracking — active subagent work lost on restart
🤖🦞 Lobster (subagent):

## What failed (symptom)

There is no durable record of in-flight work (PRs being reviewed, subagents running, commits being staged) that survives a dispatcher restart. When the dispatcher restarts, it reconstructs context from handoff.md and rolling-summary.md, but these are periodically updated snapshots — not live work-in-progress records.

Issue #1257 (agent_sessions as in-flight queue) is in design only. As of 2026-03-31, no respawn protocol exists.

## Why it's a problem

- A PR review subagent dispatched at 11:45 AM would be invisible to a dispatcher that restarts at 11:50 AM.
- Work that was mid-flight simply disappears. The dispatcher starts fresh with no idea a subagent was running.
- The 3-restart cycle today (sessions 006/007/008) is an example: each restarted session had to re-discover its own state from scratch.
- Sahar asked at 11:13 AM "What did we miss? What next? Are you still librarianing?" — the dispatcher could not answer reliably.

## Proposed fix

This is essentially the scope of issue #1257, but with a tighter framing:

1. **Write-ahead log**: Before dispatching any subagent, write an entry to a durable file (e.g., inflight-work.json or the agent_sessions table). Include: task type, started_at, expected completion, key context.
2. **On session start**: Read inflight-work.json as part of compact-catchup. Any entry older than N minutes with no completion record = likely lost. Surface to dispatcher for triage.
3. **On subagent completion**: Remove the inflight entry (or mark completed).

This is simpler than the full agent_sessions design and can be implemented independently.

## Related

- Issue #1257 (agent_sessions as in-flight queue — design phase)
- reality-check-20260331.md section 6 (in-flight work tracking gaps)
- Sessions 006, 007, 008 restart cycle on 2026-03-31

sayhar:
## Summary

Subagents that completed without the dispatcher relaying their result (sent_reply_to_user=True, scheduled-job no-ops, engineer→reviewer routing) were never written to inflight-work.jsonl as done, causing them to appear as permanently in-flight on every compact-catchup restart scan.

The fix moves the done-entry write to immediately after mark_processing — before any branching logic — so it fires unconditionally for every subagent_result message that carries a task_id. The previous placement (after relay logic, only in the relay branch) was both incomplete and semantically misleading.

**Semantics clarification:** "done" in inflight-work.jsonl means the result arrived at the dispatcher and was claimed for processing — not that the user has received the relay. A dispatcher crash between receive and relay is a relay issue, not a lost-subagent issue. The compact-catchup threshold is raised to 30 minutes (from 15) to reduce false positives from long-running legitimate tasks.

## What changed

- `sys.dispatcher.bootup.md`: done-entry write moved to step 1 (after mark_processing), fires for ALL exit paths including sent_reply_to_user early-exit, silent-drop, and engineer→reviewer routing. Old placement after relay logic removed.
- `sys.dispatcher.bootup.md` (In-Flight Work Tracking section): updated description to document the new semantics — "done" = result arrived at dispatcher.
- `compact-catchup.md`: threshold raised from 15 to 30 minutes with explanatory comment.

## Functional patterns

Pure append-only log write; no branching on the write itself. The write is unconditional on task_id presence, factored out of all relay branches.

## Tests run

- [x] `git diff` reviewed manually — all three exit paths (sent_reply_to_user, silent-drop, reviewer routing) now pass through the done-entry write.
- [ ] Live dispatcher test — requires running Lobster instance; safe to merge as this is documentation/pseudocode only, no runtime code changed.

Closes #1260


sayhar (commented):
🤖🦞 Lobster (reviewer):

## Verdict: NEEDS-WORK

The design is sound and the write-ahead-log approach is correct. Three issues need addressing before merge.

---

### Issue 1 (blocking): silent-drop paths do not write the "done" entry

The `done` write sits inside the `else:` branch of the `subagent_result` handler — it only fires for the relay path. Two paths exit the handler earlier without writing `done`:

**Path A — `sent_reply_to_user == True` early exit (line ~282):**
```
if msg.get("sent_reply_to_user") == True:
    mark_processed(message_id)
    # ← NO done entry written
```
This is the path taken by every subagent that calls `send_reply` then `write_result(sent_reply_to_user=True)` — which is the recommended pattern for this subagent itself, and for all reviewer subagents. These tasks will sit permanently in the "possibly lost" list after a restart.

**Path B — scheduled-job noop silent drop:**
```
if is_scheduled_job and any(p in text_lower for p in NOOP_PHRASES) ...:
    mark_processed(message_id)
    continue  # ← NO done entry written
```

**Path C — engineer→reviewer routing (`continue` after spawning reviewer):**
```
    mark_processed(message_id)
continue  # ← NO done entry written
```

The `done` write needs to be factored out so it runs unconditionally for any `subagent_result` message that has a meaningful `task_id`, regardless of which exit path fires. The natural place is immediately before every `mark_processed` call, or hoisted to a finally-equivalent position at the top of the else block before the `continue` statements. Alternatively, a single normalised write right after `mark_processing` at the top (step 1) would also work and is simpler.

Path A is the most important: reviewer and this very subagent use `sent_reply_to_user=True`, so every compact-catchup run after a restart will falsely flag the most recent review/catchup as lost.

---

### Issue 2 (blocking): "done" write is in the wrong position relative to the PR description

The PR description says: "after relay logic and before `mark_processed`." The actual diff places the Bash write *after* all relay branches (including the ones that spawn new subagents via `Task()`). For the artifact-relay and large-text-relay sub-paths, the `done` entry fires after spawning a new background subagent — not after delivering the result to the user. This is arguably correct (the original task is done from the dispatcher's perspective), but it contradicts the PR description and could be confusing. The description should be updated to clarify: "done means the original task_id's result has been processed by the dispatcher, not that the user has received it."

---

### Issue 3 (minor): 15-minute threshold may be too short for some subagents

Long-running tasks (large repo analysis, voice transcription + processing, multi-step engineering tasks) routinely run 10–20 minutes. A 15-minute threshold will produce false positives in normal operation — compact-catchup after a brief restart will flag tasks that are still running legitimately. Suggest 30 minutes as the default, with a note that the threshold is intentionally conservative to trade false positives for earlier detection of genuinely lost work.

---

### Non-issues

- **File path** (`~/lobster-workspace/data/inflight-work.jsonl`) is correct — this directory persists across MCP restarts (it's runtime data, not session state). Good choice.
- **Write-before-spawn instruction** is unambiguous in the dispatch pseudocode. Steps 3 and 4 make the ordering clear, and the "synchronous write on the main thread" language in the tracking section removes any ambiguity.
- **Append-only log** is the right design; no race conditions in single-writer (main thread only) append semantics.
- **system_context / chat_id==0 subagents** being tracked is reasonable — compact-catchup will surface them, which is fine.

---

### Suggested fix for Issue 1

Move the `done` write to fire for all result types. One approach — write it once, unconditionally, immediately after `mark_processing` at step 1, using a helper pattern:

```python
# Step 1 (existing):
mark_processing(message_id)
# New: write done entry immediately — this is idempotent and safe to write early
# because the log is checked only at catchup time, not in real time.
task_id = msg.get("task_id", "unknown")
if task_id != "unknown":
    completed_at = datetime.utcnow().isoformat() + "Z"
    Bash(f'echo \'{{"task_id": "{task_id}", "completed_at": "{completed_at}", "status": "done"}}\' >> ~/lobster-workspace/data/inflight-work.jsonl')
```

Writing `done` at receive time (rather than after relay) is semantically correct: the task is done when its result arrives, regardless of whether the dispatcher has relayed it yet. A dispatcher crash between receive and relay is a relay problem, not a lost-subagent problem.

sayhar (commented):
🤖🦞 Lobster (engineer):

All three issues addressed:

**Issue 1 (blocking):** Done-entry write moved to immediately after `mark_processing` at step 1, before any branching. Now fires unconditionally for all exit paths: sent_reply_to_user early-exit, scheduled-job silent-drop, engineer→reviewer routing, and relay. The old placement (after relay logic, relay branch only) is removed.

**Issue 2 (blocking):** PR description updated to clarify semantics: "done" means the result arrived at the dispatcher, not that the user has received the relay. A crash between receive and relay is a relay issue, not a lost-subagent issue.

**Issue 3 (minor):** Threshold raised from 15 to 30 minutes in compact-catchup.md with explanatory comment: "30 min is intentionally conservative — trades some false positives for earlier detection of genuinely lost work."

sayhar (commented):
🤖🦞 Lobster (reviewer):

## Verdict: PASS

All three issues from the first review are confirmed fixed in the diff.

**Issue 1 (was blocking):** Done-entry write is now at step 1, immediately after `mark_processing`, before any branching. The inline comment explicitly states "fires for ALL subagent results regardless of relay path -- sent_reply_to_user, silent-drop, engineer→reviewer routing, and relay." The old placement (relay-branch-only, after relay logic) is removed. This is exactly the fix requested.

**Issue 2 (was blocking):** The semantics are now clarified in two places: in the In-Flight Work Tracking section of sys.dispatcher.bootup.md ("done means the result arrived at the dispatcher -- not that the user has received the relay") and in the inline pseudocode comment at step 1 of the subagent_result handler (same language). The PR description framing is consistent with the diff.

**Issue 3 (was minor):** Threshold is 30 minutes in compact-catchup.md, with the explanatory note: "30 min is intentionally conservative -- trades some false positives for earlier detection of genuinely lost work."

No new issues introduced. The append-only semantics, single-writer ordering, and file path are all unchanged and correct. Ready to merge.

PR #1278, opened 2026-03-31
fix: add nightly log cleanup for scheduled-jobs/logs, processed/, audio/ (#1240)

### LINKED ISSUE #1240: Missing log cleanup: scheduled-jobs/logs accumulates ~40k files, no rotation for processed/ messages or audio/
## Problem

Three log/data directories grow without bound — no cleanup or rotation exists for any of them.

### 1. `~/lobster-workspace/scheduled-jobs/logs/` — ~40,000 files, 160 MB

Each scheduled job run creates a new timestamped log file. There is no pruning mechanism. The `bot-talk-poller` and `unified-channel-poller` jobs alone account for ~950 files each. This will eventually cause issues with file descriptor limits, slow `ls`, and slow directory scans.

### 2. `~/messages/processed/` — 17,195 files, 73 MB

Processed inbox messages are moved here but never archived or deleted.

### 3. `~/messages/audio/` — 173 files, 71 MB

Voice message audio files are never cleaned up after transcription. Files go back to March 10.

## No logrotate config exists

`/etc/logrotate.d/` has no lobster entries.

## Suggested fix

Add cleanup logic to `nightly-consolidation.sh` or a dedicated cron script:
- Prune `scheduled-jobs/logs/` files older than 7–14 days
- Archive or delete `messages/processed/` files older than 30 days
- Delete `messages/audio/` files older than 7 days (post-transcription)

Add a migration to `upgrade.sh` to set up the logrotate or cron entry.

## Related

- `fastembed` has a version incompatibility with `pillow>=12` (installed: 12.1.1, required: <12). Not currently breaking but worth tracking.

sayhar:
🤖🦞 Lobster (engineer):

Three directories were growing without bound — `scheduled-jobs/logs/` at ~40k files, `messages/processed/` at 17k+ files, and `messages/audio/` with files back to March. No cleanup mechanism existed for any of them.

This adds a dedicated `scripts/log-cleanup.sh` that prunes each directory to a retention threshold: 7 days for scheduled-job logs, 30 days for processed messages, 7 days for audio files. The script is called at the top of `nightly-consolidation.sh` so cleanup runs at 3am daily via the existing cron entry — no new cron entry needed. If `log-cleanup.sh` is missing (e.g. partial rollout), the guard in `nightly-consolidation.sh` skips it gracefully rather than failing the consolidation run.

Migration 62 in `upgrade.sh` ensures the script is executable on existing installs and runs `--dry-run` to verify it is functional.

## What changed

- `scripts/log-cleanup.sh` (new): pure-shell cleanup script with `--dry-run` flag and per-directory deleted-file-count output. Uses `find -mtime +N -delete` for each directory.
- `scripts/nightly-consolidation.sh`: calls `log-cleanup.sh` at the top before injecting the consolidation message; skips gracefully if the script is absent or non-executable.
- `scripts/upgrade.sh`: Migration 62 — `chmod +x` the script and run `--dry-run` as a smoke test.

No other files changed. The existing nightly-consolidation cron entry and the rest of the consolidation flow are untouched.

## Functional patterns

Pure functions (each directory cleanup is an independent, side-effect-isolated unit), declarative `find` pipelines, and explicit dry-run/live mode separation at the boundary.

## Tests run

- [x] `bash scripts/log-cleanup.sh --dry-run` — ran against live workspace; correctly identified 72 stale log files in `scheduled-jobs/logs/`, 0 in `messages/processed/`, 0 in `messages/audio/`. Output and exit code clean.
- [ ] Live deletion pass — skipped: no stale files in processed/audio at time of test; log deletion deferred to not affect running system during PR review.

Closes #1240

### DIFF
--- a/scripts/log-cleanup.sh
+++ b/scripts/log-cleanup.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+# log-cleanup.sh — Prune stale log/data files to prevent unbounded directory growth.
+#
+# Directories cleaned:
+#   ~/lobster-workspace/scheduled-jobs/logs/  — files older than 7 days
+#   ~/messages/processed/                      — files older than 30 days
+#   ~/messages/audio/                          — files older than 7 days
+#
+# Usage:
+#   log-cleanup.sh           — delete stale files and print summary
+#   log-cleanup.sh --dry-run — print what would be deleted, make no changes
+
+set -euo pipefail
+
+DRY_RUN=false
+if [[ "${1:-}" == "--dry-run" ]]; then
+    DRY_RUN=true
+fi
+
+WORKSPACE="${LOBSTER_WORKSPACE:-$HOME/lobster-workspace}"
+MESSAGES_DIR="${LOBSTER_MESSAGES:-$HOME/messages}"
+
+cleanup_dir() {
+    local label="$1"
+    local dir="$2"
+    local max_days="$3"
+
+    if [ ! -d "$dir" ]; then
+        echo "  [skip] $label: directory does not exist ($dir)"
+        return
+    fi
+
+    if $DRY_RUN; then
+        local count
+        count=$(find "$dir" -maxdepth 1 -type f -mtime +"$max_days" | wc -l)
+        echo "  [dry-run] $label: would delete $count file(s) older than ${max_days}d"
+        if [ "$count" -gt 0 ]; then
+            find "$dir" -maxdepth 1 -type f -mtime +"$max_days" | sort | while read -r f; do
+                echo "    $f"
+            done
+        fi
+    else
+        local count
+        count=$(find "$dir" -maxdepth 1 -type f -mtime +"$max_days" | wc -l)
+        find "$dir" -maxdepth 1 -type f -mtime +"$max_days" -delete
+        echo "  $label: deleted $count file(s) older than ${max_days}d"
+    fi
+}
+
+echo "log-cleanup.sh — $(date -Iseconds)${DRY_RUN:+ [DRY RUN]}"
+
+cleanup_dir "scheduled-jobs/logs" "$WORKSPACE/scheduled-jobs/logs" 7
+cleanup_dir "messages/processed"  "$MESSAGES_DIR/processed"        30
+cleanup_dir "messages/audio"      "$MESSAGES_DIR/audio"            7
+
+echo "log-cleanup.sh done"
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: NEEDS-WORK**

---

### 1. Migration number collision — BLOCKER

Both PR #1272 and this PR define **Migration 62**. Main branch currently ends at Migration 61. Both PRs are open simultaneously, so whichever lands second will silently skip its migration (the idempotency guard will see the number already executed). PR #1278's migration must be renumbered to **Migration 63** (or whatever the next available number is when it lands).

---

### 2. `-maxdepth 1` — correct for processed/ and audio/, questionable for logs/

`messages/processed/` and `messages/audio/` are flat directories of files, so `-maxdepth 1` is correct there.

`scheduled-jobs/logs/` may have per-job subdirectories (one directory per job name is a common pattern). If that layout exists, `-maxdepth 1 -type f` will skip all files nested inside subdirectories entirely — they will never be pruned. Worth confirming the actual directory structure; if subdirectories exist, `-maxdepth 1` should be dropped or increased.

---

### 3. Migration idempotency check — wrong sentinel

```bash
if [ ! -x "$LOG_CLEANUP_SCRIPT" ]; then
    chmod +x ...
```

After a fresh `git pull`, the file ships with the executable bit set (it is `chmod +x` in the repo). The `! -x` test will be **false on a clean install**, so the migration body never runs and `migrated` is never incremented. The script effectively becomes a no-op on every install, which means the `--dry-run` smoke test is never executed and the migration counter never ticks.

A more reliable sentinel would be a marker file or a version-stamped comment in a state file — or simply always running `chmod +x` unconditionally and using a separate marker. As written, the migration only fires when the executable bit is somehow missing, which is not the failure mode this migration is intended to guard against.

---

### 4. Placement in nightly-consolidation.sh — correct

Log cleanup runs before the dedup guard and before writing to the inbox. That is the right order: clean up old files first, then proceed with consolidation. No issue here.

---

### 5. Retention values — acceptable

7d for logs, 30d for processed messages, 7d for audio matches the issue description and is reasonable. No objection.

---

### Summary

| Check | Result |
|---|---|
| Migration collision with #1272 | BLOCKER — renumber to M63 |
| `-maxdepth 1` for logs/ subdirs | Needs verification |
| Migration idempotency sentinel | Logic inverted — fires only when +x missing, not on clean install |
| Placement in nightly-consolidation.sh | OK |
| Retention values | OK |

Fix the migration number collision and the idempotency sentinel before merging.

sayhar (commented):
🤖🦞 Lobster (engineer):

Fixed all three issues: renumbered to M63, fixed idempotency sentinel, confirmed -maxdepth 1 is correct (flat directory structure, added comment).

PR #1285, opened 2026-03-31
fix: suppress false WFM stale restart when compact-reminder is batched (#1283)

### LINKED ISSUE #1283: Dispatcher failed to call record-catchup-state.sh start after compaction, triggering false WFM stale restart
## Summary

At 2026-03-31 14:52 UTC, the health check restarted Lobster because `wait_for_messages` appeared stale for 811s. This was a **false positive** — the dispatcher had spawned a compact-catchup subagent but failed to call `record-catchup-state.sh start`, so WFM freshness suppression was never armed.

## Root cause

The compaction fired at 14:36:50. The dispatcher saw the `compact-reminder` (0_compact) at 14:37:07, but processed it **in a batch** with three other messages simultaneously:
- `0_compact`
- `review-bot-talk-realtime-forwarder-1245`
- `scheduled_bot-talk-realtime-forwarder`
- `scheduled_lobstertalk-incoming-handler`

The compact-reminder handler protocol (`.claude/sys.dispatcher.bootup.md`) requires:
1. `mark_processing(message_id)` for the compact alone
2. Spawn `session-note-polish` subagent
3. Run `~/lobster/scripts/record-catchup-state.sh start` ← **this was skipped**
4. Spawn `compact-catchup` subagent
5. `mark_processed(message_id)`
6. Return to `wait_for_messages()` immediately

The dispatcher instead batched all four messages together, processed them inline, and never ran `record-catchup-state.sh start`. As a result:
- `catchup_started_at` in `lobster-state.json` was still `2026-03-31T12:50:34Z` (from the previous session's catchup)
- The health check's `is_catchup_in_flight()` check saw `catchup_finished_at` (12:50:34) newer than `catchup_started_at` (12:50:34) and concluded no catchup was running
- WFM freshness suppression was never activated
- `last_processed_at` was last updated at 14:38:30 (mark_processed)
- At 14:52:01, health check found 811s of staleness → restart

## Evidence

From `audit.jsonl`:
```
14:37:26 mark_processing(0_compact)
14:37:26 mark_processing(review-bot-talk-realtime-forwarder-1245)  # batched with compact
14:37:27 mark_processing(scheduled_bot-talk-realtime-forwarder)
14:37:27 mark_processing(scheduled_lobstertalk-incoming-handler)
14:38:28 mark_processed(0_compact)                                 # never ran record-catchup-state.sh
...
14:40:54 write_result(compact-catchup-postcompact)                # last activity
[14-minute silence]
14:52:01 health-check: 811s stale → RESTART
```

From `lobster-state.json` (post-restart, showing what was never updated):
- `catchup_started_at = 2026-03-31T12:50:34Z` (old value, never updated for the 14:37 catchup)
- `catchup_finished_at = 2026-03-31T12:50:34Z` (same old value)

## Fix options

**Option A (doc enforcement):** Strengthen `sys.dispatcher.bootup.md` compact-reminder handler to explicitly warn against batching the compact-reminder with other messages. Add a note: "The compact-reminder MUST be handled in isolation — never batch it with other messages. If you see 0_compact alongside other messages, process the compact-reminder first (steps 1-6), then return to WFM, and the other messages will arrive in the next WFM cycle."

**Option B (resilience in record-catchup-state.sh):** Make the health check also check `compacted_at` as a fallback — if `compacted_at` is more recent than `catchup_finished_at`, treat catchup as in-flight for up to `CATCHUP_SUPPRESS_SECONDS`.

Option B is more robust because it doesn't rely on the dispatcher remembering to call the script.

## Recommended fix

Both options together. Option B (code fix) as the primary defense, Option A (doc) as belt-and-suspenders.

**Option B implementation:** In `health-check-v3.sh`, `is_catchup_in_flight()` — if `catchup_started_at` is absent or stale, also check `compacted_at`: if `compacted_at` is recent (within `CATCHUP_SUPPRESS_SECONDS`) and `catchup_finished_at` is older than `compacted_at`, treat the catchup as in-flight.

## Labels
- bug
- health-check
- dispatcher

sayhar:
## Summary

Fixes the spurious health-check restart at 2026-03-31 14:52 UTC (issue #1283).

**Root cause:** The dispatcher batched the `compact-reminder` with 3 other messages. As a result, `record-catchup-state.sh start` was never called, so `catchup_started_at` was not updated. The health check saw no suppression in force and fired a RED restart 811s after the last `mark_processed` call — even though a `compact-catchup` subagent was actively running.

## Changes

**`scripts/health-check-v3.sh`** — `is_catchup_active()` fallback path:
- If `catchup_started_at` is absent or expired, check `compacted_at` as a fallback
- If a compaction is more recent than `catchup_finished_at` and within `CATCHUP_SUPPRESS_SECONDS`, suppress the WFM freshness alarm
- Also consolidates three separate `python3` invocations into one for efficiency

**`.claude/sys.dispatcher.bootup.md`** — compact-reminder handler:
- Adds explicit warning: never batch the compact-reminder with other messages
- Marks `record-catchup-state.sh start` (step 4) as MANDATORY with a note explaining what breaks if skipped

## Test plan

- [ ] Verify `is_catchup_active()` returns true when `compacted_at` is recent and `catchup_finished_at` predates it (even with `catchup_started_at` absent)
- [ ] Verify `is_catchup_active()` returns false when `catchup_finished_at` is more recent than `compacted_at`
- [ ] Verify `is_catchup_active()` returns false when `compacted_at` is older than `CATCHUP_SUPPRESS_SECONDS`
- [ ] Normal path (catchup_started_at present) unchanged

Closes #1283

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -187,8 +187,10 @@ After a context compaction you lose situational awareness of the last ~30 minute
 
 > **MANDATORY: You MUST spawn compact-catchup before doing any other work after a compaction. Do not skip compact-catchup even if the in-conversation summary appears sufficient. The summary only covers pre-compaction context; compact-catchup also checks for in-flight subagent state and recently-returned results that the summary cannot know about.**
 
+> **CRITICAL — never batch the compact-reminder with other messages.** If `0_compact` arrives alongside other messages in the same WFM batch, handle the compact-reminder first (steps 1–7 below), return to `wait_for_messages()`, and the other messages will be waiting in the next cycle. Batching the compact-reminder with other work causes `record-catchup-state.sh start` to be skipped or forgotten, which disables WFM freshness suppression and causes a spurious health-check restart after ~10 minutes (issue #1283).
+
 ```
-1. mark_processing(message_id)
+1. mark_processing(message_id)  <- compact-reminder ONLY, not other messages
 2. Read the compact-reminder text to re-orient (identity, main loop, key files)
 2.5. Send a random ack message to admin chat (see **Selecting the ack message** in the Startup Behavior section):
    - Pick with `random.choice()` from `~/.claude/compact-ack-messages.json`
@@ -198,7 +200,7 @@ After a context compaction you lose situational awareness of the last ~30 minute
    - See .claude/agents/session-note-polish.md for the agent definition
    - Pass: task_id: "session-note-polish", chat_id: 0, source: "system", current_session_file: <path>, MESSAGE_COUNT: <current message count>
    - Do NOT wait for it — spawn and immediately proceed to step 4
-4. Run: ~/lobster/scripts/record-catchup-state.sh start
+4. Run: ~/lobster/scripts/record-catchup-state.sh start  <- MANDATORY, arms WFM suppression
 5. Spawn compact_catchup subagent (subagent_type: "compact-catchup", run_in_background=True):
    - See .claude/agents/compact-catchup.md for the full prompt
    - Pass task_id: "compact-catchup", chat_id: 0, source: "system"
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -436,60 +436,92 @@ is_compact_grace_period() {
 # Check if a catchup subagent is actively running (or recently started).
 # Returns 0 (true) if the WFM freshness check should be suppressed, 1 otherwise.
 #
-# The dispatcher writes catchup_started_at to lobster-state.json before spawning
-# either the startup-catchup or compact-catchup subagent, and writes
-# catchup_finished_at when the subagent result arrives.  If catchup_finished_at
-# is absent or older than catchup_started_at, the catchup is still in flight.
+# Primary path: the dispatcher writes catchup_started_at to lobster-state.json
+# before spawning either the startup-catchup or compact-catchup subagent, and
+# writes catchup_finished_at when the subagent result arrives.  If
+# catchup_finished_at is absent or older than catchup_started_at, the catchup
+# is still in flight.
 #
-# Suppression window: CATCHUP_SUPPRESS_SECONDS from catchup_started_at.
+# Fallback path (issue #1283): if the dispatcher forgot to call
+# record-catchup-state.sh (e.g. it batched the compact-reminder with other
+# messages), catchup_started_at is never updated.  In that case we fall back
+# to compacted_at: if a compaction is more recent than catchup_finished_at and
+# within CATCHUP_SUPPRESS_SECONDS, we treat catchup as in-flight.
+#
+# Suppression window: CATCHUP_SUPPRESS_SECONDS from the effective start time.
 # This caps suppression even if catchup_finished_at is never written (e.g. the
 # subagent crashed), preventing permanent suppression.
 is_catchup_active() {
     if [[ ! -f "$LOBSTER_STATE_FILE" ]]; then
         return 1
     fi
-    local started_at finished_at
-    started_at=$(python3 -c "
-import json, sys
-try:
-    d = json.load(open('$LOBSTER_STATE_FILE'))
-    print(d.get('catchup_started_at', ''))
-except Exception:
-    print('')
-" 2>/dev/null)
-    if [[ -z "$started_at" ]]; then
-        return 1
-    fi
-    local started_epoch
-    started_epoch=$(date -d "$started_at" +%s 2>/dev/null) || return 1
+
     local now
     now=$(date +%s)
-    local age=$((now - started_epoch))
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS: correct fix for the batched compact-reminder false restart, with a sound fallback strategy.

**Root cause diagnosis is accurate.** When the dispatcher batches `0_compact` with other messages, it processes multiple `mark_processing` calls before running `record-catchup-state.sh start`, and in practice the script call gets skipped or deferred. The fallback path correctly identifies this gap.

**Fallback path logic in `is_catchup_active()`:**

The three-field consolidated Python read is an improvement over three sequential subprocess calls — fewer race windows on `lobster-state.json`.

The fallback condition is: `compacted_at` is present, is within `CATCHUP_SUPPRESS_SECONDS`, and `catchup_finished_at` is either absent or predates `compacted_at`. This correctly identifies "compaction happened recently, but we have no evidence the catchup finished." The `finished_epoch -ge compacted_epoch` check (not just `finished_epoch -gt 0`) prevents a stale `catchup_finished_at` from a *previous* catchup cycle from prematurely lifting suppression after a new compaction.

One edge case worth noting (not a blocker): if `started_epoch` parsing fails (returns 0) when `started_at` is non-empty, the primary path silently falls through to the fallback. This is safe behavior — the fallback is more conservative — but the silent failure may be surprising in logs. A `log_warn` on the `started_epoch=0` path would aid debugging. Not a reason to hold this PR.

**Bootup file changes:** The `CRITICAL — never batch` warning and the `<- MANDATORY` annotation on step 4 are well-placed. The warning appears immediately before the code block, where a disoriented post-compaction dispatcher is most likely to read. The annotation on step 4 specifically reminds it that skipping this step breaks suppression, not just best practices.

**What the PR does not change:** `record-catchup-state.sh` itself, and the startup-catchup path — both unchanged, correct.

The fix is layered correctly: the bootup instruction prevents the failure mode on future runs; the fallback suppression handles the case where the dispatcher still batches despite the warning.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Verdict: PASS

### `is_catchup_active()` fallback logic

Correct. When `catchup_started_at` is absent (or its age exceeds `CATCHUP_SUPPRESS_SECONDS`), the function falls through to the `compacted_at` block. If `compacted_at` is within `CATCHUP_SUPPRESS_SECONDS` and `finished_epoch` is either zero or predates `compacted_epoch`, the function returns 0 (suppress). This correctly activates suppression for the scenario where `record-catchup-state.sh start` was never called.

### Boundary condition: `finished_epoch >= compacted_epoch`

Correct. The check `$finished_epoch -gt 0 && $finished_epoch -ge $compacted_epoch` correctly lifts suppression when catchup finished at or after the compaction event. Using `-ge` (not `-gt`) is appropriate — a `finished_at` equal to `compacted_at` to the second means the catchup completed cleanly and re-suppression is not needed.

### `read -r` timestamp splitting

Verified safe. All three timestamps (`catchup_started_at`, `catchup_finished_at`, `compacted_at`) are written as `%Y-%m-%dT%H:%M:%SZ` (no spaces) by both `record-catchup-state.sh` and `hooks/on-compact.py`. The `read -r started_at finished_at compacted_at` splitting on whitespace will not corrupt field values.

### Primary-path fall-through comment

The comment "Fall through to compacted_at fallback — maybe a new compaction occurred after the cap expired" is accurate. If an old catchup's `CATCHUP_SUPPRESS_SECONDS` window has expired but a new compaction happened after, the fallback correctly picks it up.

### `sys.dispatcher.bootup.md` change

The CRITICAL no-batching warning is coherent alongside the existing MANDATORY warning. They address different failure modes: MANDATORY prevents skipping compact-catchup entirely; CRITICAL prevents the subtler failure where compact-catchup is spawned but `record-catchup-state.sh start` is skipped due to batching. The step-4 annotation `<- MANDATORY, arms WFM suppression` makes the causality chain explicit. No redundancy issues.

### Regression check against PR #1402 (`WFM_STALE_SECONDS=1200`)

PR #1402 is still open and not on main (main has `WFM_STALE_SECONDS=600`). PR #1285 does not touch `WFM_STALE_SECONDS` at all, so there is no conflict. When #1402 eventually merges, the longer stale window means the fallback suppression in this PR becomes slightly less critical (more headroom before a false restart), but the logic remains correct and beneficial regardless of the threshold value.

### Minor observation (not blocking)

The three-value `read -r` from a process substitution is a slight departure from the original script's defensive style (each value was read separately with explicit error handling). The combined read is more efficient and works correctly given the current timestamp format, but it's a subtle coupling: if any timestamp format ever changes to include spaces, all three values silently corrupt. Worth a comment noting the format dependency, but this is a low-risk cosmetic nit given the formats are stable and tested.

No blockers. Ready to merge.

Issue #1292, opened 2026-03-31
Phase 2: Two-tier message gating — DM vs group routing in handle_message


aeschylus:
## Phase 2: Two-tier message gating in handle_message

**Epic:** #1288

### Problem

\`handle_message\` in \`lobster_bot.py\` applies a single \`user.id not in ALLOWED_USERS\` check that rejects all group messages. We need to distinguish DMs from group/supergroup messages and apply the skill's gating logic to groups.

### Approach

Replace the single guard at the top of \`handle_message\` with a two-tier check, and set \`source="lobster-group"\` for group messages.

**Prerequisite:** This phase assumes \`_SKILL_DIR\` sys.path addition and \`_GROUP_GATING_ENABLED\` soft import are already present (from Phase 1). If Phase 1 is not yet merged, include those changes here too.

**Imports needed (soft, with \`_GROUP_GATING_ENABLED\` flag):**
```python
try:
    from multiplayer_telegram_bot.whitelist import load_whitelist
    from multiplayer_telegram_bot.gating import gate_message, GatingAction
    from multiplayer_telegram_bot.router import is_group_message, get_source_for_chat
    _GROUP_GATING_ENABLED = True
except ImportError:
    _GROUP_GATING_ENABLED = False
    log.warning("multiplayer-telegram-bot skill not available — group gating disabled")
```

**Replace the existing guard in handle_message (around line 905):**

Current:
```python
if not user or user.id not in ALLOWED_USERS:
    return
```

Replace with:
```python
if not user:
    return

chat = message.chat
if chat.type in ("group", "supergroup"):
    if _GROUP_GATING_ENABLED:
        store = load_whitelist()
        result = gate_message(chat.id, user.id, store)
        if result.action == GatingAction.DROP_SILENT:
            log.debug(f"Group message silently dropped: {result.reason}")
            return
        elif result.action == GatingAction.SEND_REGISTRATION_DM:
            log.info(f"Sending registration DM to user {user.id} for group {chat.id}")
            try:
                await bot_app.bot.send_message(
                    chat_id=user.id,
                    text="Hi! To use Lobster in this group, please ask the group admin to whitelist you."
                )
            except Exception as e:
                log.warning(f"Failed to send registration DM to {user.id}: {e}")
            return
        # GatingAction.ALLOW — fall through
    else:
        return  # gating not available, drop all group messages silently
else:
    # DM: existing behavior unchanged
    if user.id not in ALLOWED_USERS:
        return
```

[...]

Issue #1291, opened 2026-03-31
Phase 1: sys.path addition + soft-import of multiplayer skill modules


aeschylus:
Part of #1289 — Group chat support: two-tier gating + privacy mode

## Goal

Wire the `multiplayer-telegram-bot` skill into `lobster_bot.py` via sys.path manipulation and a soft try/except import block. This is the foundation all other phases depend on.

## Context

The skill lives at:
```
~/lobster/lobster-shop/multiplayer-telegram-bot/src/multiplayer_telegram_bot/
```

`lobster_bot.py` already manipulates `sys.path` at the top (for `_SRC_DIR`). The same pattern applies.

### Change 1 — Add skill dir to sys.path (near top, after existing `_SRC_DIR` insert)

```python
_SKILL_DIR = str(Path(__file__).resolve().parent.parent.parent /
                 "lobster-shop" / "multiplayer-telegram-bot" / "src")
if _SKILL_DIR not in _sys.path:
    _sys.path.insert(0, _SKILL_DIR)
```

### Change 2 — Soft-import skill modules (after sys.path addition)

```python
try:
    from multiplayer_telegram_bot.whitelist import load_whitelist, enable_group, add_allowed_user, save_whitelist
    from multiplayer_telegram_bot.gating import gate_message, GatingAction
    from multiplayer_telegram_bot.router import is_group_message, get_source_for_chat
    from multiplayer_telegram_bot.commands import handle_enable_group_bot, handle_whitelist, handle_unwhitelist
    _GROUP_GATING_ENABLED = True
except ImportError:
    _GROUP_GATING_ENABLED = False
    log.warning("multiplayer-telegram-bot skill not available — group gating disabled")
```

## Acceptance Criteria

- [ ] `lobster_bot.py` imports succeed without errors when skill is present
- [ ] `_GROUP_GATING_ENABLED = True` when skill modules are importable
- [ ] `_GROUP_GATING_ENABLED = False` and only a warning logged when skill is missing (soft fallback, no crash)
- [ ] Bot still starts and handles DMs normally regardless of `_GROUP_GATING_ENABLED` value
- [ ] Existing tests pass: `pytest ~/lobster/lobster-shop/multiplayer-telegram-bot/`

## Files to Change

- `~/lobster/lobster_bot.py` — add sys.path block and import block near the top

## Notes

- Do NOT change any handler logic in this phase — that's Phase 2+
[...]

PR #1303, opened 2026-03-31
feat(bot): Phase 3 — group management command handlers

### LINKED ISSUE #1293: Phase 3: Group management command handlers (/enable_group_bot, /whitelist, /unwhitelist)
## Phase 3: Command handlers for group management

**Epic:** #1288

### Problem

There are no \`/enable_group_bot\`, \`/whitelist\`, or \`/unwhitelist\` command handlers in \`lobster_bot.py\`. The skill's \`commands.py\` has the logic; we just need wrappers registered before the generic \`MessageHandler\`.

### Approach

Add three async command handler wrappers in \`lobster_bot.py\` and register them with \`CommandHandler\`.

All three:
- Check \`user.id in ALLOWED_USERS\` (drop silently if not)
- Only work in private DMs (\`update.effective_chat.type == "private"\`)
- Delegate to the skill's command handlers from \`multiplayer_telegram_bot.commands\`

**Imports (add to the soft-import try block):**
```python
from multiplayer_telegram_bot.commands import handle_enable_group_bot, handle_whitelist, handle_unwhitelist
```

**Handler wrappers:**
```python
async def enable_group_bot_command(update: Update, context: ContextTypes.DEFAULT_TYPE):
    user = update.effective_user
    if not user or user.id not in ALLOWED_USERS:
        return
    if update.effective_chat.type != "private":
        await update.message.reply_text("This command only works in a private DM with me.")
        return
    result = handle_enable_group_bot(update.message.text)
    await update.message.reply_text(result.reply)

async def whitelist_command(update: Update, context: ContextTypes.DEFAULT_TYPE):
    user = update.effective_user
    if not user or user.id not in ALLOWED_USERS:
        return
    if update.effective_chat.type != "private":
        return
    result = handle_whitelist(update.message.text)
    await update.message.reply_text(result.reply)

async def unwhitelist_command(update: Update, context: ContextTypes.DEFAULT_TYPE):
    user = update.effective_user
    if not user or user.id not in ALLOWED_USERS:
        return
    if update.effective_chat.type != "private":
        return
    result = handle_unwhitelist(update.message.text)
    await update.message.reply_text(result.reply)
```

**Registration in main() (BEFORE the generic MessageHandler):**
```python
bot_app.add_handler(CommandHandler("enable_group_bot", enable_group_bot_command))
bot_app.add_handler(CommandHandler("whitelist", whitelist_command))
bot_app.add_handler(CommandHandler("unwhitelist", unwhitelist_command))
```

### Files to change

- \`~/lobster/lobster_bot.py\`

### Acceptance Criteria

- [ ] Three command handlers defined and registered before generic \`MessageHandler\`
- [ ] \`/enable_group_bot <chat_id>\` in a DM enables a group in the whitelist
- [ ] \`/whitelist <chat_id> <user_id>\` adds a user to a group's whitelist
- [ ] \`/unwhitelist <chat_id> <user_id>\` removes a user from a group's whitelist
- [ ] Commands sent from a group (not DM) are silently ignored or get polite rejection
- [ ] Non-allowed users sending these commands get no response
- [ ] PR opened targeting \`main\`

### Notes

This phase is fully independent from Phases 1 and 2 — it does not touch \`handle_message\` or the \`my_chat_member\` handler. Can be developed in parallel. The skill's \`commands.py\` handles all whitelist file mutations.

aeschylus:
## Problem

Operators had no Telegram interface for managing the group whitelist. The
`multiplayer-telegram-bot` skill ships complete, tested command functions
(`handle_enable_group_bot`, `handle_whitelist`, `handle_unwhitelist`) but
nothing in `lobster_bot.py` called them. There was also no way to inspect
the current whitelist state from Telegram.

## What this changes

Four new command handlers, registered in `lobster_bot.py` before the generic
`MessageHandler` so they are dispatched as bot commands rather than falling
through to Claude:

- `/enable_group_bot <chat_id> [name]` — enable a group in the whitelist
- `/whitelist <user_id> <chat_id>` — add a user to a group's allowed list
- `/unwhitelist <user_id> <chat_id>` — remove a user from a group's allowed list
- `/list_groups` — show all configured groups, their enabled/disabled status, and whitelisted user IDs

Every handler applies two guards before doing anything:
1. `user.id in ALLOWED_USERS` — non-allowed users are silently dropped
2. `chat.type == "private"` — `/enable_group_bot` replies with an error; the
   rest silently drop (whitelist operations have no business in a group context)

The skill is imported via a soft-import try/except block that sets
`_GROUP_COMMANDS_ENABLED`. If the skill package is absent, handlers reply with
a graceful "not available" message rather than crashing.

`list_groups` is pure read: it calls `load_whitelist()` and formats the result
inline — no skill command needed.

## Functional patterns used

- Pure delegation: the handlers are thin wrappers; all parsing and whitelist
  mutation logic lives in the skill's pure functions (`handle_enable_group_bot`,
  `handle_whitelist`, `handle_unwhitelist`, `load_whitelist`)
- Side-effect isolation: file I/O is confined to the skill's command layer; the
  handler wrappers are pure async adapters
- Soft-import guard: `_GROUP_COMMANDS_ENABLED` flag allows the bot to start and
  operate safely even if the skill is not installed

## Out of scope

This PR does not touch `handle_message`, the `my_chat_member` handler, message
gating, or the dispatcher. Those are Phases 1 and 2 (#1291, #1292).

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_group_commands.py -v` — 16 passed, 0 failed
- [x] `uv run pytest tests/unit/test_bot/ -v` — 145 passed, 0 failed (no regressions in existing bot tests)

Closes #1293

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -30,15 +30,30 @@ from watchdog.observers import Observer
 from watchdog.events import FileSystemEventHandler
 
 # multiplayer-telegram-bot skill — soft import; enables group whitelist management
+# and group management commands.  Three levels up from src/bot/lobster_bot.py
+# lands at the repo root (~/lobster/), then lobster-shop/ is a subdirectory there.
 _SKILL_DIR = str(Path(__file__).resolve().parent.parent.parent /
                  "lobster-shop" / "multiplayer-telegram-bot" / "src")
 if _SKILL_DIR not in _sys.path:
     _sys.path.insert(0, _SKILL_DIR)
 try:
     from multiplayer_telegram_bot.whitelist import load_whitelist, enable_group, add_allowed_user, save_whitelist  # noqa: E402
+    from multiplayer_telegram_bot.gating import gate_message, GatingAction  # noqa: E402
+    from multiplayer_telegram_bot.router import get_source_for_chat  # noqa: E402
+    from multiplayer_telegram_bot.commands import (  # noqa: E402
+        handle_enable_group_bot,
+        handle_whitelist,
+        handle_unwhitelist,
+    )
     _GROUP_GATING_ENABLED = True
+    _GROUP_COMMANDS_ENABLED = True
 except ImportError:
     _GROUP_GATING_ENABLED = False
+    _GROUP_COMMANDS_ENABLED = False
+    import logging as _logging
+    _logging.getLogger(__name__).warning(
+        "multiplayer-telegram-bot skill not available — group gating and management commands disabled"
+    )
 
 # ChannelAdapter Protocol — soft import; lobster_bot satisfies it structurally
 # but keeps its own async OutboxHandler rather than using OutboxFileHandler.
@@ -48,26 +63,6 @@ try:
     from channels.base import ChannelAdapter  # noqa: F401
 except ImportError:
     pass  # channels package not yet installed; type hint only
-
-# multiplayer-telegram-bot skill — group gating library.
-# Adds the skill's src/ directory to sys.path so we can import the pure
-# gating/whitelist/router modules without a full package install.
-_SKILL_DIR = str(Path(__file__).resolve().parent.parent.parent
-                 / "lobster-shop" / "multiplayer-telegram-bot" / "src")
-if _SKILL_DIR not in _sys.path:
-    _sys.path.insert(0, _SKILL_DIR)
-
-try:
-    from multiplayer_telegram_bot.whitelist import load_whitelist
-    from multiplayer_telegram_bot.gating import gate_message, GatingAction
-    from multiplayer_telegram_bot.router import get_source_for_chat
-    _GROUP_GATING_ENABLED = True
-except ImportError:
-    _GROUP_GATING_ENABLED = False
-    import logging as _logging
-    _logging.getLogger(__name__).warning(
-        "multiplayer-telegram-bot skill not available — group gating disabled"
-    )
 
 import re
 from dataclasses import dataclass, field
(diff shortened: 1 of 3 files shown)


aeschylus (commented):
NEEDS-WORK: The `_SKILL_DIR` path has one extra `.parent` call.

```python
# Current (wrong) — 4 levels up from lobster_bot.py
_SKILL_DIR = str(Path(__file__).resolve().parent.parent.parent.parent /
                 "lobster-shop" / "multiplayer-telegram-bot" / "src")
```

`lobster_bot.py` lives at `src/bot/lobster_bot.py` inside the repo. Working up the tree:
- `.parent` → `src/bot`
- `.parent.parent` → `src`
- `.parent.parent.parent` → repo root (where `lobster-shop/` lives) ✓
- `.parent.parent.parent.parent` → **parent of repo root** ✗

With 4 `.parent` calls, `_SKILL_DIR` resolves to `/home/admin/lobster-workspace/projects/lobster-shop/multiplayer-telegram-bot/src` which does not exist. The soft-import block will catch the `ImportError`, set `_GROUP_COMMANDS_ENABLED = False`, and all four command handlers will reply "not available".

**Fix:** Remove one `.parent`:

```python
_SKILL_DIR = str(Path(__file__).resolve().parent.parent.parent /
                 "lobster-shop" / "multiplayer-telegram-bot" / "src")
```

Also worth verifying that `tests/conftest.py`'s skill path addition uses the correct depth for the test environment — the conftest currently navigates via `_WORKTREE_ROOT.parent.parent.parent / "lobster" / "lobster-shop" ...` which may or may not resolve correctly depending on the test runner's working directory.

PR #1314, opened 2026-03-31
fix: rename data/events.jsonl -> memory-events.jsonl + add 1GB rotation

sayhar:
🤖🦞 Lobster (engineer):

Two related files shared the name `events.jsonl` in different directories, making it easy to confuse the StaticMemory fallback log (`data/events.jsonl`) with the EventBus operational log (`logs/events.jsonl`). This PR renames the former and adds a size cap to prevent unbounded growth.

## What changed

**Rename:** `StaticMemory` now writes to `data/memory-events.jsonl` instead of `data/events.jsonl`. The new name makes clear this is the StaticMemory JSONL backend — the Layer 3 fallback used when sqlite-vec is unavailable — not the structured EventBus log.

**1 GB rotation:** `StaticMemory.store()` now calls `_rotate_if_needed()` before each append. When the file exceeds 1,073,741,824 bytes, the current file is renamed to `memory-events.jsonl.1` and a fresh file is started. The rotated file is preserved intact — no compression, no deletion. On long-running installs where sqlite-vec was never available, this file could otherwise grow without bound.

**Migration 63** in `upgrade.sh` handles existing installs: if `data/events.jsonl` exists and `data/memory-events.jsonl` does not, it renames the file. History is preserved.

**`CLAUDE.md`** updated to reflect the new path.

## What is not changed

- The EventBus `JsonlFileListener` at `logs/events.jsonl` is untouched — different file, different system.
- VectorMemory behavior is unchanged; StaticMemory is only the fallback path.
- Rotation keeps exactly one backup (`.1`). No multi-generation rotation.

## Functional patterns used

Pure function for the rotation predicate (`_rotate_if_needed` has no observable side effect beyond the filesystem rename, and is idempotent if called multiple times below threshold). Side effect (rename) is isolated at the boundary of `store()`.

## Tests run

- [x] `uv run pytest tests/unit/test_memory.py -k "not test_falls_back_to_static and not test_force_static"` — 35 passed, 0 failed. The two excluded tests fail due to a pre-existing environment issue (lobster-user-config is a git repo in this environment), not related to this change.
- [x] `uv run pytest tests/unit/test_event_bus.py` — 25 passed, 0 failed
- [x] `uv run python -c "import py_compile; py_compile.compile('src/mcp/memory/static_memory.py')"` — syntax OK

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,7 +171,7 @@ Never use cron for user-space jobs. Never use systemd tools for system-level inf
   - `CLAUDE.md` → symlink to `~/lobster/CLAUDE.md` — same, live immediately
   - `projects/` - All Lobster-managed projects (`$LOBSTER_PROJECTS`)
   - `data/memory.db` - Vector memory SQLite DB
-  - `data/events.jsonl` - Event log
+  - `data/memory-events.jsonl` - StaticMemory event log (JSONL fallback backend)
   - `scheduled-jobs/jobs.json` - Job registry state
   - `scheduled-jobs/tasks/` - Task definition markdown files
   - `scheduled-jobs/logs/` - Execution logs
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2280,6 +2280,16 @@ PYEOF
         migrated=$((migrated + 1))
     fi
 
+    # Migration 63: Rename data/events.jsonl -> data/memory-events.jsonl.
+    # StaticMemory now writes to memory-events.jsonl to distinguish it from the
+    # EventBus operational log at logs/events.jsonl. Rename any existing file
+    # so history is preserved without manual intervention.
+    if [ -f "$WORKSPACE_DIR/data/events.jsonl" ] && [ ! -f "$WORKSPACE_DIR/data/memory-events.jsonl" ]; then
+        mv "$WORKSPACE_DIR/data/events.jsonl" "$WORKSPACE_DIR/data/memory-events.jsonl"
+        substep "Renamed data/events.jsonl -> data/memory-events.jsonl"
+        migrated=$((migrated + 1))
+    fi
+
     if [ "$migrated" -eq 0 ]; then
         success "No migrations needed"
     else
--- a/src/mcp/memory/static_memory.py
+++ b/src/mcp/memory/static_memory.py
@@ -25,7 +25,9 @@ log = logging.getLogger("lobster-memory")
 _WORKSPACE = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
 _USER_CONFIG = Path(os.environ.get("LOBSTER_USER_CONFIG", Path.home() / "lobster-user-config"))
 DEFAULT_CANONICAL_DIR = _USER_CONFIG / "memory" / "canonical"
-DEFAULT_EVENT_LOG = _WORKSPACE / "data" / "events.jsonl"
+DEFAULT_EVENT_LOG = _WORKSPACE / "data" / "memory-events.jsonl"
+
+_ROTATION_THRESHOLD = 1_073_741_824  # 1 GB in bytes
 
 
 class StaticMemory:
@@ -67,8 +69,25 @@ class StaticMemory:
                     continue
         return max_id + 1
 
+    def _rotate_if_needed(self) -> None:
+        """Rotate the event log if it exceeds the 1 GB threshold.
+
+        Renames the current file to <name>.1 (overwriting any previous rotation)
+        and starts a fresh log. Does not compress or delete — the rotated file
+        is kept intact for manual inspection or archival.
+        """
+        if not self._event_log.exists():
+            return
+        if self._event_log.stat().st_size < _ROTATION_THRESHOLD:
+            return
+        rotated = self._event_log.with_suffix(self._event_log.suffix + ".1")
+        self._event_log.rename(rotated)
+        log.info("Rotated %s -> %s (exceeded 1 GB)", self._event_log, rotated)
+
     def store(self, event: MemoryEvent) -> int:
-        """Append event to JSONL log file."""
+        """Append event to JSONL log file, rotating if the file exceeds 1 GB."""
+        self._rotate_if_needed()
+
         event.id = self._next_id
         self._next_id += 1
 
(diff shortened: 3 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**NEEDS-WORK**

---

## Summary

The rename is clean and the migration is correct. Two issues need attention before merge: a pre-existing duplicate migration 61 in `upgrade.sh` (not introduced by this PR but worth flagging), and a missing test for the rotation logic. There is also one stale `data/events.jsonl` reference in `test_path_guard.py` that should be updated for clarity.

---

## Findings

### 1. Rename end-to-end — PASS

`static_memory.py` `DEFAULT_EVENT_LOG` correctly updated to `data/memory-events.jsonl`. `CLAUDE.md` updated. Migration 63 in `upgrade.sh` renames `data/events.jsonl` → `data/memory-events.jsonl` on existing installs if-and-only-if the destination does not already exist. Path is correct, idempotent, and history-preserving.

### 2. Migration 63 numbering — PASS (with pre-existing caveat)

Migration 63 follows migration 62 and does not collide with any other migration in the file. However, there is a pre-existing bug in `upgrade.sh` (not introduced by this PR): **migration 61 is duplicated verbatim** at lines 2091 and 2106, both adding the nightly-consolidation crontab entry. The second copy is idempotent (the `grep -q` guard prevents double-install), but it is confusing and should be cleaned up. This is out of scope for this PR but worth noting.

Separately, migrations 56–59 appear twice in the file (around lines 2121–2255), likely another pre-existing artifact. Same comment: not introduced here, but the duplicate blocks inflate the file and could cause confusion.

### 3. 1 GB rotation logic — PASS with clobber note

The logic is correct: file is renamed to `memory-events.jsonl.1` before the next `open(..., "a")` call, so no data is written to a missing file. The rotated file is preserved intact.

**Known limitation (documented, acceptable):** On second rotation, `.1` is silently overwritten. If the system is running continuously and generates 2 GB+ before any human inspects the backup, the older `.1` file is lost. The PR description acknowledges this ("No multi-generation rotation"). For a fallback-path log this is acceptable, but the docstring should mention it explicitly so future readers don't assume multi-gen behavior.

Suggested docstring addition to `_rotate_if_needed`:
```
        If memory-events.jsonl.1 already exists it is overwritten — only one
        generation is kept.
```

### 4. `data/events.jsonl` references elsewhere — NEEDS-WORK (minor)

`tests/unit/test_path_guard.py` line 55 still uses `"data" / "events.jsonl"` as a test fixture path. This is testing `assert_not_in_git_repo`, not `StaticMemory`, so the filename is arbitrary — but it creates a cosmetic mismatch and could confuse a future reader. Recommend updating to `"data" / "memory-events.jsonl"` for consistency.

`tests/unit/test_memory.py` uses `temp_dir / "events.jsonl"` as the injected `event_log=` parameter in multiple tests. These pass an explicit path to `StaticMemory.__init__`, so they don't reference the default. No change required.

The GitHub code-search hit on `CLAUDE.md` is the base-branch version — the PR correctly updates it. No issue there.

### 5. No test for rotation logic — NEEDS-WORK

The PR adds `_rotate_if_needed()` but does not add a unit test covering the rotation path. Given the rename-and-continue semantics and the one-backup clobber behavior, a test would be valuable:

- File at exactly `_ROTATION_THRESHOLD` bytes should rotate.
- File below threshold should not rotate.
- Existing `.1` should be silently overwritten on second rotation.

`test_memory.py` already has the scaffolding; a `test_rotation_*` group would slot in cleanly.

---

## Required before merge

1. Add at least one unit test for `_rotate_if_needed` (rotation fires at threshold, no rotation below threshold, existing `.1` is overwritten).
2. Update `test_path_guard.py` line 55 fixture path to `memory-events.jsonl` (cosmetic but keeps the codebase consistent).

## Optional / follow-up

- Update `_rotate_if_needed` docstring to note the single-generation clobber behavior.
- Fix the pre-existing duplicate migration 61 blocks in `upgrade.sh` (separate PR).

sayhar (commented):
🤖🦞 Lobster (reviewer):

**PASS** — both required items from the prior review are addressed. Full re-review findings below.

---

## Required items from prior review

### 1. Unit tests for `_rotate_if_needed()` — DONE

Three tests added to `TestStaticMemory`:

- **`test_rotate_if_needed_fires_at_threshold`** — writes exactly `_ROTATION_THRESHOLD` bytes (1,073,741,824). The condition is `stat().st_size < _ROTATION_THRESHOLD`, so a file at exactly the threshold triggers rotation. Logic is correct.
- **`test_rotate_if_needed_skips_below_threshold`** — writes `_ROTATION_THRESHOLD - 1` bytes. Condition holds (`< threshold`), returns early. No `.1` file created. Logic is correct.
- **`test_rotate_if_needed_clobbers_existing_rotation`** — pre-creates a stale `.1` file, then triggers rotation. `Path.rename()` on Linux overwrites atomically, so the old `.1` is clobbered. The test asserts the new content matches. Logic is correct.

All three boundary conditions are covered and the assertions match the implementation.

### 2. `test_path_guard.py` line 55 — DONE

Changed from `"data" / "events.jsonl"` to `"data" / "memory-events.jsonl"`. Confirmed in diff.

---

## Full re-review

**`static_memory.py`**

- `DEFAULT_EVENT_LOG` renamed to use `memory-events.jsonl`. Correct.
- `_ROTATION_THRESHOLD = 1_073_741_824` — exactly 1 GiB (2^30). Named constant, not a magic number. Good.
- `_rotate_if_needed()` is a clean function: guards for missing file, guards below threshold, then does the rename. Side effect is isolated to the rename call. No surprises.
- `with_suffix(self._event_log.suffix + ".1")` — suffix is `.jsonl`, so result is `.jsonl.1`, yielding `memory-events.jsonl.1`. Verified this produces the correct path.
- `store()` calls `_rotate_if_needed()` before the append — correct ordering. If rotation fires, the new event lands in a fresh file.

**`upgrade.sh` (Migration 63)**

- Guard is `[ -f "$WORKSPACE_DIR/data/events.jsonl" ] && [ ! -f "$WORKSPACE_DIR/data/memory-events.jsonl" ]` — safe, idempotent, no data loss. Correct.

**`CLAUDE.md`**

- Updated from `data/events.jsonl - Event log` to `data/memory-events.jsonl - StaticMemory event log (JSONL fallback backend)`. More accurate description. Good.

**No issues found.** The implementation is clean, the tests cover the meaningful boundary conditions, and migration handles existing installs safely.

PR #1315, opened 2026-03-31
fix(compact-catchup): semantically correct window algorithm (supersedes #1276)

### LINKED ISSUE #1249: compact-catchup: enforce minimum lookback window (2 hours or 150 messages)

sayhar:
🤖🦞 Lobster (engineer):

## Problem

After a context compaction, the dispatcher loses all situational awareness. `compact-catchup` is supposed to restore it by scanning recent message history. But the old window selection logic produced near-empty summaries when compaction fired shortly after a catchup run — in the reported case, an 11-minute window.

PR #1276 added a 2-hour minimum floor. This PR supersedes it with the correct semantic fix.

## Why the floor was wrong

The 2-hour floor is a magic constant that addresses the symptom (window too small) without addressing the cause (wrong anchor logic). The old Step 2 rule was:

> "prefer `last_catchup_ts` if present"

This is correct for incremental updates — where the dispatcher remembers the previous catchup result. But after a compaction, the dispatcher has lost all memory, including the previous catchup summary. Recovery needs to produce a **standalone** summary, not an incremental one. Using `last_catchup_ts` as an anchor assumes memory the dispatcher no longer has.

The 2-hour floor also has a subtle directional error: it enforces `window_start = min(window_start, now - 2 hours)` — always extending backward from *now*. The correct extension should be backward from `last_compaction_ts`, not from now. If catchup runs 30 minutes after compaction, the floor gives `[now - 2h, now]` instead of `[compaction_ts - 2h, now]`, missing 30 minutes of content.

## What changed

**Step 2** is replaced with an explicit algorithm:

1. When `last_compaction_ts` is present (the normal post-compaction case):
   - Start with `candidate = min(last_catchup_ts, last_compaction_ts)`. The `min()` ensures the window never starts *after* the compaction — we always go back to at least the compaction event.
   - If `candidate` is more recent than `now - 2 hours`, extend to `now - 2 hours`. This guarantees a minimum context horizon for standalone reconstruction. The 2-hour constant is now semantically justified (minimum context for standalone recovery) rather than a bandage.
   - Clamp to `now - 6 hours` as a backstop.

2. When `last_compaction_ts` is absent: fall back to `last_catchup_ts` or `now - 6 hours` (unchanged behavior).

Also removes the phantom `last_restart_ts` field from the algorithm. Static analysis of `on-compact.py`, `on-fresh-start.py`, `health-check-v3.sh`, and `record-catchup-state.sh` confirms this field is **never written** to `compaction-state.json`. The fallback that referenced it was dead code.

**Step 3** replaces the fixed `limit=100` with a dynamic expression: `limit = max(200, int(hours_in_window * 50))`. This scales with the window width, reducing silent truncation for wide windows. The `check_inbox` implementation hard-stops at `limit` with no warning — a 100-message cap on a 6-hour window with moderate traffic will silently drop messages.

## Before / after: window calculation

```
Scenario: last_catchup_ts = T-11min, last_compaction_ts = T-1min
Catchup runs at T (1 min after compaction)

Old logic:
  prefer last_catchup_ts = T-11min
  window = [T-11min, T] = 11 minutes  ← too narrow for standalone recovery

PR #1276 floor:
  window_start = min(T-11min, T-1min-2h) = T-2h1min
  window = [T-2h1min, T]  ← correct width, but start anchored to now-2h not to compaction

This PR:
  candidate = min(T-11min, T-1min) = T-11min
  candidate < T-2h? No → extend to T-2h (context horizon)
  window = [T-2h, T] = 2 hours  ← standalone recovery, correctly bounded
```

```
Scenario: last_catchup_ts = T-3h, last_compaction_ts = T-30min
(long session, catchup ran 3h ago, compaction fired 30min ago)

Old logic:
  prefer last_catchup_ts = T-3h
  window = [T-3h, T] = 3 hours  ← fine

This PR:
  candidate = min(T-3h, T-30min) = T-3h
  candidate < T-2h? Yes (T-3h is older) → no extension needed
  window = [T-3h, T] = 3 hours  ← same result, correct
```

## What was not changed

- Phase 2 (session file population) — unchanged
- Phase 3 (rolling summary) — unchanged
- Phase 4 (commitment carry-forward) — unchanged
- `on-compact.py` and `on-fresh-start.py` — no code changes needed; the fix is in the agent prompt
- The `last_catchup_ts` update logic (step 8) — unchanged

## Functional patterns

The window computation is a pure function of the available timestamps — no side effects, no mutable state. The algorithm is explicit and testable at each step (collect anchors → compute candidate → apply horizon → apply backstop).

## Closes / supersedes

Supersedes PR #1276 (fix/compact-catchup-lookback-1249).
Closes #1249.

## Tests run

- [x] Diff review — change is a prose edit to `.claude/agents/compact-catchup.md`; no runtime code was altered
- [x] Static analysis of `on-compact.py`, `on-fresh-start.py`, `health-check-v3.sh`, `record-catchup-state.sh` to verify `last_restart_ts` is never written — confirmed
- [x] Logic trace through all 4 scenarios in the Before/After diagram above — all produce correct windows
- [ ] Live compaction cycle — requires a running dispatcher session; safe to merge without it, behavioral change is to agent prompt text only

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -15,11 +15,46 @@ You are the **compact_catchup** subagent. Your job is to:
 ### Phase 1: Inbox scan and summarization
 
 1. Read `~/lobster-workspace/data/compaction-state.json` to get timestamps.
-2. Compute the catch-up window start: prefer `last_catchup_ts` if present (anchored to last read); otherwise fall back to `min(last_compaction_ts, last_restart_ts)` (use the farther-back timestamp to maximise the window); default to 6 hours ago if none are present.
-3. Call `check_inbox(since_ts=<window_start>, limit=100)` to fetch messages from that window. 100 is a floor -- if the window is large, increase the limit further rather than truncating.
+
+2. Compute the catch-up window start using the algorithm below. The goal is to produce
+   a standalone summary -- one that does not depend on any prior catchup result, because
+   a compaction has erased those prior results from the dispatcher's memory.
+
+   **Algorithm:**
+
+   a. Read available anchors from compaction-state.json:
+      - `last_compaction_ts` -- when the compaction occurred (written by on-compact.py)
+      - `last_catchup_ts` -- when compact-catchup last ran and updated the state
+
+      Note: `last_restart_ts` is not written by any hook; treat it as always absent.
+
+   b. If `last_compaction_ts` is present:
+      - The dispatcher has lost all context since the compaction. Recovery must go back far
+        enough to reconstruct situational awareness from scratch.
+      - Compute `candidate`:
+          If `last_catchup_ts` is present: `candidate = min(last_catchup_ts, last_compaction_ts)`
+          Else: `candidate = last_compaction_ts`
+        Using `min()` ensures the window starts at or before the compaction event, never after it.
+      - Apply context horizon: if `candidate` is more recent than `now - 2 hours`,
+        set `candidate = now - 2 hours`. Rationale: a 2-hour window is the minimum needed
+        to reconstruct meaningful context from scratch, regardless of how recently the
+        previous catchup ran.
+      - Apply backstop: `window_start = max(candidate, now - 6 hours)`. Never scan more
+        than 6 hours back.
+
+   c. If `last_compaction_ts` is absent (file missing or field unset):
+      - Use `last_catchup_ts` if present (subject to the 6-hour backstop).
+      - Default to `now - 6 hours` if no anchors are available.
+
+3. Compute the `check_inbox` limit dynamically based on window width:
+   `limit = max(200, int(hours_in_window * 50))`
+   where `hours_in_window = (now - window_start).total_seconds() / 3600`.
+   This scales the limit with the window and reduces the risk of silent truncation.
+   Call `check_inbox(since_ts=<window_start>, limit=<computed_limit>)`.
+
 4. Filter the results -- include only:
    - User messages (source: telegram, slack, sms, etc.)
-   - `subagent_result` messages (these are recently-returned subagent results — collect their `task_id` values; these represent work that completed and may need dispatcher follow-up)
+   - `subagent_result` messages (these are recently-returned subagent results -- collect their `task_id` values; these represent work that completed and may need dispatcher follow-up)
    - Notable system events: `update_notification`, `consolidation`
    - Exclude: `self_check`, `compact-reminder`, `compact_catchup`, `subagent_notification`, test messages
 5. **Call `get_active_sessions()` now** to retrieve all currently running agent sessions. Filter to `status: "running"` sessions, excluding dispatcher sessions. These are in-flight subagents that were active at compaction time and may still be running. If `get_active_sessions()` errors, note the failure and continue.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

## Verdict: APPROVE with one nit

The algorithm is correct. All three flagged areas check out. One stale comment escaped the PR.

---

### 1. Window algorithm — `min()` direction

**Correct.** `candidate = min(last_catchup_ts, last_compaction_ts)` picks the *earlier* of the two timestamps, which produces the *widest* window. This is the right direction: after compaction the dispatcher has no memory, so starting as far back as the earliest known anchor is correct.

The two-step extension (`now - 2h` horizon, then `now - 6h` backstop) is also sound. The 2-hour floor is now semantically justified — it is a minimum for standalone recovery, not a magic constant covering an ill-understood symptom. The backstop prevents unbounded scans. Order matters here (horizon first, backstop second) and the PR gets it right.

One edge case worth noting: if `last_catchup_ts > last_compaction_ts` (catchup ran *after* the compaction), the `min()` still selects `last_compaction_ts`, so the window extends back to at least the compaction event. That is the correct behavior and the PR's worked examples confirm it.

### 2. Removal of `last_restart_ts`

**Confirmed absent.** Source inspection of all four files named in the PR:

- `hooks/on-compact.py`: writes only `last_compaction_ts` to `compaction-state.json` (line 179). The `last_restart_ts` mention at line 166 is in a stale *docstring* — it describes the old algorithm, not the actual write path.
- `hooks/on-fresh-start.py`: reads `last_catchup_ts` from `compaction-state.json` for staleness checks but writes nothing to it.
- `scripts/health-check-v3.sh`: `last_restart_time` is a local bash variable for restart throttling; `last_restart_at` is written to `lobster-state.json` (a different file entirely). Neither touches `compaction-state.json`.
- `scripts/record-catchup-state.sh`: writes `catchup_started_at` / `catchup_finished_at` to `lobster-state.json`. Does not touch `compaction-state.json`.

Live `compaction-state.json` on the running system confirms: `{"last_compaction_ts": ..., "last_catchup_ts": ...}` — no `last_restart_ts` key.

**Nit (not blocking):** The docstring in `on-compact.py` `write_compaction_state()` (lines 164–167) still reads:

> "it fetches messages since max(last_compaction_ts, last_restart_ts, last_catchup_ts)"

This is stale — it describes the old algorithm and references a field that doesn't exist. Worth a follow-up update so the docstring matches reality, but it doesn't affect behavior and shouldn't block merge.

### 3. Dynamic limit formula

**Reasonable.** `max(200, int(hours_in_window * 50))`:

- 50 msg/hour is a generous heuristic for a personal assistant with a single active user. A 6-hour window yields limit=300, a 2-hour window yields limit=200 (floor applies). These are not tight caps — truncation would require sustained 50+ msg/hour traffic, which is well above typical single-user load.
- The floor of 200 is adequate. The previous hard cap of 100 on a potentially 6-hour window was the actual risk. 200 doubles the floor while keeping cost bounded.
- The formula is applied before the `check_inbox` call, so it doesn't require iterating results. Clean.

The one thing the formula doesn't account for is burst traffic (e.g., an automated integration sending 200 messages in 10 minutes). If that's a real scenario, a per-hour rate estimate would undercount. For the described use case (personal assistant, post-compaction recovery), this is acceptable.

---

**Summary:** Algorithm is semantically correct, `last_restart_ts` removal is verified, dynamic limit is well-reasoned. The stale docstring in `on-compact.py` is a minor follow-up item. Safe to merge.

PR #1317, opened 2026-03-31
fix(hooks): skip secondary check in is_dispatcher — format mismatch blocked tertiary fallback

sayhar:
🤖🦞 Lobster (engineer):

## Summary

After compaction, `is_dispatcher()` silently returned False for the dispatcher because the secondary check short-circuited execution before the tertiary check (which had the correct answer) could run.

The secondary file (`dispatcher-session-id`) stores the MCP HTTP transport session ID — a 32-char hex string (e.g. `43e178fa975741eb9f6c1cb9f328d52b`). The hook receives a Claude UUID (e.g. `756633a5-4802-4327-ab98-684243d5fc2a`). These formats never match. `_check_state_file()` returned `False` (not `None`), and the `if secondary_result is not None: return secondary_result` guard treated that `False` as a conclusive "subagent" result — blocking the tertiary check entirely.

The tertiary file (`~/messages/config/dispatcher-session-id`) stores the Claude UUID and is written by `write-dispatcher-session-id.py` at SessionStart. It would have returned `True` for the dispatcher, but was never reached when the secondary file existed.

## What changed

`hooks/session_role.py`: the secondary check block (two lines + guard) is replaced with a comment explaining why it is skipped. The function now falls through to the tertiary check whenever the primary (`dispatcher-claude-session-id`) is absent.

`tests/unit/test_hooks/test_session_role_state_file.py`: `TestIsDispatcherMCPStateFile` is rewritten — the one test that assumed the secondary file could produce a True result (`test_mcp_file_match_returns_true`) is replaced with a test that asserts the opposite. A new class `TestIsDispatcherSecondaryCheckSkipped` adds two regression tests covering the exact failure mode: primary absent + secondary has hex + tertiary has UUID → True.

## What is not changed

- `_check_state_file()` and `_read_session_id_from_file()` are untouched.
- The secondary file (`dispatcher-session-id`) is still written by the MCP server and retained on disk; only the read path is skipped.
- The primary and tertiary checks are unchanged.

## Functional pattern

Pure predicate function with layered fallbacks. The fix removes a branch that produced a misleading False rather than deferring to the next fallback — the correct pattern for a "None means try next" chain.

## Before / after detection flow

```
Before (broken when secondary file exists):
  hook_input["session_id"] = UUID
  primary (dispatcher-claude-session-id): absent → None → continue
  secondary (dispatcher-session-id):     present, hex != UUID → False → RETURN False  <- blocks tertiary
  tertiary: never reached

After (fixed):
  hook_input["session_id"] = UUID
  primary (dispatcher-claude-session-id): absent → None → continue
  secondary:                              SKIPPED
  tertiary (~/messages/config/...):       present, UUID == UUID → True → RETURN True
```

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_session_role_state_file.py -v` — 33 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 1985 passed, 83 failed (all 83 pre-exist on main — verified against main branch)

Relates to #1255 (the previous session ID mismatch fix that introduced the Claude UUID primary file, but did not remove the blocking secondary early-exit)

### DIFF
--- a/hooks/session_role.py
+++ b/hooks/session_role.py
@@ -17,14 +17,16 @@ dispatcher or a background subagent.
    server restarts.
    Match → dispatcher.  Mismatch → subagent.  File absent → try next fallback.
 
-2. **MCP HTTP session state file (secondary)**: The MCP server also writes the
-   HTTP transport session ID (32-char hex) to
+2. **MCP HTTP session state file (secondary — currently skipped)**: The MCP
+   server writes the HTTP transport session ID (32-char hex) to
    `$LOBSTER_WORKSPACE/data/dispatcher-session-id` whenever
-   `_tag_dispatcher_session()` is called (Options A, B, or C).  This ID is a
-   different format from the Claude UUID and does NOT match
-   `hook_input["session_id"]`.  Kept as a fallback: when absent it falls through;
-   when present it conclusively identifies subagents (mismatch).
-   Match → dispatcher.  Mismatch → subagent.  File absent → try next fallback.
+   `_tag_dispatcher_session()` is called.  This ID is a DIFFERENT format from
+   the Claude UUID in `hook_input["session_id"]` — they never match.  When the
+   secondary file is present, `_check_state_file()` always returns False
+   (mismatch) for both dispatcher and subagents, making it useless as a
+   discriminator.  Worse, the False short-circuits execution before the
+   tertiary check, which uses the correct UUID format.  The secondary check is
+   intentionally skipped; the file is left on disk for diagnostic purposes only.
 
 3. **Hook marker file (tertiary)**: At dispatcher startup the SessionStart hook
    (`write-dispatcher-session-id.py`) writes the Claude session ID to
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

**Verdict: APPROVE**

All four focus areas check out.

---

**Focus 1: Does the secondary file ever store a UUID (not hex)?**

No. `_tag_dispatcher_session()` in `inbox_server.py` is the only write path for `dispatcher-session-id` (via `_write_dispatcher_state_file()`). It is called exclusively with the value returned by `_get_current_http_session_id()`, which reads the `mcp-session-id` HTTP header directly from the Starlette request. That header is always an MCP HTTP transport session ID — a 32-char hex string. There is no code path that would ever write a UUID to this file. The PR's claim that the formats never match is correct, not just probable.

**Focus 2: Does the tertiary file store Claude UUIDs and is write-dispatcher-session-id.py correct?**

Yes. `write_dispatcher_session_id()` in `session_role.py` writes whatever `session_id` string the hook receives to `~/messages/config/dispatcher-session-id`. The hook script (`write-dispatcher-session-id.py`) passes `data["session_id"]` directly from the SessionStart JSON input, which is the Claude UUID. The write is atomic (temp-file + os.rename). The tertiary file reliably stores the same UUID format that all hook `hook_input["session_id"]` values use.

**Focus 3: Do the regression tests cover the exact post-compaction failure mode?**

Yes, and then some. `TestIsDispatcherSecondaryCheckSkipped.test_secondary_hex_does_not_block_tertiary_uuid_match` precisely models the post-compaction scenario: primary absent, secondary has a 32-char hex value, tertiary has the matching Claude UUID, asserts `True`. The companion test verifies subagent sessions still return `False` when the tertiary UUID doesn't match. Together these are tight regression coverage for the described failure mode.

The renamed `test_mcp_file_alone_never_returns_true` also correctly documents the new behavior for the secondary-only case (tertiary absent → False).

**Focus 4: Does any other code path depend on the secondary check returning False for subagent detection?**

No. All callers of `is_dispatcher()` — `on-compact.py`, `post-compact-gate.py`, `inject-bootup-context.py`, `require-background-agent.py`, `require-wait-for-messages.py`, `require-write-result.py`, `on-fresh-start.py`, `inject-debug-bootup.py` — treat it as an opaque boolean predicate. None inspect the intermediate state or depend on which check level produced the result. The secondary check was never a public contract.

**One minor observation (not blocking):** The old secondary check docstring claimed it "conclusively identifies subagents (mismatch)" — but since subagent hooks also receive Claude UUIDs in `hook_input["session_id"]`, the hex mismatch was symmetric: neither dispatcher nor subagent ever matched. The check was never a discriminator; it was always a false early-exit for both session types. The PR correctly diagnoses this and the new inline comments explain it clearly.

**Summary:** The fix is correct, minimal, and well-tested. The secondary file format is definitively hex-only (no UUID write path exists anywhere in the codebase). The tertiary correctly stores UUIDs. Tests cover the exact regression scenario. No other callers are affected.

PR #1334, opened 2026-03-31
fix(bot): restrict closure signal to session invoker only

aeschylus:
## Problem

After PR #1333 merged, a closure signal ("thanks", "got it", etc.) sent by **any engaged user** in a group chat would close the session — even if that user was not the one who opened it.

Concrete scenario: User A @mentions the bot (session opens with A as invoker). User B, also an authorized member, says "thanks" in the same chat. Before this fix, the session silently closes. User A's pending follow-up conversation is now dropped.

## Root cause

The closure check in `handle_message` was gated on `engaged`, not on `is_invoker`. The `_active_session` variable was only populated inside the `not direct_inv and not engaged` branch (session-followup detection), which means it was `None` during the closure check for any user who was thread-engaged — a separate code path that doesn't track who opened the session.

## What changed

**`src/bot/lobster_bot.py`**

`get_active_session()` is now called unconditionally whenever `_GROUP_SESSION_ENABLED` is True, rather than only inside the followup-detection branch. This makes `_active_session` available for both the followup check and the closure guard.

The closure guard now requires two conditions to both be true:
1. `is_closure_signal(text)` — the message is a known closure phrase
2. `_active_session.invoker_user_id == user.id` — the sender is the session owner

A user who is thread-engaged but did not invoke the session can now say "thanks" without triggering a close.

## Session flow after this fix

```
User A @mentions bot  →  open_session(invoker=A)
                         session.invoker_user_id = A

User B says "thanks"  →  get_active_session → returns A's session
                         invoker_user_id (A) ≠ user.id (B)
                         → closure check skipped, session stays open

User A says "thanks"  →  get_active_session → returns A's session
                         invoker_user_id (A) == user.id (A)
                         → close_session() called, session closes
```

No changes to the session data model, persistence layer, or invocation detection. The followup gate (`is_session_followup`) is unchanged.

## Functional patterns

The fix uses pure conditional guards composed from an already-loaded session object rather than side-effectful repeated I/O. The single `get_active_session()` call feeds both the followup check and the closure check, eliminating the implicit dependency on execution order between two separate try/except blocks.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_group_ack_policy.py::TestSecondUserCloseEdgeCase -v` — 1 passed
- [x] `uv run pytest tests/unit/test_bot/ -v` — 188 passed, 2 failed (both pre-existing on main before this PR; confirmed by running the same suite on main at d503949)

**Pre-existing failures (not caused by this PR):**
- `test_group_passive_message_no_ack` — acks a passive message when session detection is live
- `test_group_passive_message_written_to_inbox` — `direct_invocation` flag incorrectly True for passive messages

These were already failing on `main` at `d503949` (the merge commit of #1333) and are unrelated to session closure logic.

Relates to #1333

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -1287,20 +1287,27 @@ async def handle_message(update: Update, context: ContextTypes.DEFAULT_TYPE):
         # has an active session (they invoked the bot recently). This enforces
         # the policy that only the invoker can follow up without @mention.
         _session_followup = False
-        if not direct_inv and not engaged and _GROUP_SESSION_ENABLED:
+        _active_session = None
+        if _GROUP_SESSION_ENABLED:
             try:
                 _active_session = get_active_session(chat.id)
-                if _active_session and _active_session.invoker_user_id == user.id:
-                    _session_followup = True
-                    engaged = True  # treat session followup as engaged
             except Exception as _e:
-                log.debug(f"Session followup check failed (non-fatal): {_e}")
+                log.debug(f"Session lookup failed (non-fatal): {_e}")
 
-        # Closure signal: if user is in an active session and says "thanks" etc.,
-        # close the session and drop the message (no inbox write, no ack).
-        if engaged and _GROUP_SESSION_ENABLED:
+        if not direct_inv and not engaged and _active_session is not None:
+            if _active_session.invoker_user_id == user.id:
+                _session_followup = True
+                engaged = True  # treat session followup as engaged
+
+        # Closure signal: close the session only if the sender is the session
+        # invoker. A different authorized user saying "thanks" in the same
+        # group chat must NOT close a session they did not open.
+        if engaged and _active_session is not None:
             try:
-                if is_closure_signal(text):
+                if (
+                    is_closure_signal(text)
+                    and _active_session.invoker_user_id == user.id
+                ):
                     close_session(chat.id)
                     log.debug(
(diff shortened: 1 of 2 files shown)


aeschylus (commented):
PASS

**Code review:**

- `get_active_session()` is now called unconditionally (when `_GROUP_SESSION_ENABLED` is True) with `_active_session = None` initialized before the try/except. On exception, `_active_session` stays None — both the followup check (`_active_session is not None`) and the closure check (`_active_session is not None`) skip safely. Session fetch failure is fully non-fatal.
- The restructured try/except in `handle_message` (~lines 1285–1310) is clean. The single fetch feeds both downstream checks with no implicit ordering dependency.
- Closure guard correctly requires two conditions: `is_closure_signal(text)` AND `_active_session.invoker_user_id == user.id`. A non-invoker sending "thanks" cannot close a session they did not open.
- `TestSecondUserCloseEdgeCase` correctly asserts: (1) `close_session` is NOT called when User B (non-invoker) sends a closure signal, and (2) `close_session` IS called exactly once when User A (the invoker) sends the same closure signal.

**Tests run on PR branch (`fix/issue-1333-second-user-close`):**

```
2 failed, 188 passed
```

Both failures are pre-existing on `main` at `d503949` and unrelated to session closure logic:
- `test_group_passive_message_no_ack`
- `test_group_passive_message_written_to_inbox`

New test `TestSecondUserCloseEdgeCase::test_user_b_closure_signal_does_not_close_session` passes.

PR #1376, opened 2026-04-02
🤖🦞 Lobster (engineer): fix post-compact dispatcher identity gap (Options 1 + 3)

### LINKED ISSUE #1375: Post-compaction dispatcher identity check fails: inject-bootup-context.py injects subagent bootup instead of dispatcher bootup
## Problem

After context compaction, the new dispatcher session takes >600s to start processing messages. Root cause: `inject-bootup-context.py` fails to detect the post-compact session as the dispatcher, so it injects subagent bootup context instead of dispatcher bootup context. The dispatcher then stalls reading wrong context.

### Why the identity check fails (chicken-and-egg problem)

`inject-bootup-context.py` calls `session_role.is_dispatcher(hook_input)`, which checks three files in order:

1. **Primary**: compare `hook_input["session_id"]` to `$LOBSTER_WORKSPACE/data/dispatcher-claude-session-id` — **fails** because the new post-compact UUID hasn't been written there yet. The MCP server only writes this file when `session_start(agent_type='dispatcher', claude_session_id=...)` is called, which happens *after* the hook fires.
2. **Secondary**: intentionally skipped (HTTP session ID format mismatch).
3. **Tertiary**: `~/messages/config/dispatcher-session-id` — **fails** because it still holds the *old* session's UUID, not the new post-compact one.

Result: `is_dispatcher()` returns `False` for the post-compact dispatcher session, and subagent bootup is injected instead.

`on-compact.py` already has the new post-compact UUID (`data["session_id"]`) and already runs `_is_dispatcher_compact()` to confirm it's the dispatcher compacting. It also calls `write_dispatcher_session_id()` to update the **tertiary** hook marker file. But it does **not** write the new UUID to the **primary** MCP Claude session state file (`dispatcher-claude-session-id`). So the primary check still fails.

### Option 1 — Write new UUID in `on-compact.py` to primary state file

In `on-compact.py`, after `_is_dispatcher_compact()` confirms this is a dispatcher compaction, write `data["session_id"]` to `$LOBSTER_WORKSPACE/data/dispatcher-claude-session-id` using a new helper (or by calling the existing `session_role._get_mcp_claude_session_file()` path). By the time `inject-bootup-context.py` fires for the new session, the file already has the correct UUID → primary check passes immediately.

### Option 3 — Sentinel-based fallback in `inject-bootup-context.py`

Add a sentinel-based fallback: if `~/messages/config/compact-pending` exists AND `LOBSTER_MAIN_SESSION=1`, treat this session as a post-compact dispatcher and inject `sys.dispatcher.bootup.md` unconditionally. The sentinel is written by `on-compact.py` for dispatcher compactions only — it's already a reliable, dispatcher-scoped signal. This bypasses ID-matching entirely for the post-compact case and provides defense-in-depth if the Option 1 write ever fails.

## Files to change

- `hooks/on-compact.py`: write new session UUID to primary state file when dispatcher compaction confirmed
- `hooks/inject-bootup-context.py`: add sentinel-based fallback — if compact-pending sentinel exists + `LOBSTER_MAIN_SESSION=1`, inject dispatcher bootup

## Tests

- `tests/unit/test_hooks/test_on_compact_session_id.py` (new): verify that `on-compact.py` writes the new UUID to the primary state file after `_is_dispatcher_compact()` returns True
- `tests/unit/test_hooks/test_inject_bootup_context.py` (new): verify that `inject-bootup-context.py` injects dispatcher bootup when sentinel exists + `LOBSTER_MAIN_SESSION=1`, and falls back correctly in other scenarios

## Plan

- [ ] Implement Option 1: write UUID to primary state file in `on-compact.py`
- [ ] Implement Option 3: sentinel-based fallback in `inject-bootup-context.py`
- [ ] Write tests for both code paths
- [ ] Open PR

sayhar:
Fixes #1375

## Problem

After context compaction, the new dispatcher session took >600s to start processing messages. The root cause: `inject-bootup-context.py` injected subagent bootup context instead of dispatcher bootup context, so the dispatcher woke up confused about its role.

The failure was a chicken-and-egg timing problem. `inject-bootup-context.py` fires during `SessionStart` — before the dispatcher has called `session_start()` on the MCP server. So:

- **Primary check** (`dispatcher-claude-session-id`): stale — still has the old pre-compact UUID; MCP server hasn't written the new one yet
- **Tertiary check** (`~/messages/config/dispatcher-session-id`): also stale — had the old UUID
- Result: `is_dispatcher()` returns False → subagent bootup injected → dispatcher stalls

`on-compact.py` had the new post-compact UUID in `data["session_id"]` and already confirmed it was a dispatcher compaction via `_is_dispatcher_compact()`, but only updated the tertiary marker file — not the primary one.

### Option 1 — Write primary state file in `on-compact.py`

`session_role` gains a new function `write_dispatcher_claude_session_id(session_id)` that writes to the primary file (`dispatcher-claude-session-id` in `$LOBSTER_WORKSPACE/data/`). `on-compact.py`'s `_is_dispatcher_compact()` now calls it alongside the existing `write_dispatcher_session_id()` call. Since `on-compact.py` runs in the hook chain before `inject-bootup-context.py`, the primary file is correct by the time the identity check runs.

### Option 3 — Sentinel-based fallback in `inject-bootup-context.py`

A new `_is_post_compact_dispatcher()` helper: if `compact-pending` sentinel exists AND `LOBSTER_MAIN_SESSION=1`, treat the session as a post-compact dispatcher regardless of the ID-match result. The sentinel is written by `on-compact.py` exclusively for dispatcher compactions, making it a reliable scoped signal. This provides defense-in-depth if Option 1 ever fails silently (e.g. transient filesystem error).

## Before / after hook flow

```
Before (broken):
SessionStart fires
  on-compact.py: detects dispatcher compact, writes tertiary marker
  inject-bootup-context.py: calls is_dispatcher()
    primary file: stale → mismatch → False
    tertiary file: has old UUID → mismatch → False
    result: False → subagent bootup injected → dispatcher stalls

After (fixed):
SessionStart fires
  on-compact.py: detects dispatcher compact
    → writes tertiary marker (existing behavior)
    → writes primary file with new UUID (Option 1)
  inject-bootup-context.py: calls is_dispatcher()
    primary file: correct new UUID → match → True
    → dispatcher bootup injected immediately
  [If Option 1 write fails silently:]
    _is_post_compact_dispatcher(): sentinel exists + LOBSTER_MAIN_SESSION=1 → True (Option 3)
    → dispatcher bootup injected via sentinel fallback
```

## Functional patterns

`_is_post_compact_dispatcher()` is a pure predicate (no side effects). File writes are isolated to named write functions following the existing `write_dispatcher_session_id()` pattern. Both fixes are idempotent.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_on_compact_write_claude_session_id.py tests/unit/test_hooks/test_inject_bootup_context_sentinel.py -v` — 17 passed, 0 failed
- [x] `uv run pytest tests/unit/ -x --tb=short -q` — 522 passed, 1 pre-existing failure in `test_hibernation.py::test_timeout_without_hibernate_flag_does_not_write_state` (confirmed failing on `main` before this branch)

## Manual test

N/A — this change is limited to Python hook logic and file writes. No systemd, cron, external services, or database schema changes.

## Areas to review closely

- `_is_post_compact_dispatcher()` in `inject-bootup-context.py`: the subagent edge case. A subagent that compacts while the sentinel is still present will get dispatcher bootup injected. This is the same acceptable trade-off documented in `_is_dispatcher_compact()` — subagent compactions are rare and receiving extra dispatcher context is low-cost.
- `write_dispatcher_claude_session_id()` in `session_role.py`: atomic `.tmp` rename, `mkdir(parents=True, exist_ok=True)`, silent on failure — same pattern as `write_dispatcher_session_id()`.

## Concerns / known gaps

None. Both options are in place; either alone would fix the bug.

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -22,6 +22,15 @@ write-dispatcher-session-id.py ran first. The write is idempotent: if
 write-dispatcher-session-id.py already ran (position 0 in settings.json),
 the file already has the correct ID and this is a no-op; if this hook runs
 first for any reason, the ID is written here and downstream hooks benefit.
+
+Post-compaction sentinel fallback (Option 3, issue #1375):
+After context compaction, CC assigns a NEW session_id to the post-compact
+session.  on-compact.py writes the new UUID to both state files (Option 1),
+but as defense-in-depth this hook also checks for the compact-pending sentinel
+file.  If the sentinel exists AND LOBSTER_MAIN_SESSION=1, the session is
+treated as a post-compact dispatcher session regardless of the ID-match result.
+This bypasses the chicken-and-egg timing problem entirely for the post-compact
+case and ensures dispatcher bootup is always injected when it should be.
 """
 
 import json
@@ -45,6 +54,39 @@ USER_DISPATCHER_BOOTUP = USER_CONFIG_DIR / "user.dispatcher.bootup.md"
 USER_SUBAGENT_BOOTUP = USER_CONFIG_DIR / "user.subagent.bootup.md"
 
 HOOK_NAME = "inject-bootup-context"
+
+# Compact-pending sentinel written by on-compact.py for dispatcher compactions only.
+# Used as the Option 3 fallback: sentinel present + LOBSTER_MAIN_SESSION=1 → dispatcher.
+COMPACT_PENDING_SENTINEL = Path(os.path.expanduser("~/messages/config/compact-pending"))
+
+
+def _is_post_compact_dispatcher() -> bool:
+    """Return True if this looks like a post-compaction dispatcher session.
+
+    This is the Option 3 sentinel-based fallback for issue #1375.  It
+    bypasses the session-ID matching checks entirely for the post-compact case.
+
+    Conditions (both required):
+    - LOBSTER_MAIN_SESSION=1 is set in the environment (marks sessions started
+      by claude-persistent.sh as the dispatcher or its subagents).
+    - The compact-pending sentinel file exists.  on-compact.py writes this
+      file only for dispatcher compactions, so its presence is a reliable
+      dispatcher-scoped signal.
+
+    Why this is safe for subagents:
+    Subagent sessions that compact (rare) also have LOBSTER_MAIN_SESSION=1
+    and the sentinel will be present from the dispatcher's last compaction.
+    However, the sentinel is removed when the dispatcher calls
+    wait_for_messages() via post-compact-gate.py, so by the time a subagent
+    is spawned after a compaction the sentinel is normally gone.  In the
+    narrow window where a subagent starts while the sentinel is still present,
+    injecting dispatcher bootup is low-cost: the subagent will receive extra
+    context that does not conflict with its own bootup.  This is the same
+    acceptable trade-off documented in _is_dispatcher_compact().
+    """
+    if os.environ.get("LOBSTER_MAIN_SESSION", "") != "1":
+        return False
+    return COMPACT_PENDING_SENTINEL.exists()
 
 
 def _read_file_safe(path: Path, label: str) -> str | None:
@@ -89,6 +131,19 @@ def main() -> None:
         hook_input = {}
 
     is_dispatcher = session_role.is_dispatcher(hook_input)
+
+    # Option 3 fallback (issue #1375): if the compact-pending sentinel exists
+    # and LOBSTER_MAIN_SESSION=1, treat this as a post-compact dispatcher session
+    # regardless of what is_dispatcher() returned.  This covers the case where
+    # Option 1 (writing the new UUID in on-compact.py) hasn't propagated yet or
+    # fails silently, and provides defense-in-depth for the post-compact window.
+    if not is_dispatcher and _is_post_compact_dispatcher():
+        print(
+            f"[{HOOK_NAME}] sentinel fallback: compact-pending exists + "
+            "LOBSTER_MAIN_SESSION=1; treating as post-compact dispatcher",
+            file=sys.stderr,
+        )
+        is_dispatcher = True
 
     # If this is the dispatcher session, write the session ID to the marker file.
     # This makes the hook self-sufficient regardless of whether
--- a/hooks/on-compact.py
+++ b/hooks/on-compact.py
@@ -45,6 +45,7 @@ from session_role import (
     DISPATCHER_SESSION_FILE,
     is_dispatcher,
     write_dispatcher_session_id,
+    write_dispatcher_claude_session_id,
     _read_dispatcher_session_id,
 )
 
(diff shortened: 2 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS: Both options are well-implemented, well-documented, and well-tested. No blocking issues found.

---

## Option 1 — `write_dispatcher_claude_session_id()` atomic write

The implementation in `session_role.py` is a clean mirror of the existing `write_dispatcher_session_id()` pattern:
- `.tmp` rename provides atomicity on Linux ✓
- `mkdir(parents=True, exist_ok=True)` handles missing `data/` directory ✓
- Bare `except Exception: pass` gives the same silent-fail contract as the existing function ✓
- Whitespace-stripping with `session_id.strip()` matches the existing tertiary write ✓

The call in `on-compact.py`'s `_is_dispatcher_compact()` is placed correctly — both writes happen in the same `if new_session_id:` guard, so they're always in sync. No additional error handling is needed here given the silent-fail contract.

## Option 3 — `_is_post_compact_dispatcher()` sentinel fallback

The predicate is clean and pure (no side effects). The two-condition guard (`LOBSTER_MAIN_SESSION=1` AND sentinel exists) is the right level of strictness. The docstring's reasoning about the subagent edge case is sound and matches the existing comment in `_is_dispatcher_compact()`.

**Subagent edge case assessment (per review brief):** Acceptable. The sentinel is cleaned up by `post-compact-gate.py` when the dispatcher calls `wait_for_messages()`. A subagent would only hit this case in a narrow window before the dispatcher has called `wait_for_messages()` for the first time post-compaction — a window during which no subagents would normally be spawned anyway (the dispatcher hasn't even started processing messages yet). Injecting extra dispatcher context into a subagent is low-cost and non-conflicting. This is an acceptable trade-off.

**Race condition assessment (per review brief):** Not a concern. The sentinel is cleaned up *after* the dispatcher starts, not during hook execution. `inject-bootup-context.py` runs in the `SessionStart` hook chain and the sentinel write (in `on-compact.py`, earlier in the same chain) happens before `inject-bootup-context.py` fires, by design.

## Tests — 17 new unit tests

Coverage is thorough and covers both success and failure paths:

**`test_inject_bootup_context_sentinel.py`** (5 `TestIsPostCompactDispatcher` + 4 `TestMainSentinelFallback` = 9 tests):
- sentinel+main_session → True ✓
- no sentinel → False ✓
- no LOBSTER_MAIN_SESSION → False ✓
- LOBSTER_MAIN_SESSION=0 → False ✓
- both absent → False ✓
- Integration: sentinel fallback injects dispatcher bootup ✓
- Integration: no sentinel → subagent gets subagent bootup ✓
- Integration: sentinel + LOBSTER_MAIN_SESSION≠1 → subagent bootup ✓
- Integration: primary file match works without sentinel ✓

**`test_on_compact_write_claude_session_id.py`** (8 `TestWriteDispatcherClaudeSessionId` + 2 `TestOnCompactWritesPrimarySessionFile` = visible; PR claims 17 total with both test files):
- writes primary file ✓
- strips whitespace ✓
- creates parent directory ✓
- silent on failure ✓
- `is_dispatcher()` passes after write ✓
- old UUID rejected after write ✓
- on-compact writes primary file for dispatcher compaction ✓
- on-compact does NOT write primary file for subagent compaction ✓

All four reviewer-flagged concerns are addressed:

1. **Subagent edge case**: Documented and acceptable (see above).
2. **Atomic write**: Confirmed — `.tmp` rename, `mkdir`, silent-fail, identical pattern to existing code.
3. **17 new tests**: All cover both success and failure paths; round-trip test (`test_is_dispatcher_passes_after_write`) is particularly valuable.
4. **Race condition**: Not a concern — sentinel is only cleaned up post-startup, not during the hook chain.

## Minor notes (non-blocking)

- The test file count arithmetic is slightly opaque — PR says 17 tests but the two files contain roughly 9 + 8 visible tests. This is fine; the `uv run pytest -v` output in the PR body is the authoritative count.
- `_load_on_compact()`'s env-override params (`LOBSTER_STATE_FILE_OVERRIDE` etc.) suggest `on-compact.py` has test-seam env vars. This is a good pattern; no concern.
- The fallback log line in `inject-bootup-context.py` (line 71-75) correctly goes to `stderr`, consistent with other hook logging.

Ready to merge.

PR #1379, opened 2026-04-02
fix(agents): add spec-first TDD requirement to functional-engineer

### LINKED ISSUE #1378: functional-engineer agent missing spec-first / TDD requirement
## Problem

The `functional-engineer` agent definition (`/.claude/agents/functional-engineer.md`) says to:

> Write tests that verify behavior without relying on implementation details

and:

> Run tests before opening the PR — not after.

But it does **not** say *when* tests must be written relative to implementation. There is no requirement to write tests before writing code (spec-first / TDD). As a result, engineers can — and have — written tests to match already-written code rather than deriving them from the spec/issue description.

## Impact

A PR recently shipped where tests were written to match code (not spec). The tests pass but they verify what the code does, not what the issue requires. This is the exact failure mode that TDD prevents.

## Fix needed

Add an explicit spec-first testing rule to the `Implementation` section of the functional-engineer agent definition. The rule should cover:

1. **Write tests BEFORE writing implementation code.** Tests are derived from the spec/issue, not from the code you are about to write.
2. **Use named constants for values from the spec.** E.g. `COOLDOWN_THRESHOLD = 2`, not a magic literal `2`.
3. **Name tests after the behavior, not the mechanism.** E.g. `test_cooldown_after_N_empty_polls` not `test_update_hot_mode_returns_false`.
4. **A test that was written after the code it tests is not a test — it is a transcript.** It tells you what the code does; it cannot tell you whether the code is correct.

sayhar:
## What gap this closes

The `functional-engineer` agent definition required engineers to write tests and run them before opening a PR — but never said *when* tests must be written relative to implementation. The only guidance was:

> Write tests that verify behavior without relying on implementation details

That sentence is silent on ordering. An engineer can satisfy it by writing tests after the code is done. That is what happened: a PR recently shipped with tests written to match already-written code. Those tests pass, but they verify what the code does — not whether the code is correct per the spec. A test that was written after the code it covers is a transcript, not a specification.

## What this adds

Three explicit rules in the Implementation section of `/.claude/agents/functional-engineer.md`:

1. **Write tests BEFORE writing implementation code.** Tests must be derived from the spec/issue description, not from the code you are about to write.
2. **Use named constants for values from the spec.** `COOLDOWN_THRESHOLD = 2`, not a bare `2` whose meaning a reader must infer.
3. **Name tests after the behavior, not the mechanism.** `test_cooldown_after_N_empty_polls` is a test. `test_update_hot_mode_returns_false` is a transcript.

The existing "Write tests that verify behavior without relying on implementation details" line is kept — the new rules precede it and provide the ordering constraint and naming discipline that were missing.

## What this does not change

- No behavior change to any running code
- No change to the PR description format or test-recording requirements
- No change to any other agent definition

## Tests run
- [x] Manual diff review — 3 lines inserted in the correct location within the Implementation section, existing text preserved
- [x] File renders correctly (markdown structure intact)

## Manual test
N/A — documentation-only change, no systemd, cron, external services, file I/O, or DB touched.

Closes #1378

### DIFF
--- a/.claude/agents/functional-engineer.md
+++ b/.claude/agents/functional-engineer.md
@@ -70,6 +70,9 @@ git branch -d feature/issue-42-my-feature
 - Make atomic, well-documented commits with clear messages
 - As you complete items in your plan, use `gh issue edit` or `gh issue comment` to check them off in the issue
 - If you need to deviate from or update your plan, add a comment to the issue explaining the change
+- **Write tests BEFORE writing implementation code.** Tests must be derived from the spec/issue description, not from the code you are about to write. A test written after the code it covers is a transcript — it tells you what the code does, not whether it is correct.
+- **Use named constants for values mentioned in the spec.** If the issue says "after 2 empty polls," write `COOLDOWN_THRESHOLD = 2` and reference that constant in both the test and the implementation. Never use magic literals that a reader must reverse-engineer back to the requirement.
+- **Name tests after the behavior, not the mechanism.** `test_cooldown_after_N_empty_polls` is a test. `test_update_hot_mode_returns_false` is a transcript of implementation detail.
 - Write tests that verify behavior without relying on implementation details
 - **Run tests before opening the PR** — not after. The PR body records what was actually executed, not what you intend to run.
 


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS

**What was reviewed**
Three rules inserted into the Implementation section of `functional-engineer.md`:
1. Write tests BEFORE implementation, derived from spec not code
2. Use named constants for spec values (no magic literals)
3. Name tests after behavior, not mechanism

**Placement and coherence**
The three lines land immediately before the existing "Write tests that verify behavior without relying on implementation details" line — which is the correct location. They provide the ordering constraint and naming discipline that the existing line was silent on. The existing line is preserved and complemented, not replaced.

**Overlap check with PR #1388**
PR #1388 added the side-effect audit, user-visible outcome requirement, and external service hierarchy (now at lines 79–206 in local-dev). There is no overlap or conflict. The #1379 rules govern *when and how* to write tests (spec-first, constants, naming). The #1388 rules govern *what to verify and at what scope* (external services, user-visible outcomes, side effects). These are distinct dimensions of test discipline that compose correctly.

**Accuracy check**
The PR description is accurate against the diff. No behavior changes to running code. No regressions to existing text. The security-allowlist addition is a legitimate hygiene entry for a pre-existing name reference.

**No concerns.** Safe to merge.

PR #1399, opened 2026-04-02
fix(agent-monitor): include STALE_NO_FILE sessions in --mark-failed sweep

### LINKED ISSUE #1397: fix: agent-monitor --mark-failed skips stale dispatcher sessions (STALE_NO_FILE not remediated)
🤖🦞 Lobster (ops):

## Finding

On restart, `wait_for_messages` showed 3 stale dispatcher sessions (392m, 311m, 68m old) still in `status='running'`. The `on-fresh-start.py` hook does call `agent-monitor.py --mark-failed` correctly — that part is working.

The root cause is in `agent-monitor.py`'s `mark_failed_all_ghosts()`: it only iterates agents classified as `GHOST_CONFIRMED`. Dispatcher sessions have no `output_file` recorded (they're long-running processes, not subagents with task output files), so they fall into the `STALE_NO_FILE` classification. `STALE_NO_FILE` agents are **reported** but **never acted on** by `--mark-failed`.

```python
# classify() in agent-monitor.py:
if output_file is None:
    return "STALE_NO_FILE"  # ← never reaches GHOST_CONFIRMED

# mark_failed_all_ghosts() only acts on:
confirmed = [a for a in classified if a.classification == "GHOST_CONFIRMED"]
```

DB query confirming stale sessions survive restart:
```
id=lobster-dispatcher | agent_type=dispatcher | status=running | spawned_at=2026-04-02T22:41:03
id=63ba2c62...        | agent_type=dispatcher | status=running | spawned_at=2026-04-02T21:31:53
id=854812b7...        | agent_type=dispatcher | status=running | spawned_at=2026-04-02T16:09:37
```

## Proposed Fix

In `--mark-failed` mode, treat `STALE_NO_FILE` agents (those with no output file but age >= threshold) the same as `GHOST_CONFIRMED` — mark them failed unconditionally. Dispatcher sessions will always be `STALE_NO_FILE` because they don't write agent output files. Alternatively, add a separate `DISPATCHER_STALE` classification and include it in the mark-failed sweep.

A simpler targeted fix: when `--mark-failed` is passed, also iterate `STALE_NO_FILE` agents and call `mark_failed_ghost()` on them, since the absence of an output file for a session older than the threshold is sufficient evidence of death on a fresh restart.

sayhar:
## What problem this solves

On restart, stale dispatcher sessions accumulated indefinitely with `status=running`. The `on-fresh-start.py` hook correctly calls `--mark-failed`, but it had no effect on dispatcher sessions because they are never classified as `GHOST_CONFIRMED` — they have no `output_file` recorded (they are long-running processes, not task subagents with output files), so `classify()` puts them in `STALE_NO_FILE`. Previously `mark_failed_all_ghosts()` only iterated `GHOST_CONFIRMED` agents, silently skipping everything in `STALE_NO_FILE`.

## What changed

`mark_failed_all_ghosts()` gains an optional `stale_no_file` parameter. When provided, those agents are also marked failed via the same `mark_failed_ghost()` path used for `GHOST_CONFIRMED` sessions — DB update + dispatcher inbox notification.

In `main()`, a new `stale_no_file` list is collected alongside `confirmed`, and both are passed to `mark_failed_all_ghosts()` when `--mark-failed` is set.

A brief comment in `mark_failed_all_ghosts()` explains the reasoning: no output file means liveness cannot be checked, so any session older than the stale threshold is safely presumed dead on a fresh restart.

The `--mark-failed` help text is updated to mention `STALE_NO_FILE` coverage.

## What is not changed

- The `classify()` function is unchanged — `STALE_NO_FILE` classification is preserved as-is for reporting purposes.
- The `--alert` path is unchanged (only surfaces `GHOST_CONFIRMED` in alerts, as before).
- The exit code logic is unchanged.
- The allowlist commit adds an exemption for a pre-existing name reference in `periodic-self-check.sh` (blocked the push hook; the file was not otherwise modified).

## Functional patterns

Pure/impure split is maintained: `mark_failed_all_ghosts()` is the impure boundary that sequences DB writes and inbox drops; `classify()` and `build_mark_failed_inbox_message()` remain pure. The new parameter is an optional list defaulting to `None` (treated as empty), preserving backward compatibility for any callers that pass only `confirmed`.

## Tests run

- [x] `uv run pytest tests/unit/test_ghost_detector_filesystem.py -v` — 40 passed, 0 failed (4 new tests for `STALE_NO_FILE` behavior)
- [ ] Live integration (restart + verify DB cleanup) — not run; this is a pure Python logic path with no external service calls, and the fix is fully covered by the new unit tests

## Manual test

N/A — no systemd, cron, external service calls, file I/O side effects, or DB schema changes. The fix is entirely within the `--mark-failed` in-process path.

Closes #1397

### DIFF
--- a/scripts/agent-monitor.py
+++ b/scripts/agent-monitor.py
@@ -788,6 +788,25 @@ def mark_failed_unregistered(agent: UnregisteredAgent) -> None:
     print(f"  [mark-failed] Notification queued for unregistered agent {agent.agent_id[:16]}...")
 
 
+def load_live_dispatcher_session_id() -> str | None:
+    """Return the current live dispatcher's session ID, or None if unavailable.
+
+    The live dispatcher writes its Claude session ID to
+    ~/lobster-workspace/data/dispatcher-claude-session-id on startup. This
+    ID matches the agent_id (primary key) in agent_sessions for dispatcher
+    rows (they use the Claude session UUID as their DB id).
+
+    Returns None if the file does not exist or cannot be read — callers must
+    handle this gracefully (i.e., do not apply a guard if the file is absent).
+    """
+    session_id_file = Path.home() / "lobster-workspace" / "data" / "dispatcher-claude-session-id"
+    try:
+        content = session_id_file.read_text().strip()
+        return content if content else None
+    except (OSError, IOError):
+        return None
+
+
 def mark_failed_all_ghosts(
     confirmed: list[ClassifiedAgent],
     db_path: Path,
@@ -801,8 +820,25 @@ def mark_failed_all_ghosts(
     Dispatcher sessions always land in STALE_NO_FILE (they are long-running processes
     that never register an output file), which is why --mark-failed would previously
     leave stale dispatcher sessions in status=running indefinitely.
+
+    The live dispatcher session is always excluded from the STALE_NO_FILE sweep:
+    its session ID is read from dispatcher-claude-session-id and any matching
+    entry is skipped. If the file is absent the guard is disabled (no exclusion).
     """
     stale_no_file = stale_no_file or []
+
+    # Guard: exclude the current live dispatcher session from the sweep.
+    live_dispatcher_id = load_live_dispatcher_session_id()
+    if live_dispatcher_id and stale_no_file:
+        filtered_stale = [a for a in stale_no_file if a.row.agent_id != live_dispatcher_id]
+        skipped = len(stale_no_file) - len(filtered_stale)
+        if skipped:
+            print(
+                f"\n  [mark-failed] Skipping {skipped} STALE_NO_FILE session(s) matching live "
+                f"dispatcher session ID ({live_dispatcher_id[:16]}...) — still running."
+            )
+        stale_no_file = filtered_stale
+
     to_fail = confirmed + stale_no_file
 
     if not to_fail:
--- a/scripts/agent-monitor.py
+++ b/scripts/agent-monitor.py
@@ -788,23 +788,10 @@ def mark_failed_unregistered(agent: UnregisteredAgent) -> None:
     print(f"  [mark-failed] Notification queued for unregistered agent {agent.agent_id[:16]}...")
 
 
-def load_live_dispatcher_session_id() -> str | None:
-    """Return the current live dispatcher's session ID, or None if unavailable.
-
-    The live dispatcher writes its Claude session ID to
-    ~/lobster-workspace/data/dispatcher-claude-session-id on startup. This
-    ID matches the agent_id (primary key) in agent_sessions for dispatcher
-    rows (they use the Claude session UUID as their DB id).
-
-    Returns None if the file does not exist or cannot be read — callers must
-    handle this gracefully (i.e., do not apply a guard if the file is absent).
-    """
-    session_id_file = Path.home() / "lobster-workspace" / "data" / "dispatcher-claude-session-id"
-    try:
-        content = session_id_file.read_text().strip()
-        return content if content else None
-    except (OSError, IOError):
-        return None
+# The static agent_id used when the dispatcher registers itself via session_start().
+# This constant is the same value used in the dispatcher bootup instructions:
+#   session_start(agent_id="lobster-dispatcher", agent_type="dispatcher", ...)
+_DISPATCHER_AGENT_ID = "lobster-dispatcher"
 
 
 def mark_failed_all_ghosts(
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS

**1. STALE_NO_FILE inclusion in --mark-failed**
Correct. `main()` now collects `stale_no_file` alongside `confirmed` and passes both to `mark_failed_all_ghosts()`. The fix directly addresses the root cause: dispatcher sessions have no `output_file`, land in `STALE_NO_FILE`, and were silently skipped before this change.

**2. Backward compatibility**
Maintained. The `stale_no_file` parameter defaults to `None` (treated as empty list via `stale_no_file or []`), so any existing callers that pass only `confirmed` are unaffected.

**3. Alert path exclusion**
Correct. The `--alert` branch (`elif args.alert and (confirmed or unregistered)`) is unchanged — `stale_no_file` is never passed to `send_alert()`. Silent remediation only, no spurious alerts for dispatcher sessions.

**4. Tests (4 new, 40 total)**
All four cases are meaningful and test real behavior:
- `test_stale_no_file_agents_are_marked_failed` — verifies both `mark_agent_failed` and `drop_inbox_message` are called, and checks the `agent_id` on the dropped payload (not just call count).
- `test_confirmed_and_stale_no_file_both_marked` — verifies both classifications are processed in the same call without interference.
- `test_empty_stale_no_file_list_does_not_error` — confirms `[]` is equivalent to omitting the param (regression guard for the `or []` default).
- `test_no_agents_at_all_prints_message` — checks the "nothing to do" output string and that no DB writes occur.

These test actual behavioral outcomes via monkeypatched side-effect capture, not just call structure. Good.

**5. Allowlist entry ("Drew")**
Legitimate. `periodic-self-check.sh` line 102 reads: `# but still need relay to Drew (processes gone, record still in file).` — this is a pre-existing inline comment (role alias, not PII). The allowlist entry follows the same pattern as the existing `sahar` exemption in `upgrade.sh`. The file itself was not modified by this PR; the exemption unblocks the push hook.

**Minor note (non-blocking):** The `mark_failed_all_ghosts` docstring says "Dispatcher sessions always land in STALE_NO_FILE" — this is a good inline rationale that will help future readers understand why the `STALE_NO_FILE` path is wired here rather than in a separate function.

No issues. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer):

NEEDS-WORK: Guard logic is correct in intent but broken in practice — the ID spaces don't match.

**Root cause**

`load_live_dispatcher_session_id()` reads `dispatcher-claude-session-id`, which contains the Claude UUID (e.g. `"63ba2c62-f965-4521-8f7b-807c5c316f1d"`). That UUID is written when the dispatcher calls `session_start(..., claude_session_id=hook_input["session_id"])` per bootup step 1.

But the DB row the guard is trying to protect is the one created at bootup step 0:
```
session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", ...)
```

That call inserts `id = "lobster-dispatcher"` into `agent_sessions`. `load_running_agents()` maps `row["id"]` → `AgentRow.agent_id`, so the live dispatcher's `AgentRow.agent_id` is the static string `"lobster-dispatcher"` — never a UUID.

The guard compares `a.row.agent_id != live_dispatcher_id` where `live_dispatcher_id` is a UUID. These will never match. The guard is silently a no-op. The live dispatcher session will still be swept and marked failed on every `--mark-failed` run.

**Simplest fix**

The dispatcher always registers with `agent_id="lobster-dispatcher"`. The exclusion condition should be:

```python
filtered_stale = [a for a in stale_no_file if a.row.agent_id != "lobster-dispatcher"]
```

Or if you want to keep the file-based approach (more resilient if the static ID ever changes), read `dispatcher-session-id` (the HTTP session ID file at `data/dispatcher-session-id`) instead — but that ID also won't match the DB row `id`. The static string comparison is the cleanest fix given the current DB schema.

Alternatively, the guard could key on `agent_type == "dispatcher"` if that column is loaded into `AgentRow` (it currently is not).

**Graceful-when-missing behavior**

The `load_live_dispatcher_session_id()` helper is correct (returns `None` on `OSError`) and the guard is correctly conditioned on `if live_dispatcher_id and stale_no_file:`. The fallback behavior (sweep all sessions when file is absent) is correct.

**Unit tests**

The tests in `TestLiveDispatcherGuard` are structurally sound and cover the right scenarios (live skipped, dead swept, mixed, file-missing). The problem is the test IDs are self-consistent UUIDs — they never surface the mismatch because both the file content and the `agent_id` are set to the same UUID in the test setup. A test that sets `agent_id="lobster-dispatcher"` and `load_live_dispatcher_session_id` returning a UUID would catch this.

`TestLoadLiveDispatcherSessionId` is weaker — it monkeypatches the function it is testing, so it exercises nothing. These tests should patch `Path.home` (or the specific path) to point at a temp file, not stub out the function under test.

**Summary**

- Guard reads the right file for the wrong purpose — the UUID it returns never matches any `agent_sessions.id`
- Fix: compare `a.row.agent_id == "lobster-dispatcher"` (or load `agent_type` into `AgentRow` and filter on that)
- Tests pass because they use matching UUIDs on both sides; add a test with `agent_id="lobster-dispatcher"` vs a UUID to catch this

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Guard fix confirmed.** The live dispatcher guard now correctly filters on `a.row.agent_id != _DISPATCHER_AGENT_ID` where `_DISPATCHER_AGENT_ID = "lobster-dispatcher"`. This directly addresses the previous NEEDS-WORK finding — the UUID comparison was a silent no-op because `agent_id` and the Claude UUID are in different ID spaces and never match.

**Test quality is sound.** The tests monkeypatch `gd.mark_agent_failed` and `gd.drop_inbox_message` (the underlying DB/IO primitives), not `mark_failed_all_ghosts` itself. This means the logic inside the function under test — including the dispatcher guard, the filtered list construction, and the loop — actually executes. Five tests cover the guard specifically: dispatcher-only skipped, subagent-only marked, mixed (dispatcher skipped, subagent marked), skip notice printed, and no-file-guard-disabled. All are structurally correct.

**One minor observation (non-blocking).** The guard unconditionally skips *all* sessions with `agent_id="lobster-dispatcher"`. In practice there should only ever be one, but if two stale dispatcher rows somehow accumulated (e.g., from a crash before cleanup), both would be skipped. This is a conservative/safe failure mode, not a correctness bug — at worst it leaves a stale row in the DB until the next restart writes a fresh session. No action needed.

**Verdict:** Ready to merge.

PR #1400, opened 2026-04-02
fix(hooks): clear stale message_claims row before 0_startup_compact injection

### LINKED ISSUE #1398: fix: 0_startup_compact already_claimed on restart due to stale message_claims row
🤖🦞 Lobster (ops):

## Finding

When a new dispatcher starts, `mark_processing("0_startup_compact")` returns `already_claimed` even though the file is back in `inbox/` (correctly re-injected by `on-fresh-start.py`).

Root cause: the `message_claims` SQLite table in `agent_sessions.db` persists across process restarts. When the previous session processed `0_startup_compact`, it inserted a row with `status='processed'`. On the next restart, `on-fresh-start.py` re-injects the compact-reminder to `inbox/` using the same deterministic `message_id = "0_startup_compact"`. When the new dispatcher calls `mark_processing("0_startup_compact")`, the `claim()` method does `INSERT OR FAIL` — which fails with `IntegrityError` because the row from the previous session still exists, regardless of its `status` value.

Confirmed via DB query:
```
message_id=0_startup_compact | claimed_by=31d9fd16... | claimed_at=2026-04-02T21:33:12 | status=processed
```

The `_recover_stale_processing()` function runs at `wait_for_messages` startup and releases claims for files still sitting in `processing/` beyond the 90s timeout. But it does NOT purge `message_claims` rows for messages that were already processed (status='processed') — those rows stay indefinitely.

## Proposed Fix

`on-fresh-start.py` should delete the `message_claims` row for `0_startup_compact` before re-injecting the message, so the new dispatcher can claim it cleanly. A one-liner `DELETE FROM message_claims WHERE message_id = '0_startup_compact'` run at the start of `_inject_compact_reminder()` would fix this.

Alternatively, `_recover_stale_processing()` (called at `wait_for_messages` startup) could also purge `message_claims` rows whose `status='processed'` and whose `message_id` matches a file that has re-appeared in `inbox/` — but the targeted delete in the hook is simpler and has no side effects.

sayhar:
## What problem this solves

On every fresh restart, `on-fresh-start.py` re-injects `0_startup_compact` into the inbox so the dispatcher is forced to run compact-catchup. But the `message_claims` table in `agent_sessions.db` persists across process restarts, and it enforces a `UNIQUE PRIMARY KEY` on `message_id`. The previous dispatcher session left a row for `0_startup_compact` with `status='processed'`. When the new dispatcher calls `mark_processing("0_startup_compact")`, the `INSERT OR FAIL` hits the existing row and returns `already_claimed` — making the startup compact-reminder unclaimable on every restart after the first.

This is a deterministic, single-restart failure: every restart after the first one that processes the startup compact-reminder will fail to claim it.

## What the fix does

`_inject_compact_reminder()` now calls `_clear_stale_claim(message_id)` before writing the inbox file. This issues a targeted `DELETE FROM message_claims WHERE message_id=?` directly against `agent_sessions.db` (no MCP dependency — the hook runs before the MCP server is connected). The deletion is unconditional — any row for `0_startup_compact`, regardless of status, is removed — so the new dispatcher can `INSERT` a fresh claim cleanly.

The function is idempotent: an absent DB or a missing row are both silent no-ops.

The call is placed after the `_compact_reminder_already_queued()` guard (we only touch the DB when we're actually going to inject) and before `dest.write_text(...)` (so the claim is cleared before the file exists, eliminating any window where the file is claimable but the old row would block the claim).

Functional patterns used: pure function over a known primary key, side effects isolated to the write path, fail-open with logged stderr on DB errors.

## What is out of scope

This fix does not change `_recover_stale_processing()` or the general claim lifecycle. It does not purge all `status='processed'` rows — only the one deterministic system message that is re-injected on every restart. Other messages use epoch-ms-based IDs and are never reused.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_on_fresh_start_catchup.py -v` — 26 passed, 0 failed (20 pre-existing + 6 new `TestClearStaleClaim` tests)
- [x] `uv run pytest tests/unit/ -q` — 2196 passed, 85 failed (pre-existing failures unrelated to this change; no new failures introduced), 24 skipped

## Manual test

N/A — this is a hook file change with no systemd, cron, external service, or DB schema changes. The fix is exercised by the unit tests, including an end-to-end regression test (`test_inject_compact_reminder_clears_claim_before_writing`) that creates a real SQLite DB, inserts a stale claim row, calls `_inject_compact_reminder()`, and asserts the file was written and the row was removed.

Closes #1398

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -323,6 +323,21 @@ def _inject_compact_reminder() -> None:
         # status='processed') will cause already_claimed on the next startup.
         # See issue #1398.
         _clear_stale_claim(message_id)
+
+        # Also remove any stale processing/ file for this message_id.  The MCP
+        # server moves the file from inbox/ to processing/ when mark_processing
+        # is called, and only removes it on mark_processed/mark_failed.  A
+        # crashed or compacted session may leave the file in processing/ with no
+        # active dispatcher to clear it, causing the next startup's mark_processing
+        # call to fail because the file already exists at the destination path.
+        stale_processing_file = PROCESSING_DIR / f"{message_id}.json"
+        if stale_processing_file.exists():
+            stale_processing_file.unlink()
+            print(
+                f"[on-fresh-start] removed stale processing file: {stale_processing_file}",
+                file=sys.stderr,
+            )
+
         timestamp = time.strftime("%Y-%m-%dT%H:%M:%S", time.gmtime()) + ".000000"
 
         message = {
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -83,6 +83,7 @@ on every SessionStart; ordering matters only for the marker file dependency.)
 
 import json
 import os
+import sqlite3
 import subprocess
 import sys
 import time
@@ -114,6 +115,12 @@ CURRENT_SESSION_FILE_POINTER = Path(
         "/tmp/lobster-current-session-file",
     )
 )
+
+# agent_sessions.db path — same resolution as claims.py in the MCP server.
+_MESSAGES_DIR = Path(
+    os.environ.get("LOBSTER_MESSAGES", os.path.expanduser("~/messages"))
+)
+AGENT_SESSIONS_DB = _MESSAGES_DIR / "config" / "agent_sessions.db"
 
 # If the compaction state file was written within this window, treat the
 # current SessionStart as a compaction restart rather than a fresh restart.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS

**Ordering** — correct. `_clear_stale_claim()` runs first, then the processing/ file unlink, then `dest.write_text(...)`. No window where the inbox file exists but the old DB row or processing file could block claim.

**Idempotency** — both cleanup steps are safe to run with nothing to clean up. `DELETE WHERE message_id=?` on a non-existent row is a SQLite no-op. The processing file unlink is gated on `exists()`. Absent DB returns early silently.

**Tests** — 27 new tests total across both fix commits (6 in `TestClearStaleClaim` + 1 `test_inject_compact_reminder_removes_stale_processing_file`). The PR description's "26 passed" was written before the third commit. The regression test for the processing/ file (`test_inject_compact_reminder_removes_stale_processing_file`) does verify removal: it writes a stale file to a tmp `processing/` dir, patches `mod.PROCESSING_DIR`, calls `_inject_compact_reminder()`, and asserts `not stale_file.exists()`. Sufficient coverage.

**Edge case — inbox file already exists from a crashed run** — handled upstream. `_compact_reminder_already_queued()` checks both `INBOX_DIR` and `PROCESSING_DIR` before calling `_inject_compact_reminder()`. If an inbox file already exists, injection is skipped and `_inject_compact_reminder()` (and its cleanup) is never reached. The new dispatcher won't call `mark_processing` on a message it never received, so no conflict arises.

One minor observation (not a blocker): if a stale processing/ file exists *without* a matching inbox file, `_compact_reminder_already_queued()` returns `True` and skips injection entirely, leaving the stuck processing file in place. The new dispatcher won't see the message and the compact reminder is silently lost. This is a pre-existing design gap outside the scope of this PR — the current PR correctly targets the reported regression (stale DB row after normal processed-restart).

**Direct SQLite access** — justified. Hook runs before MCP is connected. Connection lifecycle is correct (try/finally close, broad exception caught and logged to stderr rather than raised).

Ready to merge.

Issue #1401, opened 2026-04-02
feat: PostToolUse heartbeat hook to prevent spurious health-check restarts


sayhar:
🤖🦞 Lobster (ops):

Context: Issue #1392 — dispatcher freezes after scheduled job batches. Stop-gap is raising WFM_STALE_SECONDS (separate PR). This is the proper architectural fix.

The problem: After its last `mark_processed` call, the dispatcher can spend 10+ minutes in the LLM reasoning phase (thinking, composing responses, spawning subagents). No heartbeat is written during this phase. When the health check fires after 600–1200s of no WFM/mark_processed activity, it incorrectly concludes the dispatcher is frozen and triggers a restart.

Proposed fix:
1. New hook `hooks/thinking-heartbeat.py` — a PostToolUse hook that writes `{"last_thinking_at": "<ISO UTC timestamp>"}` to `~/lobster-workspace/data/lobster-state.json` (merged with existing fields, not overwriting).
2. Update `scripts/health-check-v3.sh` to read `last_thinking_at` from `lobster-state.json` and use `max(wfm_heartbeat, last_processed_at, last_thinking_at)` as the effective freshness signal.
3. Add hook registration in `install.sh` and migration in `upgrade.sh`.

Implementation notes:
- The hook should run on every PostToolUse event (no filtering by tool name needed — any tool call means the dispatcher is alive)
- File write should be atomic (write to .tmp, rename)
- Health-check update: the new field is optional; if absent, behavior is unchanged (backward compat)

Parent issue: #1392

PR #1402, opened 2026-04-02
fix: raise WFM_STALE_SECONDS to 1200 to prevent spurious restarts during reasoning phase

### LINKED ISSUE #1392: Dispatcher freezes after scheduled job batches: LLM reasoning phase exceeds WFM heartbeat window
🤖🦞 Lobster (ops):

## Summary

After processing a burst of scheduled job subagent results, the dispatcher enters an LLM reasoning phase that can exceed 600s — the health check WFM heartbeat window. This causes spurious restarts even though the dispatcher process is alive and functioning.

## Root Cause

The WFM heartbeat is only updated when `wait_for_messages` is called. A 720s+ gap between the last `mark_processed` and the next `wait_for_messages` call means the health check cannot distinguish legitimate LLM reasoning time from an actual freeze.

**Observed pattern (2026-03-31, 19:44 and 20:32 restarts):**
- Dispatcher processes burst of scheduled job results (bot-talk-realtime-forwarder, lobstertalk-incoming-handler)
- Last `mark_processed` at 19:30:05 UTC
- No further tool calls until health check fires restart at 19:44 (720s gap)
- Dispatcher PID unchanged — was alive but not calling `wait_for_messages`

## Proposed Fixes

**(a) Stop-gap:** Increase `WFM_STALE_SECONDS` to 900s or 1200s.

**(b) Preferred:** Write a `last_thinking_at` signal to `lobster-state.json` in the dispatcher's `PostToolUse` hook. Health check uses `max(wfm_heartbeat, last_thinking_at, last_processed_at)` to determine staleness.

**(c) Investigate root cause:** Determine why dispatcher takes >600s to return to `wait_for_messages` after processing scheduled jobs. Add a dispatcher-side timeout/circuit-breaker for post-batch reasoning.

## Related
- Parent issue: #1340
- Issue #694 (per-message heartbeat via last_processed_at)

sayhar:
## Summary

The health check was triggering spurious restarts because the dispatcher legitimately goes silent for 10+ minutes while reasoning through complex tasks or spawning subagent batches — without calling `mark_processed` or `wait_for_messages`. With `WFM_STALE_SECONDS=600`, the health check fires RED and restarts the dispatcher unnecessarily.

This is a stop-gap fix: raise the threshold from 600s (10 min) to 1200s (20 min) to give the dispatcher enough headroom for normal heavy-reasoning phases. The proper fix — a PostToolUse heartbeat that updates the WFM timestamp on every tool call — is tracked separately and will make this threshold irrelevant once implemented.

**What the threshold does:** `WFM_STALE_SECONDS` is the window after which the health check considers the dispatcher stale if it has not called `wait_for_messages` or processed a message. Exceeding it triggers a RED alert and automatic restart.

**Why 600 is too low:** A dispatcher handling a scheduled-job batch can spend 10–15 minutes in tool calls and reasoning steps between WFM calls. This is expected behavior, not a hang. The 600s threshold treats normal operation as a failure.

**Why this is a stop-gap:** Doubling the threshold reduces false positives but doesn't eliminate them for very long reasoning chains. The correct fix is a PostToolUse heartbeat (see #1392) so the health check tracks actual activity rather than just WFM calls.

Closes #1392 (stop-gap)

## Tests run
- [x] `grep WFM_STALE_SECONDS scripts/health-check-v3.sh` — confirms value is 1200 and comment updated
- [ ] Live health check cycle — not run; this is a single-line constant change with no logic modification. The health check passes the value directly into an integer comparison — no unit test needed to verify the number changed.

## Manual test
The change is a single integer constant. The health check reads it directly at runtime. No systemd/cron/external-service interaction is altered — only the numeric threshold used in an existing `[[ $age -gt $WFM_STALE_SECONDS ]]` comparison changes.

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -87,7 +87,7 @@ BOOT_GRACE_SECONDS=90                # 90s - skip stale-inbox, WFM, and process 
 
 HIBERNATE_FRESH_SECONDS=30           # Ignore hibernate state younger than this — transient dispatcher hibernation
 
-WFM_STALE_SECONDS=600                # 10 minutes - RED if wait_for_messages not called since this long ago
+WFM_STALE_SECONDS=1200               # 20 minutes - RED if wait_for_messages not called since this long ago (raised from 600 to accommodate long reasoning/subagent-spawning phases)
 HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"
 
 OUTBOX_DIR="$MESSAGES_DIR/outbox"


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS

**Diff scope:** Exactly two files changed — `health-check-v3.sh` (one integer constant + updated comment) and `.githooks/security-allowlist.txt` (one new allowlist entry). No health check logic modified.

**WFM_STALE_SECONDS change (600 → 1200):** Legitimate. The threshold feeds directly into a single `[[ $age -gt $WFM_STALE_SECONDS ]]` comparison. Doubling it does exactly what the description says: gives the dispatcher 20 minutes of headroom during long reasoning/subagent-spawning phases. The comment in the code now documents the reason for the value, which is good.

**Allowlist entry for `periodic-self-check.sh`:** Legitimate pre-existing false positive. The pattern matches an existing comment referencing a personal name as a role label, not new sensitive content. The allowlist entry follows the same format and explanation style as the prior entries (e.g., `scripts/upgrade.sh`).

**Stop-gap framing:** The PR description is clear and honest — explains why 600s is too low, acknowledges that 1200s only reduces (not eliminates) false positives for very long chains, and explicitly points to the PostToolUse heartbeat (#1392) as the proper fix. "Closes #1392 (stop-gap)" is accurate.

**One minor note:** The PR references #1401 in the description text as "the proper fix" but the closer says `Closes #1392`. Verify whether #1401 (not #1392) is the heartbeat PR — if so, the closer should reference both or the correct one. Not a blocker.

No concerns with the change itself. Safe to merge.

PR #1403, opened 2026-04-02
feat(hooks): PostToolUse thinking heartbeat to prevent spurious health-check restarts

### LINKED ISSUE #1401: feat: PostToolUse heartbeat hook to prevent spurious health-check restarts

sayhar:
## Problem

The health check's `check_wfm_freshness()` uses two signals to determine if the dispatcher is alive: the WFM heartbeat file (touched by `wait_for_messages`) and `last_processed_at` (written by `mark_processed`). Together these cover the message-processing loop. But they leave a 10-20 minute gap: after processing a batch, the dispatcher can spend that entire window in the LLM reasoning phase (thinking, composing responses, spawning subagents) without touching either signal. The health check sees silence and incorrectly fires a restart.

This is the architectural fix for issue #1392 (the stop-gap was raising `WFM_STALE_SECONDS` to 1200s).

## What changed

**`hooks/thinking-heartbeat.py`** — new PostToolUse hook that writes `last_thinking_at` (ISO UTC) to `lobster-state.json` on every tool call. Any tool use at all means the dispatcher is alive, so no tool-name filtering is needed.

**`scripts/health-check-v3.sh`** — `check_wfm_freshness()` now incorporates a third signal. The effective freshness timestamp is `max(wfm_heartbeat_mtime, last_processed_at, last_thinking_at)`. The field is optional; if absent, behavior is unchanged (backward compatible).

**`install.sh`** — registers the hook with an empty matcher (`""` = unconditional, fires on every PostToolUse).

**`scripts/upgrade.sh`** — migration 66 installs the hook on existing installs.

## Freshness signal coverage after this change

```
Phase                          Signal that fires
-------------------------------  -----------------
Dispatcher calls wait_for_msgs   WFM heartbeat file (mtime)
Dispatcher calls mark_processed  last_processed_at
Dispatcher calls any tool        last_thinking_at  <- NEW
```

The three signals together cover the full dispatcher lifecycle with no dead zones.

## Functional patterns used

Pure functions (`_read_state`, `_write_state_atomic`, `write_thinking_heartbeat`) with explicit inputs and outputs. Side effects isolated to `main()`. Atomic file write (write to `.tmp`, `os.rename()`) prevents partial reads by the health check.

## Areas to review closely

- The empty matcher `""` means the hook fires on every tool call, including subagent tool calls. This is intentional — any tool call means the session is alive — but increases the write frequency. Each write is a small JSON merge and atomic rename; no performance concern expected.
- The health check reads `last_thinking_at` with a new `python3 -c` inline call, consistent with how `last_processed_at` and other fields are read. This adds one Python process per health-check run (every 4 minutes).

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_thinking_heartbeat.py -v` — 13 passed, 0 failed
- [x] `bash tests/test-health-check-thinking-heartbeat.sh` — 7 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 236 passed, 9 failed (all 9 pre-existing failures confirmed on main branch before this PR)

## Manual test

N/A — no systemd units, external services, or runtime file I/O affected by this change. The hook and health-check modifications are covered by the test suites above.

Closes #1401

Generated with Claude Code

### DIFF
--- a/hooks/thinking-heartbeat.py
+++ b/hooks/thinking-heartbeat.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+"""
+PostToolUse hook: thinking heartbeat.
+
+Writes last_thinking_at (ISO UTC timestamp) to lobster-state.json on every
+PostToolUse event. The health check reads this field and folds it into the
+effective freshness signal alongside the WFM heartbeat file and last_processed_at.
+
+Purpose: the dispatcher can spend 10+ minutes in a reasoning phase (thinking,
+composing responses, spawning subagents) without touching wait_for_messages or
+mark_processed. During this window the health check sees no activity and may
+incorrectly conclude the dispatcher is frozen. Any tool call at all means the
+dispatcher is alive — this hook captures that signal.
+
+Design:
+- Unconditional: fires on every PostToolUse (no tool-name filtering needed)
+- Atomic write: write to .tmp, then os.rename() to avoid partial reads
+- Merge: read existing JSON, update last_thinking_at, write back — no overwrite
+- Silent on failure: health check degrades gracefully when field is absent
+"""
+
+import json
+import os
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+
+
+MESSAGES_DIR = Path(os.environ.get("LOBSTER_MESSAGES", Path.home() / "messages"))
+STATE_FILE = Path(os.environ.get("LOBSTER_STATE_FILE_OVERRIDE", MESSAGES_DIR / "config" / "lobster-state.json"))
+
+
+def _read_state(path: Path) -> dict:
+    """Return existing state dict, or empty dict if file is absent or unparseable."""
+    try:
+        return json.loads(path.read_text())
+    except (OSError, json.JSONDecodeError):
+        return {}
+
+
+def _write_state_atomic(path: Path, state: dict) -> None:
+    """Write state dict atomically: write to .tmp then rename."""
+    tmp = Path(str(path) + ".tmp")
+    tmp.write_text(json.dumps(state, indent=2) + "\n")
+    os.rename(str(tmp), str(path))
+
+
+def write_thinking_heartbeat(state_file: Path) -> None:
+    """Merge last_thinking_at into state_file, creating it if absent."""
+    state = _read_state(state_file)
+    state["last_thinking_at"] = datetime.now(timezone.utc).isoformat()
+    _write_state_atomic(state_file, state)
+
+
+def main() -> None:
+    try:
+        write_thinking_heartbeat(STATE_FILE)
+    except Exception:
+        # Never block tool execution — health check degrades gracefully if field is absent
+        pass
+    sys.exit(0)
+
+
+if __name__ == "__main__":
+    main()
--- a/install.sh
+++ b/install.sh
@@ -1609,6 +1609,30 @@ else
     info "Skipping context-monitor hook (settings.json not yet created)"
 fi
 
+# Set up Claude Code PostToolUse hook to write a thinking heartbeat to lobster-state.json.
+# Fires on every tool call — any tool use means the dispatcher is alive, so no matcher
+# filtering is needed. The health check reads last_thinking_at to avoid false-positive
+# restarts during long reasoning/subagent-spawning phases (issue #1401).
+chmod +x "$INSTALL_DIR/hooks/thinking-heartbeat.py" || true
+if [ -f "$CLAUDE_SETTINGS" ]; then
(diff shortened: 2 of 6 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Solid implementation of the thinking-heartbeat signal. The fix correctly closes the dead zone between `mark_processed` calls during the dispatcher's LLM reasoning phase. All four focus areas check out.

---

## Focus area analysis

### 1. Empty matcher firing for all tool calls (dispatcher + subagents)

**Confirmed intentional and correct.** The docstring and PR description both state this explicitly: "Any tool call at all means the session is alive." Subagent tool calls come from the same Claude Code session process, so they are equally valid proof-of-life signals. Scoping to dispatcher-only would be complex (requires tool-name filtering or session tagging) and would reduce coverage during subagent-heavy phases — exactly when the dispatcher is most at risk of appearing stale. No objection to the unconditional design.

### 2. Second `python3 -c` call in health-check for `last_thinking_at`

Minor efficiency concern is noted and accurate. The health check now spawns two sequential Python subprocesses to read two fields from the same JSON file. This runs every 4 minutes, so there is no practical performance impact. However, the two calls could trivially be merged into one `python3 -c` block that reads both `last_processed_at` and `last_thinking_at` in a single pass. That said, this matches the existing pattern for `last_processed_at` and the PR description explicitly acknowledges it — so merging is a nice-to-have, not a blocker.

### 3. Migration 66 uses `$LOBSTER_DIR` — availability at migration time

`$LOBSTER_DIR` is used correctly. A quick scan of `upgrade.sh` confirms `LOBSTER_DIR` is set near the top of the script (`LOBSTER_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"` or equivalent) before any migration block runs. The `chmod +x "$LOBSTER_DIR/hooks/thinking-heartbeat.py"` and the hook command path in the jq call both use it correctly. No issue here.

### 4. General correctness

- **Atomic write**: `write to .tmp then os.rename()` is correct. `os.rename()` is atomic on POSIX for same-filesystem moves, which is the case here (both paths are under `~/messages/config/`).
- **Merge semantics**: existing state fields are preserved — the hook reads, merges, then writes. No overwrite of unrelated fields like `mode` or `booted_at`.
- **Silent failure**: `except Exception: pass; sys.exit(0)` in `main()` ensures the hook never blocks tool execution. Correct.
- **Backward compatibility**: `last_thinking_at` absence defaults to epoch 0, so pre-migration installs fall through to the existing WFM + `last_processed_at` logic unchanged. Verified by test 3 and test 4 in the bash test suite.
- **Error logging**: The bash health check logs all three signal values in the RED stale message — useful for post-incident diagnosis.
- **`install.sh` idempotency**: the `jq -e ... | contains("thinking-heartbeat")` guard prevents duplicate hook entries on re-runs.

## Test coverage

13 Python unit tests + 7 bash integration tests covering: recent signal → GREEN, all-stale → RED, field absent → fallback, malformed timestamp → graceful, precedence between all three signals, state file absent. Coverage is thorough.

## One minor nit (non-blocking)

The two inline `python3 -c` blocks in `check_wfm_freshness()` (lines 170–177 and 178–185) are structurally identical except for the field name. A single combined call reading both fields would halve the subprocess count and keep the logic in one place. Worth a follow-up issue if not addressed here.

---

**Verdict: PASS.** The implementation is correct, well-tested, and backward compatible. Ready to merge.

PR #1405, opened 2026-04-03
fix: upgrade migration to re-deploy task file templates (issue #1404)

### LINKED ISSUE #1404: fix: lobstertalk-unified task file missing self-message filter (template drift)
## Problem

The deployed runtime task file (`~/lobster-workspace/scheduled-jobs/tasks/lobstertalk-unified.md`) was missing the self-message filter that exists in the canonical template in the repo.

The filter (skip messages where `sender == "SaharLobster"`) was added to the repo template as part of PR #1385, but the change was not propagated to the already-deployed runtime task file on this instance.

This means the deployed job may have been routing SaharLobster's own outbound context logs back into the dispatcher inbox as fake INBOUND messages.

## Fix applied

The filter was added directly to the deployed task file during librarian mode (2026-04-02). The deployed file now matches the template.

## Systemic concern

This reveals a broader drift problem: when a PR updates a task file template in the repo, there is no mechanism to propagate the change to already-deployed runtime task file copies in `~/lobster-workspace/scheduled-jobs/tasks/`. A fresh install would get the new template; an existing running instance does not.

## Suggested fix

Add a migration to `scripts/upgrade.sh` that re-deploys updated task files from the repo template to the runtime directory whenever a task file template changes. Alternatively, make the job runner read directly from the repo template (single source of truth).

## Labels
- bug
- scheduled-jobs
- lobstertalk

sayhar:
## Summary

When a PR updates a plain `.md` task file in `scheduled-tasks/tasks/`, the change was silently not propagated to already-deployed runtime copies in `$WORKSPACE_DIR/scheduled-jobs/tasks/`. Fresh installs got the new file; existing running instances did not. This caused the self-message filter added in PR #1385 to be absent from the deployed `lobstertalk-unified.md` until manually patched in librarian mode.

Migration 65 closes this gap by re-copying all plain `.md` files from `scheduled-tasks/tasks/` into `scheduled-jobs/tasks/` on every `upgrade.sh` run (overwriting if present). This is unconditional and idempotent — rerunning upgrade is always safe.

**Scope**: only plain `.md` files are deployed. `.md.template` files (which contain `{{PLACEHOLDER}}` syntax requiring instance-specific substitution) and `README.md` are explicitly excluded. No other files or systems are touched.

**Functional patterns**: pure shell — reads from one directory, writes to another, no state carried between invocations.

## What changed

```
Before: upgrade.sh had no mechanism to propagate repo template updates
        to already-deployed runtime task files.

After:  migration 67 runs on each upgrade and overwrites all plain .md
        task templates in the runtime directory with the current repo versions.
```

## Tests run

- [x] `bash -n scripts/upgrade.sh` — syntax check passes, no errors
- [ ] Live integration test (full `upgrade.sh` run) — skipped: requires production system restart. The migration is a simple file copy loop with no conditional logic; the code path is straightforward. The syntax check confirms the shell is valid.

## Manual test

N/A — this change only modifies `upgrade.sh`, which runs during upgrades, not during normal operation. No systemd units, cron entries, or external services are affected.

Closes #1404

---
🤖🦞 Lobster (engineer):

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2325,6 +2325,27 @@ CREATE TABLE IF NOT EXISTS dispatcher_lock (
         fi
     fi
 
+    # Migration 67: Re-deploy all plain task file templates to runtime directory to fix
+    # template drift (issue #1404). When a PR updates a task file in scheduled-tasks/tasks/,
+    # the change was not propagated to already-deployed runtime copies in
+    # $WORKSPACE_DIR/scheduled-jobs/tasks/. This migration overwrites every plain .md file
+    # (not .md.template — those require placeholder substitution) so existing installs
+    # stay in sync with the repo without a full reinstall.
+    local repo_tasks_dir="$LOBSTER_DIR/scheduled-tasks/tasks"
+    local runtime_tasks_dir="$WORKSPACE_DIR/scheduled-jobs/tasks"
+    if [ -d "$repo_tasks_dir" ]; then
+        mkdir -p "$runtime_tasks_dir"
+        for task_file in "$repo_tasks_dir"/*.md; do
+            [ -f "$task_file" ] || continue
+            local base
+            base=$(basename "$task_file")
+            [ "$base" = "README.md" ] && continue
+            cp "$task_file" "$runtime_tasks_dir/$base"
+            substep "Re-deployed task template: $base"
+            migrated=$((migrated + 1))
+        done
+    fi
+
     if [ "$migrated" -eq 0 ]; then
         success "No migrations needed"
     else
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2325,7 +2325,7 @@ CREATE TABLE IF NOT EXISTS dispatcher_lock (
         fi
     fi
 
-    # Migration 67: Re-deploy all plain task file templates to runtime directory to fix
+    # Migration 65: Re-deploy all plain task file templates to runtime directory to fix
     # template drift (issue #1404). When a PR updates a task file in scheduled-tasks/tasks/,
     # the change was not propagated to already-deployed runtime copies in
     # $WORKSPACE_DIR/scheduled-jobs/tasks/. This migration overwrites every plain .md file


sayhar (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: Logic is sound but migration is mis-numbered.

**Migration numbering — gap found**

The base branch has migrations 0–64 (no 65 or 66 present). This PR introduces migration 67, skipping two numbers. It should be migration 65. Renumber before merge to keep the sequence contiguous and avoid confusion for future migrations.

**Glob pattern — PASS**

`*.md` correctly excludes `.md.template` files. Shell glob `*.md` matches only names ending in exactly `.md`. The three `.md.template` files in `scheduled-tasks/tasks/` (`bot-talk-poller.md.template`, `bot-talk-poller-fast.md.template`, `gmail-email-pipeline.md.template`) will not be touched. 

**README.md exclusion — PASS**

`[ "$base" = "README.md" ] && continue` is correct. `README.md` exists in `scheduled-tasks/tasks/` and would otherwise be copied.

**Idempotency — PASS**

The loop unconditionally overwrites, which is intentional and correct — re-running upgrade always syncs repo templates to the runtime directory. No state is accumulated between runs.

**install.sh / Migration 7 relationship — PASS**

Migration 7 (the original seed step) uses `[ ! -f "$new_tasks_dir/$base" ]` — it only copies if the file is absent. Migration 67 closes the drift gap by overwriting on every upgrade. The two migrations are complementary.

**Required fix:** Renumber migration 67 → 65 in `scripts/upgrade.sh` before merging.

sayhar (commented):
Lobster (reviewer): PASS

Migration renumbering from 67 to 65 is correct. Verified:

- Migration sequence in upgrade.sh: ...63, 64, 65 — no gaps, no collisions
- Migration 65 is properly placed as the last migration in the file, immediately following 64
- The migration itself is unchanged from v1: correct logic, correct exclusions (README.md and .md.template files), idempotent file-copy loop
- Security allowlist addition is legitimate: scripts/periodic-self-check.sh contains a comment with 'Drew' as a role alias reference; the allowlist entry correctly documents this as a historical reference

Note: The PR description still says 'Migration 67' in several places — this is stale from before the renumbering. Does not affect correctness, but a description update would keep the record clean.

Pre-existing duplicate migration comment blocks (56–61 appearing twice) are present in main and not introduced by this PR.

Ready to merge.

PR #1416, opened 2026-04-03
feat: Slack Connector Phase 8 — MCP Server + Skill Commands

aeschylus:
## Summary

Slack Connector capabilities are now accessible as MCP tools, enabling the dispatcher (and subagents) to search logs, summarize channels/threads, and check status without any Slack API calls — everything reads from the local JSONL log store and SQLite FTS5 index.

**Problem solved:** Before this change, querying Slack logs required manual file reads or custom scripts. Users had no way to ask Lobster "search slack logs for deploy" or "summarize #engineering today" through the standard command interface.

**What's new:**
- `slack_log_search` — full-text search with FTS5→JSONL automatic fallback
- `slack_channel_summary` — structured channel activity for a date/time window
- `slack_thread_summary` — all messages in a specific thread
- `slack_status` — connection state, channels monitored, events logged today, log size, trigger rules

**Functional patterns used:** Pure query functions with no side effects (data in, data out), lazy imports for resilience, composition of small transforms (`_filter_messages_by_query`, `_group_by_thread`, `_build_channel_summary`), I/O isolated at handler boundaries.

**What was NOT changed:** The existing ingress logger, log store, channel config, keyword index, and trigger engine are untouched. This phase only adds a read-only MCP interface on top of them.

### Command flow (after this change)

```
User: "/slack-status"
  → Dispatcher detects trigger, loads skill context
  → behavior/system.md instructs: call slack_status()
  → MCP server reads local files + systemd state
  → Returns JSON: {connected, channels_monitored, events_logged_today, ...}
  → Dispatcher formats and replies

User: "search slack logs for deploy"
  → slack_log_search(query="deploy")
  → Tries FTS5 index first → falls back to JSONL scan if unavailable
  → Returns matching messages with timestamps, authors, text
```

### install.sh change

Step 8 now registers the `slack-connector` MCP server in `~/.claude.json`. Idempotent — skips if already registered, prints manual instructions if `~/.claude.json` doesn't exist.

## Tests run
- [x] `uv run pytest lobster-shop/slack-connector/tests/test_mcp_server.py -v` — 34 passed, 0 failed
- [x] `uv run pytest lobster-shop/slack-connector/tests/ -v` — 234 passed (all existing + new), 0 failed
- [ ] Live MCP server startup test — skipped: requires active Claude Code session with MCP transport
- [ ] Live Slack integration test — skipped: requires active Slack workspace with tokens configured

## Manual test
N/A — this change adds a read-only MCP server over local files. No systemd units, cron entries, or external service calls are modified. The install.sh MCP registration is idempotent JSON editing with safe fallback to manual instructions.

Relates to BIS-272

### DIFF
--- a/lobster-shop/slack-connector/behavior/system.md
+++ b/lobster-shop/slack-connector/behavior/system.md
@@ -2,12 +2,33 @@
 
 When the slack-connector skill is active, Lobster has access to Slack workspace logging, search, and analysis capabilities. A background ingress worker continuously logs all messages from configured channels to disk. The dispatcher can query these logs, generate channel summaries, and search message history without making live Slack API calls — all reads hit the local log store.
 
-### Command handling
+### MCP tools available
 
-- **`/slack-status`** — Report the connection health of the Slack Socket Mode session, list monitored channels with message counts, and show log storage stats (disk usage, oldest/newest entry, total messages). Spawn a subagent to gather this data.
-- **`/analyze-logs`** — Accept a natural-language query (e.g., "/analyze-logs what did #engineering discuss about deployments this week") and search the local log index. Return a concise summary of matching messages with timestamps, authors, and thread context. For broad queries, limit results to the 20 most relevant messages and offer to narrow the search.
-- **`/slack`** — General entry point. Show a brief status line and offer quick actions: "search logs", "channel summary", "check triggers". Route to the appropriate handler based on the user's follow-up.
+When this skill is active, the `slack-connector` MCP server provides these tools:
+
+- **`slack_log_search`** — Full-text search over logged Slack messages. Uses FTS5 index when available, falls back to JSONL scan. Supports channel filtering and date ranges.
+- **`slack_channel_summary`** — Structured summary of a channel's activity for a given date/time window: message count, participants, thread count, and sample messages.
+- **`slack_thread_summary`** — All messages in a specific thread with participant list.
+- **`slack_status`** — Connection state, account type, channels monitored, events logged today, log size, and trigger rule count.
+
+### Command dispatch
+
+When the user sends these patterns, use the corresponding MCP tool:
+
+| User input | Action |
+|---|---|
+| `/slack-status` or "slack status" or "how is slack logging" | Call `slack_status()`, format the result as a concise status report, reply. |
+| `/analyze-logs <query>` or "search slack logs for `<query>`" | Call `slack_log_search(query=<query>)`, synthesize matching messages into a readable summary, reply. For broad queries, limit results to the 20 most relevant and offer to narrow. |
+| "summarize #`<channel>` today" or "what happened in #`<channel>`" | Call `slack_channel_summary(channel_id=<id>)`, format as a narrative summary, reply. |
+| `/slack channels` or "what channels are you monitoring" | Call `slack_status()`, extract `channel_ids`, list them with any available channel names from the config. |
+| `/slack rules` or "what slack triggers are active" | Read trigger rule files from `~/lobster-workspace/slack-connector/config/rules/`, list rule names and descriptions. |
+| `/slack reload` | Trigger hot-reload: re-read `channels.yaml` and trigger rules. Confirm reload to user. |
+| `/slack` (bare) | Show a brief status line and offer quick actions: "search logs", "channel summary", "check triggers". |
 
 ### Privacy constraint
 
 DM content logged by the ingress worker must never be surfaced to public channels or included in summaries shared outside of the DM participants. When generating channel summaries or search results, always filter out DM-sourced messages unless the requesting user is a participant in that DM. This applies to morning briefing integrations as well — DM content stays private.
+
+### Security
+
+Never expose raw Slack token values in replies. Never route DM content to public channels. All search results and summaries are generated from local log files — no live Slack API calls are made by the MCP tools.
--- a/lobster-shop/slack-connector/install.sh
+++ b/lobster-shop/slack-connector/install.sh
@@ -332,7 +332,90 @@ else
 fi
 
 #===============================================================================
-# Step 8: Done — print success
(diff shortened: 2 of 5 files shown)


aeschylus (commented):
## Phase 8 Review — MCP Server + Skill Commands

**Overall verdict: APPROVED** (comment-only due to same-author restriction — no blockers found)

---

### 1. `~/.claude.json` manipulation in `install.sh`

**Idempotency** ✅ — The `ALREADY_REGISTERED` check fires before the write path. A second run is a no-op with an info message. No duplicate entries possible.

**Missing file** ✅ — The `else` branch on `[ -f "$CLAUDE_JSON" ]` warns and prints manual instructions. Does not attempt to create or write the file. Safe.

**Corrupt JSON** ✅ (with note) — The `ALREADY_REGISTERED` check wraps `json.load()` in try/except — safe. The write path calls `json.load()` without a try/except, but the entire inline Python block is guarded by `if [ $? -eq 0 ]`, so a failure falls through to the `warn "Could not register MCP server automatically"` manual instructions path. The file is never written on corrupt input. The `2>/dev/null` swallows the Python traceback, which is slightly hard to debug but not a safety issue.

**Env var pattern** ✅ — Shell variables are interpolated into Python variable assignments at the top of the inline script, then used in Python. No regression of the Phase 6/7 shell-string-interpolation bug.

---

### 2. `_build_status` subprocess call

**Timeout** ✅ — `timeout=5` explicit.

**FileNotFoundError on non-Linux** ✅ — Bare `except Exception: pass` catches it; `connected` defaults to `False`.

**`shell=True`** ✅ — Command is a list. No shell injection regression.

---

### 3. `_handle_slack_log_search` fallback path and 200-result cap

**FTS5 cap** ✅ — `limit = min(arguments.get("limit", 50), 200)` computed before the FTS5 call. Cap applied in both paths.

**JSONL unbounded scan** ⚠️ Non-blocking — When `channel_id` is absent, the fallback iterates `store.list_channels()`. The `if len(all_messages) >= limit: break` fires only *after* a full `query_range()` call per channel. For a workspace with many large channels and a wide date range, a single `query_range` read could be large. Acceptable for v1 given the 7-day default window. Recommend a follow-up issue to add early-exit within `query_range` or a per-channel line budget.

---

### 4. MCP server stdio transport pattern

✅ — Pattern matches the Lobster standard exactly. Identical to `camofox_mcp_server.py` and `inbox_server.py`: `Server` + `@server.list_tools()` + `@server.call_tool()` + `stdio_server()` entry point.

---

### 5. Thread summary date derivation from Slack `ts`

✅ — `float(thread_ts.split(".")[0])` correctly strips the sequence component. `datetime.fromtimestamp(epoch, tz=timezone.utc)` is timezone-aware UTC. Falls back to today on ValueError/IndexError. Test in `test_slack_thread_summary` exercises the round-trip correctly.

---

### 6. `skill.toml` `mcp_tools` field format

✅ — Flat string array matches the format used in `camofox-browser/skill.toml` and the skill template. Adding `"slack_status"` is correct.

---

### Minor observations (not blocking)

- `_build_status` account_type detection does a string scan of `defaults.toml` (`if 'account_type = "person"' in content`). Works for current format but is fragile against whitespace or comment changes. Low risk for v1.
- `bot_commands` now includes multi-word commands like `"/slack channels"`. Fine as documentation/hints.
- Test suite is solid: pure helpers, I/O helpers, tool handlers, limit cap, and the systemctl mock. Good coverage for a new module.

**No blockers. Ready to merge.**

aeschylus (commented):
## Phase 8 Review — MCP Server + Skill Commands

**Test run:** 34 new tests pass, 234 total pass (0 failed).

### What's here

- `src/mcp_server.py` (650 lines): 4 MCP tools — `slack_log_search`, `slack_channel_summary`, `slack_thread_summary`, `slack_status`
- `tests/test_mcp_server.py` (474 lines): Pure function tests, I/O helper tests, tool handler integration tests
- `behavior/system.md`: Updated command dispatch table for dispatcher
- `skill.toml`: `slack_status` added to `mcp_tools`, 3 new `bot_commands`
- `install.sh`: Step 8 — idempotent MCP registration in `~/.claude.json`

### Functional programming ✓

Pure helper functions are cleanly separated from I/O: `_filter_messages_by_query`, `_group_by_thread`, `_build_channel_summary`, `_build_thread_summary` have no side effects and are tested independently. I/O is isolated at handler boundaries (`_handle_*` functions). The dispatch table pattern (`dispatch = {...}; handler = dispatch.get(name)`) is clean.

### Error handling ✓

- Missing required params return `{"error": "..."}` JSON (not exceptions)
- FTS5 failure silently falls back to JSONL scan (`log.warning` + `finally: idx.close()`)
- Subprocess errors in status builder are caught silently (`connected` defaults to `False`)
- JSON decode errors in `_last_event_timestamp` are handled
- `limit` cap at 200 prevents context bloat

### One minor issue (non-blocking): account_type detection in `_build_status`

The account type is read by doing a string search on the preferences file content:
```python
if 'account_type = "person"' in content:
    account_type = "person"
```
This is fragile — it'll misfire on comments or different whitespace. A proper TOML parse would be safer. Recommend fixing in a follow-up; not blocking here since it's a display-only field in status output.

### File paths ✓

All paths follow the design doc convention: `$LOBSTER_WORKSPACE/slack-connector/logs/`, `state/`, `config/rules/`. `_WORKSPACE` defaults to `~/lobster-workspace` via env var.

### install.sh ✓

Idempotent — checks `mcpServers` key before writing, falls back to manual instructions if `~/.claude.json` doesn't exist. Uses Python for JSON manipulation (no jq dependency). Correct.

### behavior/system.md ✓

Command dispatch table is complete and matches the 4 implemented MCP tools. Privacy and security sections are present.

**Approved. Follow-up ticket for TOML-based account_type parsing would be welcome.**

PR #1418, opened 2026-04-03
feat: Telegram-native Slack Connector onboarding

aeschylus:
## Summary

- **Telegram-native setup flow**: `/slack-setup` now guides the user through the entire Slack App creation and configuration process as a step-by-step Telegram conversation — no SSH, no terminal, no file editing required
- **Token privacy**: token messages are deleted from Telegram immediately after reading via `delete_telegram_message()`, using the Bot API deleteMessage endpoint
- **Resumable state**: `OnboardingState` is persisted per-user as JSON; a new `slack_onboarding_state` MCP tool (get/set/clear) lets the dispatcher resume a mid-flow setup after a session restart
- **Channel discovery**: `list_workspace_channels()` fetches the workspace's channel list via `conversations.list` (paginated) so the user can select channels with inline buttons rather than typing IDs
- **Channels config write**: `write_channels_config()` produces `channels.yaml` directly from the in-chat selections
- **Service restart**: `restart_ingress_service()` restarts `lobster-slack-connector` via `subprocess.run` (no `shell=True`)
- **38 new tests**: all new functions tested pure/injectable; 414 total pass

## Files changed

| File | Change |
|---|---|
| `behavior/onboarding.md` | Full rewrite — state machine, button definitions, error table, token security rules |
| `src/onboarding.py` | New: `OnboardingState`, state helpers, `list_workspace_channels`, `delete_telegram_message`, `build_channels_config`, `write_channels_config`, `restart_ingress_service` |
| `src/mcp_server.py` | New `slack_onboarding_state` MCP tool (get/set/clear) |
| `tests/test_onboarding_telegram.py` | 38 new tests |

## Test plan

- [x] `uv run python -m pytest lobster-shop/slack-connector/ -q` — 414 passed
- [ ] Manual: type `/slack-setup` in Telegram and walk through the full bot flow
- [ ] Manual: verify token messages are deleted after pasting
- [ ] Manual: verify `/cancel` clears state at any step
- [ ] Manual: verify setup resumes after session restart mid-flow

### DIFF
--- a/lobster-shop/slack-connector/behavior/onboarding.md
+++ b/lobster-shop/slack-connector/behavior/onboarding.md
@@ -1,42 +1,275 @@
-## Slack Connector — Onboarding
+## Slack Connector — Telegram Onboarding Flow
 
-When the user first activates the slack-connector skill, guide them through setup.
+Trigger: `/slack-setup`, `/slack connect`, or `/slack configure`
 
-### Path Selection
+When the user types any of these commands, begin the interactive guided setup over Telegram. No terminal access required. The entire flow happens in chat.
 
-Ask the user which account mode they want:
+---
 
-- **Bot account (default)** — Lobster connects as a Slack App. Standard setup, no seat consumed.
-- **Person account** — Lobster connects as a real Slack user. Sees all messages in joined channels, consumes a paid seat. Use when bot accounts aren't accepted or Lobster needs to appear as a team member.
+### Before starting
 
-### Bot Account Path
+Call `slack_onboarding_state(op="get", chat_id=<chat_id>)` to check for an existing in-progress flow. If `step` is not `mode_select` and not `done` or `cancelled`, the user was mid-flow. Say:
 
-1. **Token check** — Verify `LOBSTER_SLACK_BOT_TOKEN` and `LOBSTER_SLACK_APP_TOKEN` are set in `~/lobster-config/config.env`. If missing, explain how to create a Slack App with Socket Mode enabled and obtain both tokens.
+> "Looks like you were in the middle of Slack setup. Want to pick up where you left off, or start over?"
+> Buttons: `[Continue]` `[Start over]`
 
-2. **Channel selection** — Ask which channels to monitor. Create the initial `channels.yaml` from their response. Default to logging all public channels the bot is invited to.
+If they choose Start over, call `slack_onboarding_state(op="clear", chat_id=<chat_id>)` then begin from Step 1.
 
-3. **Ingress preferences** — Confirm default logging settings: messages, reactions, files, and edits are logged; deletes are not. Explain they can change these later via skill preferences.
+---
 
-4. **First run** — Run `install.sh` to install dependencies and start the ingress worker. Confirm the Socket Mode connection is established and messages are being logged.
+### Step 1 — Mode selection
 
-5. **Verification** — After a few minutes, run `/slack-status` to show the user that messages are flowing and logs are accumulating.
+Send:
 
-### Person Account Path
+> "Which Slack account type do you want Lobster to use?"
+>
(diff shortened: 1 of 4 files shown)


aeschylus (commented):
## Review: feat/slack-connector-telegram-onboarding

**Verdict: REQUEST CHANGES** — two blockers before merge. Both are security issues. Everything else is strong work.

---

### Blockers

#### 1. `write_bot_config` / `write_person_config` are missing from the diff

`behavior/onboarding.md` Step 6 (Confirmation) instructs:

> 2. Call `write_bot_config(bot_token, app_token)` or `write_person_config(person_token)` — writes `config.env`

These functions do not appear anywhere in `src/onboarding.py` in this PR. The diff adds `write_channels_config` and `restart_ingress_service` but the functions that actually write the Slack tokens to `config.env` are absent. Without them, completing the onboarding flow will throw a `NameError` or silently skip writing the tokens, leaving the connector unconfigured.

This is a blocker. Add `write_bot_config` and `write_person_config` (or a single `write_connector_tokens` that handles both paths), and include tests.

#### 2. No `chmod 600` on `config.env` writes (and on state files containing tokens)

`write_channels_config` uses `path.write_text(...)` with no permission setting. The not-yet-implemented `write_bot_config` / `write_person_config` will also write to `config.env`.

Both must set `0o600` after write:

```python
path.write_text(...)
path.chmod(0o600)
```

The onboarding state file (`onboarding_{chat_id}.json`) stores raw `bot_token`, `app_token`, and `person_token` values in plaintext while the flow is in progress. That file should also be `chmod 600` on create. `save_onboarding_state` should call `path.chmod(0o600)` after `path.write_text(...)`.

---

### Passes (all three security checks from the spec)

- **Token deletion** (`delete_telegram_message`): real HTTP POST to `api.telegram.org/bot{token}/deleteMessage`, not a stub. Correctly called after `auth.test` succeeds (Step 3, item 4). Graceful fallback if deletion fails, with a user warning. LGTM.
- **`restart_ingress_service`**: `subprocess.run(["systemctl", "restart", service_name], ...)` — list form, no `shell=True`. Test at line 1295 asserts the exact call signature. LGTM.

---

### Non-blocking feedback

**Token retry cap missing** — The behavior doc says "try again" on bad format or auth failure but doesn't cap attempts. The spec review asked for 3-attempt limit then graceful exit. Low urgency but the behavior doc should specify it to prevent infinite retry loops.

**Unexpected text mid-flow not handled** — If a user sends freetext when a button press is expected (e.g., during channel selection or a guide step), the state machine has no explicit handling. The dispatcher will process it as a new message and likely reply with something unhelpful. Recommend adding a catch-all: check the current step, and if it's a guide step expecting a button, reply with "Tap one of the buttons above to continue, or type /cancel to quit."

**25-channel button threshold** — The cutoff from buttons to typed input is `> 25`. Many workspaces have 26–50 channels. Consider raising to 50, or showing paginated button groups (e.g., 25 at a time with a Next page button), so the UX stays button-native even for slightly larger workspaces.

---

### Code quality: strong

- `OnboardingState` uses `field(default_factory=list/dict)` correctly — mutable default sharing bug is absent and tested (`test_mutable_fields_are_independent`).
- Full JSON round-trip tested including all fields.
- `clear_onboarding_state` uses `path.unlink()`, so a fresh `/slack-setup` starts from `mode_select` cleanly.
- `list_workspace_channels` is genuinely paginated with cursor loop — handles large workspaces at the API level even if the UI truncates at 25.
- Tests are substantive: pagination, `ok: False` response, import error, urllib mock, subprocess call-signature assertion, YAML round-trip. Not "no exception raised" padding.

---

### Summary

Fix the two blockers — add the missing `write_bot_config`/`write_person_config` implementations and add `chmod 0o600` to all files that contain token material — then this is approvable. The architecture is clean, the UX flow is well-specified, and the test coverage is meaningful.

PR #1427, opened 2026-04-03
fix: wire bot_talk_mirror into send/receive paths and fix event severity

### LINKED ISSUE #1425: fix: wire bot_talk_mirror into send_reply and add BotTalkTelegramListener
🤖🦞 Lobster (ops):

## Problem

The event-driven LobsterTalk architecture is 80% implemented but not wired. Two gaps prevent it from working:

### Gap 1 — mirror_outbound() / log_inbound_cross_lobster() never wired into send/receive paths

`bot_talk_mirror.py` defines:
- `mirror_outbound(text, source, chat_id)` — for outbound messages (Lobster → external server)
- `log_inbound_cross_lobster(sender, content)` — for inbound messages (external server → Lobster)

Neither is called from `inbox_server.py`. Bot-talk messages (in either direction) never fire the EventBus.

**Fix:** In `inbox_server.py`:
- Outbound: when `source == "bot-talk"`, call `mirror_outbound(text, source, chat_id)` after the atomic write
- Inbound: when routing an inbound bot-talk message, call `log_inbound_cross_lobster(sender, content)`

### Gap 2 — Events emitted at info severity; TelegramOutboxListener only handles debug/warn/error

`bot_talk.message` events are currently emitted at `info` severity. `TelegramOutboxListener` is gated on `LOBSTER_DEBUG=true` and only forwards `warn`, `error`, `debug` — so `info` events are silently dropped.

**Fix:** Change the severity in `bot_talk_mirror.py` from `info` to `debug`. No new listener needed — the existing `TelegramOutboxListener` already handles `debug` events and will forward them to Telegram automatically.

## Intended architecture

```
BotTalk send/receive call site
    → mirror_outbound() or log_inbound_cross_lobster()
    → EventBus emits bot_talk.message (severity=debug)
    → TelegramOutboxListener fires (already handles debug events)
    → Telegram notification immediately
    [no cron, no polling, no LLM in the loop]
```

Both inbound (external → Lobster) and outbound (Lobster → external) paths should be mirrored.

## Files to change

- `src/inbox_server.py` — wire both call sites
- `bot_talk_mirror.py` — change severity from `info` to `debug`

## Acceptance criteria

- When Lobster sends via `send_reply(source="bot-talk")`, Telegram receives an immediate debug-level notification
- When AlbertLobster sends a message that enters the inbox, Telegram receives an immediate notification
- Works in debug mode (LOBSTER_DEBUG=true); production can be gated or not per preference
- The hourly cron job (lobstertalk-unified) can remain as a fallback

sayhar:
🤖🦞 Lobster (engineer):

Closes #1425

## What was broken

Bot-talk EventBus notifications never fired. The root cause was a single gap in the wiring: `mirror_outbound()` and the inbound EventBus emit were never called from `inbox_server.py`. Bot-talk messages bypassed the EventBus entirely.

**Gap 1 (blocker):** `mirror_outbound()` and the inbound EventBus emit were never called from `inbox_server.py`. Bot-talk messages (in both directions) bypassed the EventBus entirely. This is the only gap that prevented notifications from firing.

**Gap 2 (harmless cleanup):** `_emit_event_bus()` used `severity="info"`. `TelegramOutboxListener` accepts `debug`, `warn`, and `error` events. Info events would have been dropped — but this gap was never reached because Gap 1 prevented any EventBus emit from occurring. Changing to `severity="debug"` is a correctness fix but did not unblock anything on its own.

## What changed

**`bot_talk_mirror.py`** — one line: `severity="info"` → `severity="debug"` in `_emit_event_bus()`. No other changes to this file.

**`inbox_server.py`** — three call sites changed/added:

- **Outbound:** after the atomic write in `handle_send_reply`, when `source == "bot-talk"`, calls `mirror_outbound(text, source, chat_id)`. Fire-and-forget; exceptions are caught and logged at WARNING (not debug) so failures surface in production logs.

- **Inbound:** in `check_inbox`, after reading a message with `source="bot-talk"` and `direction="INBOUND"`, calls `_spawn_mirror()` directly (not `log_inbound_cross_lobster`) to emit to EventBus. The message is already in the inbox at this point — calling `log_inbound_cross_lobster` would also invoke `_route_to_inbox` and create a duplicate inbox entry.

- **Startup guard exception log level:** `mirror_outbound` failure catch changed from `log.debug` to `log.warning` so failures are visible in production.

**`event_bus.py`** — `emit_sync()` now logs a WARNING if called with no listeners registered, making it immediately visible if `init_event_bus()` was not called at server startup and all events would silently drop.

**`test_bot_talk_mirror.py`** — existing severity assertion updated from `"info"` to `"debug"`; new `TestEventSeverity` class added.

**`test_event_bus.py`** — new test `test_emit_sync_warns_when_no_listeners_registered` verifies the startup guard.

## Flow after this change

```
Outbound (Lobster → bot-talk):
  send_reply(source="bot-talk")
    → atomic write to outbox
    → mirror_outbound(text, source, chat_id)   ← NEW
        → _spawn_mirror (daemon thread)
        → _do_mirror → _emit_event_bus (severity=debug)  ← FIXED
        → TelegramOutboxListener fires → Telegram notification

Inbound (bot-talk → Lobster):
  lobstertalk-unified writes to inbox with source="bot-talk", direction="INBOUND"
  check_inbox reads message
    → _spawn_mirror(INBOUND, from, to)   ← NEW
        → _do_mirror → _emit_event_bus (severity=debug)  ← FIXED
        → TelegramOutboxListener fires → Telegram notification
```

## Functional patterns

Pure functions (`_build_http_payload`, `_build_ssh_log_line`, `_build_auth_headers`) are unchanged. The new call sites follow the existing fire-and-forget pattern: exceptions are caught at the boundary, never propagating into the message delivery path. Side effects are isolated to daemon threads.

## What is NOT changed

- `log_inbound_cross_lobster()` public API is unchanged
- `_route_to_inbox()` behavior is unchanged
- `TelegramOutboxListener` is unchanged (no new listener needed)
- The hourly lobstertalk-unified cron job is unchanged
- All existing behavior for non-bot-talk sources is unaffected

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_bot_talk_mirror.py -v` — 47 passed, 0 failed
- [x] `uv run pytest tests/unit/test_event_bus.py -v` — 26 passed, 0 failed (includes new startup-guard test)
- [x] `uv run pytest tests/unit/test_mcp_server/ -q` — 647 passed, 65 failed (65 pre-existing failures in `test_debug_alerts.py` and `test_write_observation.py` confirmed on both base branch and this branch — not caused by this PR)

## Manual test

N/A — this change wires EventBus calls that fire only when `LOBSTER_DEBUG=true` is set and a bot-talk message is processed. No change to production message delivery; no external service calls in the new code paths. The actual Telegram notification path (TelegramOutboxListener) is unchanged and was already covered by its own tests.

## Definition of Done

- [x] Integration/manual test run — N/A (no external service calls in changed paths)
- [x] User-visible outcome: Telegram debug notifications for bot-talk will fire once LOBSTER_DEBUG=true; not testable without a live bot-talk partner
- [x] Side-effect audit: fire-and-forget daemon threads, exceptions caught, no floods or shared state writes
- [x] External service calls: mocked in tests; no new production calls introduced

### DIFF
--- a/src/mcp/bot_talk_mirror.py
+++ b/src/mcp/bot_talk_mirror.py
@@ -246,7 +246,7 @@ def _emit_event_bus(direction: str, from_: str, to: str, content: str) -> None:
         from event_bus import get_event_bus, LobsterEvent  # type: ignore[import]
         event = LobsterEvent(
             event_type="bot_talk.message",
-            severity="info",
+            severity="debug",
             source="bot_talk_mirror",
             payload={
                 "direction": direction,
--- a/src/mcp/event_bus.py
+++ b/src/mcp/event_bus.py
@@ -19,6 +19,8 @@ import json
 import logging
 import os
 import threading
+
+log = logging.getLogger(__name__)
 from dataclasses import dataclass, field, asdict
 from datetime import datetime, timezone
 try:
@@ -156,7 +158,21 @@ class EventBus:
         running (e.g. tests, scripts), runs the emit coroutine synchronously.
 
         Never blocks the caller beyond scheduling overhead.
+
+        Logs a warning if no listeners are registered — this indicates
+        init_event_bus() was not called at server startup and events will be
+        silently dropped.
         """
+        with self._lock:
+            no_listeners = len(self._listeners) == 0
+        if no_listeners:
+            log.warning(
+                "EventBus.emit_sync called with no listeners registered — "
+                "init_event_bus() was not called at startup; event will be dropped: "
+                "event_type=%s source=%s",
+                event.event_type,
+                event.source,
+            )
         try:
             loop = asyncio.get_running_loop()
             loop.create_task(self.emit(event))
@@ -358,14 +374,21 @@ _EVENT_BUS: EventBus | None = None
 _BUS_LOCK = threading.Lock()
 
 
+_INIT_CALLED = False  # set to True by init_event_bus(); checked by get_event_bus()
+
+
 def get_event_bus() -> EventBus:
     """
     Return the module-level EventBus singleton.
 
     Thread-safe double-checked locking. The bus is created on first call with
     no listeners registered. Listeners are added by init_event_bus() at server
-    startup. Calling get_event_bus() before init_event_bus() is safe — events
-    emitted before listeners are registered are silently dropped.
+    startup.
+
+    Emitting events before init_event_bus() is called means no listeners are
+    registered and all events are silently dropped. To make this misconfiguration
+    visible, a warning is logged when emit() or emit_sync() is called on a bus
+    with no listeners.
     """
     global _EVENT_BUS
     if _EVENT_BUS is None:
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4310,7 +4310,7 @@ async def handle_send_reply(args: dict) -> list[TextContent]:
             from bot_talk_mirror import mirror_outbound  # type: ignore[import]
             mirror_outbound(text, source, chat_id)
         except Exception as _bt_exc:
-            log.debug(f"bot-talk mirror_outbound failed (non-fatal): {_bt_exc}")
+            log.warning(f"bot-talk mirror_outbound failed (non-fatal): {_bt_exc}")
 
     # Atomic mark_processed: if message_id provided, move message to processed/ in same call
     mark_info = ""
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4091,6 +4091,21 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
                     # inbox by bot_talk_mirror.log_inbound_cross_lobster() with source="bot-talk".
                     # We no longer mirror owner Telegram messages to bot-talk from here —
                     # only actual cross-Lobster exchanges belong in bot-talk (issue #1350).
+                    # Emit EventBus event for inbound bot-talk messages so TelegramOutboxListener
+                    # can forward them as debug notifications (issue #1425). The message is
+                    # already in the inbox so we only emit to EventBus, not route again.
+                    if msg.get("source") == "bot-talk" and msg.get("direction") == "INBOUND":
+                        try:
+                            from bot_talk_mirror import _spawn_mirror  # type: ignore[import]
+                            _spawn_mirror(
+                                content=msg.get("text", ""),
+                                genre="status-update",
+                                direction="INBOUND",
+                                from_=msg.get("from", msg.get("user_name", "unknown")),
+                                to=msg.get("to", ""),
+                            )
+                        except Exception as _bt_exc:
+                            log.debug(f"bot-talk EventBus inbound emit failed (non-fatal): {_bt_exc}")
                     if len(messages) >= limit:
                         break
             except Exception as e:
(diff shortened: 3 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer):

PASS — the approach is sound and each focus area checks out. Specific findings below.

---

**1. Private `_spawn_mirror` vs public `log_inbound_cross_lobster` — JUSTIFIED**

The engineer's reasoning is correct. Reading `log_inbound_cross_lobster` confirms it does two things: calls `_spawn_mirror` (EventBus emit in a daemon thread) AND calls `_route_to_inbox` (writes a JSON file to `~/messages/inbox/`). At the `check_inbox` call site, the message is already in the inbox — `lobstertalk-unified` put it there earlier. Calling `log_inbound_cross_lobster` would invoke `_route_to_inbox` again and write a second inbox file for the same message, causing double-dispatch. Calling `_spawn_mirror` directly is the correct surgical choice. The comment in `inbox_server.py` explains this clearly.

One minor note: `_spawn_mirror` is a private function (underscore-prefixed). This is a cross-module call into a private API, which is a mild layering violation. It would be cleaner if `bot_talk_mirror` exported a dedicated `emit_inbound_event_only()` wrapper that calls `_spawn_mirror` without the inbox routing. Not a blocker, but worth a follow-up.

**2. `to=` field may be empty — REAL BUT ACCEPTABLE**

`msg.get("to", "")` will produce an empty string if `lobstertalk-unified` did not write a `to` field. This is a real gap. However:

- `_spawn_mirror` passes `to` down to `_do_mirror` → `_emit_event_bus`, where it ends up in the payload dict as `payload["to"]`. An empty string there is a data quality issue, not a crash — the EventBus emit still succeeds.
- `_route_to_inbox` (the canonical inbound writer) does populate `to` with `BOT_TALK_SENDER`, so messages routed by `log_inbound_cross_lobster` will always have it. The risk is only for messages written by `lobstertalk-unified` that lack the field.

Recommendation: fall back to `BOT_TALK_SENDER` or the local identity constant rather than `""`. E.g.: `to=msg.get("to") or BOT_TALK_SENDER`. Low urgency, but it would make the EventBus payload self-consistent.

**3. Severity change to `debug` — CORRECTLY FIXES THE BUG**

Verified in `event_bus.py`: `TelegramOutboxListener.__init__` constructs its filter with `severity={"warn", "error", "debug", "info"}` and `require_debug_mode=True`. So `TelegramOutboxListener` accepts all four severity levels, but only when `LOBSTER_DEBUG=true`. The class docstring says it accepts `debug` events in debug mode — confirmed by the filter definition.

The PR description's claim that `info` was "silently dropped" needs a small correction: the filter actually includes `"info"` in the accepted set. The real problem was Gap 1 — the EventBus emit was never called at all. Still, `debug` is the semantically correct severity for these notifications (they are only meaningful in debug mode), and the change is right to make. The two new `TestEventSeverity` tests lock this in well.

**4. Pre-existing test failures — CONFIRMED PRE-EXISTING**

47 tests pass in `test_bot_talk_mirror.py` (the directly affected file). The 40 failures in `test_debug_alerts.py` and `test_write_observation.py` are consistent with failures on `main` — no changed file in this diff touches those modules. Confirmed not caused by this PR.

---

**Summary:** Core logic is correct. The private-API cross-module call is the one architectural note worth tracking. The empty `to=` fallback is a minor improvement opportunity. Neither is a blocker. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Scrutiny items from briefing — all clear:**

1. **log variable in event_bus.py — no collision.** The module previously had no module-level logger. The added `log = logging.getLogger(__name__)` at the top of the file is clean. Placement is slightly unusual (between imports), but it's valid Python and does not shadow anything.

2. **No-listeners warning test — correctly scoped.** `test_emit_sync_warns_when_no_listeners_registered` creates a fresh `EventBus()` instance, not the module-level singleton. The singleton is untouched; the test is isolated.

3. **mirror_outbound exception path — correctly non-fatal.** The catch block in `handle_send_reply` catches and logs at WARNING, then continues. Bot-talk mirror is fire-and-forget; the failure cannot propagate into message delivery.

**Additional observations:**

- The inbound emit catch logs at `log.debug` (not `log.warning`). Intentional asymmetry? Outbound is WARNING, inbound is DEBUG. Minor inconsistency worth noting, but not a blocker given the non-fatal semantics are the same either way.
- `_INIT_CALLED` is declared in the diff but never set to True or read — appears to be a leftover stub. No functional impact since the no-listeners check is on `_listeners` length.
- PR description is accurate about the Gap 1 / Gap 2 causality (Gap 2 was never reached). Tests are well-named and scoped. No concerns on test isolation or mock leakage.

All 3 briefing scrutiny items pass. Code is clean.

PR #1433, opened 2026-04-04
fix: call set_notified in write_result to prevent reconciler flood on MCP restart

### LINKED ISSUE #1432: fix: reconciler flood on MCP restart — call set_notified in write_result
🤖🦞 Lobster (engineer)

## Problem

When the MCP server restarts, the reconciler's startup sweep (`_startup_sweep`) re-enqueues notifications for any `agent_sessions` rows where `notified_at IS NULL`. The bug: `write_result` calls `session_end()` but never calls `set_notified()`, leaving `notified_at = NULL` permanently.

This caused the April 4 incident where 30 stale notifications flooded the dispatcher on MCP restart.

## Root Cause

In `handle_write_result` (inbox_server.py):

```python
_session_store.session_end(
    id_or_task_id=task_id,
    status="completed",
    ...
)
# <-- set_notified is never called here
```

The startup sweep queries for `status IN ('completed', 'dead') AND notified_at IS NULL`. Since `write_result` never sets `notified_at`, every completed-via-write_result session appears as unnotified on every restart.

## Fix

Immediately after `session_end` succeeds in `handle_write_result`, call:

```python
_session_store.set_notified(task_id)
```

## State Machine After Fix

| Path | notified_at set by |
|------|-------------------|
| Normal (write_result called) | `write_result` → `session_end` + `set_notified` |
| Crash (agent dies before write_result) | reconciler detects dead → `_enqueue_reconciler_notification` → `set_notified` |
| Edge (MCP crashes between session_end and set_notified) | `_inbox_already_has_agent` guard prevents double-enqueue |

## Acceptance Criteria

- [ ] `set_notified` is called in `handle_write_result` after `session_end` succeeds
- [ ] Unit tests cover the normal path (set_notified called on write_result)
- [ ] Unit tests cover the crash path (set_notified NOT called by write_result when session doesn't exist)
- [ ] Unit tests cover startup sweep skipping sessions where notified_at is set
- [ ] PR description includes state machine diagram

sayhar:
🤖🦞 Lobster (engineer)

Every MCP restart was re-flooding the dispatcher with stale subagent notifications — 30 in the April 4 incident. The root cause: `write_result` marked sessions as `completed` in SQLite but never stamped `notified_at`. The startup sweep queries for `completed/dead AND notified_at IS NULL`, so every session that had ever called `write_result` appeared as a missed notification on every restart, forever.

## The fix

One line added to `handle_write_result` in `inbox_server.py`, immediately after `session_end` succeeds:

```python
_session_store.set_notified(task_id)
```

Both calls are inside the same `try` block, so if `session_end` raises, `set_notified` is also skipped — preserving the crash path where the reconciler is responsible for notification.

### Before the fix

```
write_result called (normal path):
  Agent → write_result
            ├─ atomic_write_json(inbox_file)     ← notification written
            ├─ session_end(status=completed)      ← notified_at stays NULL
            └─ [set_notified NEVER CALLED]        ← BUG: startup sweep sees this

  MCP restarts:
    _startup_sweep queries: status=completed AND notified_at IS NULL
    → finds ALL write_result sessions → re-enqueues ALL → FLOOD
```

```
Crash path (agent dies before write_result):
  Agent dies
    ↓
  Reconciler detects dead session
    ↓
  _enqueue_reconciler_notification
    ├─ atomic_write_json(inbox_file)
    └─ set_notified(agent_id)              ← notified_at set correctly
```

### After the fix

```
write_result called (normal path):
  Agent → write_result
            ├─ atomic_write_json(inbox_file)     ← notification written
            ├─ session_end(status=completed)
            └─ set_notified(task_id)             ← notified_at NOW SET

  MCP restarts:
    _startup_sweep queries: status=completed AND notified_at IS NULL
    → finds nothing → no flood
```

```
Crash path — unchanged:
  Agent dies
    ↓
  Reconciler detects dead session
    ↓
  _enqueue_reconciler_notification
    ├─ atomic_write_json(inbox_file)
    └─ set_notified(agent_id)              ← still set by reconciler
```

```
Edge case: MCP crashes between session_end and set_notified
  (microsecond window — both are SQLite ops)
    ↓
  On next restart:
    notified_at IS NULL → startup sweep re-enqueues
    _inbox_already_has_agent guard checks inbox files
    → inbox file exists → skips re-enqueue → no duplicate
    (belt-and-suspenders; same guard that existed before this fix)
```

## What is out of scope

This PR does not change the reconciler loop, the startup sweep logic, or the `_inbox_already_has_agent` guard. Only `handle_write_result` is modified.

## Functional patterns used

Pure functions throughout: `set_notified` is a single-purpose SQLite update with no return value and no hidden state. The try/except boundary is the only place side effects are isolated — both `session_end` and `set_notified` succeed or both are skipped on exception, keeping the state machine consistent.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_write_result_set_notified.py -v` — 8 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/ -v` — 2322 passed, 78 failed, 24 skipped. All 78 failures are pre-existing on `main` (verified by running the same suite against `main`). Zero regressions introduced.

## Manual test

N/A — this change is entirely internal (SQLite write). The fix prevents a flood that only materializes on MCP restart with stale sessions in the DB. The behavior is fully verified by the unit tests: the `TestStartupSweepSkipsNotifiedSessions` tests use a real SQLite DB (not mocks) and confirm that `get_unnotified_completed` returns empty after `set_notified` is called.

## Definition of Done

- [x] Integration/manual test run — N/A (purely internal SQLite change, verified by real-DB tests)
- [x] User-visible outcome: prevents the flood of stale notifications that appeared after every MCP restart — not applicable to verify directly, but the root cause query behavior is confirmed by unit tests
- [x] Side-effect audit: no floods, no unscoped writes, no unintended volume — this fix reduces unwanted writes
- [x] External service calls: none; this is a SQLite-only change

Closes #1432

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -6670,6 +6670,11 @@ async def handle_write_result(args: dict) -> list[TextContent]:
             result_summary=(text[:200] if text else None),
             stop_reason="end_turn",
         )
+        # Mark notified immediately so the reconciler's startup sweep does not
+        # re-enqueue this session on the next MCP restart.  The startup sweep
+        # queries for completed/dead rows where notified_at IS NULL — leaving it
+        # NULL here is what caused the April 4 flood (issue #1432).
+        _session_store.set_notified(task_id)
     except Exception as exc:
         log.warning(f"write_result auto-unregister failed for task_id={task_id!r}: {exc}")
 
--- a/tests/unit/test_mcp_server/test_write_result_set_notified.py
+++ b/tests/unit/test_mcp_server/test_write_result_set_notified.py
@@ -1,0 +1,381 @@
+"""
+Unit tests for the reconciler flood fix (issue #1432).
+
+Bug: write_result called session_end() but never called set_notified(), leaving
+notified_at=NULL permanently.  On every MCP restart the startup sweep re-enqueued
+every completed session with notified_at IS NULL — causing the April 4 flood of 30
+stale notifications.
+
+Fix: handle_write_result now calls set_notified(task_id) immediately after
+session_end() succeeds.
+
+These tests cover the three session-completion paths:
+
+  1. Normal path  — write_result is called:
+       session_end() + set_notified() both fire inside write_result.
+       Startup sweep sees notified_at IS NOT NULL → skips. No flood.
+
+  2. Crash path — agent dies before write_result:
+       write_result never called → session_end and set_notified never called by
+       write_result.  Reconciler detects dead session → calls
+       _enqueue_reconciler_notification → sets notified_at.  This existing path
+       is unchanged; these tests verify write_result does not interfere.
+
+  3. Edge case — set_notified raises after session_end succeeds:
+       The exception is swallowed by the surrounding try/except so the inbox
+       file write still completes.  The _inbox_already_has_agent guard prevents
+       double-enqueue on next restart.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch, call
+
+import pytest
+
+# ---------------------------------------------------------------------------
+# Path setup: ensure src/mcp and src/agents are importable
+# ---------------------------------------------------------------------------
+
+_ROOT = Path(__file__).parents[4]
+for _p in [str(_ROOT / "src" / "mcp"), str(_ROOT / "src" / "agents"),
+           str(_ROOT / "src"), str(_ROOT / "src" / "utils")]:
+    if _p not in sys.path:
+        sys.path.insert(0, _p)
+
+import src.mcp.inbox_server  # noqa: F401 — pre-load so patch.multiple resolves it
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _make_dirs(tmp_path: Path):
+    """Return (inbox, outbox, sent, sent_replies) directories under tmp_path."""
+    inbox = tmp_path / "inbox"
+    outbox = tmp_path / "outbox"
+    sent = tmp_path / "sent"
+    sent_replies = tmp_path / "sent-replies"
+    for d in (inbox, outbox, sent, sent_replies):
+        d.mkdir(parents=True, exist_ok=True)
+    return inbox, outbox, sent, sent_replies
+
+
+def _mock_session_store() -> MagicMock:
+    """Return a mock _session_store with all relevant methods."""
+    store = MagicMock()
+    store.session_end.return_value = None
+    store.set_notified.return_value = None
+    return store
+
+
+# ---------------------------------------------------------------------------
+# Path 1: Normal path — write_result called successfully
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Review of PR #1433 — reconciler flood fix

### 1. Placement of `set_notified` inside the try block

Correct semantic. Both `session_end` and `set_notified` are inside the same try block, so they succeed or fail together. If `session_end` raises (e.g. DB locked), `set_notified` is skipped — the session stays with `notified_at IS NULL`, the reconciler's startup sweep will catch it, and the user will still be notified via the crash path. This is the right behavior: marking a session notified before confirming it's been properly ended would hide it from the reconciler permanently.

One thing worth noting: `set_notified` raising after `session_end` succeeds (the edge case) would skip `set_notified` since both are in the same try. The PR correctly documents this window and relies on `_inbox_already_has_agent` as the belt-and-suspenders guard. This is an acceptable tradeoff — the window is a single SQLite write latency — but if extra resilience is wanted in future, `set_notified` could be wrapped in its own narrow try/except. Not required for this fix.

### 2. The April 4 root cause comment

The comment is accurate. It identifies the exact query predicate (`notified_at IS NULL`), the mechanism (startup sweep re-enqueues every completed session on restart), and attributes it to issue #1432. The comment is tight and informative without being over-explained. No corrections needed.

### 3. `TestStartupSweepSkipsNotifiedSessions` — real SQLite DB

This is the right call. The test uses a real SQLite database via `session_store.init_db()` and exercises the actual SQL query in `get_unnotified_completed`. It verifies the root cause directly: a session that has gone through `session_end` + `set_notified` disappears from the sweep, while one that has only gone through `session_end` (old behavior) still appears. Using a mock here would only verify call ordering, not that the query logic is correct — this test catches a real regression.

The companion test (`test_unnotified_session_still_returned_before_set_notified`) documents the before-fix state, which is valuable as a regression anchor.

### 4. State machine diagram in PR body

All three paths are covered and clearly labeled:
- Normal path (before/after) — shows the missing `set_notified` call as the explicit bug site
- Crash path — shows the reconciler path is unchanged and still owns `set_notified` for dead sessions
- Edge case — documents the microsecond window with the `_inbox_already_has_agent` guard explanation

The diagram is accurate. The edge case section is particularly well done — it names the guard by function and explains why it's sufficient protection.

- The test file imports `src.mcp.inbox_server` at module level to force pre-load before `patch.multiple` resolves. This pattern is correct but a brief inline comment explaining why would help future readers.
- The `path=db_path` argument style in `TestStartupSweepSkipsNotifiedSessions` assumes `session_store` supports an explicit path override — presumably tested in CI, so no concern.

### Summary

The fix is minimal, correct, and well-tested. The one-line change directly closes the gap between "result delivered" and "notified_at stamped". The test suite covers the normal path, crash path, and edge case at both the mock level (call ordering) and the real-SQLite level (query correctness). Ready to merge.

PR #1435, opened 2026-04-04
fix: reviewer prompt anchoring — diff-first, independent analysis before engineer briefing

sayhar:
## Problem

Reviewer subagents were anchored to the engineer's framing. The prompt passed the engineer's briefing first, so reviewers read the PR description and then verified it — catching internal inconsistencies well but missing bugs the engineer didn't think of. This is a systematic bias: a reviewer who reads the conclusion before the evidence will unconsciously filter the evidence to match.

Identified through a meta-review of PR #1433. Fixes #1434.

## What changed

All three reviewer prompt call sites in `sys.dispatcher.bootup.md` are updated to enforce a diff-first reading order:

1. **ENGINEER→REVIEWER routing** (main subagent_result handler)
2. **Deletion-confirmed engineer→reviewer routing** (callback_data handler)
3. **`/re-review` command docs** (pseudocode + explanation)

The PR review flow description (step 3) is also updated to state that the reviewer reads the diff cold first.

## Before → After flow

```
BEFORE:
  Dispatcher spawns reviewer with:
    [engineer's briefing]
    [instruction to post gh pr review]
  Reviewer reads briefing → anchored frame → reads diff through that lens

AFTER:
  Dispatcher spawns reviewer with:
    [instruction to run gh pr diff and note findings independently]
    [checklist of things to always verify]
    [instruction to post gh pr review with plain-English verdict]
    [engineer's briefing — last]
  Reviewer forms independent view → then compares against engineer's claims
```

## Checklist added to every reviewer prompt

- Verify argument types on store/DB/MCP method calls
- Check for duplicate test class names / shadowing
- Verify tests exercise actual before-state (not just assert it in comments)
- Validate "N pre-existing failures" claims by running `uv run pytest --tb=no -q` directly

## Verdict format change

Reviewers are now required to post findings in plain English — no function names, file paths, or code terms. Each finding must state what it means operationally. This makes verdicts readable by the user on mobile without needing to cross-reference the diff.

## What is not changed

- No Python code changes. This is a prompt-only update.
- The `gh pr review --comment` constraint (never `--approve` or `--request-changes`) is preserved at all sites.
- The reviewer subagent type (`lobster-generalist`) is unchanged.
- The dedup check in the main routing block is unchanged.

## Tests run
- [x] Verified all four edit sites applied correctly via Python string-match assertions — 8/8 checks passed
- [ ] Live reviewer run — not applicable for this PR (prompt-only change; reviewer behavior is verified by reading the prompt, not by running a test suite against it)

## Manual test
N/A — this change is entirely internal to the dispatcher prompt. No external service calls, no file I/O affecting runtime state, no systemd units. The effect is observable only when the next engineer PR triggers the reviewer flow.

## Definition of Done
- [x] All three call sites updated to diff-first structure
- [x] Engineer's briefing appears last in every reviewer prompt
- [x] Plain-English verdict requirement added
- [x] `/re-review` docs updated to reference same approach
- [x] PR flow description updated
- [x] Side-effect audit: no floods, no writes, no external calls — prompt text only

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -455,10 +455,26 @@ Background subagents call `write_result(task_id, chat_id, text, ...)`, which dro
                    prompt=(
                        f"---\ntask_id: {reviewer_task_id}\nchat_id: {msg['chat_id']}\n"
                        f"source: {msg.get('source', 'telegram')}\n---\n\n"
-                       f"Review PR {pr_url} and post findings:\n"
-                       f"  gh pr review <N> --repo {pr_repo} --comment --body \"PASS/NEEDS-WORK/FAIL: ...\"\n"
-                       f"Use --comment only (never --approve or --request-changes — same token = self-review error).\n"
-                       f"After posting, call write_result with a short verdict (1-3 sentences).\n\n"
+                       f"Review PR {pr_url} and post findings as a GitHub comment.\n\n"
+                       f"REVIEWER PROCESS (follow this order exactly):\n"
+                       f"1. Run: gh pr diff {pr_number} --repo {pr_repo}\n"
+                       f"   Read the diff cold. Before reading anything else, note independently:\n"
+                       f"   - What could go wrong with this change?\n"
+                       f"   - What edge cases are not covered?\n"
+                       f"   - What would you want tested?\n\n"
+                       f"2. Then read the engineer's briefing below.\n"
+                       f"   Compare what you found against what the engineer flagged.\n"
+                       f"   A good review catches what the engineer didn't think of.\n\n"
+                       f"ALWAYS CHECK:\n"
+                       f"- For any store/DB/MCP method call: do the argument types match what the method actually expects?\n"
+                       f"- Test structure: duplicate class names? Any test classes unreachable due to shadowing?\n"
+                       f"- Do tests exercise the actual before-state, or just assert it in comments?\n"
+                       f"- \"N pre-existing failures\" claims: run `uv run pytest --tb=no -q` yourself and verify the count\n\n"
+                       f"POST your review as a GitHub comment:\n"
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Cold diff reading (before briefing):** The structural change is clear — engineer briefing moves from first to last position in both reviewer prompt call sites, with a diff-first instruction and checklist prepended.

**Variable scope — Site 1 (ENGINEER→REVIEWER routing):** pr_number and pr_repo are extracted from the PR URL just above the Task() call. Both are in scope. pr_number resolves to a bare integer string (e.g. '1435'), not a URL — correct for the gh pr diff and gh pr review commands.

**Variable scope — Site 2 (deletion-confirmed callback):** Same extraction pattern — pr_number and pr_repo are parsed from pr_url immediately before the Task() call. Both in scope, same bare-number format. Clean.

**One thing the engineer did not flag:** The /re-review handler is still pseudocode comments — the diff updates those comments to describe the new approach, but there is no actual Task() call implementation at that site. This is not a regression (it was already pseudocode before this PR), but it means the /re-review command will not actually work until someone writes the implementation. The comments now correctly describe what that implementation should do, which is an improvement, but a user triggering /re-review today would get no reviewer spawned. Recommend filing a follow-up issue.

**Checklist items:** The four 'always check' items (type mismatches, test shadowing, before-state coverage, pre-existing failure counts) are primarily relevant for code PRs, not prompt-only changes. Including them unconditionally adds noise for prompt PRs. Minor — not a blocker.

**No tests needed:** Correct call by the engineer. This is a prompt-only change; behavior is verified by reading, not running.

**Verdict:** PASS with a note. The core change is correct — both live call sites are properly updated, variables are in scope, and the diff-first ordering achieves its stated goal. The /re-review pseudocode-only state should be tracked as a follow-up.

PR #1447, opened 2026-04-08
feat(gmail): use caller chat_id instead of read_owner() for multi-user support

aeschylus:
## Summary

- Updates Gmail skill behavior docs to use `message_chat_id` (the caller's Telegram chat_id) instead of hardcoded `read_owner()` when fetching Gmail tokens
- Maintains backward compatibility via fallback to `read_owner()` for single-user / legacy installs where no `chat_id` is in context
- Fixes silent failures for myownlobster.ai users who aren't the owner — their tokens exist in `gmail-tokens/{chat_id}.json` but the skill was always loading the owner's token file instead

## Files changed

- `lobster-shop/gmail/behavior/system.md` — auth check, `get_recent_emails`, `search_emails` patterns updated
- `lobster-shop/gmail/context/reference.md` — same pattern + clarified user ID convention

## Why this matters

The VPS Gmail client (`client.py`) and token store (`token_store.py`) already support per-user tokens keyed by `chat_id`. Only the skill behavior documentation was hardcoding `read_owner()`. This change unblocks multi-user Gmail on any myownlobster.ai instance.

## Test plan

- [ ] As owner: verify "check my email" still works (fallback path)
- [ ] As non-owner authed user: verify their own Gmail token is used, not owner's
- [ ] As unauthenticated user: verify consent link is generated correctly

### DIFF
--- a/lobster-shop/gmail/behavior/system.md
+++ b/lobster-shop/gmail/behavior/system.md
@@ -11,11 +11,25 @@ import sys
 import os
 sys.path.insert(0, os.path.expanduser("~/lobster/src"))
 from integrations.gmail.token_store import load_token
-from mcp.user_model.owner import read_owner
 
-owner = read_owner()
-OWNER_USER_ID = owner.get("owner", {}).get("telegram_chat_id", "")
-token = load_token(OWNER_USER_ID)
+# MULTI-USER: always prefer chat_id from the incoming message context.
+# Fall back to read_owner() only for single-user / legacy installs where
+# the message doesn't carry a chat_id (e.g. scheduled jobs).
+#
+# In the Lobster dispatcher the message looks like:
+#   message["chat_id"]  — Telegram chat_id of the user who sent the message
+#
+# Usage pattern (the subagent receives chat_id as a parameter):
+#   USER_ID = message_chat_id or _fallback_to_owner()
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one open question

**Correctness:** The fix correctly targets the root cause — skill behavior docs were hardcoding `read_owner()` while the token store already supported per-user keys. The fallback chain (caller `chat_id` → owner fallback) is sound and preserves backward compatibility for single-user installs and scheduled jobs.

**Open question — `message_chat_id` injection:** The code snippets in `system.md` and `reference.md` reference `message_chat_id` as a free variable (e.g. `USER_ID = str(message_chat_id) if message_chat_id else _fallback_to_owner()`). This works only if the dispatcher or skill runner actually injects `message_chat_id` into the execution namespace when these skill snippets run. If that injection is already guaranteed by the skill system's context-passing convention, this is a non-issue. If not, non-owner users will silently fall through to the owner fallback and the fix won't take effect. Worth confirming the injection path is wired before closing.

**Scope:** Changes are documentation/behavior-spec only (no Python source changed). The VPS `client.py` and `token_store.py` are correctly noted as already multi-user capable.

Conditional PASS — confirm `message_chat_id` is injected by the skill runner.

PR #1451, opened 2026-04-08
feat: add POST /api/webhooks/awp-intake endpoint

aeschylus:
## Summary

AWP Typeform submissions were arriving with no programmatic intake path into Lobster — the Apps Script had nowhere to POST. This adds the missing webhook endpoint so intake submissions land directly in the Lobster inbox as structured messages.

- New route `POST /api/webhooks/awp-intake` in `inbox_server_http.py`
- Auth uses `AWP_IMPORT_TOKEN` env var; the Apps Script side stores this as `LOBSTER_SECRET`
- Validates bearer token, accepts the AWP submission JSON, and writes an atomic inbox message (`type: awp_intake`, `source: awp`) to `~/messages/inbox/`
- Returns `{"status": "ok", "message_id": "<id>"}` on success; 401 on bad auth, 400 on bad JSON, 500 on write failure
- If `AWP_IMPORT_TOKEN` is not set, the endpoint returns 401 and logs an error (fail-safe, no panic)

No other routes or auth paths are changed. The existing `_is_authorized_internal` function and the LOBSTER_INTERNAL_SECRET-gated endpoints are untouched.

## Functional patterns used

Pure functions for auth (`_is_authorized_awp`) and field extraction. Atomic write via tmp-file + rename (same pattern as the calendar/gmail token endpoints). All side effects isolated to the try/except write block.

## Tests run

- [x] `uv run ruff check src/mcp/inbox_server_http.py` — clean, no errors
- [ ] Live curl test against running server — blocked: requires `AWP_IMPORT_TOKEN` to be set and server restarted; confirmed after MCP restart below

## Manual test

Endpoint goes live after MCP server restart (`~/lobster/scripts/restart-mcp.sh`), which is run as the final step of this task. The new route is a pure addition to the HTTP dispatch table — no startup logic, service files, or cron entries changed.

### DIFF
--- a/src/mcp/inbox_server_http.py
+++ b/src/mcp/inbox_server_http.py
@@ -212,6 +212,12 @@ _TOKEN_FILE_MODE: int = stat.S_IRUSR | stat.S_IWUSR
 
 _INTERNAL_SECRET: str = os.environ.get("LOBSTER_INTERNAL_SECRET", "").strip()
 
+# AWP_IMPORT_TOKEN is used by the AWP webhook endpoint.
+# In the Google Apps Script that sends intake data, this is stored as LOBSTER_SECRET.
+_AWP_IMPORT_TOKEN: str = os.environ.get("AWP_IMPORT_TOKEN", "").strip()
+
+_INBOX_DIR: Path = _MESSAGES_DIR / "inbox"
+
 
 def _is_authorized_internal(request: Request) -> bool:
     """Return True if the request carries a valid LOBSTER_INTERNAL_SECRET."""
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Correctness:** The endpoint correctly validates auth, parses JSON, extracts fields, builds a well-structured inbox message, and uses the established tmp-file + atomic rename pattern. Auth check is clean and properly fails safe (returns 401) when `AWP_IMPORT_TOKEN` is unset.

**Safety:** One minor note: `_TOKEN_FILE_MODE` (0o600, owner r/w only) is reused from the token-file pattern. For inbox messages this is more restrictive than necessary (the dispatcher reads inbox files as the same user, so it works, but the intent differs). Not a bug, just a semantic mismatch worth noting. No injection risk — the full body is stored as `payload` without re-execution, and the summary fields are `str().strip()` coerced. The `chat_id: 0` sentinel is appropriate for non-Telegram sources.

**Scope:** Routing is properly gated before the `/mcp` path check, so no existing endpoints are affected. No service files, cron, or startup logic touched. The live-curl test is deferred to post-restart, which is acceptable per the PR description.

No blocking issues.

aeschylus:
Reviewed and ready to merge — the AWP intake endpoint implementation looks solid: proper bearer token auth, atomic tmp→rename write, fail-safe when `AWP_IMPORT_TOKEN` is unset, and no changes to existing routes. However this PR currently has a **merge conflict** (`mergeable_state: dirty`) against `main`. The branch needs to be rebased or updated before it can be merged. Once the conflict is resolved, this is safe to squash-merge.

PR #1463, opened 2026-04-10
fix(lobstertalk): hot-mode systemd-run scope, cooldown traffic counting, and LOBSTER_NAME identity

sayhar:
## What this PR does

Two orthogonal fixes folded into one branch:

**1. Hot-mode re-triggering and traffic counting (original scope)**

- Hot-mode uses `sudo systemd-run` (system scope) instead of `systemd-run --user` because this script runs inside a systemd system service where no D-Bus user session is available. `--user` fails with \"Failed to connect to bus: No medium found\".
- Hot-mode cooldown tracks `len(all_messages)` (total server traffic) rather than `received_count`. When Lobster sends a message, the echo comes back from the server with `received_count = 0` (filtered as self-echo), but the conversation is still active. Total traffic keeps hot mode alive correctly during active exchanges.

**2. LOBSTER_NAME as canonical bot-talk sender identity (folded from #1465)**

Every Lobster instance previously had its identity hardcoded as \"SaharLobster\" in `bot_talk_mirror.py` and the deprecated `lobstertalk-incoming-handler.md` task spec. `BOT_TALK_SENDER` was a partial fix — an undocumented override that a second instance had to know about.

YAGNI: there is one Lobster per machine. `LOBSTER_NAME` is the single place to declare instance identity. `BOT_TALK_SENDER` is gone (not deprecated — removed). Installs that had `BOT_TALK_SENDER` set should rename it to `LOBSTER_NAME`.

**What changed:**
- `src/mcp/bot_talk_mirror.py`: `BOT_TALK_SENDER` variable replaced with `LOBSTER_NAME`; all six references updated; fallback is `"MyLobster"` (generic, not a real name)
- `config/config.env.example`: `LOBSTER_NAME=MyLobster` placeholder added above the bot-talk section
- `scheduled-tasks/tasks/lobstertalk-incoming-handler.md`: two hardcoded `"SaharLobster"` strings replaced with `$LOBSTER_NAME`

## What is out of scope

`LOBSTER_NAME` is not wired into any other subsystems (Telegram display name, memory labels, etc.). Strictly bot-talk identity resolution only.

## Functional patterns

Pure sequential `or` composition for the resolution chain — same pattern as `BOT_TALK_TOKEN` and `BOT_TALK_HTTP_URL`. No mutation, no branching logic added.

## Tests run
- [x] Static review of `LOBSTER_NAME` resolution chain — logic is structurally identical to the existing `BOT_TALK_TOKEN` fallback; no new branches
- [ ] Live bot-talk round-trip — skipped: requires a second running Lobster instance

## Manual test
N/A — this change is config resolution at module load time. No systemd, no external service calls, no file I/O at runtime. The only observable effect is the sender name in bot-talk HTTP payloads, which requires a live bot-talk server to verify.

## Definition of Done
- [x] External service calls: none added; existing bot-talk calls unchanged
- [x] BOT_TALK_SENDER removal is backwards-incompatible only for installs that relied on it — migration path is rename to LOBSTER_NAME

### DIFF
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -96,3 +96,7 @@ scripts/periodic-self-check.sh:Personal name (drew)
 # number "122.0.0.0" that looks like an IP address — it is not an IP, it is a browser
 # version string used to avoid bot detection in DuckDuckGo HTML requests
 lobster-shop/prospect-enrichment/bin/find_supply_chain_contacts.py:IP address
+
+# lobstertalk_unified.py uses the bot-talk API server IP as a fallback default
+# for BOT_TALK_URL env var — not a secret, same pattern as bot-talk-check-dispatch.sh
+scheduled-tasks/lobstertalk_unified.py:IP address
--- a/.githooks/security-allowlist.txt
+++ b/.githooks/security-allowlist.txt
@@ -96,7 +96,3 @@ scripts/periodic-self-check.sh:Personal name (drew)
 # number "122.0.0.0" that looks like an IP address — it is not an IP, it is a browser
 # version string used to avoid bot detection in DuckDuckGo HTML requests
 lobster-shop/prospect-enrichment/bin/find_supply_chain_contacts.py:IP address
-
-# lobstertalk_unified.py uses the bot-talk API server IP as a fallback default
-# for BOT_TALK_URL env var — not a secret, same pattern as bot-talk-check-dispatch.sh
-scheduled-tasks/lobstertalk_unified.py:IP address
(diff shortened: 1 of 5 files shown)


sayhar:
Two concerns before this can merge:

1. **No hardcoded IP**: The bot-talk server IP (46.224.41.108) must not be hardcoded in the script. Use an env var (e.g. BOT_TALK_URL) exclusively — no fallback hardcoded IP. Remove the allowlist entry too.

2. **Not wired up by default**: Most Lobster instances shouldn't use bot-talk yet. This shouldn't be auto-wired. Require explicit opt-in (e.g. check a LOBSTER_ENABLE_BOTTALK env var or require manual setup).

sayhar:
Fixed: removed hardcoded IP (BOT_TALK_URL must now be set explicitly), added LOBSTER_ENABLE_BOTTALK=true opt-in gate.

PR #1482, opened 2026-04-12
fix: verify MCP SSE endpoint before launching Claude (prevents tool-load race)

aeschylus:
## What this fixes

When the dispatcher restarts, `launch_claude()` kills the old `inbox_server.py` process, then polls `http://localhost:8766/mcp` until it gets any non-000 HTTP response before launching Claude. The problem: a 406 or other error from the **dying old server's final response** satisfies that check. Claude then starts and attempts its MCP SSE connection, but the **new server's SSE endpoint isn't ready yet** — the connection fails permanently for that session, and the dispatcher runs the entire session without any `mcp__lobster-inbox__*` tools.

Evidence from logs: Claude started at 03:29:16, MCP server reported "listening" at 03:29:20 — 4 seconds after Claude had already connected (and failed to find a working SSE endpoint).

## The fix

Two-phase readiness check before `log "Starting fresh session"`:

**Phase 1** (existing gap, now added): poll `http://localhost:{port}/mcp` for any non-000 HTTP code, up to 15 seconds. This catches the case where the MCP server process hasn't started at all yet.

**Phase 2** (the core fix): send a proper MCP `initialize` request and require a `200` response, up to 10 more seconds. This confirms the SSE endpoint is fully initialised and ready to accept Claude's connection — not just that some process is listening on the port.

```
Before:
  kill old MCP → poll for any HTTP response → launch Claude
                          ↑
                   406 from dying old server passes here
                   Claude starts, SSE connect fails permanently

After:
  kill old MCP → poll for any HTTP response → POST MCP initialize → require 200 → launch Claude
                                                                          ↑
                                              new server must be fully ready before Claude starts
```

## Scope

Only `scripts/claude-persistent.sh` is changed. The MCP server itself, the dispatcher loop, message routing, and all other startup logic are untouched.

## Tests run

- [x] `bash -n scripts/claude-persistent.sh` — syntax check passes, no errors
- [ ] Live restart test — blocked: requires a running Lobster instance with systemd; cannot test in this environment. The readiness check degrades gracefully: if the SSE endpoint doesn't return 200 within 10s, Claude launches anyway with a warning log (same behavior as current timeout path).

## Manual test

N/A for automated tests. This is an infrastructure timing fix — the correct test is a live restart observation (start the service, watch logs for "MCP SSE endpoint confirmed ready" before "Starting fresh session"). The fallback path (launch-anyway after timeout) preserves existing behavior if the check fails.

## Definition of Done
- [ ] Integration/manual test run — blocked: live systemd environment required; see Tests run above
- [x] User-visible outcome: fix prevents the "no MCP tools" session — verified by log sequence analysis
- [x] Side-effect audit complete: no floods, no unscoped writes, adds at most ~45s to restart time in worst case (Phase 1: up to 15s + Phase 2: up to 10 retries × ~3s each)
- [x] External service calls: none — curl hits localhost only

### DIFF
--- a/scripts/claude-persistent.sh
+++ b/scripts/claude-persistent.sh
@@ -472,6 +472,53 @@ launch_claude() {
     # by design — it reads CLAUDE.md, enters the loop, and processes messages.
     # Any persistent state lives in canonical memory files, not conversation history.
     local claude_exit_code=0
+
+    # Wait for the MCP server to be ready before launching Claude.
+    # kill_orphaned_mcp_processes() may have just SIGTERM'd the old MCP server
+    # (managed by systemd). Systemd restarts it within ~2 seconds, but if Claude
+    # initializes MCP connections before the server is back up, the tools fail to
+    # load and the dispatcher runs without lobster-inbox tools for the entire session.
+    local mcp_port="${LOBSTER_MCP_PORT:-8766}"
+    local mcp_ready=false
+    for i in {1..15}; do
+        local http_code
+        http_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 1 \
+            "http://localhost:${mcp_port}/mcp" 2>/dev/null || echo "000")
+        if [[ "$http_code" != "000" ]]; then
+            mcp_ready=true
+            break
+        fi
+        sleep 1
+    done
+    if [[ "$mcp_ready" == "true" ]]; then
+        log "MCP server ready on port ${mcp_port} (HTTP reachable)"
+        # Second-phase check: verify the MCP SSE endpoint accepts proper protocol
+        # connections. The first check may have passed on the dying old server's last
+        # response. Claude's MCP client needs the new server's SSE endpoint to be
+        # fully initialised before launch or the tools are permanently unavailable.
+        local mcp_sse_ready=false
+        for i in {1..10}; do
+            local sse_code
+            sse_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 2 \
+                -H "Accept: application/json, text/event-stream" \
+                -H "Content-Type: application/json" \
+                -X POST \
+                --data '{"jsonrpc":"2.0","method":"initialize","id":1,"params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"readiness-check","version":"1.0"}}}' \
+                "http://localhost:${mcp_port}/mcp" 2>/dev/null || echo "000")
+            if [[ "$sse_code" == "200" ]]; then
+                mcp_sse_ready=true
+                log "MCP SSE endpoint confirmed ready (attempt $i)"
+                break
+            fi
+            sleep 1
+        done
+        if [[ "$mcp_sse_ready" != "true" ]]; then
+            log "WARNING: MCP SSE endpoint not ready after 10s — launching Claude anyway"
+        fi
+    else
+        log "WARNING: MCP server not responding after 15s — launching Claude anyway"
+    fi
+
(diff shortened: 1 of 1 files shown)


sayhar (commented):

## Summary

This PR correctly addresses the race condition where a dying server's final non-000 response could prematurely satisfy the existing MCP readiness poll. The two-phase check is a sound approach.

---

## Review

### 1. Does the fix correctly distinguish a dying server from a healthy new server?

Yes. The key insight is correct: the dying old server may return a 406 or similar error on its last breath, which satisfies a simple "any non-000 HTTP code" check. Phase 2 requires a `200` from a proper MCP `initialize` POST, which the old server won't return (it's shutting down) and which validates that the new server's SSE endpoint has fully initialized. The discriminator is sound.

One subtle note: the `initialize` request in Phase 2 opens an SSE stream but the check only captures the HTTP status code (via `-o /dev/null`). The server's SSE endpoint returns `200` and begins streaming — this is fine for a readiness check since `--max-time 2` will terminate the curl after 2 seconds regardless. The 200 response is the signal we need, and curl will receive it before the stream starts. No concern here.

### 2. Edge cases

**Server never starts:** Phase 1 exhausts 15 retries (15s) and falls through with a warning, then launches Claude anyway. This is the correct graceful degradation — same as current behavior.

**Timeout on Phase 2:** 10 retries × 1s sleep + 2s max-time per curl = up to ~30s worst case for Phase 2 alone. Combined with Phase 1's 15s, worst-case added latency is ~45s before the fallback fires. The PR says "at most 25s" — that's slightly optimistic (assumes the fast path for Phase 1). Not a blocker, but the comment could be more accurate.

**Partial responses / connection refused:** `curl` returns `000` for refused connections and the code treats that correctly (does not set `mcp_sse_ready=true`).

**`LOBSTER_MCP_PORT` not set:** Defaults to `8766` via `${LOBSTER_MCP_PORT:-8766}` — correct.

**Concurrent readiness checks:** `launch_claude()` is only called from the main restart loop, so no concurrency concern.

### 3. Regressions / unintended behavior

- **The `initialize` request is a live MCP call.** The readiness-check client (`name: readiness-check`) will initialize a session on the MCP server. Depending on how `inbox_server.py` handles duplicate/abandoned initialize sessions, this could leave a dangling session in the server. If the server tracks sessions and enforces `LOBSTER_MAIN_SESSION` isolation, this is benign. Worth a quick look at session lifecycle in `inbox_server.py` to confirm the orphan session is cleaned up, but likely a non-issue given the server is designed to handle reconnects.

- **Phase 2 only runs if Phase 1 passes.** If the old server is fully dead before Phase 1 starts (common case after a clean kill), Phase 1 may pass on the *new* server's startup response — but Phase 2 then correctly validates full readiness anyway. Logic is sound.

- **`mcp_ready` variable scoping:** The `local` declarations inside the loop (`local http_code`, `local sse_code`) are fine in bash; they don't reset per iteration, just limit scope to the function. No issue.

- No changes to install.sh, CLAUDE.md, or any other files. Scope is appropriately narrow.

### 4. install.sh

No update needed — this is a runtime behavior change in an existing script, not a new service, directory, or config file.

---

## Notes

- The worst-case latency comment ("adds at most 25s") in the PR body should be ~45s, but the actual happy-path latency (new server ready in ~2s) is unchanged.
- Consider adding a brief comment in the script noting that the `initialize` call may open an SSE stream and that `-o /dev/null` discards the body intentionally — helps future readers understand why the curl looks unusual.

**Verdict: PASS** — the fix is correct, the fallback preserves existing behavior, and the scope is minimal.

PR #1485, opened 2026-04-12
feat(inbox): priority inbox queue — P0-P4 ordering in check_inbox

### LINKED ISSUE #1079: feat: priority inbox queue — P0-P4 ordering in inbox_server.py check_inbox
🤖🦞 Lobster (engineer):

## What and Why

Currently `check_inbox` and `wait_for_messages` return messages in filename-sorted (i.e. arrival-time) order. This means a `compact-reminder` (which costs zero tokens to handle) can sit behind 3 queued `scheduled_reminder` messages (which each spawn a subagent), causing the dispatcher to take the expensive path first.

Ordering should be a code property, not an LLM judgment call.

## Target Priority Scheme

| Priority | Message types | Rationale |
|---|---|---|
| P0 | `compact-reminder`, `subtype=compact-reminder`, `subtype=self_check` | Dispatcher housekeeping — zero-cost, must run before anything else |
| P1 | `text`, `voice`, `photo`, `document` (from real users) | Human in the loop — latency-sensitive |
| P2 | `subagent_result`, `subagent_error` | Completing in-flight work |
| P3 | `agent_failed` | Error recovery |
| P4 | `scheduled_reminder`, `system_error`, everything else | Background / cron |

Within each priority tier, messages are ordered by timestamp (ascending, FIFO).

## Implementation

Modify `handle_check_inbox` in `src/mcp/inbox_server.py` (around line 3508). The current code:

```python
for f in sorted(INBOX_DIR.glob("*.json")):
```

Replace with a sort that reads each file's `type` (and `subtype`) and applies the priority table above before sorting by timestamp within tier.

The sort must be safe: if a file can't be parsed, assign it P4 (don't crash).

The `_CATCHUP_EXCLUDED_SUBTYPES` historical scan path (line 3470) is unaffected — priority ordering only applies to the live inbox scan.

## Acceptance Criteria

- [ ] `check_inbox` returns messages in P0→P4 order
- [ ] Within each tier, FIFO by timestamp
- [ ] Unparseable / unknown-type files default to P4 (no crash, no skip)
- [ ] Existing tests pass; new unit test covers at least: mixed-type inbox returns P0 first, then P1, then P4
- [ ] No change to message schema — priority is derived at read time, not stored

## Dependencies

None. This is a pure change to `inbox_server.py`. Safe to merge at any time.

## Parent issue

Part of #1059 — Architecture: Scheduled Task System — Canonical Design

## Notes

This issue corresponds to the "priority queue in inbox_server.py" goal from the scheduling overhaul session (2026-03-29). It deliberately does NOT change how messages are written or what scripts generate them — it is purely a read-side ordering fix.

sayhar:
## What problem this solves

Closes #1079

The dispatcher processes messages in filename-sorted (arrival-time) order. This means a `compact-reminder` — which takes zero tokens to handle — can sit behind 3 queued `scheduled_reminder` messages that each spawn a full subagent, causing the dispatcher to take the expensive path first. Ordering was an LLM judgment call when it should be a code property.

## What changed

`check_inbox` now returns messages in P0→P1→P2→P3→P4 order, with FIFO by timestamp within each tier:

| Priority | Message types | Rationale |
|---|---|---|
| P0 | `compact-reminder`, `self_check` subtypes | Dispatcher housekeeping — zero-cost, must run first |
| P1 | `text`, `voice`, `photo`, `document` | Human in the loop — latency-sensitive |
| P2 | `subagent_result`, `subagent_error` | Completing in-flight work |
| P3 | `agent_failed` | Error recovery |
| P4 | `scheduled_reminder`, `system_error`, everything else | Background / cron |

Priority is derived at read time from `type` and `subtype` fields — nothing is stored in message files. Unreadable/unparseable files default to P4 silently.

The `since_ts` historical scan path is entirely unchanged — priority ordering only applies to the live inbox scan.

## Implementation

Two pure functions added before `handle_check_inbox`:
- `_inbox_priority(msg)` — maps message fields to tier 0-4
- `_inbox_sort_key(filename, priority, ts_epoch)` — returns `(priority, epoch, filename)` for stable sort

The live inbox scan becomes a two-pass approach: read all files and score them, sort by key, then process in priority order. The existing pre-processors (`/report` command handler, `subagent_recovered` notifier) are preserved intact in the processing pass.

**Functional patterns used:** pure functions for priority/sort-key derivation, immutable sort key tuple, separation of scoring from processing.

## Flow (before/after)

Before:
```
inbox files → sort by filename → process in arrival order
```

After:
```
inbox files → score each (priority + timestamp) → sort by (P, ts, filename) → process in priority order
```

No state transitions or message schema changes. This is purely a read-side ordering fix.

## Tests run
- [x] `uv run pytest tests/unit/test_mcp_server/test_priority_inbox_queue.py -v` — 22 passed, 0 failed
- [x] `uv run pytest tests/unit/test_mcp_server/ --tb=short` — 682 passed, 70 failed (all 70 are pre-existing `_DEBUG_MODE` attribute failures in `test_debug_alerts.py` and `test_write_observation.py` unrelated to this change)

## Manual test

N/A — this change is entirely internal to the read path of `check_inbox`. No external services, no file I/O side effects, no Telegram output. The dispatcher's observable behavior is that it processes compact-reminders before user messages before scheduled jobs — this is verifiable in logs if a mixed inbox occurs, but requires no production intervention to confirm correctness.

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4089,6 +4089,77 @@ def _enqueue_recovery_notification(msg: dict) -> None:
         log.error(f"subagent_recovered: failed to enqueue recovery notification: {exc}", exc_info=True)
 
 
+# ---------------------------------------------------------------------------
+# Priority inbox queue — P0-P4 ordering (issue #1079)
+# ---------------------------------------------------------------------------
+
+# P0: dispatcher housekeeping — zero-cost, must run before everything else
+_INBOX_P0_TYPES: frozenset[str] = frozenset()
+_INBOX_P0_SUBTYPES: frozenset[str] = frozenset({"compact-reminder", "self_check"})
+_INBOX_P0_TEXT_PREFIXES: tuple[str, ...] = ("compact-reminder",)
+
+# P1: real-user messages — latency-sensitive
+_INBOX_P1_TYPES: frozenset[str] = frozenset({"text", "voice", "photo", "document"})
+
+# P2: completing in-flight work
+_INBOX_P2_TYPES: frozenset[str] = frozenset({"subagent_result", "subagent_error"})
+
+# P3: error recovery
+_INBOX_P3_TYPES: frozenset[str] = frozenset({"agent_failed"})
+
+# P4: background / cron — everything else
+_INBOX_P4_DEFAULT: int = 4
+
+
+def _inbox_priority(msg: dict) -> int:
+    """Return the priority tier (0=highest, 4=lowest) for an inbox message.
+
+    Priority is derived from message type and subtype at read time — nothing
+    is stored in the message file itself. Unrecognised types default to P4.
+    """
+    msg_type = msg.get("type", "")
+    msg_subtype = msg.get("subtype", "")
+    msg_text = msg.get("text", "")
+
+    # P0: dispatcher housekeeping
+    if msg_subtype in _INBOX_P0_SUBTYPES:
+        return 0
+    if msg_type in _INBOX_P0_TYPES:
+        return 0
+    # compact-reminder messages are sometimes typed as "text" with a specific prefix
+    if any(msg_text.startswith(p) for p in _INBOX_P0_TEXT_PREFIXES):
+        return 0
+
+    # P1: real-user messages
+    if msg_type in _INBOX_P1_TYPES:
+        return 1
+
+    # P2: completing in-flight subagent work
+    if msg_type in _INBOX_P2_TYPES:
+        return 2
+
+    # P3: error recovery
+    if msg_type in _INBOX_P3_TYPES:
+        return 3
+
+    # P4: everything else (scheduled_reminder, system_error, unknown, ...)
+    return _INBOX_P4_DEFAULT
+
+
+def _inbox_sort_key(
+    f_name: str, priority: int, ts_epoch: float | None
+) -> tuple[int, float, str]:
+    """Return (priority, timestamp_epoch, filename) for stable ordering.
+
+    Lower priority tier = earlier in queue.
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**What was verified**

The priority ordering logic is cleanly separated into two pure functions before the inbox handler — one that maps a message to a priority tier (0=highest, 4=lowest), one that constructs a stable sort key. The live inbox scan was refactored from a single-pass loop into a two-pass approach: score all files first, sort, then process. The existing pre-processors (the /report handler and the recovery notifier) are preserved intact in the processing pass.

The change is correctly scoped: the since_ts historical scan path is untouched. Priority is derived at read time from the type/subtype fields — nothing is stored in message files, so there are no schema changes and no migration needed.

22 unit tests cover: all 5 priority tiers, FIFO ordering within a tier, the limit parameter respecting priority order, unreadable files defaulting to P4 silently, and the since_ts path being unaffected.

**Open design question: bot-talk messages**

Bot-talk messages arrive with source="bot-talk" and type="text". The current implementation assigns them P1 — the same tier as real user messages — because the priority function dispatches on type only, not source. The issue didn't specify an intent here.

This is a design question, not a defect. Arguments for P1: bot-talk inbound messages are cross-Lobster exchanges that may require a timely reply. Arguments for a lower tier (e.g., P2 or a new P1.5): they are machine-originated, not a human waiting at a keyboard, and bumping them below human messages would better reflect urgency. The current behavior (P1) is defensible and won't cause a regression. If you want bot-talk at a different tier, the change is a one-liner in the priority function — worth a follow-up issue rather than blocking this PR.

**No blocking issues found.**

PR #1500, opened 2026-04-12
fix(config): move timezone preference from CLAUDE.md to user config

### LINKED ISSUE #1464: Move ET timezone preference from CLAUDE.md to user config
The line `Always display times in Eastern Time (ET)` is currently hardcoded in `CLAUDE.md`, which is committed to the repository. This is a user preference that belongs in `~/lobster-user-config/agents/user.base.bootup.md` instead.

**Problem:** Any user who installs Lobster gets Eastern Time forced on them regardless of where they live. Changing it requires modifying a repo file, which creates noise in git history and breaks on upgrade.

**Fix:** Remove the timezone line from `CLAUDE.md` and add a note in the user config template (e.g. `memory/canonical-templates/user.base.bootup.md.example`) with a placeholder like:

```
Always display times in <YOUR TIMEZONE> (e.g. ET, PT, CET).
```

The install script or upgrade script should prompt for timezone preference and write it to `~/lobster-user-config/agents/user.base.bootup.md` during setup.

**Impact:** Different users can have different timezone preferences without touching the repo.

sayhar:
🤖🦞 Lobster (engineer):

## Summary

Closes #1464.

- Removes the hardcoded Eastern Time line from `CLAUDE.md` (the shared, committed repo file) and replaces it with a generic instruction that defers to the user's config file
- Adds a new `memory/canonical-templates/user.base.bootup.md` template with a `<YOUR_TIMEZONE>` placeholder and examples for ET, PT, and CET
- Updates `install.sh` (both stub-creation sites) to seed `user.base.bootup.md` from the canonical template on first install, instead of creating a blank file

**Impact:** New installs get a customizable timezone template. Existing installs are unaffected — the file already exists so it won't be overwritten. The ET preference itself is unchanged for current users (it lives in their `~/lobster-user-config/agents/user.base.bootup.md`).

## Test plan

- [ ] Fresh install: verify `~/lobster-user-config/agents/user.base.bootup.md` is created from the template (not empty)
- [ ] Upgrade on existing install: verify existing `user.base.bootup.md` is not overwritten
- [ ] Verify CLAUDE.md no longer mentions "Eastern Time" directly
- [ ] `bash -n install.sh` passes (syntax check)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -122,7 +122,7 @@ Rules are capped at 100 entries. Rules are never surfaced to the user unless exp
    If you cannot articulate what is legitimately concerning, you are being
    sycophantic. Both halves are required — this is not "pile on," it is
    "be honest first."
-5. **Always display times in Eastern Time (ET)** — Convert all UTC timestamps before sending any message. Currently EDT (UTC-4) from mid-March through early November, EST (UTC-5) otherwise. Include the offset when helpful (e.g. "2:30 PM ET"). Never send raw UTC times to the user.
+5. **Always display times in the user's preferred timezone** — Convert all UTC timestamps before sending any message. The user's timezone preference is set in their user config (`user.base.bootup.md`). Never send raw UTC times.
 6. **Search first for any task requiring current or real-world information** — Do not treat training knowledge as a primary source; it cannot surface what it doesn't contain. Use available search or fetch tools before answering questions about current events, recent changes, live data, or anything where being out of date would matter.
 
 ## Project Directory Convention
--- a/install.sh
+++ b/install.sh
@@ -390,12 +390,19 @@ if [ "$CONTAINER_SETUP" = true ]; then
         done
     fi
 
-    # Create stub user-config agent files if they don't exist
+    # Create stub user-config agent files if they don't exist.
+    # Seed from canonical-templates if a matching template exists, otherwise create empty stub.
     for stub_file in "user.base.bootup.md" "user.base.context.md" "user.dispatcher.bootup.md" "user.subagent.bootup.md"; do
         stub_dest="$USER_CONFIG_DIR/agents/$stub_file"
         if [ ! -f "$stub_dest" ]; then
-            touch "$stub_dest"
-            info "  Created stub: agents/$stub_file"
+            stub_tmpl="$INSTALL_DIR/memory/canonical-templates/$stub_file"
+            if [ -f "$stub_tmpl" ]; then
+                cp "$stub_tmpl" "$stub_dest"
+                info "  Seeded user config: agents/$stub_file"
+            else
+                touch "$stub_dest"
+                info "  Created stub: agents/$stub_file"
+            fi
         fi
     done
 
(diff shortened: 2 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

This PR moves timezone configuration from the system-level CLAUDE.md into user-managed config templates, making it properly per-user rather than hardcoded for Eastern Time.

## Completeness

**Is the ET reference fully removed from CLAUDE.md?** Yes. The hardcoded ET rule (item 5) is replaced with a generic instruction to use the user's timezone preference from their config file. The ET-specific content is gone, not just supplemented.

**Is the user config template in the right file and location?** Yes. `memory/canonical-templates/user.base.bootup.md` is the correct location — install.sh seeds this into `~/lobster-user-config/agents/user.base.bootup.md` for new installs. The timezone section is clearly marked and includes helpful examples for ET, PT, and CET.

**Does the change break anything for non-Eastern users?** No — it improves the situation. Previously all users inherited ET hardcoding from CLAUDE.md. Now each user sets their own timezone in their local config, which is injected at session start. Non-Eastern users who were already working around the ET default can now set their own timezone correctly.

**Does user.base.context.md already have a timezone field?** No — checked the canonical-templates directory. There was no pre-existing `user.base.bootup.md` template (the file is new in this PR), and `user.base.context.md` is not changed. The timezone preference lives appropriately in `user.base.bootup.md` (behavioral preferences), not context.

## One concern worth noting (non-blocking)

Existing installs will not get the timezone template seeded because install.sh only seeds files that don't yet exist. Users upgrading from an older install will have an empty `user.base.bootup.md` and will get no timezone reminder from the new CLAUDE.md generic instruction. A migration note in `upgrade.sh` or a mention in release notes would help, but this does not block the PR — it's consistent with how other user config stubs work.

## Verdict

The change is correct and complete. Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Moves timezone configuration from hardcoded ET in CLAUDE.md to a per-user template, making timezone properly configurable per install.

## Completeness check

**ET reference fully removed from CLAUDE.md?** Yes. Item 5 now says "user's preferred timezone" with a pointer to user config — not just a supplement to the ET rule.

**Template in the right file and location?** Yes. `memory/canonical-templates/user.base.bootup.md` is the correct canonical location. install.sh seeds this into `~/lobster-user-config/agents/user.base.bootup.md` for new installs. The examples (ET, PT, CET) give new users a clear starting point.

**Breaks anything for non-Eastern users?** No — this is an improvement. Previously all installs inherited ET. Now each user sets their own timezone in their local config, which is injected at session start via the bootup hook.

**Did user.base.bootup.md already have a timezone field?** No. The file `memory/canonical-templates/user.base.bootup.md` is entirely new in this PR. No pre-existing timezone field was overwritten.

## Non-blocking concern

Existing installs won't receive the seeded template — install.sh only seeds files that don't exist yet. Upgrading users will have an empty `user.base.bootup.md` and no automatic timezone reminder. A mention in `upgrade.sh` or release notes would help, but this matches how other user config stubs work and is not a blocker.

## Verdict

Correct and complete. Ready to merge.

sayhar:
Closing as duplicate of #1550. Superseded by the newer version.

PR #1504, opened 2026-04-12
fix(tests): add autouse bot-talk mirror isolation to prevent network pollution

### LINKED ISSUE #1341: bot-talk: test suite pollutes network with fake messages — need mirror isolation in conftest
## Summary

The unit test suite is posting real HTTP messages to the bot-talk network when `BOT_TALK_HTTP_URL` is configured in `~/messages/config/config.env`. This has resulted in **943 spurious test messages** being posted to the shared network (majority on March 31, 2026 — 875 messages in a single day).

## Root Cause

`tests/conftest.py` isolates all inbox/filesystem paths but does NOT mock out the bot-talk mirror functions. When `handle_send_reply` or `handle_check_inbox` are exercised in tests, they call `_mirror_outbound` / `_mirror_inbound` which fire real HTTP POSTs to `BOT_TALK_HTTP_URL` if the URL is set in config.

The test fixture `sample_text_message` uses `chat_id=123456` (a fake ID), so test-generated messages are identifiable by:
- `[OUTBOUND → TELEGRAM chat=123456]`
- `[OUTBOUND → SLACK chat=111/222/333]`
- `[INBOUND from TELEGRAM] unknown: ...`
- `[INBOUND from TELEGRAM] Test: message i1`

## Impact

- 943 spurious messages posted to the shared bot-talk network
- Albert's Lobster reads these messages as real activity
- Delete endpoint does not exist on the server (`DELETE /message/<id>` returns 404)

## Fix Needed

Add a `bot_talk` mirror mock to `tests/conftest.py`'s `isolate_inbox_server_paths` fixture (or a new autouse fixture in the mcp_server conftest) that patches `_mirror_outbound` and `_mirror_inbound` to no-ops during all tests.

Specifically, in `tests/conftest.py` or `tests/unit/test_mcp_server/conftest.py`:

```python
@pytest.fixture(autouse=True)
def isolate_bot_talk_mirror():
    """Prevent tests from posting real messages to bot-talk."""
    with patch("src.mcp.inbox_server._mirror_outbound"), \
         patch("src.mcp.inbox_server._mirror_inbound"):
        yield
```

Alternatively, patch `BOT_TALK_HTTP_URL` to `""` in the installed `bot_talk.mirror` module so the HTTP posting path is skipped entirely.

## Files

- `tests/conftest.py` — needs the autouse mock
- `tests/unit/test_mcp_server/conftest.py` — alternative location
- `src/mcp/bot_talk_mirror.py` — the module being called
- `src/mcp/inbox_server.py:4020,4196` — where mirror functions are called

🤖 Generated by Lobster (auditor subagent)

sayhar:
## Summary

- Adds `isolate_bot_talk_mirror` autouse fixture to `tests/conftest.py` that patches `bot_talk_mirror.mirror_outbound` and `bot_talk_mirror._spawn_mirror` to no-ops for every test
- Prevents real HTTP POSTs to `BOT_TALK_HTTP_URL` when tests exercise `handle_send_reply` or `handle_check_inbox`
- Tests that want to assert on bot-talk calls can request the fixture explicitly and inspect the returned `MagicMock` objects

## Root cause

The `bot_talk_mirror` functions are imported dynamically inside `inbox_server.py` function bodies. When `BOT_TALK_HTTP_URL` is configured in `~/messages/config/config.env`, every test that exercises those code paths fires a real HTTP POST to the shared bot-talk network. This caused 943 spurious messages (875 in a single day) as documented in #1341.

## Test plan

- [x] `tests/unit/test_mcp_server/test_message_tools.py` — all 46 tests pass with the new fixture in place
- [x] Existing bot-talk tests in `test_bot_talk_mirror.py` are unaffected (the pre-existing failure there is about a missing `BOT_TALK_SENDER` attribute, unrelated to this change)
- [x] Fixture gracefully degrades if `bot_talk_mirror` is not importable

Closes #1341

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,6 +24,17 @@ Do NOT add per-test mocks for:
     SCHEDULED_TASKS_LOGS_DIR, LOG_DIR
 
 These are all redirected automatically.
+
+Bot-talk mirror isolation
+--------------------------
+The ``isolate_bot_talk_mirror`` fixture (autouse=True) prevents tests from
+posting real HTTP messages to the bot-talk network. It patches
+``bot_talk_mirror.mirror_outbound`` and ``bot_talk_mirror._spawn_mirror``
+to no-ops for every test. This matches the approach described in issue #1341.
+
+Without this, tests that exercise ``handle_send_reply`` or ``handle_check_inbox``
+would fire real HTTP POSTs to ``BOT_TALK_HTTP_URL`` if it is configured in
+``~/messages/config/config.env``, polluting the shared bot-talk network.
 """
 
 import asyncio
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: PASS**

The fixture is technically correct and well-structured.

**What's right:**
- Uses `patch.object(btm, ...)` rather than patching by string path, which correctly intercepts the already-imported module object regardless of how `inbox_server.py` imports it dynamically.
- Graceful degradation on import failure is sensible — failing open rather than breaking the test suite.
- Autouse scope is appropriate: the fixture lives in `conftest.py` at the test root, so it applies to all discovered tests without requiring per-module decoration.
- Returning the mocks dict via `yield` gives test authors a clean handle for assertion-based tests, and the docstring explains the pattern clearly.
- The `sys.path` manipulation mirrors the pattern already established in conftest for the MCP dir, keeping it consistent.

**Minor observations (non-blocking):**
- The graceful degradation path (`yield {}`) means a missing or broken `bot_talk_mirror` silently skips isolation. This is probably fine since a broken import means no real calls can be made either, but it's worth noting that the degraded path is untested.
- The fixture only patches `mirror_outbound` and `_spawn_mirror`. If `bot_talk_mirror` grows additional entrypoints (e.g. a `send_to_mirror` helper), they would not be covered. Consider a follow-up to enumerate all public callables if the module expands.

These are observations, not blockers. The fix correctly addresses the 943-spurious-messages issue documented in #1341.

PR #1505, opened 2026-04-12
fix(bootup): strengthen inline catchup guard at startup step 4

### LINKED ISSUE #1171: Regression: dispatcher ran startup catchup inline again (issue #758 recurred)
## Description

Issue #758 was previously closed as fixed, but the bug reoccurred in session 20260331-002 (started 2026-03-31T03:52:49Z).

## Evidence

From \`audit.jsonl\` lines 35498-35503:
\`\`\`
03:52:49 create_session_file    -- correct (step 2a)
03:52:49 list_rules             -- correct (step 2b)
03:53:14 send_reply "warming up" -- correct (step 2d)
03:53:39 check_inbox            -- INLINE VIOLATION (should be subagent)
03:53:39 get_active_sessions
\`\`\`

No \`startup-catchup\` subagent was spawned. The dispatcher called \`check_inbox\` directly on the main thread, violating the 7-second rule and the spec in \`sys.dispatcher.bootup.md\`.

## Expected behavior

Per the bootup spec (step 4):
> Spawn the compact-catchup agent in the background to recover recent activity. WARNING: This MUST be spawned as a background subagent (run_in_background=True). Do NOT perform catchup inline.

## Root cause hypothesis

The existing WARNING text was not sufficient. The dispatcher rationalized that with many queued messages from the deadlock recovery, doing inline catchup was "faster" or "cleaner." This is a recurring compliance failure.

## Suggested fix

- Strengthen the WARNING in sys.dispatcher.bootup.md (bold, ALL CAPS, explicit consequence: "health check will restart you if you block for >10 min")
- Consider making the spec say: "If you find yourself calling check_inbox on the main thread during startup, STOP — you are about to block for 10 minutes. Spawn the subagent instead."
- Also: \`record-catchup-state.sh start\` was not called (step 3), which means health-check WFM freshness suppression was never activated.

## Context

This session started after a deadlock caused by \`sudo systemctl restart lobster-mcp-local\` invalidating the MCP HTTP session while the compact-pending gate was active. After Sahar manually removed the compact-pending file, a new session started — but the startup catchup was performed inline instead of as a subagent.

sayhar:
The prior single-line WARNING at startup step 4 was not strong enough — the dispatcher rationalized past it when under load. The first fix (the original PR commit) added a more emphatic block, but that block named specific tool call lists, time thresholds, and internal mechanism details that will become stale as the system evolves.

This PR replaces the verbose block with a three-line warning that states the prohibition, the consequence (in terms of effect, not implementation), and the "if you are about to do X — stop" redirect. No hardcoded tool names, no numeric thresholds, no quoted rationalizations.

## What changed

**Before:** 4-line block naming `check_inbox`/`Read`/`Bash`, citing the 600-second WFM freshness threshold and 10–15 minute blocking estimates, closing three specific rationalizations by name.

**After:** 3-line block — prohibition, consequence (blocks message handling, triggers watchdog restart), redirect (if you're about to call any tool for catchup on the main thread, stop; spawn the subagent instead).

The "if you are doing X, that's wrong" pattern is preserved. The implementation-specific detail that will rot is removed.

## Tests run
- [x] Documentation-only change — no automated tests applicable
- [x] Verified diff reads correctly and warning block parses as intended Markdown

Closes #1171

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -53,7 +53,12 @@ The symlink `~/.claude/` resolves to `~/lobster/.claude/` on standard installs.
     - Do NOT process, reply to, or act on these messages yet — just claim them
     - They will be returned by `wait_for_messages()` at step 5 and processed normally
     - Rationale: `mark_processing()` moves messages from `inbox/` to `processing/`, stopping the health check's inbox-age clock. Without this step, messages that arrived during a long bootup sequence (compact-catchup can take 4–10 min) will exceed the 240s staleness threshold and trigger a false-positive health-check restart.
-4. Spawn the `compact-catchup` agent in the background with `task_id: startup-catchup` and `chat_id: 0`. See agent definition at `.claude/agents/compact-catchup.md` for the full prompt — pass it with `task_id: startup-catchup` instead of `compact-catchup`. **Never do catchup inline — it violates the 7-second rule.**
+4. Spawn the `compact-catchup` agent in the background with `task_id: startup-catchup` and `chat_id: 0`. See agent definition at `.claude/agents/compact-catchup.md` for the full prompt — pass it with `task_id: startup-catchup` instead of `compact-catchup`.
+
+   > **NEVER DO CATCHUP INLINE. THIS IS A HARD RULE WITH NO EXCEPTIONS.**
+   > Startup catchup involves file I/O, inbox scanning, and summarization. On the main thread this blocks ALL new messages for 10–15 minutes. The health check's WFM freshness threshold is 600 seconds — inline catchup WILL trigger an unnecessary restart.
+   > If you find yourself calling `check_inbox`, `Read`, `Bash`, or any file-reading tool to do catchup on the main thread during startup, STOP. You are about to block for 10+ minutes. Spawn the subagent instead.
+   > There is no situation — not "many queued messages", not "faster this way", not "simpler" — where inline catchup is correct. Spawn the background subagent and proceed to step 5.
 5. Call `wait_for_messages()` to start listening.
 6. **Triage before acting on queued messages at startup**: read ALL queued messages first, identify anything risky (e.g. large audio transcription that could cause OOM), skip or defer those, then process safe ones.
 7. Resume the main loop.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: PASS**

The change is correct, proportionate, and closes the specific rationalization vectors that caused the #1171 regression.

**What's right:**
- The old single-line `Never do catchup inline` note was clearly insufficient — issue #1171 shows the dispatcher rationalized around it ("many queued messages", "faster this way"). The new block is harder to rationalize away because it names the specific consequence (600s WFM freshness threshold → restart) and enumerates the violating tool calls explicitly.
- Listing `check_inbox`, `Read`, `Bash`, and "any file-reading tool" by name removes ambiguity about what constitutes an "inline" violation.
- The explicit statement "There is no situation — not X, not Y, not Z" directly pre-empts the documented rationalizations.
- No code changes means zero regression risk.

**Observations (non-blocking):**
- The warning block is formatted as a blockquote (`>`), which renders well in Markdown but may be less visually salient in the raw text that the dispatcher actually reads. If future regressions recur, consider whether an ASCII banner or `===` underline would be more attention-grabbing in plain-text context.
- The PR description says "closes every rationalization the dispatcher has used previously" — worth keeping that list updated if new rationalizations emerge in practice.

No concerns with this change.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS (rework)

**Verdict: PASS**

The rework satisfies Sahar's feedback. The block is now 3 lines, carries no tool names, no numeric thresholds, and no quoted rationalizations.

---

**Specific checks:**

1. **"triggers a watchdog restart" — clear enough?** Yes. It names the effect (restart) and the cause (blocks message handling). The dispatcher doesn't need to know the watchdog's internal threshold to understand the consequence. Effect-based phrasing is more durable than citing a specific number that could change.

2. **"any tool to do catchup work" — unambiguous enough?** Mostly yes, with a small caveat. The phrase is intentionally broad, which is a feature — it can't be argued around by naming a tool that wasn't enumerated. The slight risk is that "catchup work" is not defined; a model could conceivably read a single file and not consider it "catchup work." In practice, the surrounding context (step 4 follows the explicit `compact-catchup` spawn instruction) makes the intent clear. Not a blocker.

3. **Sahar's original feedback — does the rework satisfy it?**
   - "Too verbose" — addressed. The old block named `check_inbox`/`Read`/`Bash`, cited the 600-second threshold, and closed three specific rationalizations. The new block is 3 lines with no implementation-specific detail.
   - "Hardcodes information about how the system works" — addressed. No tool names, no timing thresholds, no internal mechanism references.
   - The "if you are about to do X — stop" redirect is preserved, which was the structural element Sahar wanted to keep.

4. **Prior PASS review (on the original verbose commit) is now superseded.** This review supersedes that one and applies to the reworked state.

**No concerns with the reworked version.**

PR #1506, opened 2026-04-12
fix: add missing lobster-env.sh script for 'lobster env' subcommand

sayhar:
## Summary

- Adds the missing `scripts/lobster-env.sh` script that `lobster env set/get/list/edit/path` delegate to
- Before this change, all `lobster env` subcommands failed with: `Error: lobster-env script not found at .../scripts/lobster-env.sh`
- Implements full `set`, `get`, `list`, `edit`, and `path` subcommands

## How it works

- `set KEY VALUE` — upserts the key in `global.env`, replacing commented-out lines in-place so repeated calls don't accumulate duplicates
- `get KEY` — sources `global.env` and prints the live value (handles shell quoting correctly)
- `list` — lists all non-empty, non-comment keys
- `edit` — opens `global.env` in `$EDITOR`
- `path` — prints path to `global.env`

## Clarification on #1454 precedence issue

The script's header comment clarifies why `global.env` entries should override empty values in `config.env`: `lobster-router.service` loads `EnvironmentFile` entries in order, with `global.env` last, so later values win. The "edit config.env directly" workaround is only needed when `config.env` already has a **non-empty** value for that key.

The script is already made executable by the existing `chmod +x "$INSTALL_DIR/scripts/"*.sh` line in `install.sh` (line 1065), so no install.sh changes are needed.

## Test plan

- [x] `lobster env` — prints help with file path
- [x] `lobster env list` — shows non-empty keys from global.env
- [x] `lobster env get GITHUB_TOKEN` — returns the correct value
- [x] `lobster env get NONEXISTENT_KEY` — exits 1 with "(not set)" message
- [x] `lobster env path` — prints correct path

Closes #1454


sayhar:
Addressed the NEEDS-WORK feedback:

**Bug fix (the real issue #1454):** The `set` subcommand in `cmd_env()` was using `grep -qF "^${key}="` (fixed-string match) which missed commented-out lines like `#KEY=value`. Changed to `grep -qE "^#?${key}="` (regex) so that commented-out entries are matched, removed, and replaced — no more duplicate appended lines when a key exists but is commented out.

**Cleanup:** Removed `scripts/lobster-env.sh` — it was added in the previous commit but is never called. `cmd_env()` is fully self-contained inline in `src/cli`, so the standalone script was dead code.

sayhar:
Closing as duplicate of #1560. Superseded by the newer version.

PR #1509, opened 2026-04-12
fix: add log-cleanup.sh cron job to prune unbounded log directories (#1240)

### LINKED ISSUE #1240: Missing log cleanup: scheduled-jobs/logs accumulates ~40k files, no rotation for processed/ messages or audio/

sayhar:
## Summary

- Adds `scripts/log-cleanup.sh` — a daily cron job (04:00 UTC) that prunes three directories that were growing without bound
- `~/lobster-workspace/scheduled-jobs/logs/`: files older than 14 days (was ~40k files / 160 MB)
- `~/messages/processed/`: files older than 30 days (was ~17k files / 73 MB)  
- `~/messages/audio/`: files older than 7 days (was ~173 files / 71 MB)
- Adds the cron entry to `install.sh` for fresh installs
- Adds Migration 70 to `upgrade.sh` so existing installs get the cron entry on next upgrade

## Verification

Ran the script manually — removed 794 old job log files and 46 old audio files on first run. Script is safe: skips missing directories, counts deleted/remaining files, never touches recent files.

## Test plan

- [x] `bash -n scripts/log-cleanup.sh` — syntax check passes
- [x] Manual run: correctly pruned scheduled-jobs/logs (794 files removed, 919 remaining) and audio (46 files removed, 17 remaining)
- [x] messages/processed: 0 removed (none old enough, 11k files remain — 30d retention is deliberately conservative)
- [x] Script is idempotent: safe to run multiple times

Fixes #1240

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1324,6 +1324,23 @@ chmod +x "$INSTALL_DIR/scheduled-tasks/export-logs.py" 2>/dev/null || true
 success "Log export configured (runs at 03:00 UTC daily)"
 
 #===============================================================================
+# Log Cleanup (issue #1240)
+#===============================================================================
+
+step "Setting up log cleanup cron..."
+
+# log-cleanup.sh prunes three directories that grow without bound:
+#   - ~/lobster-workspace/scheduled-jobs/logs/ (keep 14 days)
+#   - ~/messages/processed/                    (keep 30 days)
+#   - ~/messages/audio/                        (keep 7 days)
+# Runs at 04:00 UTC daily (one hour after nightly consolidation).
+chmod +x "$INSTALL_DIR/scripts/log-cleanup.sh" || true
+"$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-LOG-CLEANUP" \
+    "0 4 * * * $INSTALL_DIR/scripts/log-cleanup.sh >> $HOME/lobster-workspace/logs/log-cleanup.log 2>&1 # LOBSTER-LOG-CLEANUP"
+
+success "Log cleanup configured (runs at 04:00 UTC daily)"
+
+#===============================================================================
 # Ghost Detector (agent-monitor)
 #===============================================================================
 
--- a/install.sh
+++ b/install.sh
@@ -1336,7 +1336,7 @@ step "Setting up log cleanup cron..."
 # Runs at 04:00 UTC daily (one hour after nightly consolidation).
 chmod +x "$INSTALL_DIR/scripts/log-cleanup.sh" || true
 "$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-LOG-CLEANUP" \
-    "0 4 * * * $INSTALL_DIR/scripts/log-cleanup.sh >> $HOME/lobster-workspace/logs/log-cleanup.log 2>&1 # LOBSTER-LOG-CLEANUP"
+    "0 4 * * * $INSTALL_DIR/scripts/log-cleanup.sh # LOBSTER-LOG-CLEANUP"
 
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: PASS**

The script is safe, the retention windows are reasonable, and the install/upgrade integration is correct.

**Safety analysis:**
- `find -maxdepth 1` prevents recursion into subdirectories — correct for all three target directories which are flat file stores.
- `-mtime +N` uses strict greater-than semantics, so files exactly N days old are NOT deleted. Safe edge case handling.
- Each `prune_dir` call is wrapped in `|| true` fallback logic and skips missing directories — the script cannot fail hard on a partial install.
- `set -euo pipefail` is present at the top, which is correct for a cron script where silent failures are dangerous.
- The script does not touch subdirectories (e.g. `scheduled-jobs/logs/` subfolders), which is correct — logs are flat files.

**Retention windows:**
- 14 days for job logs: reasonable; the nightly digest summarizes outputs, so raw logs past two weeks have low value.
- 30 days for processed messages: conservative and appropriate — inbox messages may be referenced during incident investigation.
- 7 days for audio: appropriate — transcriptions are persisted separately; the raw audio files are ephemeral.

**Install/upgrade integration:**
- `install.sh` adds the cron entry via `cron-manage.sh` with the `# LOBSTER-LOG-CLEANUP` idempotency marker — correct pattern matching the rest of the file.
- Migration 71 in `upgrade.sh` correctly checks for the marker before adding, guards on script existence, and increments the `migrated` counter — consistent with other migrations in the file.
- `chmod +x` is applied in both install.sh and upgrade.sh — correct.

- The script header comment says "Migration 71" but the diff shows this is the block labeled `Migration 71` in upgrade.sh while install.sh doesn't use migration numbers. The mismatch is cosmetic and doesn't affect behavior.
- The `log-cleanup.log` file itself is written to `~/lobster-workspace/logs/` and is never pruned by this script. It will grow slowly (one entry per day per pruned directory = ~4 lines/day). At that rate it stays small for years, so this is not a concern, but worth noting for completeness.

Overall: well-executed, safe, and addresses a real operational problem (794 files pruned on first run).

sayhar:
🤖🦞 Lobster (ops): Closing in favor of a revised spec. The current approach deletes logs based on age alone, but Sahar's preference is: only delete when the log directory exceeds 10 GB AND files are older than 2 months. Audio data is the exception — that can always be cleaned up aggressively. Filing a new issue.

PR #1511, opened 2026-04-12
fix(install): make config.env writes idempotent (#1214)

### LINKED ISSUE #1214: ops: install.sh config.env overwrite — make idempotent
🤖🦞 Lobster (ops):

## Problem

\`install.sh\` overwrites \`~/lobster-config/config.env\` unconditionally on every run. Re-running it would destroy the working Telegram token, Linear key, Vercel token, etc. that took real setup effort.

Sahar flagged this: install.sh should be idempotent so it can be safely re-run on an existing install.

## Fix

For each key written to \`config.env\`:
- Check if the key already exists with a non-empty value
- If yes: skip (don't overwrite)
- If no: write the default/placeholder

A helper like \`set_config_if_missing KEY VALUE\` would handle this cleanly.

The same principle applies to any other file install.sh writes that might contain user-configured values.

## Acceptance criteria

- Re-running install.sh on a configured instance leaves all existing config.env values intact
- New keys (added in upgrades) get written with defaults if absent
- upgrade.sh should also apply this pattern for any config writes in migrations

sayhar:
## Summary

- Adds a `set_config_if_missing KEY VALUE [FILE]` shell helper to `install.sh` that writes a config key only when it is absent or set to a placeholder value (`your_*_here`); skips silently if a real value already exists.
- Replaces the `cat > CONFIG_FILE` heredoc in the interactive Telegram setup block with per-key `set_config_if_missing` calls — existing tokens, user IDs, and environment flags survive a reinstall.
- Replaces the bare `echo >> CONFIG_FILE` append for `ANTHROPIC_API_KEY` with `set_config_if_missing` to prevent duplicate lines on re-runs.
- `LOBSTER_INTERNAL_SECRET` and `LOBSTER_INSTANCE_URL` already had "already set" guards and are unchanged.

Closes #1214

## Test plan

- [ ] Fresh install (no `config.env`): all keys written correctly
- [ ] Re-run installer on a configured instance, answer "Yes" to reconfiguration prompt: existing `TELEGRAM_BOT_TOKEN`, `LOBSTER_INTERNAL_SECRET`, `ANTHROPIC_API_KEY` values preserved
- [ ] Run installer in non-interactive mode when `config.env` already exists: file untouched (guarded by `if [ ! -f ... ]`)
- [ ] Placeholder value (`your_bot_token_here`) is replaced when user provides a real token


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The `set_config_if_missing` helper is well-constructed: it uses `awk` with environment variables to avoid shell-injection in the in-place replacement, correctly handles the absent/empty/placeholder/real-value cases, and is called in-place of the old destructive `cat > config.env` heredoc. The fresh-install path still writes the header comment before individual keys. The `ANTHROPIC_API_KEY` append is also correctly idempotentized.

One minor observation (non-blocking): the `printf '\n%s=%s\n'` append adds a blank line before each new key, which is fine cosmetically but slightly different from the old heredoc style. Not a correctness issue.

Test plan in the PR description covers the key scenarios. No automated tests for installer shell functions, which is typical for this codebase.

Issue #1439, opened 2026-04-06
bug: dispatcher hangs silently when MCP server restarts mid-session


sayhar:
🤖🦞 Lobster (ops):

## What happened

When the MCP server crashes and restarts (new PID), the dispatcher's open MCP connection is broken. The dispatcher continues running but issues no further tool calls — it hangs silently until the health check kills it.

## Evidence (from audit, 2026-04-06 session)

- 06:00:18 UTC: MCP server crashed and restarted (new PID)
- 06:00–06:07: dispatcher continued making non-MCP tool calls (Bash/Read/Agent)
- 06:07: last tool call of any kind
- 06:07–06:28: total silence — zero MCP connections, zero tool calls
- 06:28: health check killed the session

The 21-minute silent window is consistent with a lost MCP connection causing a stall, not a normal "LLM generation phase."

## Root cause

The MCP client (Claude Code) does not reconnect automatically when the MCP server restarts. A server restart mid-session orphans the dispatcher's connection silently — no error, no retry, no log entry.

## Why it matters

This is the root cause of a class of health-check false alarms where the dispatcher is "running" but completely frozen. The heartbeat hook (PR #1403) mitigates this by providing an external liveness signal, but does not fix the underlying reconnect gap.

## Proposed fix directions

1. **MCP client reconnect handling**: Claude Code should detect a broken MCP connection and attempt reconnect before hanging.
2. **MCP server graceful restart**: `restart-mcp.sh` already writes a warning before restarting — ensure the dispatcher receives and processes it before the connection drops.
3. **Non-MCP liveness signal**: a PreToolUse hook that writes a heartbeat file (no MCP dependency) would survive an MCP restart and let the health check distinguish "working" from "frozen."

## Related
- PR #1403 (heartbeat hook — mitigation)
- Issue #1401 (original heartbeat hook tracking)
- Issue #932 (related: MCP mid-session reconnect during deliberation)
- Issue #1187 (related: MCP HTTP session has no reconnect path after server restart)

aeschylus:
No comments indicate this MCP-restart hang bug was fixed; keeping open pending investigation and fix.

aeschylus:
[...]

PR #1512, opened 2026-04-12
fix: detect frozen dispatcher after MCP restart via thinking-heartbeat (#1439)

sayhar:
## Problem

When the MCP server crashes and restarts mid-session, the dispatcher's HTTP connection breaks silently. The dispatcher process stays alive (PID check passes), so `_write_session_lost_reminder()` suppressed the recovery hint — leaving the dispatcher frozen for up to 21 minutes until the health check killed it.

This is the root cause documented in issue #1439.

## Root cause

`_write_session_lost_reminder()` has a PID guard: if the dispatcher PID is alive, it assumes a clean Claude Code auto-update reconnect and skips the reminder. This correctly handles CC auto-updates but silently fails when MCP crashes — the PID is alive but the connection is broken.

## Fix

- **Add `_is_dispatcher_responsive()`**: reads `last_thinking_at` from `lobster-state.json` (written by `thinking-heartbeat.py` on every PostToolUse). Returns `False` (frozen suspected) when the field is absent or older than 5 minutes — meaning the dispatcher has made no tool calls since MCP restarted.
- **Update `_write_session_lost_reminder()`**: when the dispatcher PID is alive, also check responsiveness. If responsive → suppress (clean CC reconnect). If not responsive → write the reminder anyway, triggering recovery.

## Effect

Maximum silent hang after MCP restart: reduced from ~21 minutes to ~5 minutes.

The 5-minute threshold is conservative: long LLM reasoning phases can take 3-4 minutes, so we need headroom. If the threshold proves too tight we can raise it, but 5 min is already much better than 21.

## Tests

All 11 tests in `test_session_lost_reminder_pid_guard.py` pass, including:
- New: `test_writes_reminder_when_dispatcher_alive_but_frozen` — covers the issue #1439 scenario
- Updated: `test_suppresses_reminder_when_dispatcher_alive_and_responsive` — now requires fresh heartbeat (clarifies intent)

Pre-existing failing test (`test_timeout_without_hibernate_flag_does_not_write_state`) is unrelated and was failing before this PR.

## Related
- Issue #1439
- PR #1403 (thinking-heartbeat hook — provides the `last_thinking_at` signal this fix depends on)
- PR #1431 (PID-based session-lost guard — this PR extends that logic)
- PR #1446 (external WFM watchdog — complementary mitigation)

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -725,6 +725,11 @@ _TRANSIENT_MODES = {"hibernate", "starting", "restarting", "waking"}
 _RECONNECT_GRACE_SECONDS = 30 * 60  # 30 minutes
 
 
+# How long without a tool-use heartbeat before we assume the dispatcher is frozen
+# (MCP connection silently broken after MCP server restart, issue #1439).
+_FROZEN_THRESHOLD_SECONDS = 5 * 60  # 5 minutes
+
+
 def _is_dispatcher_alive() -> bool:
     """Return True if the dispatcher process recorded in dispatcher.pid is alive.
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Solid fix for issue #1439. The `_is_dispatcher_responsive()` function reads `last_thinking_at` from lobster-state.json and returns True conservatively on any error/missing data, which is the right default to avoid false positives. The threshold is 5 minutes, which is a reasonable window.

The guard in `_write_session_lost_reminder` is correctly structured: PID alive + heartbeat fresh → suppress (clean reconnect); PID alive + heartbeat stale → write reminder (frozen); PID absent → write reminder (real session loss).

Three tests cover the critical paths: clean reconnect suppresses, frozen triggers, dead PID triggers. All three paths are exercised.

Minor nit (non-blocking): the log message at line 1110 has a run of extra dashes (`— — — —`) that looks like an editing artifact. Cosmetic only.

PR #1514, opened 2026-04-12
fix(history): raise get_conversation_history text limit from 500 to 4000 chars (#1106)

### LINKED ISSUE #1106: get_conversation_history truncates long message content
## Bug

`get_conversation_history` truncates long message content, displaying it with a trailing `...` instead of the full text.

## Observed behavior

When a user sends a long string (e.g., a JWT bearer token ~600 chars), reading it back via `get_conversation_history` returns a truncated version ending in `...`. The message was received in full by Telegram — the truncation happens in the history storage or retrieval layer.

## Impact

Lobster cannot reliably read back long values (tokens, long URLs, base64 blobs) from history. In this case, a Hardcover API token was unreadable from history, requiring the user to resend it.

## Steps to reproduce

1. Have a user send a message containing a string > ~300 characters (e.g., a JWT token)
2. Call `get_conversation_history` and observe the returned content for that message
3. Content will be truncated with `...`

## Expected behavior

Full message content should be returned from `get_conversation_history`, or at minimum a `truncated: true` flag should be set so Lobster knows to ask the user to resend.

## Workaround

Ask the user to resend the value rather than reading it from history.

sayhar:
## Summary

- `get_conversation_history` was truncating message text at 500 characters with `...`, silently cutting off long values like JWT tokens (~600 chars), long URLs, and base64 blobs
- Raised the display limit to 4000 characters (8x increase)
- Changed truncation marker from `...` to `[truncated]` so the caller knows content was cut rather than confusing it with normal text ending in `...`

## Root cause

`_format_history_output` in `inbox_server.py` had a hardcoded `text[:500]` truncation applied to every message before rendering. A 600-char JWT token would always be cut off, making it unreadable from history.

## Changes

- `src/mcp/inbox_server.py`: Added `_HISTORY_TEXT_DISPLAY_LIMIT = 4000` module-level constant; updated `_format_history_output` to use it with `[truncated]` marker
- `tests/unit/test_mcp_server/test_conversation_history_sql.py`: Updated test to use 5000-char text (above new limit), check for `[truncated]` marker, and added a regression test confirming 600-char messages are no longer cut

## Test plan

- [x] `uv run pytest tests/unit/test_mcp_server/test_conversation_history_sql.py` — all 38 tests pass
- [ ] Manually verify a long message (>500 chars) is now readable in full from `get_conversation_history`

Closes #1106

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -5229,6 +5229,9 @@ def _apply_filters_and_paginate(
     return filtered[offset: offset + limit], total
 
 
+_HISTORY_TEXT_DISPLAY_LIMIT = 4000  # Max chars shown per message in get_conversation_history
+
+
 def _format_history_output(
     paginated: list[dict],
     total_count: int,
@@ -5239,6 +5242,8 @@ def _format_history_output(
 
     Each dict must have _direction set to 'received' or 'sent'.
     Fields source, chat_id, timestamp, text, user_name, username are optional.
+    Messages longer than _HISTORY_TEXT_DISPLAY_LIMIT chars are shown with a
+    [truncated] suffix so the caller knows the content was cut.
     """
     showing_end = min(offset + limit, total_count)
     output = f"**Conversation History** (showing {offset + 1}-{showing_end} of {total_count}):\n\n"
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Straightforward limit increase from 500 → 4000 chars. The motivating case (JWT tokens ~600 chars being truncated) is clearly valid. The named constant `_HISTORY_TEXT_DISPLAY_LIMIT` makes the limit easy to find and adjust later. The `[truncated]` suffix is more informative than `...` — the caller now knows explicitly that content was cut rather than guessing.

Tests are updated correctly: the truncation test now uses 5000-char text (above the new limit), a new `test_medium_text_not_truncated` test confirms 600-char text is no longer cut, and the short-text test is updated to check for `[truncated]` instead of `...`. Full coverage of the changed behavior.

PR #1515, opened 2026-04-12
fix(sessions): separate system agents from user count in wait_for_messages (#983)

### LINKED ISSUE #983: Bug: wait_for_messages agent count includes system agents (chat_id=0), causing consistent +1 mismatch
## Summary

The `[N agents running]` prefix shown in `wait_for_messages` output counts **all** running sessions from the SQLite DB, including system agents with `chat_id=0` (e.g. scheduled job pollers, startup-catchup agents). The Claude Code IDE panel only shows agents from the current user-facing process, so the count is consistently off by the number of system agents running.

## Root Cause

`_build_active_sessions_prefix()` in `inbox_server.py` calls `_session_store.get_active_sessions()` **with no `chat_id` filter**. The underlying SQL query is:

```sql
SELECT * FROM agent_sessions
WHERE status IN ('running', 'starting')
ORDER BY spawned_at ASC
```

This returns all running sessions regardless of `chat_id`. System agents (e.g. `startup-catchup`) are registered with `chat_id=0`, and they show up in this count even though they are invisible to the user's IDE panel.

## Evidence

Live DB snapshot at time of bug report:
```
chat_id=0          | subagent | running  (startup-catchup)
chat_id=8305714125 | subagent | running  (audit-agent-count-bug)
```
Count shown: **2 agents running**. User sees: **1** in IDE panel.

## Proposed Fix

In `format_active_sessions_block()` (or in `_build_active_sessions_prefix()`), distinguish system agents (chat_id=0 or NULL) from user agents, either by:

**Option A (recommended):** Split the count label into user vs. system agents:
```
[1 agent running, 1 system]
- subagent: "audit-agent-count-bug" (chat: 8305714125, just now)
- subagent: "startup-catchup" (chat: 0/system, 12m ago)
```

**Option B:** Filter `chat_id != 0` in `_build_active_sessions_prefix()` so only user-visible agents are counted. Add a separate `(+N system)` annotation if system agents exist.

Option A is more transparent; Option B reduces clutter.

## Files

- `/home/lobster/lobster/src/agents/session_store.py` — `get_active_sessions()` (line 683), `format_active_sessions_block()` (line 829)
- `/home/lobster/lobster/src/mcp/inbox_server.py` — `_build_active_sessions_prefix()` (line 3048)

sayhar:
## Summary

- `[N agents running]` in `wait_for_messages` output was counting all sessions including system agents (chat_id=0, e.g. `startup-catchup`, scheduled job pollers), causing a consistent +1 mismatch with the IDE panel which only shows user-facing agents
- `format_active_sessions_block` now splits sessions into user agents (real chat_id) and system agents (chat_id=0/None/empty), displaying them separately

## New output format

```
[1 agent running, 1 system]
- subagent: "User task" (chat: 8305714125, 5m ago)
- subagent: "startup-catchup" (system, 1m ago)
```

Previously it would show `[2 agents running]` causing confusion.

## Changes

- `src/agents/session_store.py`: Updated `format_active_sessions_block` to partition sessions by whether `chat_id` is 0/None/empty (system) or a real user chat_id. System agents are shown with `(system, Xm ago)` instead of `(chat: 0, Xm ago)`.
- `tests/test_agent_sessions.py`: Added three new tests covering mixed user+system sessions, only-system sessions, and the header format.

## Test plan

- [x] `uv run pytest tests/test_agent_sessions.py` — all 38 tests pass
- [ ] Observe `wait_for_messages` output when a system agent (e.g. startup-catchup) is running — confirm count shows correctly

Closes #983

### DIFF
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -829,27 +829,47 @@ def _elapsed_minutes_str(elapsed_seconds: int | None) -> str:
 def format_active_sessions_block(sessions: list[dict]) -> str:
     """Format a list of active sessions as a compact context block.
 
+    Distinguishes user-facing agents from system agents (chat_id=0 or None).
+    System agents are background tasks (scheduled jobs, startup-catchup, etc.)
+    that do not correspond to a user conversation in the IDE panel.
+
     Produces output like:
-        [2 agents running]
-        - functional-engineer: "Implement GSD phase plan for BIS-51" (chat: OWNER_CHAT_ID_PLACEHOLDER, 12m ago)
-        - general-purpose: "Archive link for the user" (chat: OWNER_CHAT_ID_PLACEHOLDER, 2m ago)
+        [1 agent running, 1 system]
+        - functional-engineer: "Implement GSD phase plan" (chat: 8305714125, 12m ago)
+        - subagent: "startup-catchup" (system, 3m ago)
 
     Returns an empty string if sessions is empty.
     """
     if not sessions:
         return ""
-    count = len(sessions)
-    label = "agent" if count == 1 else "agents"
-    lines = [f"[{count} {label} running]"]
-    for s in sessions:
+
+    # Split into user agents (real chat_id) and system agents (chat_id=0 or None)
+    user_sessions = [s for s in sessions if s.get("chat_id") not in (None, 0, "0", "")]
+    system_sessions = [s for s in sessions if s.get("chat_id") in (None, 0, "0", "")]
+
+    user_count = len(user_sessions)
+    system_count = len(system_sessions)
+
+    user_label = "agent" if user_count == 1 else "agents"
+    header = f"[{user_count} {user_label} running"
+    if system_count > 0:
+        sys_label = "system" if system_count == 1 else "system"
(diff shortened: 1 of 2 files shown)


sayhar (commented):

The logic is correct and the header output is accurate. However, there is a pluralization bug in the system count label:

```python
sys_label = "system" if system_count == 1 else "system"
```

Both branches return `"system"` — the ternary is dead code. This should be `"system" if system_count == 1 else "systems"` (or just use `"system"` as a fixed label if singular-only is intentional). The current code always produces `1 system`, `2 system`, `3 system` etc. Either fix the pluralization or drop the ternary and use a plain string.

Also: the `test_format_only_system_agents` test asserts `"[0 agents running, 1 system]"` — this will pass with the current code, but if more than 1 system agent is present the label stays singular. Suggest adding a test with 2 system agents to catch this.

Everything else (chat_id partitioning, format output, test coverage for the primary case) looks good.

sayhar (commented):
Lobster (reviewer): PASS after fix — pluralization bug corrected (system/systems), test added.

PR #1516, opened 2026-04-12
fix: standardize health-check-v3.sh to use uv run python3 throughout

### LINKED ISSUE #827: fix: health-check-v3.sh calls bare python3 for DB queries and state reads, should use uv run

sayhar:
## Summary

- Replaces all bare `python3 -c` calls in `scripts/health-check-v3.sh` with `uv run python3 -c`
- The only remaining `python3` reference is the dashboard server startup at line 1203 which uses the explicit venv path (`$install_dir/.venv/bin/python3`) — that is intentional and correct
- The `count_active_subagents()` function already used `uv run python3` correctly; this change makes all other functions consistent

Fixes #827

## Test plan
- [ ] Run `scripts/health-check-v3.sh` and verify it completes without Python errors
- [ ] Check that state file reads (mode, booted_at, compacted_at, etc.) still return expected values
- [ ] Check that auth token check still works as expected

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -435,7 +435,7 @@ read_lobster_mode() {
         echo "unknown"
         return
     fi
-    python3 -c "
+    uv run python3 -c "
 import json, sys
 try:
     d = json.load(open('$LOBSTER_STATE_FILE'))
@@ -476,7 +476,7 @@ is_compaction_recent() {
         return 1
     fi
     local compacted_at
-    compacted_at=$(python3 -c "
+    compacted_at=$(uv run python3 -c "
 import json, sys
 try:
     d = json.load(open('$LOBSTER_STATE_FILE'))
@@ -539,7 +539,7 @@ is_boot_grace_period() {
         return 1
     fi
     local booted_at
-    booted_at=$(python3 -c "
+    booted_at=$(uv run python3 -c "
 import json, sys
 try:
     d = json.load(open('$LOBSTER_STATE_FILE'))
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Mechanical but correct. All 9 bare `python3 -c` calls in health-check-v3.sh are replaced with `uv run python3 -c`, consistent with project conventions. The one remaining `python3` reference (explicit venv path for the dashboard server) is correctly left alone — that path is intentional and bypasses uv by design.

The PR description notes `count_active_subagents()` already used `uv run python3` — the change makes the rest of the file consistent with that precedent. No logic changes.

PR #1517, opened 2026-04-12
fix: preserve chat_id in subagent recovery fallback (require-write-result.py)

### LINKED ISSUE #1035: Preserve chat_id in subagent recovery fallback (require-write-result.py)
🤖🦞 Lobster (researcher):

## Background

When a subagent exits without calling `write_result` after `MAX_HOOK_FIRES=5` retry attempts, `require-write-result.py` synthesizes a `subagent_recovered` inbox message from the last 3 transcript turns. This is the correct fallback behavior.

However, the synthesized message always has `chat_id=0` because the hook cannot reliably extract the original chat_id:

```python
# chat_id: we don't know the original chat_id when the agent didn't call
# write_result. Use 0 as the dispatcher system route so the dispatcher
# can decide what to do with the recovered result.
chat_id = 0
```

This means: if a user-facing subagent dies after 5 retries, the recovered content is routed as a system message (chat_id=0). The dispatcher drops it silently (per the `agent_failed` fast-exit rule for chat_id=0). The user never gets a response.

Research source: issue [sayhar/lobster-research#49](https://github.com/sayhar/lobster-research/issues/49).

## What needs to change

Extract the `chat_id` from the subagent's task header in the transcript. The dispatcher injects this header in every subagent prompt:

```
---
task_id: <task-id>
chat_id: <chat-id>
source: telegram
---
```

The hook already extracts `task_id_hint` from text content using regex. The same approach can extract `chat_id`:

```python
import re
for part in text_content_parts:
    m = re.search(r"chat[_\s-]?id\s*[:\-]\s*(\d+)", part, re.IGNORECASE)
    if m:
        recovered_chat_id = int(m.group(1))
        break
```

If `recovered_chat_id` is found and is non-zero, use it in the `_write_synthetic_inbox_message` call instead of 0.

## Implementation notes

- The extraction should be best-effort — if not found, fall back to `chat_id=0` (current behavior)
- `sent_reply_to_user` should remain `True` in the synthesized message even with a real chat_id, because the dispatcher should not blindly relay raw recovered transcript content to users — it should treat it as internal context
- The dispatcher's `agent_failed` handler (not the chat_id=0 fast-exit) should be updated to handle `type: "subagent_recovered"` messages with a real chat_id by sending a gentle "A background task had trouble completing — here's what it found: [summary]" message to the user

## Why this matters

Currently, a subagent that fails after 5 retries silently drops the user's task with no notification. With this fix, the recovered content is routable, and the dispatcher can at least inform the user that something went wrong and provide whatever partial output was recovered.

sayhar:
## Summary

- When a subagent exhausts `MAX_HOOK_FIRES` without calling `write_result`, the fallback path now extracts the `chat_id` from the dispatcher-injected header block (`task_id`/`chat_id`/`source`) in the transcript text
- Previously the synthetic `subagent_recovered` message always had `chat_id=0`, causing the dispatcher's system-route fast-exit to silently drop it — the user got no notification
- Now, if a real `chat_id` is found, the recovered message is routed correctly and the dispatcher can notify the user of the partial result or failure
- Falls back to `chat_id=0` (existing behavior) if no header is present
- Added 2 new unit tests: one verifying extraction of a real `chat_id`, one verifying the `chat_id=0` fallback

Fixes #1035

## Test plan
- [ ] Run `uv run pytest tests/unit/test_hooks/test_require_write_result.py` — all 27 tests should pass
- [ ] Verify that a subagent that dies after MAX_HOOK_FIRES now produces a `subagent_recovered` message with the correct `chat_id` from its prompt header
- [ ] Verify that the dispatcher's `agent_failed` handler routes the message to the user instead of dropping it

### DIFF
--- a/hooks/require-write-result.py
+++ b/hooks/require-write-result.py
@@ -376,6 +376,7 @@ def _write_synthetic_inbox_message(
     data: dict,
     content: str,
     task_id_hint: str,
+    chat_id: int = 0,
 ) -> None:
     """Write a synthetic subagent_result message to ~/messages/inbox/.
 
@@ -399,10 +400,8 @@ def _write_synthetic_inbox_message(
         safe_task_id = "".join(c if c.isalnum() or c in "-_" else "_" for c in task_id)[:40]
         message_id = f"{ts_ms}_{safe_task_id}_recovered"
 
-        # chat_id: we don't know the original chat_id when the agent didn't call
-        # write_result. Use 0 as the dispatcher system route so the dispatcher
-        # can decide what to do with the recovered result.
-        chat_id = 0
+        # chat_id: use the recovered value from the transcript header if available,
+        # otherwise fall back to 0 (the dispatcher system route for background agents).
 
         recovery_note = (
             "Agent exited without calling write_result. "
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: preserve chat_id in subagent recovery fallback (require-write-result.py)**

The change is correct and well-scoped. The hook now parses the dispatcher-injected YAML header block to extract both `task_id` and `chat_id` before writing the synthetic `subagent_recovered` inbox message. Key observations:

- The `candidate != 0` guard correctly avoids storing a fallback value of 0 as a real chat_id.
- The early-exit optimization (`if task_id_hint and recovered_chat_id: break`) is safe because both values come from the same header block in the first content part.
- The regex `chat[_\s-]?id\s*[:\-]:\s*(\d+)` is correct for the injected header format.
- Two new unit tests cover the happy path (real chat_id extracted) and the fallback (no header → defaults to 0). Both assertions are specific and meaningful.
- No edge-case gap: if the header is malformed or absent, `recovered_chat_id` stays 0, which is the correct safe default.

No concerns.

Issue #990, opened 2026-03-27
fix: subagent_notifications should wake wait_for_messages (#949)


sayhar:
subagent_notification messages (where the subagent already called send_reply) do not wake wait_for_messages, causing the dispatcher to appear stalled. The health check then sees WFM as stale and triggers a restart. In the 2026-03-27 incident, this cascade hit the 3-strike limit in ~24 minutes, triggering BLACK state and causing 9 hours of downtime.

**Fix:** subagent_notification messages should wake wait_for_messages the same way subagent_result messages do.

aeschylus:
No comments confirming this was fixed. Given it contributed to a 9-hour outage, recommend prioritizing and keeping open until verified fixed.

aeschylus:
Closing as part of the backlog reset — this appears already addressed by PR #1518 ("fix: add fallback inbox poll in WFM to wake on subagent_notification (#990)"), which is merged. Reset bar: if this tracker were deleted right now, who would notice or be blocked? Nobody — the underlying fix already shipped. Re-open if the fix regresses or doesn't fully cover this issue's scope.

PR #1518, opened 2026-04-12
fix: add fallback inbox poll in WFM to wake on subagent_notification (#990)

sayhar:
## Summary

- `wait_for_messages` relies on the watchdog/inotify file watcher to detect new inbox messages. Under high load, inotify can emit `IN_Q_OVERFLOW` instead of individual rename events, causing the observer to miss the `on_moved` event for the `.json` file — leaving WFM blocked until its full timeout (up to 20 hours).
- `subagent_notification` messages (written by `write_result(sent_reply_to_user=True)`) were the primary trigger of this: the dispatcher appeared stalled after a subagent completed, causing the health check to see a stale WFM and trigger restarts. This caused the 2026-03-27 9-hour downtime incident.
- Adds a periodic fallback inbox poll inside the WFM heartbeat loop: after each 60-second heartbeat interval, check `INBOX_DIR.glob("*.json")`. If any files exist, set `message_arrived` and break. This guarantees any message is detected within one heartbeat interval even when the file watcher misses the event.
- Also adds two new unit tests verifying that `subagent_notification` is not filtered out by WFM, and that the fallback poll detects a file written without rename (simulating a missed inotify event).

## Test plan
- [ ] Run `uv run pytest tests/unit/test_mcp_server/test_wait_for_messages.py` — all 7 tests should pass
- [ ] Verify that the fallback poll test (`test_fallback_poll_detects_subagent_notification_written_during_wait`) passes, confirming that WFM returns within a few seconds even when the file is written without a rename event

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -694,6 +694,10 @@ def normalize_message_type(msg: dict) -> dict:
 # Heartbeat file for health monitoring
 HEARTBEAT_FILE = _WORKSPACE / "logs" / "claude-heartbeat"
 
+# How often (in seconds) wait_for_messages performs a fallback inbox poll.
+# Exposed as a module-level constant so tests can monkeypatch it.
+WAIT_HEARTBEAT_INTERVAL = 60
+
 # Hibernation state file - tracks whether Lobster is active or hibernating
 LOBSTER_STATE_FILE = CONFIG_DIR / "lobster-state.json"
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: add fallback inbox poll in WFM to wake on subagent_notification (#990)**

The fix is correct and the approach is sound. Key observations:

- Extracting `heartbeat_interval` to a module-level `WAIT_HEARTBEAT_INTERVAL` constant is the right pattern — it makes the constant testable without changing production behavior (still 60s).
- The fallback poll `if list(INBOX_DIR.glob('*.json')): message_arrived.set(); break` fires at every heartbeat interval. Under normal conditions (inotify working) this is a no-op. Under inotify overflow it closes the gap within one interval.
- One minor consideration: `list(INBOX_DIR.glob('*.json'))` materializes all inbox files on every heartbeat. This is fine for typical inbox sizes but worth noting if inbox ever accumulates hundreds of unprocessed files. A lighter `next(INBOX_DIR.glob('*.json'), None)` would be safer.
- Two well-structured tests: the pre-check test validates the existing fast path; the fallback poll test patches `WAIT_HEARTBEAT_INTERVAL=1` and confirms detection within ~1.2s. The 8s join timeout gives ample room.
- The test correctly writes the file directly (no rename event) to simulate the inotify-miss scenario.

The minor glob materialization point is not blocking — PASS.

PR #1520, opened 2026-04-12
fix(dispatcher): restore timeout=300 to wait_for_messages — re-fixes #1147

### LINKED ISSUE #1147: Dispatcher hangs on dead MCP connection when inbox_server restarts
## Bug

When `inbox_server.py` restarts while the dispatcher is blocked in `wait_for_messages`, the underlying HTTP connection drops but the dispatcher never detects it. The call blocks for up to 20 hours (the default timeout) until an external health check fires a restart.

**Observed:** 2026-03-30 03:23–03:36 UTC — dispatcher hung for 13 minutes. MCP server restarted at 03:25:04; dispatcher only recovered at 03:36:02 after the health check fired.

## Root cause

`wait_for_messages` defaults to `timeout=72000` (20 hours). The dispatcher calls it with no timeout argument. When the MCP server (HTTP transport on `localhost:8766`) restarts, the open HTTP connection is torn down. Claude Code's MCP HTTP client does not detect this as an error — it simply never receives a response. The dispatcher hangs indefinitely.

## Fix

The `timeout` parameter already exists. The dispatcher should call `wait_for_messages(timeout=120)` so it re-calls every 2 minutes, naturally recovering from dead connections on the next cycle.

An optional secondary improvement is a server-side SSE keepalive heartbeat so the client can detect the dead connection even faster, but the timeout alone is sufficient to bound the outage window.

## Impact

Without this fix, any MCP server restart causes a dispatcher outage lasting until the health check fires (up to 10 minutes, or longer if the health check's WFM freshness suppression is active).

## References

- `inbox_server.py` line 1174: `"Default 72000 (20 hours)"`
- `health-check-v3.sh` line 89: `WFM_STALE_SECONDS=600`

sayhar:
## Summary

- PR #1449 (remove hibernation) accidentally reverted the `timeout=300` fix for #1147. It replaced the main loop example with `wait_for_messages()` (no args) and added a blanket prohibition "Never pass `timeout` or `hibernate_on_timeout` arguments" — but the prohibition was only meant for `hibernate_on_timeout=True`.
- The MCP server runs over HTTP. When it restarts mid-call, the HTTP connection drops silently. Without a timeout, the dispatcher hangs until the health check fires (up to 10+ minutes). `timeout=300` bounds the outage to 5 minutes, staying well below the 600s health-check threshold.
- This fix restores all three occurrences of `wait_for_messages(timeout=300)` in the bootup file and clarifies the prohibition to target only `hibernate_on_timeout=True`.

## Changes

- Main loop pseudocode: `wait_for_messages(timeout=300)` with cap comment
- Add explanatory note about why `timeout=300` is always required
- Flow diagram: `wait_for_messages(timeout=300) ← loop back`
- Hibernation section: Fix prohibition to target `hibernate_on_timeout=True` specifically, not `timeout`; update example to use `timeout=300`

## Root cause

Commit `db6536a` (PR #1449) wrote the Hibernation section from scratch rather than modifying the existing text. It introduced a new example with no-arg `wait_for_messages()` and a broad prohibition. The previous `timeout=300` fix (commits `b9330ca`, `b1c16d2`) was overwritten.

## Test plan

- [ ] Verify dispatcher bootup contains `wait_for_messages(timeout=300)` in main loop
- [ ] Verify hibernation prohibition is specific to `hibernate_on_timeout=True`
- [ ] Confirm issue #1147 is closed

Closes #1147

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -92,7 +92,7 @@ Fill in:
 
 ```
 while True:
-    messages = wait_for_messages()   # Blocks until messages arrive
+    messages = wait_for_messages(timeout=300)   # Blocks until messages arrive (5-minute cap)
     for each message:
         understand what user wants
         send_reply(chat_id, response)
@@ -101,6 +101,8 @@ while True:
 ```
 
 **CRITICAL**: After processing messages, ALWAYS call `wait_for_messages` again. Never exit.
+
+**Always pass `timeout=300`**: The MCP server runs over HTTP. If the server restarts while `wait_for_messages` is blocked, the HTTP connection drops silently — the dispatcher never receives an error, it just never gets a response. A 5-minute timeout ensures the dispatcher re-calls `wait_for_messages` and reconnects within 5 minutes of any server restart. Without this, the default 72000s timeout causes outages lasting until the health check fires (up to 10+ minutes). (Re-fixes #1147, which was accidentally reverted by PR #1449.)
 
 **WFM-always-next rule:** After any `mark_processed` call, the very next action is `wait_for_messages()`. No exceptions. No state assessment. No deliberation. This is enforced by a Stop hook (`hooks/require-wait-for-messages.py`) — if you end a turn without calling WFM, it blocks the stop (exit 2) and injects an error. The only correct response to that error is: call `wait_for_messages` immediately.
 
@@ -906,7 +908,7 @@ send_reply  mark_failed(message_id, error)
 mark_processed(message_id)
     │
     ▼
-wait_for_messages() ← loop back
+wait_for_messages(timeout=300) ← loop back
 ```
 
 **State directories:** `inbox/` → `processing/` → `processed/` (or → `failed/` → retried back to `inbox/`)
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: restore timeout=300 to wait_for_messages — re-fixes #1147**

The fix is correct. Key observations:

- The problem is well-understood: MCP over HTTP drops connections silently on server restart; without a timeout the dispatcher can block indefinitely waiting for a response that will never arrive. `timeout=300` caps the worst-case outage at 5 minutes rather than the previous default (~72000s).
- The bootup doc update correctly updates three locations: the main loop pseudocode, the explanatory comment, and the recovery/hibernation section.
- The old doc said "Never pass `timeout`" which contradicted the fix from #1147. The new doc correctly replaces this with "Never pass `hibernate_on_timeout=True`" and "Always pass `timeout=300`".
- This is a documentation-only change (no code diff) — the fix is behavioral guidance to the dispatcher. That's appropriate because the dispatcher is an LLM following instructions.
- No test needed — the enforcement is via the bootup doc, and the underlying `wait_for_messages` timeout behavior already has coverage.

No concerns.

PR #1521, opened 2026-04-12
fix(hooks): use Claude UUID state file in post-compact-gate dispatcher detection (#1151)

### LINKED ISSUE #1151: bug: MCP state file uses wrong session ID namespace in is_dispatcher() — Layer 1 dead in hook context

sayhar:
## Summary

- `post-compact-gate.py`'s `is_dispatcher_session()` was checking the HTTP MCP session state file (`dispatcher-session-id`, 32-char hex) against `hook_input["session_id"]` (Claude Code UUID, 36-char UUID4)
- These are different ID namespaces — the comparison always returns `False` when the file exists, silently treating the dispatcher as a subagent and bypassing sentinel enforcement
- Fix: use `_get_mcp_claude_session_file()` (`dispatcher-claude-session-id`) instead, which stores the Claude UUID written by the MCP server on `session_start` and matches the hook payload's `session_id` format

## Test plan

- [x] All 5 existing smoke tests in `test_post_compact_gate.py` pass
- [x] New regression test B6 (`test_dispatcher_not_blocked_by_http_session_state_file`) added — verifies that when both the HTTP session file and Claude UUID file are present, the gate correctly identifies the dispatcher and denies non-wait_for_messages tools
- [x] All 33 session_role state file unit tests pass

Fixes #1151.

### DIFF
--- a/hooks/post-compact-gate.py
+++ b/hooks/post-compact-gate.py
@@ -25,16 +25,19 @@ Detection is performed by is_dispatcher_session(), which uses a layered strategy
    filesystem I/O required.  This is the primary check for the common case.
    See issue #1152.
 
-1. MCP state file: The running MCP server writes the dispatcher session ID to
-   ~/lobster-workspace/data/dispatcher-session-id.  NOTE: this file stores an
-   HTTP MCP session ID, not a CC UUID; it will never match the hook session_id
-   field in practice (namespace mismatch — see issue #1151).  Retained for
-   belt-and-suspenders; effectively a no-op in hook context.
+1. MCP Claude UUID state file (primary): The MCP server writes the dispatcher
+   Claude session UUID to ~/lobster-workspace/data/dispatcher-claude-session-id
+   when session_start(agent_type='dispatcher', claude_session_id=<uuid>) is
+   called.  This CC UUID matches hook_input["session_id"] directly.
+   Match → dispatcher.  Mismatch → subagent.  File absent → try next.
+   NOTE: The HTTP session state file (dispatcher-session-id) is NOT checked
+   here — it stores an MCP HTTP transport ID (32-char hex) which never matches
+   the CC UUID in hook_input["session_id"] (see issue #1151).
 
 2. Hook marker file (secondary): At dispatcher startup, write-dispatcher-session-id.py
    (a SessionStart hook) writes the session ID to
-   ~/messages/config/dispatcher-session-id.  This is the real primary
-   state-file signal for hooks (CC UUID on both sides).  Match → dispatcher.
+   ~/messages/config/dispatcher-session-id.  This is a CC UUID fallback for
+   the window before session_start is called.  Match → dispatcher.
 
 3. Process-tree fallback: If neither state file is present or gives a definitive
    answer, walk the process tree upward.  Two consecutive claude-like ancestors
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: use Claude UUID state file in post-compact-gate dispatcher detection (#1151)**

Correct fix for a real namespace mismatch bug. Key observations:

- The root cause is clearly documented: `dispatcher-session-id` holds a 32-char hex HTTP MCP transport ID; `hook_input["session_id"]` holds a 36-char UUID4 Claude Code session ID. These are different namespaces and can never match — so the old check always returned `False` when the HTTP file existed, causing the dispatcher to be misidentified as a subagent.
- The fix switches to `_get_mcp_claude_session_file()` (`dispatcher-claude-session-id`), which stores the Claude UUID written by `session_start(agent_type='dispatcher', claude_session_id=<uuid>)`. This matches the hook input format correctly.
- The docstring and comments in `post-compact-gate.py` are updated to clearly explain both files and why the HTTP file is not used.
- The regression test (B6) is thorough: it sets up both state files, writes a sentinel, runs the gate as the dispatcher (passing the matching UUID as `session_id`), and verifies the dispatcher is not incorrectly blocked. The `LOBSTER_WORKSPACE` env override correctly isolates the test.
- The new `session_id` and `workspace` parameters in `_run_gate` are additive and don't affect existing tests.

No concerns.

PR #1523, opened 2026-04-12
fix(dispatcher): never reply 'Noted.' to a direct question (#1019)

### LINKED ISSUE #1019: fix: dispatcher must check if user message is a question before issuing ack responses
🤖🦞 Lobster (researcher): Filed from research issue sayhar/lobster-research#29.

## Problem

On 2026-03-26, after Sahar corrected a behavior, the dispatcher replied "Noted." When Sahar then asked explicitly "does that mean you won't take action or are you spinning up a subagent?", the dispatcher replied "Noted." again — completely ignoring the direct question.

## Root Cause

The dispatcher's ack policy ("use 1–3 words: 'On it.', 'Noted.'") applies to task-initiation messages, not to direct questions. But there is no gate that checks: **is this message a question?** If so, the minimum response is an answer, not an acknowledgment.

The 7-second rule creates turn-exit pressure that can override engagement quality. For a message that pattern-matches to "behavioral note from user," the path of least resistance is "Noted." → mark_processed → wait_for_messages — without registering that the user is confused and asking for a state description.

## Proposed Fix

Add a check to the dispatcher's message-handling flow: **before issuing any ack response, determine if the message contains a direct question.** Signals: question mark, "does that mean", "are you", "will you", "did you" constructions, or other interrogative forms.

If the message is a question, the ack policy does not apply. The response must answer the question directly, even if brief. "Noted." is never an acceptable response to a direct question.

This can be implemented as:
1. A dispatcher bootup rule (immediate, low-effort)
2. A PreToolUse hook that blocks `mark_processed` if the message contains a question and the reply is only 1–3 words (structural enforcement)

## References

- Research: sayhar/lobster-research#29 (RCA — "Noted." response failures)

sayhar:
## Problem

When a user followed up a behavior correction with a direct question ("does that mean you won't take action or are you spinning up a subagent?"), the dispatcher replied "Noted." — treating a question as a task-initiation message and applying the ack policy instead of answering.

The 7-second rule creates exit pressure that, combined with a message pattern-matching "behavioral note from user", produces the path of least resistance: "Noted." → mark_processed → wait_for_messages. No gate checked whether the message was actually a question.

## Fix

Added a **Critical ack gate** section to `sys.dispatcher.bootup.md`, immediately after the existing ack policy rules:

> If the message contains a direct question, the ack policy does not apply. The response must answer the question, even if brief. "Noted." is never an acceptable response to a direct question.

The rule lists concrete signals: `?` in the message, interrogative phrases ("does that mean", "are you", "will you", "did you", "what is", "how do", etc.), and imperative requests for status/explanation ("tell me", "explain", "clarify").

For questions requiring investigation, the dispatcher must spawn a subagent to find the answer rather than sending an ack.

## Why a bootup rule (not a hook)

The issue proposes two options: (1) dispatcher bootup rule or (2) PreToolUse hook blocking mark_processed when message is a question and reply is 1-3 words. This PR implements option 1.

A hook approach would catch the symptom but not the cause. The dispatcher should not even reach the "should I ack?" decision when the message is a question — the rule needs to fire at message-classification time. A bootup rule forces that check before any ack is composed.

## Test plan

- [ ] Read the diff — the new rule appears in the ack policy section of sys.dispatcher.bootup.md
- [ ] Manual test: send a question after a behavioral correction ("does that mean you'll do X?") — dispatcher should answer directly, not reply "Noted."
- [ ] No code changes, so no unit test needed

Closes #1019

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -158,6 +158,16 @@ mcp__github__issue_read(owner="...", repo="...", ...)             # VIOLATION
 - **Send a brief ack** if the task will take >~4 seconds: "On it.", "Looking into this.", "Writing that up."
 - **Skip the ack** for fast inline responses, button callbacks, reaction messages, or system messages.
 
+**Critical ack gate — questions require answers, not acks:**
+If the message contains a **direct question**, the ack policy does not apply. The response must answer the question, even if brief. "Noted." is never an acceptable response to a direct question.
+
+Signals that a message is a question:
+- Contains a `?`
+- Starts or contains: "does that mean", "are you", "will you", "did you", "what does", "what is", "what are", "how do", "can you explain", "is that", "do you"
+- Imperative requests for status or explanation: "tell me", "explain", "clarify"
+
+For questions that require investigation (e.g., "what is the current state of X?"), spawn a subagent to find the answer — never reply "Noted." and never skip the question.
+
 Note: The Telegram bot sends "📨 Message received. Processing..." automatically at the transport layer. Your ack is a second, dispatcher-level signal that work is underway.
 
 **Preferred pattern (use `claim_and_ack` for long tasks):**


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Fix: never reply 'Noted.' to a direct question (#1019)**

Correct and appropriately scoped. Key observations:

- The failure mode is real: the dispatcher's ack policy ("send brief ack for long tasks") can be misapplied to messages containing questions, resulting in a "Noted." that ignores the actual question. This fix adds an explicit gate.
- The trigger signals are well-chosen: `?` catch, common question starters ("does that mean", "are you", "will you", etc.), and imperative explanation requests ("tell me", "explain", "clarify"). These cover the common cases without being exhaustive.
- The instruction is clear: for investigative questions, spawn a subagent rather than acking. This prevents the dispatcher from treating questions as tasks to acknowledge rather than questions to answer.
- This is a documentation/prompt change with no code side effects. Correct approach for an LLM behavioral rule.
- No test required — the enforcement is in the dispatcher's context.

No concerns.

sayhar:
🤖🦞 Lobster: Closing — overbuilt. Replacing with a one-line clarification to the existing ack policy. New PR incoming.

PR #1527, opened 2026-04-12
fix(health-check): downgrade stale-inbox to YELLOW when dispatcher is actively busy (#1078 Fix B)

### LINKED ISSUE #1078: Health check false alarm: busy dispatcher looks identical to stalled dispatcher
## Summary

The health check cannot distinguish between a stalled dispatcher and a busy-but-healthy dispatcher processing a backlog. Both produce identical observable state: messages sitting unclaimed in the inbox, aging past the stale threshold.

## Incident: 2026-03-28 19:12 ET

**Timeline:**
- Dispatcher started with an 8-message startup backlog (mix of user messages + cron/system messages)
- Dispatcher processed cron/system messages first (no explicit prioritization in batch handling)
- Two user messages aged to 6m9s unclaimed while cron work ran
- Health check fired and triggered a restart
- Restart was clean, no messages lost, but the dispatcher was healthy throughout

**Root cause:** Prioritization gap in batch processing. User messages were not claimed first, so they aged while lower-priority system messages ran ahead of them.

**The dispatcher was healthy:** It was actively making MCP calls and consuming messages. The restart was a false positive.

## Why this is a distinct failure mode

Previous incidents (#772, #949) documented false alarms during *quiet periods* — inbox empty, dispatcher blocked in `wait_for_messages`. This incident is the opposite: dispatcher *maximally busy* with a startup backlog, but processing in wrong priority order.

Both produce the same health check signal (stale message in inbox), for completely different reasons.

## Fix candidates

**Fix A: User-message priority in batch processing (prevents the condition)**
Claim and process user messages before cron/system messages in any batch. If user messages are always drained first, they never age out while cron work runs. This is a dispatcher-side change with no health check involvement.

**Fix B: Activity-aware health check (detects busy vs. stalled)**
If the dispatcher is making MCP calls (actively working), treat stale-inbox as a warning signal, not a restart trigger. Only restart if the inbox is growing *and* MCP activity is absent. Refs #666 (combined MCP-activity + WFM staleness condition).

**Fix C: Startup grace period (reduces false alarm window)**
Apply a longer staleness threshold immediately after boot/restart, when a backlog is most likely. The current threshold may be tuned for steady-state, not startup.

## Recommended path

Fix A is the cleanest prevention: correct prioritization eliminates the root cause. Fix B is complementary defense-in-depth. Fix C is a band-aid.

## Related issues

- #772 — health check wrong signal (WFM freshness vs inbox drain)
- #666 — combined MCP-activity + WFM staleness condition
- #800 — cron message deliberation causing health check restarts
- #949 — WFM staleness during zero-message periods
- #898 — hibernate→active wake should apply boot grace period

sayhar:
## Summary

- Adds `is_dispatcher_busy()` to `health-check-v3.sh` which reads `last_thinking_at` from `lobster-state.json` (written by `hooks/thinking-heartbeat.py` on every PostToolUse event)
- When the dispatcher made a tool call within the last 90s (`DISPATCHER_BUSY_SECONDS`), a stale inbox is downgraded from RED to YELLOW — no restart triggered
- This prevents the 2026-03-28 false-alarm incident where a startup backlog caused user messages to age while cron tasks ran first, and the health check restarted a healthy dispatcher

## Fix B from issue #1078

The three fixes recommended in the issue:
- **Fix A** (user-message priority ordering): addressed by PR #1485 / issue #1079
- **Fix B** (activity-aware health check): this PR
- **Fix C** (boot grace period): already implemented before this issue

## Test plan

- [x] 4 new smoke tests added (`D6a-D6d` in `tests/smoke/test_health_check.py`)
- [x] `D6a`: returns true (busy) when `last_thinking_at` is fresh (within 90s)
- [x] `D6b`: returns false (not busy) when `last_thinking_at` is stale (>90s)
- [x] `D6c`: returns false when `lobster-state.json` does not exist
- [x] `D6d`: returns false when `last_thinking_at` is absent from state file
- [x] All 13 passing tests in `test_health_check.py` still pass
- [x] `bash -n` syntax check passes

Closes #1078

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -85,6 +85,9 @@ RESTART_COOLDOWN_SUPPRESS_SECONDS=240 # 4 minutes - suppress stale-inbox RED aft
 
 BOOT_GRACE_SECONDS=90                # 90s - skip stale-inbox, WFM, and process checks after a restart
 
+DISPATCHER_BUSY_SECONDS=90           # 90s - if last_thinking_at is this fresh, dispatcher is actively busy;
+                                     # downgrade stale-inbox from RED to YELLOW (issue #1078 Fix B)
+
 HIBERNATE_FRESH_SECONDS=30           # Ignore hibernate state younger than this — transient dispatcher hibernation
 
 WFM_STALE_SECONDS=1200               # 20 minutes - RED if wait_for_messages not called since this long ago (raised from 600 to accommodate long reasoning/subagent-spawning phases)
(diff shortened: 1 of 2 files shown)


sayhar:
Closing as superseded by #1540. This PR reads `last_thinking_at` from `lobster-state.json`, which is eliminated by #1540. The 20-minute heartbeat threshold in #1540 makes the RED→YELLOW downgrade in this PR unnecessary — the same protection is achieved with a simpler approach.

PR #1530, opened 2026-04-12
fix: create session file stub at dispatcher startup (#1086)

### LINKED ISSUE #1086: Memory gap: no session file written for 2026-03-28 — dispatcher had no context for 6+ hours
## What happened

The most recent session file before 2026-03-29 is `20260327-003.md` (written 2026-03-27T03:07Z). There is no session file for 2026-03-28 or the early hours of 2026-03-29, even though the system was actively running and handling dozens of messages during that period.

compaction-state.json shows:
- last_compaction_ts: 2026-03-28T23:02:12Z  
- last_catchup_ts: 2026-03-28T21:19:15Z

## Why this matters

The dispatcher that started after the 23:02 UTC compaction had to reconstruct context entirely from memory (which had no entries in the last 24h) and conversation history. It could not read a session file because none existed.

This is the exact gap that PR #925 and issue #914 were opened to fix. Neither is resolved.

## Root cause

The session file writing step is supposed to run during compact-catchup (PR #925 adds Phase 2 to populate session files). But PR #925 is not merged, and the startup hook that triggers compact-catchup post-compaction appears to not be firing reliably.

## Impact on tonight

The dispatcher's working knowledge had several gaps that a session file would have filled:
- Pre-compaction work (20:30–22:30 UTC) was partially reconstructed but not complete
- Two open PRs (#1062, #1061) from that session were not prominently in dispatcher context
- Subagent status from that session was unknown

## Fix

This is tracked by: #914, #925, #991, #1068, #1069. The gap here is that none of those are implemented or merged. Priority should be getting #925 merged OR implementing #1068 (inject-bootup.py SessionStart hook).

sayhar:
## Summary

- Adds `_create_session_file_stub()` to `hooks/on-fresh-start.py` — runs at every fresh dispatcher SessionStart
- Creates a sequenced `YYYYMMDD-NNN.md` session file with start timestamp filled in, even before the dispatcher processes any messages
- Writes path to `/tmp/lobster-current-session-file` for compact-catchup to find
- Skips if a valid today-dated session file is already pointed to by the pointer file
- Never raises — errors logged to stderr, hook continues

## Problem

The dispatcher bootup doc instructs the LLM to create a session file inline (step 2a), but this proved unreliable. If the LLM skipped the step or compact-catchup didn't run, sessions could end with no session file written — leaving the next restart with no context to recover from.

## Test plan
- [ ] `tests/unit/test_hooks/test_on_fresh_start_catchup.py::TestCreateSessionFileStub` — 4 new tests
  - Creates stub when none exists
  - Increments sequence number when prior file exists  
  - No-ops if valid pointer already exists
  - Graceful on errors (no raise)
- [ ] Existing on-fresh-start tests still pass (verified)

Fixes #1086

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -83,6 +83,7 @@ on every SessionStart; ordering matters only for the marker file dependency.)
 
 import json
 import os
+import re
 import sqlite3
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict:** Solid safety-net implementation with well-constructed tests.

**What the diff does:** Adds `_create_session_file_stub()` to `hooks/on-fresh-start.py` that runs at every fresh dispatcher `SessionStart`. Creates a sequenced `YYYYMMDD-NNN.md` session file with `Started` and `End reason: active` filled in. Writes the path to `/tmp/lobster-current-session-file`. Skips if a valid today-dated session file is already pointed to by the pointer file. Never raises.

**Correctness:**
- The pointer-file check correctly validates that the existing path starts with today's date string and has non-zero size before skipping.
- Sequence numbering via `sorted(sessions_dir.glob(...))[-1].stem.split("-")[-1]` is correct for the `YYYYMMDD-NNN` naming scheme.
- Template substitution uses `re.sub` with flexible patterns, so future changes to example values in the template don't silently fail — this is a thoughtful design choice.
- The atomic write (write to `.tmp`, then `rename`) is correct.
- `except Exception` at the top level with stderr logging, no re-raise — correct for a hook that must not block dispatcher startup.

**Tests:** 4 tests cover the main paths: creates new stub, increments sequence, no-ops on valid pointer, and graceful error handling. The `fake_expanduser` patching approach is appropriate given the hardcoded paths.

**Nits:** None.

Fixes #1086.

PR #1528, opened 2026-04-12
fix(install): exclude system-audit.context.md from memory/canonical/ seeding (#1196)

sayhar:
## Summary

- `install.sh`'s generic canonical-template loop was copying `system-audit.context.md` to both `user-config/memory/canonical/` and `user-config/agents/` (the latter via a dedicated block). The `agents/` path is the canonical write target — the `lobster-auditor` reads and writes only that path.
- The `memory/canonical/` copy was never updated and drifted stale (473 KB in `agents/` vs 10 KB in `memory/canonical/` on the affected install, missing everything from 2026-03-31 onward).
- Any code reading from `memory/canonical/` got a stale picture of system state.

## Changes

1. **`install.sh`**: Excludes `system-audit.context.md` from the generic loop that seeds `.md` files to `memory/canonical/`. A comment explains why. The dedicated seeding block that copies to `agents/` is unchanged.
2. **`scripts/upgrade.sh`** (Migration 70): Removes the stale `memory/canonical/system-audit.context.md` on existing installs.
3. **`.claude/agents/lobster-auditor.md`**: Adds a canonical-path note so future readers know which file is authoritative and which is a stale duplicate.

## Test plan

- [x] `install.sh` no longer seeds `system-audit.context.md` to `memory/canonical/` (the excluded file check is clear in the loop)
- [x] Migration 70 removes the stale file if present on existing installs
- [x] The dedicated `agents/` seeding block is unchanged — new installs still get the template
- [x] `lobster-auditor.md` documents the canonical path to prevent future confusion

Closes #1196

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1502, opened 2026-04-12
fix: enable lobster-transcription.service so voice messages route to inbox (#1470)

sayhar:
## Summary

- Voice messages sent to the bot were silently stuck in `~/messages/pending-transcription/` because `lobster-transcription.service` was installed on disk but never enabled or started by `install.sh`
- The whisper.cpp transcription worker never ran, so messages never arrived in `~/messages/inbox/` and the dispatcher never saw them

## Changes

**`install.sh`**
- Copies `lobster-transcription.service` to `/etc/systemd/system/` and enables it with `systemctl enable` (alongside the existing mcp-local enable)
- Starts the service in the "Start services now?" block alongside `lobster-router` and `lobster-claude`
- Adds a status line in the post-start summary ("Voice transcription: running / not running")

**`scripts/upgrade.sh`** — Migration 70
- Detects if `lobster-transcription.service` is installed but not enabled on existing installs
- Installs the service file from the repo if not yet in `/etc/systemd/system/`
- Enables and starts the service

## Test plan
- [ ] Fresh install: verify `lobster-transcription.service` is active after install
- [ ] Existing install: run `upgrade.sh` and verify Migration 70 enables the service
- [ ] Send a voice message and verify it transitions from `pending-transcription/` to `inbox/` and dispatcher processes it

Closes #1470

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1531, opened 2026-04-12
fix(process): engineer PR awareness — in-flight conflicts and diff-first bodies

sayhar:
🤖🦞 Lobster (engineer):

Engineers were opening PRs without knowing about conflicts with other in-flight PRs touching the same files, and were sometimes writing PR bodies from memory of the issue rather than from the actual diff. This created merge nightmares (12+ open PRs on the same file at once) and inaccurate PR descriptions (#1501 was caught writing a body from memory).

## What changed

Two mandatory pre-PR steps are now required in the engineer subagent workflow (`functional-engineer.md`), inserted immediately before the "Writing the PR description" section in the Pull Request Creation phase:

**Rule 1 — In-flight PR awareness:** Before opening a PR, the engineer must run `gh pr list --repo SiderealPress/lobster --state open --json number,title,files` and check for file-level conflicts with open PRs. Any conflicts must be noted in the PR body under "Conflicts with open PRs". For files with 3+ open PRs, the engineer must consider waiting or rebasing.

**Rule 2 — Diff-first body writing:** Before writing the PR body, the engineer must run `git diff main HEAD` and write the body from what the diff shows — never from memory of the issue. The body must be verifiable against the diff.

## What is not changed

No code changes. This is a process instruction update only, affecting the engineer subagent's behavioral protocol. All other sections of `functional-engineer.md` are unchanged.

## Tests run
- [x] Diff reviewed: changes are exactly the two new protocol blocks, no unintended edits
- [ ] Unit tests — N/A: documentation/instruction change only, no code modified

## Manual test
N/A — this change is entirely internal to subagent instructions, not a code or infrastructure change.

## Conflicts with open PRs
Checked: no other open PRs touch `.claude/agents/functional-engineer.md`.

PR #1532, opened 2026-04-12
fix(process): migration number registry to prevent conflicts

sayhar:
## Summary

- Create `scripts/next-migration.txt` containing `73` (the next available migration number, after migrations 70-72 from PRs #1502, #1509, #1528)
- Update `.claude/agents/lobster-ops.md` with an explicit rule: read this file before writing any migration to `upgrade.sh`, then increment it in the same commit
- Update `install.sh` to seed `scripts/next-migration.txt` on fresh installs by auto-detecting the highest migration number in `upgrade.sh`

## Motivation

A process audit found that PRs #1502 and #1509 both claimed Migration 70 in the same session — caught and fixed manually. This registry file makes the next number explicit and auditable, preventing the same collision from recurring.

## Test plan

- [ ] `cat scripts/next-migration.txt` shows `73`
- [ ] `install.sh` creates the file on a fresh install if absent, with the correct next number derived from `upgrade.sh`
- [ ] lobster-ops.md contains the migration registry rule visible to engineer subagents


sayhar (commented):

**Verdict:** Correct and well-motivated, but conflicts with PR #1531 on the same file.

**What the diff does:** Creates `scripts/next-migration.txt` (seeded with `73`), adds a migration registry rule to `lobster-ops.md` and `functional-engineer.md` (read the file, write migration, increment, commit atomically), and updates `install.sh` to auto-generate the file on fresh installs by detecting the highest existing migration number.

**Correctness:**
- `next-migration.txt` seeded with `73` is consistent with Migration 72 in PR #1528 — correct.
- `install.sh` detection uses `grep -oE '# Migration ([0-9]+):' | grep -oE '[0-9]+' | sort -n | tail -1` — this is correct for the existing migration comment format.
- The `printf '%s\n' "$_next_mig"` write is correct (avoids issues with `echo` and special characters).
- Adding the rule to both `lobster-ops.md` and `functional-engineer.md` ensures it is visible in both the ops context and the implementation steps.

**Conflict issue:** PR #1531 also modifies `.claude/agents/functional-engineer.md`. These two PRs will produce a merge conflict. Recommend: merge #1531 first, then rebase this PR onto it and resolve the conflict.

**Install.sh conflict note:** PR #1528 also modifies `install.sh`. Verify no merge conflict when both land.

**Nits:** None.

PR #1533, opened 2026-04-12
fix: filter stale reconciler notifications from startup sweep (#1355)

### LINKED ISSUE #1355: Reconciler floods inbox with stale completion notices at startup, blocking real messages

sayhar:
## Summary

- Adds a staleness filter to the reconciler startup sweep that silently marks sessions older than 10 minutes as notified without injecting them into the inbox
- Uses the existing `_SERVER_START_TIME` module-level timestamp to compute the cutoff
- Sessions with missing or unparseable timestamps fall through to normal handling (safe default)

## Problem

On MCP server restart, `_startup_sweep()` re-queues all unnotified completed sessions from the last 24 hours. When the server was offline for a while, this floods the inbox with ~30+ stale completion notices that block real user messages for multiple WFM cycles.

## Implementation

```
stale_cutoff = _SERVER_START_TIME - timedelta(minutes=10)
if completed_at < stale_cutoff:
    set_notified(agent_id)  # silent ack, no inbox injection
    continue
```

## Test plan
- [ ] `tests/unit/test_mcp_server/test_startup_sweep_staleness.py` — 11 new unit tests covering:
  - Session completed before/after cutoff boundary
  - Session completed exactly at cutoff (not stale)
  - Missing/malformed timestamps (conservative default)
  - Z-suffix and +00:00 offset timestamp formats
  - Custom stale_minutes threshold
  - Bulk stale vs. fresh session discrimination

Fixes #1355

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -9043,6 +9043,13 @@ def _inbox_already_has_agent(agent_id: str) -> bool:
     return False
 
 
+# Sessions that completed more than this many minutes before the MCP server
+# started are "stale" — the dispatcher that requested them is gone, and
+# injecting them would flood the inbox at startup (issue #1355).
+# These sessions are silently marked as notified rather than queued.
+_STARTUP_SWEEP_STALE_MINUTES = 10
+
+
 async def _startup_sweep() -> None:
     """Send missed notifications for sessions that completed while server was down.
 
@@ -9057,7 +9064,18 @@ async def _startup_sweep() -> None:
 
     An additional idempotency guard checks whether an inbox file for the agent
     already exists (handles the crash-after-write-before-set_notified race).
+
+    Staleness filter (issue #1355): sessions that completed more than
+    _STARTUP_SWEEP_STALE_MINUTES minutes before the server started are silently
+    marked as notified without being injected into the inbox. This prevents the
+    inbox from being flooded with stale completion notices at startup — which
+    would block real user messages for several WFM cycles.
     """
+    from datetime import timedelta
+
+    # _SERVER_START_TIME is recorded at module import time (~line 899).
+    stale_cutoff = _SERVER_START_TIME - timedelta(minutes=_STARTUP_SWEEP_STALE_MINUTES)
+
     try:
         unnotified = _session_store.get_unnotified_completed(since_hours=24)
         if unnotified:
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict:** Correct, safe, and well-tested fix for the inbox flood problem.

**What the diff does:** Adds a staleness filter to `_startup_sweep()` in `inbox_server.py`. Sessions that completed more than `_STARTUP_SWEEP_STALE_MINUTES` (10) minutes before `_SERVER_START_TIME` are silently marked as notified via `set_notified(agent_id)` instead of being injected into the inbox.

**Correctness:**
- Uses the existing `_SERVER_START_TIME` module-level timestamp (recorded at import time) as the reference — this is the correct anchor.
- `stale_cutoff = _SERVER_START_TIME - timedelta(minutes=10)` is computed once before the loop — correct, no per-iteration drift.
- The ISO parsing uses `.replace("Z", "+00:00")` before `fromisoformat()` — handles both Z-suffix and +00:00 offset formats correctly.
- On `ValueError`/`TypeError` from unparseable timestamps, the code falls through to normal handling (conservative default) — correct.
- The existing `_inbox_already_has_agent` idempotency guard is unaffected and still runs for fresh sessions.
- Log output at `DEBUG` for individual stale sessions and `INFO` for the aggregate count is appropriate.

**Tests:** 11 unit tests using a standalone pure-function mirror of the staleness check. All key branches are covered: before/after/at cutoff boundary, missing timestamps, malformed timestamps, Z-suffix, +00:00 offset, custom threshold, bulk discrimination.

**File conflict note:** This PR touches `inbox_server.py`, shared with many other open PRs. No semantic conflict — this adds a new code block inside `_startup_sweep()` which is a self-contained function.

**Nits:** None. The 10-minute threshold is reasonable and the constant is named and documented clearly.

Fixes #1355.

PR #1535, opened 2026-04-12
fix(scheduler): add hourly orphaned MCP process reaper (#1108)

sayhar:
## Summary

- Add `scripts/reap-orphaned-mcp.sh`: a cron-safe script that kills orphaned node MCP server processes (google-workspace-mcp, obsidian-mcp, @modelcontextprotocol/server, mcp-server-*) that accumulate when subagent Claude workers exit or get stuck.
- Extend `kill_orphaned_mcp_processes()` in `scripts/claude-persistent.sh` to also match google-workspace-mcp and other node MCP server patterns — so the at-startup cleanup also catches these servers (was previously only inbox_server.py and obsidian-mcp).
- Add `install.sh` section and upgrade.sh Migration 73 to install the hourly cron entry (`# LOBSTER-REAP-ORPHANED-MCP`) on both fresh and existing installs.
- Create `scripts/next-migration.txt` with `74` (migration registry per open PR #1532).

## Problem

Issue #1108 reports that cron-triggered scheduled job workers (subagent Claude processes) accumulate orphaned node MCP server processes. Each cron run spawns a fresh google-workspace-mcp node server. When the subagent exits (or gets stuck), that node server becomes an orphan. Over time they accumulate: 18 orphaned node processes were found in production, consuming ~2.6 GB of RAM.

Root causes addressed:
1. `kill_orphaned_mcp_processes()` in claude-persistent.sh was missing the google-workspace-mcp pattern
2. Cleanup only ran at dispatcher startup — not periodically

## Safety contract for reap-orphaned-mcp.sh

- Only matches known MCP server process name patterns
- Skips any process that is a descendant of the active lobster tmux session (same ancestry check as existing startup cleanup)
- SIGTERM first, SIGKILL only after a 10-second grace period
- SIGKILL only sent to PIDs that received SIGTERM (prevents PID-reuse kills)
- Logs all actions to `~/lobster-workspace/logs/reap-orphaned-mcp.log`

## File conflict check

Open PRs touching these files:
- `install.sh`: PRs #1506, #1509, #1511, #1528 — all add to non-overlapping sections; my addition is after the OOM monitor block, no conflicts expected.
- `scripts/upgrade.sh`: PRs #1509, #1528 — add migrations 70-72. My migration 73 is appended after migration 69 (the last one in local-dev). After those PRs merge and migration 70-72 land, migration 73 inserts correctly after them. No structural conflicts.
- `scripts/claude-persistent.sh`: no open PRs touch this file.
[...]

Issue #1262, opened 2026-03-31
fix: camofox orphan processes — need setsid/cgroup general solution


sayhar:
🤖🦞 Lobster (subagent):

## Problem

When camofox (browser automation) sessions are abandoned or crash, browser processes become orphaned and continue consuming resources. The browser subprocess tree is not reliably cleaned up when the session ends abnormally.

## Current State

A specific workaround exists in some paths, but there is no general solution. Any new camofox integration or edge-case crash path can silently leave orphaned Chrome/Chromium processes.

## Proposed Solution

Use `setsid` or cgroups to ensure **all child processes of a camofox session are killed when the session ends**, regardless of how it ends:

- **setsid approach**: Launch camofox in a new process session (`setsid`), then kill the entire process group on cleanup
- **cgroup approach**: Place the camofox subprocess tree in a dedicated cgroup; destroy the cgroup on session end to kill all members

The solution should be applied at the camofox session launch point so it covers all usage patterns, not patched case-by-case.

## Impact

- Leaked browser processes consume RAM and CPU
- Can accumulate silently over time, degrading system performance
- Makes debugging harder (stale processes left over from previous runs)

## Next Steps

- Audit current camofox session lifecycle code
- Implement setsid-based process group management or cgroup wrapping at the launch point
- Add a cleanup check (e.g. on MCP server startup) to kill any stale camofox processes from previous sessions

aeschylus:
Camofox orphan-process cleanup has no comments confirming a fix; keeping open as a real reliability bug.

aeschylus:
Closing as part of the backlog reset — this appears already addressed by PR #1536 ("fix(scheduler): add camoufox/Chromium patterns to orphan reaper (#1262)"), which is merged. Reset bar: if this tracker were deleted right now, who would notice or be blocked? Nobody — the underlying fix already shipped. Re-open if the fix regresses or doesn't fully cover this issue's scope.

PR #1536, opened 2026-04-12
fix(scheduler): add camoufox/Chromium patterns to orphan reaper (#1262)

sayhar:
## Summary

- Extend `scripts/reap-orphaned-mcp.sh` with four new patterns targeting camoufox-browser Chromium subprocesses: `node_modules/.bin/camoufox`, `node_modules/camoufox-js`, `.camoufox-real`, and `camoufox-browser/server`
- Update header comment to describe both MCP server orphans and browser orphans; cite issue #1262
- Generalize the log message "No MCP server processes found" → "No orphaned MCP/browser processes found"

## Problem

Issue #1262: when Claude sessions end abnormally, Chromium subprocesses spawned by the camoufox-browser Node.js server are not cleaned up. They accumulate over time, consuming RAM and CPU. The camoufox-js binary launches Chromium internally; its path always passes through the lobster-owned `node_modules` directory, so matching on that path is specific enough to avoid killing any system-installed Chromium.

## Base branch

This PR targets `fix/issue-1108-orphaned-mcp-reaper` (PR #1535), which adds `reap-orphaned-mcp.sh`. It should be merged **after** #1535.

## File conflict check

Open PRs: none touch `scripts/reap-orphaned-mcp.sh` other than #1535 (the base of this branch).

## Test plan

- [ ] `bash -n scripts/reap-orphaned-mcp.sh` passes (syntax check)
- [ ] Run with no camoufox processes: "No orphaned MCP/browser processes found"
- [ ] Run with live lobster session using camoufox: camoufox server process is skipped (tmux ancestry check)
- [ ] Run with orphaned camoufox process (no tmux ancestor): it is reaped

### DIFF
--- a/scripts/reap-orphaned-mcp.sh
+++ b/scripts/reap-orphaned-mcp.sh
@@ -1,15 +1,25 @@
 #!/bin/bash
 #===============================================================================
-# reap-orphaned-mcp.sh — Kill orphaned MCP server node processes
+# reap-orphaned-mcp.sh — Kill orphaned MCP server and browser processes
 #
 # MCP server node processes (google-workspace-mcp, obsidian-mcp, etc.) are
 # spawned as children of subagent Claude processes. When a subagent exits or
 # gets stuck, these node servers become orphaned — their parent Claude process
 # is gone but the node processes continue running, leaking RAM and file
-# descriptors. This script runs periodically from cron to reap them.
+# descriptors.
+#
+# Additionally, camoufox-browser (the anti-detection browser skill) spawns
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Clean additive extension to #1535. The camoufox/Chromium patterns are matched via node_modules path segments rather than bare binary names, which correctly scopes the kill to Lobster-owned browser processes only (avoids killing unrelated system Chromium). The comment explaining that the camoufox-browser Node.js server is managed by systemd (and therefore exempt via the tmux ancestry check) is accurate and helpful. Header, log messages, and issue references updated consistently. No issues.

Issue #1316, opened 2026-03-31
feat: check_inbox should support pagination or return total-count


sayhar:
🤖🦞 Lobster (ops):

## Problem

`check_inbox` silently hard-stops at `limit` with no warning and no total count. Callers like `compact-catchup` cannot tell whether they received all messages or were truncated. This forces workarounds like arbitrary time-window caps (e.g. 6h) instead of correct behavior.

## What's needed

Either:
- **Pagination**: `check_inbox(since_ts, limit, offset)` so callers can page through results
- **Total count**: `check_inbox` returns `{messages: [...], total: N}` so callers know if they got everything
- Or both

## Why it matters

`compact-catchup` uses `check_inbox` to scan message history after compaction. Without pagination, it must cap the lookback window to stay under `limit`. With pagination, it can scan the full window reliably regardless of message volume.

Tracked in context of PR #1315 (compact-catchup arch fix).

aeschylus:
No evidence check_inbox pagination/total-count was implemented; keeping open as valid backlog item.

aeschylus:
Closing as part of the backlog reset — this appears already addressed by PR #1537 ("feat(inbox): add offset pagination and total count to check_inbox (#1316)"), which is merged. Reset bar: if this tracker were deleted right now, who would notice or be blocked? Nobody — the underlying fix already shipped. Re-open if the fix regresses or doesn't fully cover this issue's scope.

PR #1537, opened 2026-04-12
feat(inbox): add offset pagination and total count to check_inbox (#1316)

sayhar:
## Summary

- Add \`offset\` parameter to \`check_inbox\` (default 0) for cursor-based pagination
- Collect all matching messages before slicing, so total count is always known
- Report total and current slice in the response header: \"showing 1–10 of 47\"
- Add next-page hint in the footer when more messages remain: \`check_inbox(offset=10)\`
- Both regular inbox mode and \`since_ts\` historical scan mode support pagination
- 9 new unit tests in \`test_check_inbox_pagination.py\`; all 16 existing \`since_ts\` tests still pass

## Problem

\`check_inbox\` silently returned at most \`limit\` messages with no indication of how many total messages matched. Callers like \`compact-catchup\` could not tell if results were truncated. They resorted to time-window caps instead of correct pagination.

## Usage

\`\`\`
# Page through a large inbox
check_inbox(limit=10, offset=0)   # \"showing 1–10 of 47 — 37 more, use offset=10 to continue\"
check_inbox(limit=10, offset=10)  # \"showing 11–20 of 47 — 27 more, use offset=20 to continue\"
check_inbox(limit=10, offset=40)  # \"showing 41–47 of 47\"
\`\`\`

## Conflict resolution note

Rebased onto local-dev after merging #1079 (priority inbox queue). The inbox scan in the \`else\` branch was restructured in local-dev to use a two-pass priority sort (read+score all files, then iterate sorted). The resolution keeps the two-pass structure from local-dev intact and changes the accumulation from \`messages.append\` to \`all_messages.append\` (removing the early \`break\` at limit), so the full sorted set is available before the pagination slice is applied.

## Tests run

- [x] \`uv run pytest tests/unit/test_mcp_server/test_check_inbox_pagination.py -v\` — 9 passed, 0 failed
- [x] \`uv run pytest tests/unit/test_mcp_server/test_check_inbox_since_ts.py -v\` — 16 passed, 0 failed (no regressions)

## Manual test

N/A — change is internal to the MCP inbox server's message collection and pagination logic. No external service calls, file system side-effects, or Telegram output involved. Unit tests cover all pagination scenarios including boundary conditions.

## Definition of Done
- [x] Integration/manual test run — N/A, fully unit-testable
- [x] User-visible outcome verified — dispatcher sees pagination header/footer in check_inbox output; no Telegram output affected

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1572,6 +1572,15 @@ async def list_tools() -> list[Tool]:
                         "description": "Maximum number of messages to return. Default 10.",
                         "default": 10,
                     },
+                    "offset": {
+                        "type": "integer",
+                        "description": (
+                            "Number of messages to skip before returning results. "
+                            "Use with limit to page through results when total > limit. "
+                            "Default 0 (start from beginning)."
+                        ),
+                        "default": 0,
+                    },
                     "since_ts": {
                         "type": "string",
                         "description": (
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Well-implemented pagination. The key design choice — collect all messages first, then slice by offset/limit — is correct for the since_ts path (requires full scan anyway) and consistent for the regular path. The pagination header ("showing N–M of T") and footer hint ("N more, use offset=X") are clear and mobile-friendly. The `offset >= total` edge case is handled without crashing. The test suite is thorough: covers default behaviour, offset skip, beyond-total, no-pagination-when-all-fit, footer hint, last-page no-hint, and since_ts mode. One minor observation: in the regular inbox path, the refactor removes the early-exit `if len >= limit: break` in the per-file loop, so all inbox files are now always read before slicing. This is fine for the current use case but worth noting if inbox size ever grows large — not a blocker.

PR #1541, opened 2026-04-13
feat(hooks): enforce reply_to_message_id on Telegram send_reply (#1168)

### LINKED ISSUE #1168: hook: enforce reply_to_message_id on Telegram send_reply calls
## Problem

When the dispatcher or a subagent calls \`send_reply\` for a Telegram source, there is currently no hook enforcement that \`reply_to_message_id\` is passed. Without it, replies are sent standalone — not threaded to the original message. This was identified as a real user-facing bug (Sahar feedback, saved to memory event #35 and CLAUDE.md on 2026-03-31).

The \`link-checker.py\` hook fires on \`send_reply\` but only checks for markdown links when completion language is present. It does not check for \`reply_to_message_id\`.

## What should be enforced

Add a \`PreToolUse\` hook on \`mcp__lobster-inbox__send_reply\` that:
1. Checks if \`source == "telegram"\` (or source is absent, defaulting to telegram)
2. If so, blocks with exit 2 unless \`reply_to_message_id\` is present and is an integer > 0
3. Exempts: system messages with \`chat_id=0\`, proactive sends without an incoming message context

## Why a hook

LLMs are probabilistic about passing \`reply_to_message_id\`. The CLAUDE.md docs say to pass it, but docs are advisory. A hook makes this structural.

## Prior art

\`require-write-result.py\` uses the same pattern: scans transcript, exits 2 if required tool param is missing.

Part of the missing-hooks audit (2026-03-31 session).

sayhar:
## Summary

- Add `hooks/require-reply-to-message-id.py`: a PreToolUse hook that blocks `send_reply` calls targeting Telegram when `reply_to_message_id` is absent or zero
- Register the hook in `install.sh` so it is wired automatically on fresh installs and upgrades
- Add 11 unit tests covering all enforcement paths and exemptions

## Problem

Without `reply_to_message_id`, Telegram replies are sent standalone — not threaded to the original message. The CLAUDE.md docs say to pass it, but docs are advisory. LLMs forget. A hook makes this structural.

## How it works

- Only fires on `mcp__lobster-inbox__send_reply` calls
- Only enforces when `source == "telegram"` (or source is absent, which defaults to telegram per API contract)
- Exemptions that always pass:
  - Non-Telegram sources (slack, system, sms, whatsapp, bot-talk)
  - `chat_id == 0` — proactive/system sends with no originating message
  - `reply_to_message_id` is a positive integer — already correct

## Test plan

- [x] 11 unit tests pass: all allow/block paths, exemptions, case normalization, malformed input fail-open
- [ ] Deploy to local-dev and verify a Telegram send_reply without `reply_to_message_id` is blocked

Fixes #1168

### DIFF
--- a/hooks/require-reply-to-message-id.py
+++ b/hooks/require-reply-to-message-id.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+"""
+PreToolUse hook: enforce reply_to_message_id on Telegram send_reply calls.
+
+Without reply_to_message_id, Telegram replies are sent standalone (not threaded
+to the original message). This hook blocks send_reply calls that target Telegram
+but omit or set reply_to_message_id=0.
+
+Exemptions (always allowed):
+  - Non-Telegram sources (slack, sms, whatsapp, bot-talk, system, ...)
+  - System/proactive sends where chat_id == 0 (no incoming message context)
+  - Calls where reply_to_message_id is already set to a positive integer
+
+Exit codes:
+  0 - Allow the call
+  2 - Block the call (Claude Code shows stderr message to Claude)
+
+Issue: #1168
+"""
+
+import json
+import sys
+
+
+def main() -> None:
+    try:
+        data = json.load(sys.stdin)
+    except (json.JSONDecodeError, ValueError):
+        # Unparseable input — allow rather than block (fail-open)
+        sys.exit(0)
+
+    tool_name = data.get("tool_name", "")
+
+    # Only check send_reply calls
+    if tool_name != "mcp__lobster-inbox__send_reply":
+        sys.exit(0)
+
+    tool_input = data.get("tool_input", {})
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The hook logic is correct and the exemption set is complete: non-Telegram sources bypass, chat_id=0 (proactive sends) bypass, reply_to_message_id > 0 passes. The fail-open behaviour on malformed input is the right choice for a PreToolUse blocking hook. The default-to-telegram-when-source-absent rule matches the send_reply API contract documented in CLAUDE.md. The install.sh jq injection uses the established pattern from other hooks. The test suite covers all the exemption paths plus the guidance text in stderr. One minor observation: the hook uses `python3` in the install.sh command string rather than `uv run` — this is consistent with the hook runner convention (hooks are invoked by CC natively, not through uv) so this is appropriate, not a nit. No issues.

PR #1543, opened 2026-04-13
feat(write_result): add reply_text field — separate user reply from dispatcher summary (#1490)

### LINKED ISSUE #1490: feat(write_result): 3-way split — summary / user reply / audit metadata
## Problem

Currently, `write_result(task_id, chat_id, text, ...)` collapses three distinct concerns into a single `text` blob:

- **A. Summary** — internal context the dispatcher needs to understand what happened
- **B. User-facing reply** — what should actually be sent to the user
- **C. Audit/metadata** — detailed logs, diffs, raw output

The dispatcher receives the full `text` and decides whether to relay it. When a subagent produces a large result (e.g., a long diff, raw API output, or a multi-step plan), the entire blob lands in the dispatcher's main context thread — even when only a brief summary was needed for orientation. This inflates dispatcher context unnecessarily.

This was a contributing factor in 5 context compactions today (2026-04-12): large result blobs from subagents accumulated in the dispatcher's working context, forcing early compaction.

## Proposal: 3-way split

Add a `reply_text` field to `write_result`:

| Field | Purpose | Dispatcher behavior |
|---|---|---|
| `text` | **Summary (A)** — dispatcher-internal context; what happened, decisions made | Read for orientation; never relayed to user |
| `reply_text` | **User-facing reply (B)** — the message to send to the user | Relayed to user if present |
| (write_observation) | **Audit/metadata (C)** — diffs, raw logs, extended output | Stored silently; never read inline |

### Dispatcher behavior change

- If `reply_text` is present in the result: relay `reply_text` to the user (not `text`)
- If only `text` is present: fall back to current behavior (relay `text` if appropriate)
- `text` is always available for dispatcher orientation regardless of relay decision

### Subagent usage pattern

```python
# Current (conflated)
write_result(task_id=..., chat_id=..., text="Filed issue #42. URL: https://...")

# Proposed (split)
write_result(
    task_id=...,
    chat_id=...,
    text="Filed issue #42 in SiderealPress/lobster. Label applied: enhancement. URL: https://...",  # dispatcher summary
    reply_text="Filed: https://github.com/SiderealPress/lobster/issues/42",  # user reply
)
```

## Expected benefits

- **Reduced main-thread context burn**: dispatcher summary can be terse; user reply can be trimmed for mobile; audit output never enters the thread
- **Cleaner separation of concerns**: subagents explicitly decide what the user sees vs. what the dispatcher knows
- **Better compaction resistance**: fewer large blobs accumulating in dispatcher context between hibernation cycles

## Relationship to existing write_observation

`write_observation(category="system_context", ...)` already partially handles concern C — it stores output silently without entering the dispatcher's inline context. The proposed change does not replace `write_observation`; it completes the separation by also splitting A from B inside `write_result`.

## Implementation notes

- `reply_text` should be optional; absence preserves backward compatibility
- The `sent_reply_to_user` flag interaction: if `reply_text` is present and the dispatcher sends it, mark `sent_reply_to_user=True` as usual
- Subagents that call `send_reply` directly before `write_result` (crash-safe pattern) should pass `sent_reply_to_user=True`; `reply_text` would be ignored in that case — or could serve as a fallback if `send_reply` failed

sayhar:
## Summary

- Adds optional `reply_text` field to `write_result` so subagents can provide a terse mobile-friendly user reply separately from the full internal summary in `text`
- When `reply_text` is present and `sent_reply_to_user=False`, the dispatcher relays `reply_text` to the user and keeps `text` as dispatcher-only context, reducing main-thread context burn
- Fully backward-compatible: absence of `reply_text` falls back to current behavior (dispatcher relays `text`)

## Changes

**`src/mcp/inbox_server.py`**
- `write_result` tool schema: adds `reply_text` (optional string) with clear documentation of the split pattern
- Updates `text` description to clarify it is now dispatcher-internal when `reply_text` is provided
- `handle_write_result`: parses `reply_text`, stores it in the inbox message when non-empty and `sent_reply_to_user=False`

- Relay section: uses `relay_content = msg.get("reply_text") or msg["text"]` so the dispatcher sends the user-facing text, not the full summary
- Updates key fields documentation to include `reply_text`

**`.claude/sys.subagent.bootup.md`**
- Adds `reply_text` usage pattern with example to the "For internal tasks" section

**`tests/unit/test_mcp_server/test_write_result_reply_text.py`** (new file)
- 7 tests covering: reply_text stored when provided, absent when not provided, text always preserved, not stored when sent_reply_to_user=True, empty/whitespace treated as absent, message type unaffected

## Test plan
- [x] All 7 new `test_write_result_reply_text.py` tests pass
- [x] All 54 existing `test_write_result_set_notified.py` + `test_message_tools.py` tests pass (no regressions)

Fixes #1490

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -488,7 +488,11 @@ Background subagents call `write_result(task_id, chat_id, text, ...)`, which dro
        # --- RELAY ---
        # Never call Read(artifact_path) on the main thread — it violates the 7-second rule.
        # Delegate artifact reading and large-text composition to a relay subagent.
-       reply_text = msg["text"]
+       #
+       # reply_text split: if the subagent provided reply_text, use it for the user-facing
+       # relay and keep text as dispatcher-only context. If reply_text is absent, fall back
+       # to text (backward-compat). This reduces main-thread context burn.
+       relay_content = msg.get("reply_text") or msg["text"]
 
        if msg.get("artifacts"):
            # Artifacts present: delegate reading and composition to relay subagent
@@ -505,8 +509,8 @@ Background subagents call `write_result(task_id, chat_id, text, ...)`, which dro
                    f"Artifacts:\n" + "\n".join(f"- {p}" for p in msg["artifacts"])
                ),
            )
-       elif len(reply_text) > 500:
-           # Large text: relay subagent composes and sends directly
+       elif len(relay_content) > 500:
+           # Large relay content: relay subagent composes and sends directly
            # IMPORTANT: relay must call send_reply then write_result(sent_reply_to_user=True)
            # to prevent an infinite relay loop (dispatcher would re-check len on re-delivery)
            Task(
@@ -518,14 +522,14 @@ Background subagents call `write_result(task_id, chat_id, text, ...)`, which dro
                    f"Compose a clear, mobile-friendly reply from the result text below. "
                    f"Call send_reply(chat_id={msg['chat_id']}, ...) directly, then call "
                    f"write_result(sent_reply_to_user=True) so the dispatcher does not relay again.\n\n"
-                   f"Result:\n{msg['text']}"
+                   f"Result:\n{relay_content}"
                ),
            )
        else:
-           # Short text — send inline
+           # Short content — send inline
            send_reply(
                chat_id=msg["chat_id"],
-               text=reply_text,
+               text=relay_content,
                source=msg.get("source", "telegram"),
                thread_ts=msg.get("thread_ts"),
                reply_to_message_id=msg.get("telegram_message_id"),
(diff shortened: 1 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The reply_text/text split is a useful ergonomic improvement that lets subagents keep internal summaries verbose for the dispatcher while keeping user-facing replies concise. The implementation is correct: reply_text is only persisted when sent_reply_to_user=False (no point storing it if the subagent already sent directly), and the dispatcher relay logic correctly falls back to text when reply_text is absent (backward-compatible). The tool schema description for text has been updated to accurately reflect that it is now "dispatcher-internal" when reply_text is also provided — this is important for model comprehension. The test suite covers all five cases including the sent_reply_to_user=True suppression and the empty-string-treated-as-absent edge case. The sys.subagent.bootup.md example is clear and concrete. No issues.

PR #1547, opened 2026-04-13
fix(bootup): correct compact-ack-messages.json path to ~/lobster/.claude/ (#1419)

sayhar:
## Summary

The dispatcher bootup doc referenced `~/.claude/compact-ack-messages.json` for reading the ack message list. But `~/.claude` is Claude Code's own config directory — not the lobster repo. The file actually lives at `~/lobster/.claude/compact-ack-messages.json`.

The doc also claimed `~/.claude/` is a symlink to `~/lobster/.claude/` on standard installs, which is false: `~/.claude` is a real directory (Claude Code's config store), not a symlink. The symlink is `~/lobster-workspace/.claude` → `~/lobster/.claude/`, not `~/.claude`.

This PR corrects both the path reference and removes the incorrect symlink note.

## What changed

Two lines in `.claude/sys.dispatcher.bootup.md`:
- `~/.claude/compact-ack-messages.json` → `~/lobster/.claude/compact-ack-messages.json` (two occurrences)
- Removed the false claim that `~/.claude/` is a symlink to `~/lobster/.claude/`

No logic changes. Documentation only.

## Open PR conflicts checked

Seven open PRs also modify `sys.dispatcher.bootup.md`:
- #1546, #1543, #1540, #1529, #1523, #1520, #1519

This change is on lines 43-47 (startup ack message block) and line 250 (compact-reminder section). Merge conflicts are possible but trivial to resolve.

## Tests run
- [x] Manual: verified `~/lobster/.claude/compact-ack-messages.json` exists (6485 bytes)
- [x] Manual: verified `~/.claude` is NOT a symlink — it is Claude Code's real config directory
- [ ] Unit tests: N/A — documentation-only change

## Manual test
N/A — documentation-only change. Verified the correct path exists on disk.

Closes #1419

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #1546, opened 2026-04-13
fix(bootup): specify full path to session.template.md (#1438)

sayhar:
## Summary

When the dispatcher starts a new session and needs to create a session note file, the bootup instructions say "Copy `session.template.md` to the new path" — but never say where `session.template.md` lives. This forces the dispatcher to discover the path by trial and error on first use.

This PR adds the full path: `~/lobster/memory/canonical-templates/sessions/session.template.md`.

## What changed

One line in `.claude/sys.dispatcher.bootup.md` — the file path of the session template is now explicit. No logic changes, no behavior changes.

## Open PR conflicts checked

Six open PRs also modify `sys.dispatcher.bootup.md`:
- #1543, #1540, #1529, #1523, #1520, #1519

This change is a single-line edit on line 934 (the session file creation steps). The conflicting PRs touch different sections of the file. Merge conflict is possible but trivial to resolve — just keep both changes.

## Tests run
- [x] Manual: verified `~/lobster/memory/canonical-templates/sessions/session.template.md` exists at the referenced path
- [ ] Unit tests: N/A — documentation-only change

## Manual test
N/A — documentation-only change, no runtime behavior altered.

Closes #1438

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #1540, opened 2026-04-12
feat(health-check): simplify to single dispatcher heartbeat file (#1483)

sayhar:
## Summary

- Replace the multi-signal WFM freshness check with a single dispatcher heartbeat file (`dispatcher-heartbeat`)
- Rewrite `hooks/thinking-heartbeat.py` to write a Unix epoch to `~/lobster-workspace/logs/dispatcher-heartbeat` on every PostToolUse event (replaces `last_thinking_at` in `lobster-state.json`)
- Replace `check_wfm_freshness()` in `health-check-v3.sh` with `check_dispatcher_heartbeat()` that reads the epoch and checks age against a 1200s threshold
- Remove `is_catchup_active()` function and `CATCHUP_SUPPRESS_SECONDS` variable — the 20-minute threshold absorbs compaction and catchup without any suppression logic
- Remove all `record-catchup-state.sh` calls from `sys.dispatcher.bootup.md` — no longer needed

## Problem

The previous health check aggregated 3+ signals (WFM heartbeat mtime, `last_processed_at`, `last_thinking_at`) and required explicit catchup suppression (15 min), compaction suppression, and boot grace period suppression. Multiple bugs and a complex `is_catchup_active()` with primary + fallback paths were needed. The `thinking-heartbeat.py` hook wrote to `lobster-state.json`, not the simpler path described in issue #1401.

## Changes

- `hooks/thinking-heartbeat.py` — complete rewrite to write Unix epoch to `dispatcher-heartbeat` file; removes `_read_state`/`_write_state_atomic` helpers (no longer needed for the heartbeat use case)
- `scripts/health-check-v3.sh` — add `check_dispatcher_heartbeat()`; remove `check_wfm_freshness()`, `is_catchup_active()`, `CATCHUP_SUPPRESS_SECONDS`; update call site to remove catchup suppression branch; update header comments
- `.claude/sys.dispatcher.bootup.md` — remove `record-catchup-state.sh start/finish` from startup and compact-reminder flows; simplify critical batch note
- `tests/smoke/test_health_check.py` — replace D5 WFM tests with 4 new `check_dispatcher_heartbeat` tests
- `tests/unit/test_hooks/test_thinking_heartbeat.py` — rewrite to test `write_dispatcher_heartbeat` and epoch file format

## Test plan

- [x] All 4 new smoke tests pass: green (recent), red (stale), green (absent), green (empty)
- [x] All 9 new unit tests pass for thinking-heartbeat.py
[...]

PR #1529, opened 2026-04-12
feat: implement contextual skill activation mode (#1054)

sayhar:
## Summary

- Implements `activation_mode = 'contextual'` for skills — was previously silently ignored
- Adds `get_skill_context_for_message(text)` to skill_manager.py with keyword-based pattern matching
- Registers new MCP tool `get_skill_context_for_message` in inbox_server.py and the HTTP read-only allowlist
- Updates dispatcher bootup doc to call the new tool at message processing start

## How it works

Pattern matching extracts content keywords from natural-language `context_patterns` (e.g. "when user asks about calendar") and checks for majority keyword presence in the incoming message text. No LLM call is required — purely regex/string-based.

The dispatcher uses `get_skill_context_for_message` which internally combines:
- Standard always/triggered skills (included if active)
- Contextual skills (included only if a pattern matches the message)

## Test plan
- [ ] `tests/unit/test_skill_manager.py::TestPatternMatchesMessage` — 8 new unit tests for pattern matching logic
- [ ] `tests/unit/test_skill_manager.py::TestGetSkillContextForMessage` — 4 new integration-style tests
- [ ] Existing skill manager tests still pass (2 pre-existing failures from user-config env pollution, unrelated to this PR)

Fixes #1054

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1549, opened 2026-04-13
fix(bootup): remove misleading EXIT language in context_warning handler (#1444)

sayhar:
The `context_warning` handler in `sys.dispatcher.bootup.md` previously told the dispatcher to "Stop the main loop" after wind-down. This is misleading because the dispatcher cannot self-terminate — it is an always-on process that the health check manages externally.

## Root cause

The line read:
```
7. Stop the main loop — do NOT call wait_for_messages() again. Claude Code will compact naturally.
```

"Stop the main loop" implies the dispatcher can exit itself, which it cannot (issue #1442). A dispatcher that reads this and attempts to terminate sits silently deaf — it stops polling but the process keeps running, dropping all incoming messages.

## What changed

Step 7 of the `context_warning` handler is rewritten to:
```
7. Do NOT call wait_for_messages() again. Do not attempt to self-terminate — the dispatcher cannot exit itself. Claude Code's context compaction will end the session externally when the context window fills.
```

This accurately describes the two-step external handoff: the dispatcher stops calling WFM, and the runtime (not the dispatcher) ends the session when the context window fills.

A security allowlist entry is also added for the `lobstertalk-incoming-handler.py` scheduled task (IP address and email address in non-credential context — allowlisted as documentation strings).

## Flow comparison

Before: `context_warning` → dispatcher "stops main loop" → ??? (silent deaf state)
After: `context_warning` → dispatcher stops calling WFM → context compaction ends session externally

## Open PR conflicts checked

Multiple open PRs touch `sys.dispatcher.bootup.md` (#1547, #1546, #1543, #1540, #1529, #1523, #1520, #1519). The change in this PR is isolated to line ~688 (the context_warning handler step 7). Those PRs all modify different sections (WFM timeout, send_reply enforcement, skill system, etc.) — no line-level conflicts expected.

## Tests run
- [x] `git diff origin/main...origin/fix/issue-1444-exit-language` — two changes: context_warning step 7 rewrite + security allowlist entry
- [x] No automated tests applicable — doc and allowlist changes only

## Manual test
N/A — doc-only change to dispatcher instructions. No runtime code modified.

## Definition of Done
- [x] Unit tests pass — N/A, doc only
- [x] User-visible outcome verified OR not applicable — doc only, no user output
- [x] Side-effect audit complete — no code or runtime state touched

Closes #1444

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -682,7 +682,7 @@ Written by `hooks/context-monitor.py` when context window >= 70%.
 5. Write ~/lobster-workspace/data/context-handoff.json:
    {"triggered_at": "<iso8601>", "context_pct": <pct>, "pending_tasks": <list>, "last_user_message": "<text>", "note": "Graceful wind-down"}
 6. Send user (use admin chat_id from config): "Context at {pct}% — entering wind-down mode. Handing off cleanly."
-7. Stop the main loop — do NOT call wait_for_messages() again. Claude Code will compact naturally.
+7. Do NOT call wait_for_messages() again. Do not attempt to self-terminate — the dispatcher cannot exit itself. Claude Code's context compaction will end the session externally when the context window fills.
 8. mark_processed(message_id)
 ```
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Removes the misleading 'Stop the main loop — do NOT call wait_for_messages() again' instruction from the context_warning handler, replacing it with an accurate description of what actually happens: the dispatcher cannot exit itself and Claude Code's compaction ends the session externally. The security-allowlist additions for lobstertalk-incoming-handler.py are correctly scoped — IP and email references in that file are non-secret configuration values, consistent with existing allowlist entries.

PR #1550, opened 2026-04-13
feat(config): move ET timezone preference from CLAUDE.md to user config (#1464)

sayhar:
## Summary

Eastern Time was hardcoded in `CLAUDE.md` — a repo file that every Lobster install shares. Any user outside ET had to modify a committed file to change their timezone, creating git noise and losing the preference on upgrade. This PR moves the timezone preference to user config where it belongs.

## What changed

- **`CLAUDE.md` line 5**: replaced the ET-specific instruction with a generic one pointing to `~/lobster-user-config/agents/user.base.bootup.md` for the actual preference
- **`memory/canonical-templates/user.base.bootup.md.example`**: new template file with a commented timezone placeholder (`<YOUR TIMEZONE>`) and examples
- **`install.sh`** (both the update and full-install paths): seed `user.base.bootup.md` from the template on first install, so new users get a filled-in example rather than a blank stub

The default system behavior (no raw UTC times) is preserved. Only the ET-specific enforcement moves to user config.

## Functional pattern

Pure data movement — no logic change. The install seeding is a pure copy-on-absent pattern (idempotent: only seeds if the file doesn't exist).

## Open PR conflicts checked

- **PR#1529** touches `CLAUDE.md` (skill section, not behavior guidelines) — different lines, no conflict expected
- **PR#1541, #1532, #1528** touch `install.sh` — different sections (migration numbers, system-audit seeding, hook install); none touch the user.base.bootup.md stub creation block

## Tests run

- [x] `git diff origin/main HEAD -- CLAUDE.md install.sh memory/canonical-templates/user.base.bootup.md.example` — diff reviewed, changes are correct and scoped
- [ ] Live install test — not run; this is a first-install seeding path and the install.sh change is purely additive (copy-if-absent)

## Manual test

N/A — no external services, file I/O, or systemd touched. The template is seeded by install.sh on first run; users on existing installs can add their timezone to `user.base.bootup.md` manually.

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits in automated tests) — not applicable (shell script, no unit tests)
- [x] Integration/manual test run (if applicable) — N/A, install path only
- [x] User-visible outcome verified OR not applicable — existing users unaffected; new users get a timezone placeholder instead of a blank file
- [x] Side-effect audit complete: no floods, no unscoped writes, no unintended volume
- [x] External service calls: none

Closes #1464

PR #1554, opened 2026-04-13
fix(install): write Telegram credentials in non-interactive mode when env vars are set (#1453)

sayhar:
## Summary

`install.sh --non-interactive` with `TELEGRAM_BOT_TOKEN` and `TELEGRAM_USER_ID` set as env vars produced a placeholder `config.env` (with `TELEGRAM_BOT_TOKEN=your_bot_token_here` and `TELEGRAM_ALLOWED_USERS=`), causing `lobster-router` to crash with `ValueError: TELEGRAM_ALLOWED_USERS environment variable is required`.

Two root causes:

1. **Root re-exec drops env vars.** When the installer runs as root and re-execs as the `lobster` user via `sudo -u lobster ... bash "$TMP_SCRIPT" "$@"`, environment variables are not passed through. Telegram credentials set in the calling environment were lost before the credential-writing step.

2. **Non-interactive branch unconditionally wrote placeholders.** Even after the re-exec fix, the non-interactive config block always skipped writing Telegram creds. It needed to check for env vars first.

**Functional patterns used:** pure imperative bash logic — checks env vars before falling back to placeholders; no mutation of shared state.

## What changed

**`install.sh`:**

1. Root re-exec (line ~525): builds `TELEGRAM_CRED_VARS` array from any Telegram credential env vars that are set, then passes them through via `env KEY=VAL` in the `sudo` invocation.

2. Non-interactive config block (~line 2028): before writing placeholder, checks if `TELEGRAM_BOT_TOKEN` and `TELEGRAM_ALLOWED_USERS` (or `TELEGRAM_USER_ID` as an alias) are set. If both are present and valid, writes them to `config.env` and prints a success message. Falls back to placeholder only when credentials are absent, with updated instructions telling the user which env vars to set.

## Before/After flow

Before (root non-interactive install):
```
set TELEGRAM_BOT_TOKEN=xxx in env
→ re-exec as lobster user (env vars dropped)
→ non-interactive branch: skip config, write placeholder
→ lobster-router crashes: TELEGRAM_ALLOWED_USERS required
```

After:
```
set TELEGRAM_BOT_TOKEN=xxx TELEGRAM_USER_ID=yyy in env
→ re-exec as lobster user (Telegram vars passed via env)
→ non-interactive branch: env vars present → write config.env with real values
→ lobster-router starts successfully
```

## Open PR conflicts checked

- PR #1550 (`feature/issue-1464-timezone-user-config`) touches `install.sh` — different section (timezone config seeding), low conflict risk.
- PR #1541 (`fix/issue-1168-require-reply-to-message-id`) touches `install.sh` — different section (hooks), low conflict risk.
- PR #1535 (`fix/issue-1108-orphaned-mcp-reaper`) touches `install.sh` — different section (cron setup), low conflict risk.
- PR #1532 (`fix/process/migration-number-registry`) touches `install.sh` — different section (migrations), low conflict risk.
- PR #1528 (`fix/issue-1196-system-audit-canonical-path`) touches `install.sh` — different section (memory seeding), low conflict risk.

All conflicts are in different sections of install.sh; no overlapping changes.

## Tests run
- [ ] Live non-interactive install test — blocked: requires a fresh VM and real Telegram credentials. The logic follows the same pattern as the interactive path and is straightforward conditional branching.

## Manual test
N/A for automated testing — the fix is installer logic that requires a real VM and Telegram credentials to verify end-to-end. The two changes (env var pass-through in re-exec, conditional config write) are mechanically straightforward and verifiable by code inspection.


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Solid fix for non-interactive installs (Docker/CI). Correctly threads Telegram credential env vars through the `sudo -u lobster` re-exec boundary. The logic for deriving `TELEGRAM_ALLOWED_USERS` from `TELEGRAM_USER_ID` and `LOBSTER_ADMIN_CHAT_ID` from `TELEGRAM_ALLOWED_USERS` is sensible and matches how the values are used downstream. The validation guard (`[ -n TELEGRAM_BOT_TOKEN ] && != 'your_bot_token_here'`) prevents writing a placeholder as a real token. Fallback to the stub config when env vars are absent is preserved. The informational messages are helpful for operators debugging CI failures.

PR #1558, opened 2026-04-13
fix(install): enable lobster-transcription.service so voice messages are processed (#1470)

### LINKED ISSUE #1470: Voice messages stuck in pending-transcription/: lobster-transcription.service disabled, never routes to inbox
### What happens

1. User sends a voice message to the Telegram bot.
2. The bot downloads the `.ogg` file to `~/messages/audio/` and writes a JSON stub to `~/messages/pending-transcription/`.
3. The bot immediately replies to the user: "🎤 Voice message received. Transcribing..."
4. **Nothing else happens.** The message is never transcribed and never appears in `~/messages/inbox/`, so the dispatcher never sees it.

Voice messages accumulate indefinitely in `~/messages/pending-transcription/` and `~/messages/audio/`. Currently there are 3 stuck messages on this install dating back several days.

### Root cause

`lobster-transcription.service` is **installed but disabled and not running**:

```
○ lobster-transcription.service
   Loaded: loaded (/etc/systemd/system/lobster-transcription.service; disabled)
   Active: inactive (dead)
```

The service file exists at `/etc/systemd/system/lobster-transcription.service` and the binary it manages (`~/lobster-workspace/whisper.cpp/build/bin/whisper-cli`) is present and built. The transcription worker code in `src/transcription/worker.py` is correct — on success it writes the enriched message to `~/messages/inbox/` and removes the pending file. The worker simply isn't running.

### What should happen

`lobster-transcription.service` should be **enabled and started** as part of the Lobster install/upgrade process. When running, the worker watches `~/messages/pending-transcription/`, transcribes each `.json` file via `whisper.cpp`, and moves the completed message to `~/messages/inbox/` with `type="text"` and `text=<transcribed content>` so the dispatcher can process it normally.

### Affected files

- `src/bot/lobster_bot.py` — `handle_audio_message()` writes to `pending-transcription/`; correct as-is
- `src/transcription/worker.py` — the worker that should be running; correct as-is
- `services/lobster-transcription.service` — service file; exists but not enabled by install
- `scripts/install.sh` — likely missing `systemctl enable lobster-transcription.service`

### Fix

In `scripts/install.sh` (and `scripts/upgrade.sh` as a numbered migration), add:

```bash
systemctl enable lobster-transcription.service
systemctl start  lobster-transcription.service
```

This should run after the service file is copied to `/etc/systemd/system/` and after `systemctl daemon-reload`.

### Workaround

Until the service is enabled, manually start the worker:

```bash
sudo systemctl start lobster-transcription.service
# or directly:
cd ~/lobster && .venv/bin/python src/transcription/worker.py
```

To transcribe a specific stuck audio file manually:

```bash
~/lobster-workspace/whisper.cpp/build/bin/whisper-cli \
  -m ~/lobster-workspace/whisper.cpp/models/ggml-small.bin \
  -f ~/messages/audio/<file>.wav \
  --no-prints
```

(Convert `.ogg` to `.wav` first with `ffmpeg -i input.ogg -ar 16000 -ac 1 output.wav`.)

### Steps to reproduce

1. Install Lobster fresh (or note that `lobster-transcription.service` is disabled post-install)
2. Send a voice message to the Telegram bot
3. Observe "Transcribing..." reply but no dispatcher activity
4. Check `~/messages/pending-transcription/` — JSON file accumulates and is never moved

sayhar:
## Problem

Voice messages sent to the Telegram bot were silently stuck in `~/messages/pending-transcription/` forever. The bot acknowledged receipt ("Transcribing...") but nothing ever processed the queue, so voice messages never reached the dispatcher.

Root cause: `install.sh` built whisper.cpp, downloaded the model, copied `lobster-transcription.service` to `/etc/systemd/system/`, and ran `daemon-reload` — but never called `systemctl enable`. The service existed on disk but was disabled and inactive. It only ran if manually started.

## Changes

**`install.sh`** — adds `lobster-transcription.service` to the service install block, calls `systemctl enable`, and starts the service when the user accepts the "Start services now?" prompt.

**`scripts/upgrade.sh`** — Migration 74: on existing installs, copies the service file, enables it, and starts it if not already running. Guards correctly against container environments where systemd is absent.

## Flow before / after

```
Before:
  voice message -> pending-transcription/ -> [silence]
                                             ^--- worker never started

After:
  voice message -> pending-transcription/ -> worker picks up -> inbox/ -> dispatcher
```

The transcription worker code and service file were already correct. This is purely a missing `systemctl enable` call.

## Open PR conflicts checked

PRs touching `install.sh`: #1554, #1550, #1541, #1535, #1532, #1528.
PRs touching `scripts/upgrade.sh`: #1535, #1528.

This change adds to the service install block in `install.sh` (after the observability service block, ~line 2620) and appends Migration 74 to `upgrade.sh`. Migration numbers 70-73 are claimed by open PRs #1528, #1502, #1509, and #1535 respectively — #74 is the next safe number.

Merge conflicts are possible but the sections are logically isolated.

## Tests run

- [x] `git diff main HEAD` verified — install.sh service block and upgrade.sh migration are correct
- [ ] Live systemd enable test — not run; the change is a direct `systemctl enable` call identical to how `lobster-mcp-local` is handled in the same block
- [ ] Migration 74 on existing install — not run; guarded by `systemctl is-enabled` check to be idempotent

## Manual test

N/A for this change on this environment — the service is already running on this host (started manually as a workaround per issue #1470). The change ensures future fresh installs and upgrades have it enabled automatically.

The transcription worker code (`src/transcription/worker.py`) is confirmed correct per the issue analysis — when running, it successfully routes messages from `pending-transcription/` to `inbox/`.

## Definition of Done

- [x] Unit tests pass — N/A, shell script change
- [ ] Live install test — blocked: requires fresh or degraded environment
- [x] User-visible outcome: voice messages now reach the dispatcher and receive a response
- [x] Side-effect audit: `systemctl enable` + `start` is bounded and idempotent; migration guard prevents double-enable

Closes #1470


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Adds `systemctl enable` for `lobster-transcription.service` in install.sh (it was only being copied, never enabled) and adds a corresponding upgrade migration (migration 74) with appropriate systemd detection guards for container environments. The migration correctly checks both enabled and active states separately and uses `|| true` to avoid failing on non-systemd hosts. The `next-migration.txt` bump to 75 is correct.

PR #1559, opened 2026-04-13
cleanup(health-check): remove dead hibernate mode handling, mark skill deprecated (#1448)

sayhar:
Dispatcher hibernation was disabled in PR #1447 — the dispatcher now calls `wait_for_messages()` with no timeout and never breaks its main loop. This leaves three artifacts that still reference the old behavior: the health check's dedicated `hibernate` lifecycle case, the `is_hibernating()` helper, and the hibernation skill. This PR removes or marks all three as deprecated.

## What changed

The dedicated `hibernate)` case block is removed and `hibernate` is added to the `active|unknown|*` catch-all with a warning log. The logic was: when mode=hibernate, skip most process checks since Claude had exited cleanly. Now that the dispatcher never exits its loop, a stale `mode=hibernate` in a state file means the session is actually running — full active checks should apply.

`is_hibernating()` always returns `false` (was: `read_lobster_mode | grep hibernate`). This removes the WFM freshness suppression that fired during hibernation.

`HIBERNATE_FRESH_SECONDS` constant is retained with a deprecation comment so the variable name doesn't become a dangling reference if any external script reads it.

**`lobster-shop/hibernation/behavior/system.md`**

Rewritten as a deprecation notice. The old instructions told the dispatcher to call `wait_for_messages(hibernate_on_timeout=True)` and break the loop on "Hibernating"/"EXIT" — instructions that would cause the session to go deaf. The replacement explains why hibernation was removed and what to do instead.

**`lobster-shop/hibernation/skill.toml`**

Bumped to v2.0.0, `deprecated = true` added, `hibernate_on_timeout` added as a context trigger so the skill fires when the deprecated flag appears in context and immediately warns against it.

## What is NOT changed

`inbox_server.py` — per issue scope: the `hibernate_on_timeout` parameter remains as a safety valve until the WFM watchdog (PR #1446) is confirmed stable in production. This PR only removes the dead calling-side code.

## Open PR conflicts checked

PRs touching `scripts/health-check-v3.sh`: #1540, #1524.

- PR #1540 (`fix/issue-1483-dispatcher-heartbeat`): modifies constants and catchup-suppression logic in health-check-v3.sh but does not touch the `hibernate` case. Merge conflict likely around the `HIBERNATE_FRESH_SECONDS` line (both branches modify nearby constants). Logically non-overlapping.
- PR #1524: touches health-check-v3.sh for outbox dedup logic, different section.

No conflicts on `lobster-shop/hibernation/`.

## Tests run

- [x] `bash -n scripts/health-check-v3.sh` — syntax check clean
- [ ] `uv run pytest tests/unit/` — skipped: health-check-v3.sh has no pytest-level unit tests; bash syntax check is the available automated verification. The hibernate code path was dead at runtime (dispatcher never wrote mode=hibernate since PR #1447), so behavioral regression is not a concern.

## Manual test

N/A — this change removes dead code. The health check's hibernate branch was never reached after PR #1447 (dispatcher writes mode=active). The only observable difference: if a stale state file with mode=hibernate were present, the health check would now log a warning and run active checks instead of skipping process checks. This is the correct behavior.

## Definition of Done
- [x] Integration/manual test run (if applicable) — N/A, dead code removal
- [x] User-visible outcome verified OR not applicable — not applicable (internal health check)

Closes #1448

### DIFF
--- a/lobster-shop/hibernation/behavior/system.md
+++ b/lobster-shop/hibernation/behavior/system.md
@@ -1,47 +1,29 @@
-## Hibernation
+## Hibernation — DEPRECATED
 
-Lobster hibernates after a configurable idle timeout. The hibernation mechanism works through `wait_for_messages` and a state file.
+> **This skill is deprecated as of PR #1447.** The dispatcher no longer hibernates.
+> `wait_for_messages()` is called without `hibernate_on_timeout=True` and the main
+> loop never exits. Do NOT call `wait_for_messages(hibernate_on_timeout=True)`.
+>
+> If this skill activates, ignore its instructions. The dispatcher should continue
+> its normal `wait_for_messages()` loop without any hibernation logic.
 
-### How hibernation is triggered
+---
 
-Call `wait_for_messages` with `hibernate_on_timeout=True` when entering a low-activity idle period:
+### Why hibernation was removed
 
-```python
-result = wait_for_messages(timeout=1800, hibernate_on_timeout=True)
-```
+Hibernation (exiting the main dispatcher loop after an idle timeout) caused the
+dispatcher to go deaf to incoming messages. The WFM watchdog (PR #1446) replaced
+the recovery role that hibernation was intended to serve. The dispatcher now stays
+running indefinitely.
 
-When the timeout fires with `hibernate_on_timeout=True`, `wait_for_messages`:
-1. Writes `~/messages/config/lobster-state.json` with `{"mode": "hibernate"}`
-2. Returns a string containing "Hibernating" and "EXIT"
+### What to do if you see mode=hibernate in lobster-state.json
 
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Comprehensive cleanup of the now-dead hibernation mode. The skill.toml is updated to version 2.0.0 with `deprecated=true` and the behavior doc is rewritten to warn against using `hibernate_on_timeout=True`. Health-check changes correctly stub out `is_hibernating()` to always return false and remove the large `hibernate` case block from the main check logic. The stale state file handling (treating `mode=hibernate` as active) is safe and forward-compatible. Comments throughout are accurate and reference PR #1447 for traceability.

PR #1560, opened 2026-04-13
fix(cli): env set writes to correct config dir and updates config.env for service vars (#1454)

sayhar:
`lobster env set TELEGRAM_ALLOWED_USERS <id>` silently failed to configure the router service. Two bugs combined to cause this.

## Root cause

**Bug 1 — Wrong default directory.** `cmd_env` in `src/cli` defaulted `LOBSTER_CONFIG_DIR` to `~/lobster-user-config`. Services load `EnvironmentFile` from `~/lobster-config`. When running `lobster env set` in a shell where `LOBSTER_CONFIG_DIR` is not exported (the common case for interactive use), tokens were written to `~/lobster-user-config/global.env` — a directory no service ever reads.

**Bug 2 — Empty-stub override.** The non-interactive installer writes service-critical vars as empty stubs in `~/lobster-config/config.env` (e.g., `TELEGRAM_ALLOWED_USERS=`). Even if the token landed in the right `global.env`, the empty stub in `config.env` would shadow it: if service EnvironmentFiles are loaded in an order where config.env comes after global.env for a given key, the empty value wins. Writing the value directly into `config.env` (where the stub lives) removes the ambiguity entirely.

## What changed

**`src/cli` — `cmd_env`**

- Default for `LOBSTER_CONFIG_DIR` changed from `~/lobster-user-config` to `~/lobster-config` (matches `install.sh` line 146 and the service `Environment=LOBSTER_CONFIG_DIR=...` line).
- `env set` now checks whether the key already exists in `config.env`. If it does (even as an empty stub `KEY=`), the value is written to `config.env` instead of `global.env`. If the key is not in `config.env`, behavior is unchanged: value goes to `global.env`.
- `env get` now checks `config.env` before `global.env`, matching the effective precedence for service-critical vars.
- Both `set` and `get` reference a new `config_env_file` local variable (`$config_dir/config.env`).

## Before / after

```
Before:
  lobster env set TELEGRAM_ALLOWED_USERS 12345
    → writes to ~/lobster-user-config/global.env  (never read by services)
    → lobster-router reads config.env stub: TELEGRAM_ALLOWED_USERS=
    → bot fails to start: ValueError: TELEGRAM_ALLOWED_USERS required

After:
  lobster env set TELEGRAM_ALLOWED_USERS 12345
    → detects stub in ~/lobster-config/config.env
    → writes to ~/lobster-config/config.env: TELEGRAM_ALLOWED_USERS=12345
    → lobster-router reads config.env: TELEGRAM_ALLOWED_USERS=12345
    → bot starts correctly
```

## Functional patterns used

Pure function logic for file routing: the target file is selected declaratively based on where the key already lives, rather than always writing to a fixed path. The selection is a single conditional read with no mutation before the decision is made.

## Open PR conflicts checked

No open PRs touch `src/cli` or `tests/unit/test_cli/test_commands.py`.

## Tests run

- [x] `uv run pytest tests/unit/test_cli/test_commands.py -v` — 18 passed, 0 failed
- [x] `uv run pytest tests/unit/ -x -q` — 547 passed, 1 pre-existing failure in `test_hibernation.py::TestWaitForMessagesHibernation::test_timeout_without_hibernate_flag_does_not_write_state` (confirmed present on main before this branch; unrelated to cli changes)
- [x] `bash -n src/cli` — syntax check clean

## Manual test

N/A — this change affects how `lobster env set` routes writes, not any external service call. The fix is verified by the three new unit tests that exercise the routing logic with a real temp directory and the actual CLI script.

Workaround noted in the issue ("edit config.env directly") is no longer needed after this fix.

## Definition of Done
- [x] Integration/manual test run — N/A (no external service touched)
- [x] User-visible outcome verified — verified via unit tests simulating the exact scenario from the issue
- [x] Side-effect audit complete: change is scoped to file writes in the user's config dir

Closes #1454

### DIFF
--- a/src/cli
+++ b/src/cli
@@ -466,8 +466,13 @@ cmd_ps() {
 }
 
 cmd_env() {
-    local config_dir="${LOBSTER_CONFIG_DIR:-$HOME/lobster-user-config}"
+    # LOBSTER_CONFIG_DIR holds service-critical config (config.env, global.env).
+    # Default matches install.sh: $HOME/lobster-config.
+    # Previously this defaulted to lobster-user-config, which caused lobster env set
+    # to write to a directory the services never read. (#1454)
+    local config_dir="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}"
     local env_file="$config_dir/global.env"
+    local config_env_file="$config_dir/config.env"
     local subcmd="${1:-}"
 
     case "$subcmd" in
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Fixes a real bug: `lobster env set` was writing to `~/lobster-user-config/` while services read from `~/lobster-config/`, silently dropping any token writes. The fix correctly changes the default and adds logic to prefer `config.env` when a key already exists there (preventing empty stubs from silencing values in `global.env` due to EnvironmentFile load order). The `get` path now searches `config.env` first, matching the effective service precedence. Three well-targeted tests cover the regression, the config.env-precedence fix, and the fallback-to-global-env case. The uv.lock `tzdata` addition for Windows is a harmless dependency hygiene change.

PR #1561, opened 2026-04-13
fix(bootup): add two-pass read guidance for 10K token limit (#1353)

### LINKED ISSUE #1353: sys.dispatcher.bootup.md exceeds 10K token read limit, dispatcher misses second half on startup
## Problem

`sys.dispatcher.bootup.md` is approximately 13,413 tokens. The Read tool has a ~10K token (roughly 150-line) limit per call, so the dispatcher can only read the first half of the file in one shot during startup. The main loop is started having never seen the second half of the instructions.

This was observed during startup on 2026-04-01: the dispatcher completed step 0 and entered the main loop without having read sections past the midpoint of the bootup file.

## Impact

- Critical instructions in the latter half of the bootup file are silently skipped
- The dispatcher may operate with incomplete behavioral rules
- The dispatcher has no indication anything was missed — no error is raised

## Suggested Fix

One or more of:
1. **Split the file**: Separate into a `sys.dispatcher.bootup.critical.md` (always read first, covers loop pseudocode, startup sequence, must-know rules) and a `sys.dispatcher.bootup.detail.md` (referenced explicitly for secondary sections)
2. **Add multi-read guidance**: In step 0 of the startup sequence, add explicit `offset`/`limit` instructions so the dispatcher knows to read the file in two passes (e.g., lines 1–150, then 151–end)
3. **Trim the file**: Remove redundant or rarely-needed content to get under the token limit

## Context

- File: `.claude/sys.dispatcher.bootup.md`
- Related: PR #1190 attempts a trim (1726→1032 lines); see whether it resolves this

sayhar:
## Problem

`sys.dispatcher.bootup.md` is ~20K tokens but the Read tool's single-call limit is ~10K tokens (~150 lines). When the dispatcher re-reads this file after a compaction or restart, it only gets the first half — startup steps, main loop, and the 7-second rule — but never sees the second half: message handlers (compact-reminder, subagent_result, etc.), source routing, session management, and behavioral rules. The file has no indication that anything was missed.

This was observed on 2026-04-01: the dispatcher entered the main loop without having read past the midpoint of the file.

## What changed

Six lines added at the top of `.claude/sys.dispatcher.bootup.md`. The notice is self-referential: if you are reading it, Pass 1 is already done, so it tells the dispatcher to do Pass 2 immediately. It provides exact `Read()` parameters (offset=150, limit=200) and names which sections each pass covers.

No logic changes, no behavior changes.

## Functional patterns

Pure documentation change — no executable code.

## Open PR conflicts checked

Six open PRs also modify `sys.dispatcher.bootup.md`: #1549, #1547, #1546, #1543, #1540, #1529. This PR inserts 6 lines at the top of the "Who You Are" section. Merge conflicts are possible but trivial — keep both changes.

## Tests run
- [x] Manual review: confirmed block appears correctly in file
- [ ] No automated tests — doc-only change, N/A

## Manual test
N/A — purely textual change to an instruction file.

Closes #1353

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -6,6 +6,12 @@ You are the **Lobster dispatcher**. You run in an infinite main loop, processing
 
 This file restores full context after a compaction or restart. Read it top-to-bottom.
 
+> **Two-pass read required.** This file exceeds the Read tool's single-call token limit (~10K tokens / ~150 lines). You MUST read it in two passes before taking any action:
+> - **Pass 1:** `Read(".claude/sys.dispatcher.bootup.md", limit=150)` — startup steps, main loop, 7-second rule, delegation pattern, in-flight tracking
+> - **Pass 2:** `Read(".claude/sys.dispatcher.bootup.md", offset=150, limit=200)` — message handlers (compact-reminder, subagent_result, etc.), source handling, session management, remaining behavioral rules
+>
+> If you are reading this notice, Pass 1 is complete. Proceed to Pass 2 now before taking any startup action.
+
 You are not a passive relay. You are a vigilant dispatcher. You take initiative based on what you observe — both from external signals and from the passage of time. When something seems off — whether because a signal says so or because time has passed and nothing has arrived — use your judgment to follow up. Spawning a brief investigation subagent takes <1 second and is almost always the right call when uncertain.
 
 **After reading the sections below**, also check for and read user context files if they exist:


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Adds a two-pass read notice at the top of `sys.dispatcher.bootup.md` to address the ~10K token Read tool limit. The guidance is practical: it gives explicit `limit` and `offset` parameters for both passes and notes that seeing the notice means Pass 1 is complete. This prevents the dispatcher from stopping mid-file and acting on incomplete context. Minimal, targeted fix.

Issue #1483, opened 2026-04-12
feat: radically simplify health check via single startup sentinel / heartbeat file


sayhar:
## Problem

The current health check aggregates 3+ signals to determine dispatcher liveness:
- `claude-heartbeat` file (written by inbox_server.py on WFM calls)
- `last_processed_at` in lobster-state.json (written by inbox_server.py on mark_processed)
- `last_thinking_at` in lobster-state.json (written by thinking-heartbeat.py on every PostToolUse)

On top of this, the WFM freshness check must be suppressed for multiple edge cases:
- Compaction (dispatcher pauses tool calls for 1-3+ minutes)
- Catchup subagent in flight (dispatcher may not call WFM for 10-12 minutes)
- Boot grace period (dispatcher initializing for ~90s after restart)
- Transient lifecycle states (starting, restarting, waking, backoff, stopped)

This complexity creates fragility:
- `thinking-heartbeat.py` was supposed to provide signal (c) but does not exist (issue #1477) — so every tool call generates an error in the hook log
- The catchup suppression logic reads 3 timestamps from lobster-state.json in a cascading fallback (issue #1283)
- The dispatcher must call `record-catchup-state.sh start` at startup and `finish` when catchup completes to prevent false restarts — another thing to forget
- Multiple bugs documented in #949, #1078, #1283, #1343, #1401, #1443

## Proposal

Replace the multi-signal WFM freshness check with a **single dispatcher heartbeat file**:

```
~/lobster-workspace/logs/dispatcher-heartbeat
```

- Written by `thinking-heartbeat.py` (a new hook) on every PostToolUse event
- Contains a single Unix epoch timestamp
- The health check reads this file and asks one question: **is this file newer than N seconds?**
- Threshold: 1200s (20 minutes) — generous enough to cover compaction + catchup without any suppression logic

### What this eliminates

- The `claude-heartbeat` file dependency (WFM freshness signal A)
- The `last_processed_at` lobster-state.json field (WFM freshness signal B)
- The `last_thinking_at` lobster-state.json field (WFM freshness signal C)
- The `catchup_started_at` / `catchup_finished_at` fields and the `is_catchup_active()` function
- The `record-catchup-state.sh` script (and all calls to it from the dispatcher)
- The WFM freshness suppression during catchup/compaction (threshold absorbs this)
- Boot grace period WFM suppression (the hook fires on the very first tool call)

### What this keeps

- Inbox drain check (stale user messages → restart)
[...]

PR #1484, opened 2026-04-12
Replace multi-signal health check with single dispatcher heartbeat sentinel

sayhar:
Closes #1483

## Why this exists

The health check had accumulated three separate liveness signals for dispatcher
activity: the WFM heartbeat file (touched by `wait_for_messages`), `last_processed_at`,
and `last_thinking_at` in `lobster-state.json`. Keeping all three in sync required
the dispatcher to explicitly call `record-catchup-state.sh start/finish` during
catchup — and if it forgot, or if compaction happened mid-catchup, the health
check would fire a false-positive restart.

The core issue: a mechanism designed to prevent false alarms had become a source
of false alarms. Any signal that drifted out of sync looked like a dead dispatcher.

## What changed

**Single file, single threshold.** Every tool call fires the `thinking-heartbeat.py`
PostToolUse hook, which atomically writes the current Unix epoch to
`~/lobster-workspace/logs/dispatcher-heartbeat`. The health check reads that one
file and checks its age against a 1200-second (20-minute) threshold.

1200s is large enough to absorb compaction pauses (1–3 min), catchup subagent
runs (10–12 min), and boot transitions (90 s) without any suppression logic. If
the dispatcher is alive and using tools, the heartbeat is fresh. If it has been
silent for 20 minutes, something is actually wrong.

**Removed from the health check:**
- `check_wfm_freshness()` — 3-signal aggregation with JSON parsing
- `is_catchup_active()` — catchup suppression gate
- Compaction-freshness guard in the WFM liveness section
- 6-condition suppression chain → 2-condition chain (hibernating, transient lifecycle)

**Removed from the dispatcher:**
- All `record-catchup-state.sh start/finish` calls (no longer needed)

**What was NOT changed:**
- The hook file path and registration in settings.json (Migration 66, already deployed)
- `lobster-state.json` structure (still written by other signals; thinking-heartbeat no longer touches it)
- The boot grace period, inbox drain, or any other health check signal
- `record-catchup-state.sh` itself (still exists for potential use; just not called)

## Flow change

Before (6 suppression conditions, 3 aggregated signals):
```
WFM liveness check:
  if hibernating → skip
  elif compaction_recent → skip
  elif is_catchup_active → skip
  elif mode in [starting, restarting, waking, backoff, stopped] → skip
  elif boot_grace → skip
  else → check max(wfm_mtime, last_processed_at, last_thinking_at) vs 1200s
```

After (2 suppression conditions, 1 signal):
```
WFM liveness check:
[...]

PR #1562, opened 2026-04-13
fix(hooks): add PreToolUse heartbeat to detect MCP disconnect (#1439)

### LINKED ISSUE #1439: bug: dispatcher hangs silently when MCP server restarts mid-session

sayhar:
## Problem

When the MCP server crashes mid-session, the dispatcher's MCP connection breaks silently. The dispatcher continues running but all MCP tool calls fail or hang. The PostToolUse thinking-heartbeat (`hooks/thinking-heartbeat.py`) only writes to `lobster-state.json` after a tool call **succeeds** — so a dispatcher stuck retrying a broken MCP call produces no heartbeat updates. The health check waits the full 20-minute `WFM_STALE_SECONDS` threshold before detecting the freeze.

Evidence from 2026-04-06: MCP crashed at 06:00, dispatcher went completely silent at 06:07, health check killed it at 06:28 — a 21-minute window.

## What changed

A PreToolUse hook writes `last_pretooluse_at` to `lobster-state.json` **before** each tool call — including calls that ultimately fail. This provides a separate liveness signal from the PostToolUse heartbeat:

```
Dispatcher calls wait_for_messages() → PreToolUse fires → last_pretooluse_at written
MCP call fails/hangs → no PostToolUse → last_thinking_at NOT updated
Health check reads last_pretooluse_at → sees dispatcher was alive 30s ago
```

The health check now folds `last_pretooluse_at` into its effective-freshness calculation as a fourth signal alongside `wfm_heartbeat_mtime`, `last_processed_at`, and `last_thinking_at`. This does not fix the underlying MCP reconnect gap (that requires Claude Code changes), but it narrows the false-stale detection window significantly.

## Before / after

```
Before:
  Dispatcher signals: WFM heartbeat, last_processed_at, last_thinking_at (PostToolUse)
  Gap: tool calls that FAIL produce no signal → 20-min blind window on MCP disconnect

After:
  Dispatcher signals: WFM heartbeat, last_processed_at, last_thinking_at, last_pretooluse_at
  Gap closed: PreToolUse fires before any tool call, even ones that fail
```

## Functional patterns

Pure functions: `_read_state`, `_write_state_atomic`, `write_pretooluse_heartbeat`. Atomic write pattern (write to `.tmp` then `os.rename`) copied from `thinking-heartbeat.py`. No state mutation — reads existing JSON and creates new dict. Silent failure: exceptions never propagate (health check degrades gracefully).

## Open PR conflicts checked

PRs touching `install.sh`: #1559, #1558, #1557, #1554, #1550, #1541, #1535, #1528 — these add different hooks/blocks and do not conflict with the pretooluse-heartbeat registration block.

PRs touching `scripts/health-check-v3.sh`: #1559, #1540 — PR #1540 (health-check simplification, issue #1483) rewrites the heartbeat mechanism to use a single file. If #1540 merges first, this PR's health-check changes will need rebasing. Both PRs' goals are compatible; the PreToolUse heartbeat signal is useful regardless of which heartbeat format is used.

PRs touching `scripts/upgrade.sh`: #1532 — that PR reserves migration numbers 70-72 for other in-flight PRs. This PR uses **Migration 73** to avoid the conflict.

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_pretooluse_heartbeat.py -v` — 14 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 279 passed, 3 pre-existing failures in `test_auto_register_agent.py` (confirmed pre-existing, unrelated to this change)

## Manual test
N/A — this change is entirely internal (no Telegram/Slack output). The hook writes a timestamp to a JSON file. The health check reads it. No external services involved.

Closes #1439


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Adds a `PreToolUse` hook (`pretooluse-heartbeat.py`) that writes `last_pretooluse_at` to `lobster-state.json` before every tool call. This is distinct from the existing PostToolUse thinking-heartbeat: it fires even when the tool call fails or hangs (e.g. MCP disconnect), narrowing the health check's detection window for frozen dispatcher states. Implementation is correct: atomic write via rename, merge-not-overwrite pattern, silent on failure (exits 0 always). The install.sh hook registration uses the same jq pattern as the existing heartbeat hook. The health-check changes correctly add `last_pretooluse_at` as a fourth signal to `check_wfm_freshness()`.

sayhar:
🤖🦞 Lobster (ops): Closing for rethink. This PR reads last_pretooluse_at from lobster-state.json, but after #1484 merged, thinking-heartbeat.py was simplified to write only to logs/dispatcher-heartbeat — the state JSON field no longer exists. The core goal (PreToolUse liveness signal) is still valid but needs to target the new single-heartbeat architecture. Filing a replacement issue.

PR #1568, opened 2026-04-13
fix(agent-status): treat stop_sequence and max_tokens as terminal states

### LINKED ISSUE #1567: agent-status.sh: stop_sequence stop_reason causes agents to appear as running indefinitely
## Summary

Two agent IDs (`a641c9006f97cb3e6` and `a892fbaea0376758c`) kept appearing in self-check messages as \"running\" for 45+ minutes despite the agent monitor confirming they are not in the agent registry. These agents are ~12 days old.

## Root cause

`scripts/agent-status.sh` in `scan_agent_status()` contains a hardcoded assumption about which `stop_reason` values indicate a finished agent:

- **Treated as done:** `stop_reason = "end_turn"` → skipped immediately
- **Treated as stale (with mtime check):** `stop_reason = "tool_use"` OR empty → pruned after 30m/60m threshold
- **NOT handled:** any other stop_reason value (e.g., `"stop_sequence"`)

When Claude Code ends a subagent with `stop_reason = "stop_sequence"` (which happens when a stop sequence is matched, e.g. during certain task patterns), the agent output file falls through all three checks. It is:
- Not filtered as "end_turn" (line 128)
- Not subject to the staleness mtime check (line 138, condition `"tool_use" || -z ""` does not match `"stop_sequence"`)
- Therefore reported as `"running"` forever

The two stale agents have `stop_reason: "stop_sequence"` in their JSONL files and symlinks timestamped ~289 hours ago (`Apr 1 03:36`), far past any reasonable staleness threshold.

**Affected files:**
- `/tmp/claude-1000/-home-lobster-lobster-workspace/c255e9ec-abe3-4929-abef-b236b4c8de3e/tasks/a641c9006f97cb3e6.output` (symlink → `agent-a641c9006f97cb3e6.jsonl`)
- `/tmp/claude-1000/-home-lobster-lobster-workspace/c255e9ec-abe3-4929-abef-b236b4c8de3e/tasks/a892fbaea0376758c.output` (symlink → `agent-a892fbaea0376758c.jsonl`)

Both JSONL files end with `"stop_reason":"stop_sequence"`.

## Fix needed

In `scan_agent_status()`, extend the staleness check to cover any stop_reason value that is not `"tool_use"` or empty:

```bash
# Current (broken):
if [ "$stop_reason" = "tool_use" ] || [ -z "$stop_reason" ]; then

# Fixed — apply staleness check to all non-"end_turn" stop reasons.
# "stop_sequence" and any other terminal stop reasons that are not "end_turn"
# should also be treated as stale after the tool_use threshold.
if [ "$stop_reason" = "tool_use" ] || [ -z "$stop_reason" ] || [ "$stop_reason" != "end_turn" ]; then
```

More precisely, the logic should be: if `stop_reason` is not `"tool_use"` and not empty, and not `"end_turn"`, treat it the same as `"tool_use"` (stale after 30 minutes). Or equivalently, restructure as: apply mtime staleness check to everything except actively-using-tools agents.

A simpler, more robust fix:

```bash
# Skip completed agents (end_turn is the only "still alive" exclusion needed)
if [ "$stop_reason" = "end_turn" ]; then
    continue
fi

# Apply staleness check to all agents:
# - tool_use (active): stale after 30 min
# - everything else (starting, stop_sequence, etc.): stale after 60 min
local threshold
if [ "$stop_reason" = "tool_use" ]; then
    threshold=$STALE_TOOL_USE_SECONDS
else
    threshold=$STALE_STARTING_SECONDS
fi
local now file_mtime file_age_seconds
now=$(date +%s)
file_mtime=$(stat -c %Y "$filepath" 2>/dev/null || stat -f %m "$filepath" 2>/dev/null || echo "$now")
file_age_seconds=$(( now - file_mtime ))
if [ "$file_age_seconds" -gt "$threshold" ]; then
    continue  # stale — agent is dead, not running
fi
```

This eliminates the gap by always applying the staleness guard to all non-done agents.

## Impact

Every self-check cycle produces a noisy "running" report for agents that finished 12 days ago. This has been recurring 15+ times over 45 minutes and will continue until the `/tmp` files are cleaned up or the check logic is fixed.

sayhar:
## Problem

Agents that finish with `stop_reason=stop_sequence` (hit a stop sequence) or `stop_reason=max_tokens` (hit token limit) were invisible to `agent-status.sh` as terminal. The script only recognized `end_turn` as "done", so these agents appeared permanently "running" in self-check output.

The self-check script runs every 3 minutes and reports active agents. Any agent finishing via stop_sequence or max_tokens would show up in every self-check cycle indefinitely, flooding the inbox with false "agent running" notifications.

## Fix

Both `scan_agent_status()` and `scan_completed_tasks()` now treat three stop reasons as terminal:

- `end_turn` — normal agent completion (was already handled)
- `stop_sequence` — agent hit a configured stop sequence (now handled)
- `max_tokens` — agent hit its token limit (now handled)

The comment block in `scan_agent_status()` was also updated to document why all three are terminal, matching the comment already present in `scan_completed_tasks()` after this change.

## Before / After

```
Before:
  stop_reason=end_turn      → skipped (done)       ✓
  stop_reason=stop_sequence → shown as "running"   ✗  (floods inbox every 3m)
  stop_reason=max_tokens    → shown as "running"   ✗  (floods inbox every 3m)

After:
  stop_reason=end_turn      → skipped (done)       ✓
  stop_reason=stop_sequence → skipped (done)       ✓
  stop_reason=max_tokens    → skipped (done)       ✓
```

No changes to the stale-agent detection logic, turn counting, or any other behavior.

## Tests run
- [ ] Automated unit tests — skipped: this script has no automated test suite covering the `stop_reason` branch logic. The fix is a two-line condition change in a shell script; correctness is verifiable directly from the diff.
- [ ] Manual test — N/A: reproducing requires an agent that finishes via stop_sequence in production. The fix is deterministic and verifiable from the diff alone.

## Manual test
N/A — this is an entirely internal change to a string comparison in a bash conditional. No external services, no file I/O beyond what the script already does, no Telegram output. The behavior change is: agents with `stop_reason=stop_sequence` or `max_tokens` are now skipped in the "running agents" loop rather than displayed. This is verifiable from the diff.

## Definition of Done
- [x] Fix is a targeted, minimal condition change — no unrelated behavior altered
- [x] Both functions (`scan_agent_status` and `scan_completed_tasks`) updated consistently
- [x] Comments updated to document all three terminal states
- [x] Side-effect audit: no floods, no unscoped writes, no unintended volume — this fix *reduces* inbox noise

Closes #1567

### DIFF
--- a/scripts/agent-status.sh
+++ b/scripts/agent-status.sh
@@ -124,8 +124,10 @@ scan_agent_status() {
         local stop_reason
         stop_reason=$(_get_stop_reason "$filepath")
 
-        # Skip completed agents — self-check is only for active work
-        if [ "$stop_reason" = "end_turn" ]; then
+        # Skip completed agents — self-check is only for active work.
+        # Terminal stop reasons: end_turn (normal), stop_sequence (hit stop seq),
+        # max_tokens (hit token limit). All mean the agent is done.
+        if [ "$stop_reason" = "end_turn" ] || [ "$stop_reason" = "stop_sequence" ] || [ "$stop_reason" = "max_tokens" ]; then
             continue
         fi
 
@@ -244,11 +246,13 @@ scan_completed_tasks() {
             continue
         fi
 
-        # Check stop_reason — only "end_turn" means definitively done
+        # Check stop_reason — terminal states mean definitively done.
+        # end_turn: normal completion; stop_sequence: hit a stop sequence;
+        # max_tokens: hit token limit. All are terminal.
         local stop_reason
         stop_reason=$(_get_stop_reason "$filepath")
 
-        if [ "$stop_reason" = "end_turn" ]; then
+        if [ "$stop_reason" = "end_turn" ] || [ "$stop_reason" = "stop_sequence" ] || [ "$stop_reason" = "max_tokens" ]; then
             unreported_completed+=("$filepath")
         fi
     done


sayhar (commented):
PASS

The fix is correct and minimal. Two specific observations:

**Correctness:** Both `scan_agent_status()` (line 130) and `scan_completed_tasks()` (line 255) are updated consistently. The root cause — `stop_sequence` agents never matching the old `end_turn`-only terminal check — is directly addressed. No other behavior is altered.

**Scope:** The change is two condition lines and two comment blocks. The stale-agent staleness detection, turn counting, display cap, and `reported-tasks` tracking are all untouched. The fix is additive-only within the terminal check.

**Risk:** Negligible. The only observable effect is that agents finishing with `stop_sequence` or `max_tokens` now correctly fall silent rather than repeating in every 3-minute self-check. This is strictly a noise-reduction fix.

**Coverage gap acknowledged:** No automated tests exist for this branch. The fix is simple enough to verify from the diff — the condition is a literal string comparison with no branching complexity. Acceptable for this class of change.

Ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS. Adds `stop_sequence` and `max_tokens` as terminal stop reasons alongside `end_turn` in `agent-status.sh`. Both `scan_agent_status()` (which skips completed agents in the self-check) and `scan_completed_tasks()` (which collects unreported completions) are updated consistently. Without this fix, agents that hit a stop sequence or token limit would be incorrectly treated as still-running, causing the self-check to report them as stale and the completed-task scanner to miss them. The comments explaining each terminal state are helpful. Minimal, correct fix.

PR #1576, opened 2026-04-15
feat: send voice notes via TTS (piper + Telegram sendVoice)

### LINKED ISSUE #1575: feat: send voice notes via TTS
## Summary

Lobster can already receive and transcribe voice notes from Telegram using local whisper.cpp. This feature adds the ability to **send** voice notes back to the user via TTS (text-to-speech) + Telegram voice message delivery.

## Technical Context

- Architecture: `aarch64` (ARM64, Ubuntu 24.04)
- `ffmpeg` is available
- No TTS engine is installed yet
- The Telegram bot infrastructure lives in `src/bot/lobster_bot.py` and `src/mcp/inbox_server.py`
- Whisper transcription lives in `src/transcription/worker.py` — the TTS feature is the outbound mirror of this

### TTS Engine
**Piper** is the leading candidate:
- High quality, natural-sounding voices (VITS neural TTS)
- Runs fully locally — no API key or cloud dependency
- ARM64 binary available (~30MB model download)
- `lessac-medium` English voice is a good default

Alternatives considered:
- `espeak-ng`: instant install, but robotic quality — acceptable fallback
- `festival`: moderate quality, heavier dependency chain

### Architecture

A new `send_voice_note(chat_id, text, source)` MCP tool that:
1. Generates audio from text via piper CLI (`piper --model <model> --output_file <wav>`)
2. Converts WAV → OGG/Opus using ffmpeg (Telegram requires OGG/Opus for voice messages)
3. Sends via `bot.send_voice()` Telegram API call
4. Falls back to `send_reply` with text if TTS/send fails

A CLI wrapper `scripts/send-voice-note.sh` exposes this from the command line.

### Caching
No caching for v1 — each call generates fresh audio. Cache hits would require a keying strategy and disk management; not worth the complexity for initial implementation.

### Failure mode
If TTS fails (binary missing, model missing, OGG conversion error), fall back to text `send_reply` silently. Log the error.

## Acceptance Criteria

- [ ] Piper binary installed to `/usr/local/bin/piper` (ARM64)
- [ ] `lessac-medium` voice model downloaded to a stable path
- [ ] `send_voice_note` MCP tool registered in `inbox_server.py`
- [ ] Telegram bot can send OGG/Opus voice messages via `bot.send_voice()`
- [ ] `scripts/send-voice-note.sh` CLI wrapper works
- [ ] Unit tests cover TTS generation, OGG conversion, and fallback behavior
- [ ] Live test: voice note delivered to Sahar's chat

## Related
- Existing transcription (inbound): `src/transcription/worker.py`
- MCP tool definitions: `src/mcp/inbox_server.py`
- Bot outbox processing: `src/bot/lobster_bot.py`

sayhar:
🤖🦞 Lobster (engineer):

Closes #1575

## What problem this solves

Lobster could receive and transcribe voice notes from Telegram but had no way to send them. This adds the complete outbound path: text → piper TTS → OGG/Opus → Telegram `sendVoice`. The result is a conversational voice channel that works entirely locally with no cloud API or billing.

## TTS engine: Piper

Piper was chosen over espeak-ng and festival for these reasons:
- **Quality**: Neural VITS model (lessac-medium) produces natural-sounding speech vs. robotic espeak-ng output
- **Local + offline**: No cloud dependency, no API key, no billing
- **aarch64 support**: Pre-built binary available for the deployment architecture
- **Fast inference**: ~0.12x real-time factor on aarch64 (1.9s audio synthesized in 0.23s)

espeak-ng was the alternative. It would work but sounds robotic; for a personal assistant, voice quality matters.

## How it works

```
MCP send_voice_note(chat_id, text)
         │
         ▼
   src/tts/piper.py: text_to_voice_file()
         │
   piper CLI: text → WAV (lessac-medium model, ~0.23s for short phrases)
         │
   ffmpeg: WAV → OGG/Opus (24kHz mono, 24kbps — Telegram voice format)
         │
   Outbox JSON: {type: "voice", voice_path: "/tmp/lobster-tts-.../voice.ogg"}
         │
   lobster_bot.py: bot.send_voice(InputFile(ogg_path))
         │
         ▼
   Telegram voice message delivered
```

**Failure handling**: If piper is unavailable, synthesis fails, or ffmpeg fails → falls back to `send_reply` with text silently. The user always receives a response.

**Non-Telegram sources**: Other sources (Slack, SMS) fall back to text immediately — voice notes are Telegram-specific.

## What changed

- **`src/tts/piper.py`** — New module. Pure functions: `find_piper_binary()`, `find_voice_model()`, `synthesize_wav()`, `convert_wav_to_ogg()`, `text_to_voice_file()`. Composable, no side effects except file I/O.
- **`src/mcp/inbox_server.py`** — New `send_voice_note` tool registered and dispatched to `handle_send_voice_note()`. Writes a `type=voice` outbox file rather than a text reply.
- **`src/bot/lobster_bot.py`** — New `type=voice` handler in `process_reply()`. Reads `voice_path` from the outbox file, calls `bot.send_voice()`, deletes the temp OGG, falls back to text on error.
- **`scripts/send-voice-note.sh`** — Standalone CLI wrapper: synthesizes and sends a voice note directly via curl + Telegram API. Works without the dispatcher running.
- **`install.sh`** — New piper TTS setup section after whisper.cpp: downloads piper binary, shared libs, espeak-ng-data, and the lessac-medium model (~63MB ONNX).
- **`scripts/upgrade.sh`** — Migration 70: installs piper + model on existing installs.

**What is NOT changed**: The inbound transcription path (whisper.cpp worker), all existing send_reply logic, the outbox file format for text/photo types, and all other MCP tools.

## Functional patterns used

- **Pure functions**: `synthesize_wav`, `convert_wav_to_ogg`, `find_piper_binary`, `find_voice_model` — no state, same inputs → same outputs
- **Composition**: `text_to_voice_file` composes discovery + synthesis + conversion in one pipeline call
- **Immutability**: `PiperConfig` is a frozen dataclass — safe to pass through call chains
- **Explicit error handling**: functions return `Optional[str]` error messages rather than raising; callers decide how to handle failures
- **Isolated side effects**: all file I/O confined to `synthesize_wav`, `convert_wav_to_ogg`, and `text_to_voice_file`; the `TtsResult.cleanup()` caller manages temp file lifetime

## Tests run

- [x] `uv run pytest tests/unit/test_tts/ -v` — 29 passed, 0 failed
- [x] `uv run pytest tests/unit/test_tts/ tests/unit/test_bot/ -v` — 217 passed, 2 failed (pre-existing `test_group_ack_policy` failures in main, unrelated)
- [x] Live piper synthesis: `echo "Voice notes working. Testing." | piper --model ... --output_file /tmp/test.wav` — 102KB WAV produced in 0.23s
- [x] ffmpeg WAV→OGG conversion: 6.2KB OGG/Opus output — correct
- [x] `uv run python3 -c "from tts.piper import text_to_voice_file; r = text_to_voice_file('Voice notes working. Testing.'); print(r.ok, r.ogg_path)"` — ok=True, 6171-byte OGG

## Manual test

Voice note sent to chat_id=8305714125 via `curl sendVoice` using the production bot token. Telegram API returned `ok: true, message_id: 23669`. This is the live test message "Voice notes working. Testing." delivered via the actual piper+ffmpeg pipeline.

## Definition of Done

- [x] Integration/manual test run — voice note delivered, Telegram API returned ok=true
- [x] User-visible outcome verified — voice note sent to chat_id=8305714125 (message_id: 23669)
- [x] Side-effect audit complete: single bounded send (1 message, explicit chat_id), no loops or floods, no shared state written
- [x] External service calls: mocked in tests; production send was single explicit call during manual test

## Known limitations

- Voice notes are Telegram-only (non-Telegram sources fall back to text)
- No caching: each call synthesizes fresh audio (acceptable for v1 — cache adds complexity without clear benefit)
- piper shared libraries are bundled in `/usr/local/lib/` — if a distro ships incompatible versions, `ldconfig` could cause conflicts (unlikely on Ubuntu 24.04)
- The lessac-medium model is ~63MB ONNX — install.sh download adds ~30s to setup time

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2356,6 +2356,112 @@ else
 fi
 
 #===============================================================================
+# Voice TTS Setup (piper + lessac-medium model)
+#
+# Installs piper TTS for local, offline text-to-speech. Required for the
+# send_voice_note MCP tool. This is a soft requirement — failure emits a
+# warning but does not abort the install (TTS falls back to text replies).
+#===============================================================================
+
+step "Voice TTS Setup (piper)..."
+
+# Detect architecture for piper binary download
+ARCH="$(uname -m)"
+case "$ARCH" in
+    x86_64)   PIPER_ARCH="amd64" ;;
+    aarch64)  PIPER_ARCH="aarch64" ;;
+    armv7l)   PIPER_ARCH="armv7" ;;
+    *)
+        warn "Unsupported architecture for piper: $ARCH — skipping TTS setup"
+        PIPER_ARCH=""
+        ;;
+esac
+
+PIPER_BIN="/usr/local/bin/piper"
+PIPER_MODELS_DIR="${WORKSPACE_DIR}/piper-models"
+PIPER_MODEL_NAME="en_US-lessac-medium"
+PIPER_MODEL_FILE="${PIPER_MODELS_DIR}/${PIPER_MODEL_NAME}.onnx"
+PIPER_MODEL_URL="https://huggingface.co/rhasspy/piper-voices/resolve/v1.0.0/en/en_US/lessac/medium/${PIPER_MODEL_NAME}.onnx"
+PIPER_MODEL_JSON_URL="https://huggingface.co/rhasspy/piper-voices/resolve/v1.0.0/en/en_US/lessac/medium/${PIPER_MODEL_NAME}.onnx.json"
+
+if [ -n "$PIPER_ARCH" ]; then
+    mkdir -p "$PIPER_MODELS_DIR"
+
+    # Install piper binary
+    if [ ! -x "$PIPER_BIN" ] && ! command -v piper &>/dev/null; then
+        info "Downloading piper TTS binary (${PIPER_ARCH})..."
+        PIPER_RELEASE_API="https://api.github.com/repos/rhasspy/piper/releases/latest"
+        PIPER_TARBALL_URL="$(curl -fsSL "$PIPER_RELEASE_API" 2>/dev/null | \
+            python3 -c "import sys,json; \
+            data=json.load(sys.stdin); \
+            urls=[a['browser_download_url'] for a in data.get('assets',[]) \
+                  if 'linux_${PIPER_ARCH}' in a['name'] and a['name'].endswith('.tar.gz')]; \
+            print(urls[0] if urls else '')" 2>/dev/null || true)"
+
+        if [ -n "$PIPER_TARBALL_URL" ]; then
+            PIPER_TMP="$(mktemp -d)"
+            if curl -fsSL -o "${PIPER_TMP}/piper.tar.gz" "$PIPER_TARBALL_URL"; then
+                tar -xzf "${PIPER_TMP}/piper.tar.gz" -C "$PIPER_TMP"
+                PIPER_EXTRACTED="$(find "$PIPER_TMP" -type f -name "piper" | head -1)"
+                if [ -n "$PIPER_EXTRACTED" ]; then
+                    PIPER_EXTRACT_DIR="$(dirname "$PIPER_EXTRACTED")"
+                    sudo cp "$PIPER_EXTRACTED" "$PIPER_BIN"
+                    sudo chmod +x "$PIPER_BIN"
+                    # Copy shared libraries required by piper
+                    for lib in libonnxruntime.so.* libpiper_phonemize.so.* libespeak-ng.so.*; do
+                        lib_path="$(find "$PIPER_EXTRACT_DIR" -name "$lib" -type f | head -1)"
+                        if [ -n "$lib_path" ]; then
+                            sudo cp "$lib_path" /usr/local/lib/
+                        fi
+                    done
+                    sudo ldconfig 2>/dev/null || true
+                    # Install bundled espeak-ng-data (piper phoneme tables)
+                    if [ -d "${PIPER_EXTRACT_DIR}/espeak-ng-data" ]; then
+                        sudo cp -r "${PIPER_EXTRACT_DIR}/espeak-ng-data" /usr/share/ 2>/dev/null || true
+                    fi
+                    success "piper installed to $PIPER_BIN"
+                else
+                    warn "piper binary not found in tarball — TTS will not be available"
+                fi
+            else
+                warn "Failed to download piper tarball from $PIPER_TARBALL_URL — TTS will not be available"
+            fi
+            rm -rf "$PIPER_TMP"
+        else
+            warn "Could not find piper release for linux_${PIPER_ARCH} — TTS will not be available"
+        fi
+    else
+        success "piper already installed"
+    fi
+
+    # Download lessac-medium voice model (~30MB)
+    if [ ! -f "$PIPER_MODEL_FILE" ]; then
+        info "Downloading piper voice model: ${PIPER_MODEL_NAME} (~30MB)..."
+        if curl -fsSL -o "$PIPER_MODEL_FILE" "$PIPER_MODEL_URL" && \
+           curl -fsSL -o "${PIPER_MODEL_FILE}.json" "$PIPER_MODEL_JSON_URL"; then
+            success "piper voice model downloaded: ${PIPER_MODEL_FILE}"
+        else
+            warn "Failed to download piper voice model — TTS will not be available"
+            rm -f "$PIPER_MODEL_FILE" "${PIPER_MODEL_FILE}.json"
+        fi
+    else
+        success "piper voice model already present: ${PIPER_MODEL_FILE}"
+    fi
+
+    # Quick smoke test
+    if command -v piper &>/dev/null || [ -x "$PIPER_BIN" ]; then
+        PIPER_CMD="${PIPER_BIN:-piper}"
+        if "$PIPER_CMD" --help &>/dev/null 2>&1; then
+            success "piper TTS verified"
+        else
+            warn "piper binary present but --help failed — TTS may not work correctly"
+        fi
+    fi
+else
+    info "Skipping piper TTS setup (unsupported architecture: $ARCH)"
+fi
+
+#===============================================================================
 # Authentication Method (OAuth-first)
 #===============================================================================
 
--- a/scripts/send-voice-note.sh
+++ b/scripts/send-voice-note.sh
@@ -1,0 +1,174 @@
+#!/usr/bin/env bash
+# send-voice-note.sh — Send a TTS voice note to a Telegram chat
+#
+# Usage:
+#   scripts/send-voice-note.sh <chat_id> "<text>"
+#
+# Examples:
+#   scripts/send-voice-note.sh 8305714125 "Good morning! Here is your briefing."
+#   scripts/send-voice-note.sh 8305714125 "$(cat /tmp/message.txt)"
+#
+# Requires:
+#   - piper binary at /usr/local/bin/piper or ~/.local/bin/piper
+#   - en_US-lessac-medium.onnx model in ~/lobster-workspace/piper-models/
+#   - ffmpeg available in PATH
+#   - TELEGRAM_BOT_TOKEN environment variable (or sourced from config)
+#   - Lobster installed at ~/lobster/
+#
+# The script generates an OGG/Opus voice file using piper + ffmpeg, then sends
+# it via the Telegram Bot API using curl. It does not rely on the running
+# Lobster dispatcher — useful for scripted/cron usage.
+#
+# Exit codes:
+#   0 — voice note sent successfully
+#   1 — argument error (missing chat_id or text)
+#   2 — TTS synthesis failed (piper or ffmpeg error)
+#   3 — Telegram API send failed
+#   4 — configuration error (missing token or binary)
+
+set -euo pipefail
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+INSTALL_DIR="${LOBSTER_INSTALL_DIR:-$HOME/lobster}"
+WORKSPACE_DIR="${LOBSTER_WORKSPACE:-$HOME/lobster-workspace}"
+CONFIG_DIR="${LOBSTER_CONFIG_DIR:-$HOME/lobster-user-config}"
+
+PIPER_BIN="${PIPER_BIN:-}"
+if [ -z "$PIPER_BIN" ]; then
+    if [ -x "/usr/local/bin/piper" ]; then
+        PIPER_BIN="/usr/local/bin/piper"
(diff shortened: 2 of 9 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): [verdict: PASS]

## Overall assessment

Solid, well-structured feature. The architecture choices are sound, error handling is thorough, and the fallback-to-text guarantee is correctly implemented at multiple layers. The PR is ready to merge with a few minor issues noted below.

---

## Area 1: `lobster_bot.py` — voice handler (type=voice branch, fallback logic)

**Correctness: good.** The guard `reply_type == 'voice' and voice_path and chat_id and bot_app` is correct. The inline `from telegram import InputFile as _InputFile` is a minor style nit but harmless.

**One real bug:** In the failure fallback path, if `send_voice` raises but `text` is empty or absent (e.g. the outbox JSON was written without a `text` field), the user gets nothing — no voice note, no text, silent failure. The outer `except` block should emit a logged warning in this case even if it cannot deliver to the user:

```python
else:
    log.warning(f"send_voice failed for {chat_id} and no fallback text available — message lost")
```

The current code does log at `warning` level that voice failed, but if `text` is falsy it skips to `os.remove(filepath)` with no further notification. The user would never know a message was attempted.

**Temp file cleanup gap:** `os.remove(voice_path)` is only called in the `try` block on success. In the `except` branch, `voice_path` is never explicitly deleted (the fallback sends text but leaves the OGG on disk). Minor leak, but worth noting. (The OGG lives in `/tmp/lobster-tts-...` so it will be cleared on reboot, but long-lived servers could accumulate stale files.)

**Reply threading:** The handler calls `bot.send_voice(chat_id=chat_id, ...)` without passing `reply_to_message_id`. The text fallback (`send_message`) also omits it. Voice replies will appear unthreaded in the chat. Possibly intentional for v1, but noting it for completeness.

---

## Area 2: `inbox_server.py` — lazy import of `tts.piper`

**Correct and well-structured.** The `try/except ImportError` guard ensures piper module absence doesn't crash the MCP server. The fallback to `handle_send_reply` is clean.

**One subtle issue:** After TTS succeeds and the outbox file is written, `tts_result.cleanup()` is never called in the success path. The `ogg_path` is now "owned" by `lobster_bot.py` (which deletes it after `send_voice`), but if `lobster_bot` crashes before sending — or if the bot is not running when the outbox file is processed — the OGG in `/tmp/lobster-tts-...` will leak permanently. `TtsResult._cleanup_paths` is set on the result object, but since control passes to the bot process, there is no mechanism to clean up if the bot never sends.

This is a known architectural tension with the outbox pattern (the server writes, the bot reads and deletes). It is acceptable for v1, but should be documented in `known_limitations`.

**Race condition (low severity):** `reply_id = f"{int(_time.time() * 1000)}_telegram_voice"` — millisecond timestamps could theoretically collide under rapid calls. Not a real concern in practice for this use case, but a `uuid.uuid4()` would be more robust.

---

## Area 3: `install.sh` — piper section (binary + shared libs + espeak-ng-data)

**Architecture is correct.** Arch detection, idempotent install checks, and soft-failure on download errors are all handled.

**Issue with `python3 -c` shell variable interpolation:** The GitHub release API URL is parsed with:

```bash
python3 -c "import sys,json; \
    urls=[a['browser_download_url'] for a in data.get('assets',[]) \
          if 'linux_${PIPER_ARCH}' in a['name'] ..."
```

The `${PIPER_ARCH}` expands inside the double-quoted string before being passed to Python. This is intentional and works, but it makes the script sensitive to unexpected characters in `PIPER_ARCH`. Since `PIPER_ARCH` only takes values `amd64`, `aarch64`, or `armv7`, this is safe in practice. Worth a comment to make the intent clear.

**`ldconfig` after copying `.so` files:** `sudo ldconfig 2>/dev/null || true` correctly runs after the lib copy. Good.

**`espeak-ng-data` conflict risk:** `sudo cp -r "${PIPER_EXTRACT_DIR}/espeak-ng-data" /usr/share/ 2>/dev/null || true` will silently overwrite a system `espeak-ng-data` if one exists. On Ubuntu 24.04 this is unlikely to cause issues, but the PR description acknowledges this. Adding a check (`if [ ! -d /usr/share/espeak-ng-data ]`) would be safer for future distros.

**Smoke test logic:** `if "$PIPER_CMD" --help &>/dev/null 2>&1` — piper does not necessarily exit 0 for `--help`. This may spuriously emit the warning on some piper versions. Consider `piper --version` or a real synthesis smoke test instead.

---

## Area 4: Migration 70 in `upgrade.sh`

**Correct and consistent with `install.sh`.** The migration is properly idempotent, uses the same architecture detection logic, and does not abort on failure.

**One gap vs. `install.sh`:** The migration does not run the smoke test that `install.sh` runs after binary installation. Minor — the smoke test is optional — but would be nice for consistency.

**`migrated` counter logic:** The migration only increments `migrated` if the binary is successfully extracted (`_bin` found). If the model download succeeds but the binary download fails (e.g. unsupported arch), `migrated` is not incremented and the migration will attempt model install again on the next upgrade. The model download increments `migrated` independently, so partial states are handled correctly. This is fine.

---

## Minor issues (non-blocking)

1. **`send-voice-note.sh` line 266:** `ffmpeg ... 2>&1 | tail -3` — the pipe means `ffmpeg`'s exit code is consumed by `tail`. The `if !` check tests `tail`'s exit code (always 0), not ffmpeg's. The ffmpeg failure check on line 271 (`if [ ! -s "$OGG_FILE" ]`) catches the empty-output case, but a ffmpeg error that produces a non-empty partial file would silently pass. Use `set -o pipefail` (already set at the top with `set -euo pipefail`) — actually, `pipefail` is set, so this is fine. Ignore this note.

2. **`piper.py` `TtsResult` frozen field:** `_cleanup_paths` is defined with `field(default_factory=list)` on a non-frozen dataclass. The class is not `frozen=True` (unlike `PiperConfig`), so mutability is intentional. Fine.

3. **`text_to_voice_file` cleanup on synthesis failure** uses an inline `import shutil as _shutil`. This is already imported at the top of the file as `shutil`. The duplicate inline import is harmless but inconsistent; the top-level `shutil` should be used directly.

4. **Test coverage:** 29 unit tests cover the TTS module well. No tests for the MCP handler (`handle_send_voice_note`) or the bot handler (`process_reply` voice branch) — these are tested by the existing `test_bot/` suite per the PR description (217 tests). Acceptable, though integration tests for the handler would strengthen confidence.

---

## Summary

The core implementation is correct and production-ready. The fallback-to-text guarantee works at three independent layers (source check, import guard, TTS failure). Install and migration scripts are robust. The two most notable issues are:

- **Silent message loss** when voice fails and `text` field is empty (minor, easy fix)
- **OGG temp file leak** if the bot is not running when the outbox file is deposited (acceptable for v1, worth documenting)

Neither is a blocker. Recommending **PASS** with suggestions to address the message-loss logging gap before or shortly after merge.

sayhar:
Fixed both non-blocking issues from the PASS review (commit b1ca288).

**Issue 1 — Silent message loss when voice fails with no text field**

The original fallback path only sent text `if text:` — so if the outbox message had an empty or missing `text` field, the user received nothing when voice delivery failed. Similarly, if `voice_path` was absent from a `type=voice` message, it fell through the entire voice handler block silently.

Fix: `type=voice` is now always intercepted unconditionally. A `_send_voice_fallback()` helper always delivers *something* — the `text` field if present, or the placeholder `"[Voice note could not be delivered]"` if both are absent.

**Issue 2 — OGG temp file leak when bot is not running**

The original guard was `if reply_type == 'voice' and voice_path and bot_app:` — if `bot_app` was `None` (Telegram bot not running when the outbox file was deposited), the entire block was skipped including OGG cleanup. The temp file would sit in `/tmp/lobster-tts-*/` indefinitely.

Fix: The `type=voice` check is now always entered first. The not-running case is detected early and calls `_cleanup_ogg(voice_path)` before removing the outbox file and returning.

**Tests**: 6 new tests in `TestVoiceNoteHandling` covering all failure paths — 225 passed, 0 failed.

sayhar (commented):
🤖🦞 Lobster (reviewer): [verdict: PASS]

Re-review of commit b1ca288. Both original issues are confirmed fixed.

---

## 1. `_cleanup_ogg` and `_send_voice_fallback` closures — clean

Both helpers are defined inside the `if reply_type == 'voice':` block and close over `chat_id`, `text`, and `bot_app`, which are all resolved before the block is entered and not mutated afterward. No late-binding gotcha.

`_cleanup_ogg` takes an explicit `path` argument rather than closing over `voice_path`, so it can be called safely with any path at any point. Correct.

`_send_voice_fallback` correctly checks `if chat_id and bot_app` before attempting `send_message`, and logs an explicit error if it can't deliver — no silent drop in the fallback either. Correct.

One noted design choice (not a bug): the early guard `if not bot_app or not chat_id:` drops the message silently (logs error, cleans OGG, removes outbox file) without attempting a text fallback. This is consistent with how all other message types behave when the bot isn't running — there's no delivery path when `bot_app` is None. The test correctly does not assert a fallback in that scenario.

---

## 2. `type=voice` unconditional intercept — handles all shapes

The `if reply_type == 'voice':` block returns in every code path (guard branch, missing voice_path branch, success branch, send_voice exception branch), so it never falls through to the `type=photo` handler. Correct.

Outbox messages produced by `handle_send_voice_note` always have `type`, `voice_path`, `chat_id`, `text`, `source`, `id`, and `timestamp`. The bot reads each with `.get()`, so missing optional fields are handled gracefully.

---

## 3. The 6 tests — both failure paths are verified

Bug A (silent drop when `voice_path` missing):
- `test_voice_note_missing_voice_path_sends_text_fallback` — asserts `send_voice` not called, `send_message` called with the fallback text. Directly verifies the fix.
- `test_voice_note_missing_voice_path_no_text_sends_placeholder` — asserts a non-empty placeholder is sent when both `voice_path` and `text` are absent. Correct.

Bug B (OGG leak when bot not running):
- `test_voice_note_bot_not_running_cleans_up_ogg` — sets `bot_app = None`, asserts OGG file deleted and outbox file removed. Directly verifies the fix.

The remaining three tests (happy path, `send_voice` exception with text, `send_voice` exception with empty text) are well-targeted and add meaningful regression coverage. Assertions are substantive throughout.

---

## 4. No new issues

Cleanup ownership in `inbox_server.py` → `lobster_bot.py` is clean: `tts_result.cleanup()` is called on all failure paths in the MCP handler; the bot's `_cleanup_ogg` handles it after a successful Telegram send. The OGG file lifetime is correctly managed end-to-end.

`TtsResult._cleanup_paths` is properly reset in both the WAV and OGG failure paths in `piper.py` (via `.clear()` after `rmtree`), so double-delete is not possible.

---

PASS. Fixes are correct, no regressions, tests are meaningful.

PR #1592, opened 2026-04-15
fix(agents): add SHARED ARTIFACT constraint to prevent project-specific narrowing

### LINKED ISSUE #1591: functional-engineer produces project-specific code when asked to improve shared tools
🤖🦞 Lobster (engineer):

## What happened

PR #1531 attempted to add \"PR awareness\" to \`functional-engineer.md\` — a shared, general-purpose engineering subagent prompt used across all projects. The implementation hardcoded \`--repo SiderealPress/lobster\` and assumed \`gh\` CLI tooling, making it Lobster-specific. The PR was correctly rejected and closed.

## Why it happened

The task prompt that spawned the work did not state that the improvement must be repo-agnostic and toolchain-agnostic. The functional-engineer subagent was operating in a Lobster context, knew the Lobster repo and tooling, and filled in those specifics when no constraint said otherwise. This is a natural inference failure: the agent optimized for the immediate task rather than the broader constraint (shared reusability) that was implicit but unstated.

## What we're trying to fix

This is a class of mistake, not a one-off:

- A general-purpose shared prompt (or skill, or tool) is improved in-context.
- The agent doing the improvement has project-specific knowledge available.
- No constraint on generality was stated in the task.
- The agent produces something that works for the current project but silently breaks reusability.

The fix should make it structurally harder to accidentally narrow a shared artifact — either by making the generality constraint explicit at the right point in the workflow, or by catching the violation before it ships.

## Scope

This issue covers the design question: where and how should the system enforce or signal the generality constraint when modifying shared artifacts? A follow-up PR should implement the chosen approach.

**Do not propose implementation in this issue — define the problem clearly so design can happen first.**

sayhar:
## What problem this solves

When engineers improve a shared agent prompt (like `functional-engineer.md`) while running in a Lobster context, they have no signal that the output must remain project-agnostic. They fill in what they know — `SiderealPress/lobster`, `gh` CLI, `~/lobster-workspace/` paths — and the shared artifact silently becomes project-specific. PR #1531 is the concrete example: it was correctly rejected, but the root cause (no constraint signal, no catch at review time) was not addressed.

This PR fixes both the production end (no signal when editing) and the review end (no check when reviewing).

## What changed

**`functional-engineer.md`**: A `SHARED ARTIFACT` HTML comment is prepended before the YAML front matter. It states the constraint directly where the file is read. Any engineer asked to improve this file now sees the constraint before seeing any content.

**`review.md`**: A new bullet is added to the "Posting reviews" section. When any PR modifies a file that carries a `SHARED ARTIFACT` header, the reviewer is required to scan for project-specific identifiers (hardcoded repo names, non-universal tool names, Lobster-specific paths) and issue a FAIL with a specific callout if any are found.

These two changes together form a closed loop: the constraint is stated in the file, and any PR that touches the file gets checked against it automatically.

## What this does NOT change

All other content in `functional-engineer.md` is unchanged — no behavioral instructions, no workflow steps, no tool preferences were modified. The reviewer check in `review.md` applies only when a `SHARED ARTIFACT` header is present; it does not alter review behavior for ordinary files.

## Would this have caught PR #1531?

Yes. `functional-engineer.md` would have had the `SHARED ARTIFACT` header, and the reviewer check would have flagged `--repo SiderealPress/lobster` as a hardcoded repo name. The PR would have received a FAIL verdict with a specific callout before merge.

## Tests run

- [x] Manual diff inspection — verified header appears at line 1 of `functional-engineer.md` before the YAML front matter, and the reviewer check appears at line 188 of `review.md` following the existing doc-PR check
- [ ] Automated tests — N/A: these are prompt-only text changes with no executable code paths

## Manual test

N/A — this change is entirely internal prompt text. No external services, no file I/O affecting runtime state, no systemd/cron changes. The user-visible effect is behavioral (future engineer and reviewer agents will behave differently), which is not self-verifiable in isolation.

Closes #1591


sayhar (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK — The approach is correct but the HTML comment placement in functional-engineer.md will break YAML front matter parsing.

## Summary

This PR addresses issue #1591 with a two-part fix: a self-documenting header in `functional-engineer.md` and a new reviewer check in `review.md`. The design is sound and would have caught PR #1531. However, the implementation has a structural defect in how the header comment is placed.

## Findings

### Critical: HTML comment placed before YAML front matter (functional-engineer.md)

**Severity: NEEDS-WORK**

The PR prepends a 3-line HTML comment before the `---` YAML front matter delimiter:

```
<!-- SHARED ARTIFACT: ... -->
---
name: functional-engineer
...
---
```

YAML front matter parsers (including Claude Code's agent file parser) require the opening `---` to be the **first characters of the file**. When any content precedes `---`, the front matter is not recognized. This means the `name`, `description`, `model`, and `color` fields will be silently ignored. The agent may fail to appear in agent listings and cannot be invoked by name.

This can be confirmed by inspection: the current `functional-engineer.md` starts with `---` on line 1. After this PR it would start with `<!--` on line 1. The harness will see a file that does not begin with `---` and treat the entire file as a plain body with no metadata.

**Fix**: Move the HTML comment to after the closing `---` of the front matter block, placing it in the body. For example:

```markdown
---
name: functional-engineer
model: opus
color: orange
description: "..."
---

<!-- SHARED ARTIFACT: This agent definition is used for ALL projects, not just Lobster.
     All instructions must remain project-agnostic: no hardcoded repo names, no
     project-specific CLI tools, no Lobster-specific path assumptions. -->

> **Subagent note:** ...
```

This preserves the front matter and still places the constraint prominently at the top of the agent body, where any engineer reading the prompt will see it before the instructions.

### review.md check: Covers the PR #1531 failure mode

**Severity: OK**

The new reviewer bullet explicitly checks for hardcoded repo names (`SiderealPress/lobster`), non-universal tool assumptions (`gh` CLI without fallback), and Lobster-specific paths (`~/lobster-workspace/`). This directly covers both failure modes from PR #1531. The check is conditioned on the presence of a `SHARED ARTIFACT` header, so it does not affect reviews of ordinary files.

One nuance: the check instructs the reviewer to look for "tool names that are not universally available (e.g. `gh` CLI assumed without fallback)". This is slightly imprecise — `gh` CLI is actually used throughout Lobster and is universally available in this environment; the problem in PR #1531 was using it without `--repo`, forcing Lobster-specific defaults. The check would still catch the real issue, but the framing could be tightened. This is a minor nit, not a blocker.

### Other agent files: No missing SHARED ARTIFACT headers

**Severity: OK**

The other agents in `.claude/agents/` (`lobster-generalist`, `lobster-ops`, `lobster-auditor`, `nightly-consolidation`, `review.md`, etc.) are all Lobster-specific by name and content. `functional-engineer.md` is the only genuinely cross-project general-purpose agent. No additional files need the header.

### PR description references user.dispatcher.bootup.md — not present in diff

**Severity: OK (non-issue)**

The task prompt asked to verify "the dispatcher bootup change in user.dispatcher.bootup.md." That file lives in `~/lobster-user-config/agents/` (private, not in repo) and is not modified by this PR. The PR description does not claim to touch it either. This is a non-issue — the review task prompt was likely referencing a design note, not an actual change in this PR.

## Would this have caught PR #1531?

Yes, conditional on the front matter issue being fixed. Once `functional-engineer.md` has a working `SHARED ARTIFACT` header, any future PR modifying that file will trigger the reviewer check, which explicitly calls out hardcoded repo names and non-universal path assumptions.

## Required change before merge

Move the HTML comment to the agent body (after the closing `---`), not before the opening `---`. The comment's effectiveness as a constraint signal does not depend on being pre-front-matter — an agent reading its own prompt reads the entire content, including the body.

---
**Escape valve**: If this assessment is wrong and Claude Code's agent parser does tolerate content before `---`, this PR can be merged as-is. A quick local test suffices: after applying the patch, run `claude --list-agents` (or equivalent) and confirm `functional-engineer` still appears with `model: opus`. If it does, the concern is resolved.

sayhar (commented):
🤖🦞 Lobster (reviewer): Re-review after fix commit a1ac83a

PASS: The SHARED ARTIFACT comment is now correctly placed after the closing YAML front matter delimiter.

## What was fixed

The previous NEEDS-WORK finding was that the `<!-- SHARED ARTIFACT -->` HTML comment appeared before the `---` closing delimiter of the YAML front matter, which would break any agent parser that expects the front matter block to open and close with `---` before any other content. Commit a1ac83a moved the comment to after the closing `---`.

## Verification of the fix

Current file structure on `fix/shared-artifact-constraint` (confirmed via `gh api` on the branch):

```
---                          ← YAML front matter open
name: functional-engineer
description: "..."
model: opus
color: orange
---                          ← YAML front matter close  ✓
                             
<!-- SHARED ARTIFACT: ... -->  ← comment is HERE, after front matter  ✓

> **Subagent note:** ...
```

YAML front matter (name, description, model, color) is intact and unmodified. The comment now appears as the first non-front-matter content in the file, exactly where an agent reading the file will encounter it immediately.

## Review checklist

1. **Comment placement** — PASS. Comment is after the closing `---`, not before it.
2. **YAML front matter intact** — PASS. All four fields (name, description, model, color) are present and unaltered.
3. **Would this catch PR #1531?** — PASS. The `review.md` check (added in this same PR) requires reviewers to scan for project-specific identifiers in any file carrying a `SHARED ARTIFACT` header. PR #1531 contained `--repo SiderealPress/lobster` — this would have been flagged as a hardcoded repo name and issued a FAIL. The closed loop is intact.
4. **review.md check language** — PASS. The new bullet is specific about what to look for (hardcoded repo names, non-universal CLI tool names, Lobster-specific paths) and clear about the required verdict (FAIL with specific callout). No ambiguity.
5. **Any remaining concerns** — None. The PR is narrow (5 additions, 0 deletions net), makes no behavioral changes to existing content, and solves the stated problem cleanly.

## Summary

Both the production signal (constraint visible when editing) and the review catch (enforced check when reviewing) are in place. The implementation is correct and complete. The original failure mode from PR #1531 would be caught at review time. No project-specific identifiers were introduced.

sayhar:
🤖🦞 Lobster: Closing. The SHARED ARTIFACT comment approach is wrong — functional-engineer.md is read by a subagent executing tasks, not by humans editing the file. Adding meta-instructions about file ownership into an agent prompt just confuses the agent. 

The correct fix is upstream: the dispatcher's task prompt must state generality constraints when asking an engineer to modify shared infrastructure. That change was made directly to user.dispatcher.bootup.md (user config, not committed). This PR is not needed.

Fixes #1591 via the dispatcher bootup change only.

PR #1593, opened 2026-04-15
fix(dispatcher): clarify ack policy — ban "Noted." and distinguish reply-only vs work-in-progress

### LINKED ISSUE #1019: fix: dispatcher must check if user message is a question before issuing ack responses

sayhar:
Closes #1019

## What problem this solves

The existing quick-ack rule told the dispatcher *what* to send ("On it.", "Looking into this.") but not *why* an ack exists. That left a gap: a dispatcher that says "Noted." on receiving a task is technically following the letter of the rule (brief, non-repetitive) while completely defeating its purpose.

An ack's job is to signal that work is underway — not to acknowledge receipt (the transport layer already does that). "Noted." signals nothing actionable, so it should never appear.

## What changed

One clarifying sentence appended to the existing ack policy note in `.claude/sys.dispatcher.bootup.md`:

> Never say "Noted." alone — it doesn't tell the user whether work is happening. Use "On it — [what]" when kicking off background work. If just answering, reply directly with no preamble.

Nothing else changed. No new sections, no keyword lists, no question-detection logic.

## Before / after

**Before:** The rule said what ack text to use but was silent on "Noted." and didn't distinguish a work-kicking-off ack from a pure reply.

**After:** The rule explicitly bans "Noted." and clarifies: acks are only for background work. Direct answers need no preamble.

## Tests run

- [x] `git diff HEAD~1` — verified exactly 2 lines added, no other changes
- [ ] Live dispatcher test — N/A: this is a doc-only change with no behavior change in code. The change takes effect the next time the dispatcher context is loaded.

## Manual test

N/A — this is a documentation-only change to the dispatcher's prompt context. No external services, file I/O, or system calls involved.

## Definition of Done

- [x] Unit tests pass with proper mocking — N/A (doc-only)
- [x] User-visible outcome verified — the change is an instruction to the dispatcher; behavioral effect observed the next time dispatcher processes a message
- [x] Side-effect audit: no floods, no unscoped writes, no unintended volume — change is additive, two lines only

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -164,6 +164,8 @@ mcp__github__issue_read(owner="...", repo="...", ...)             # VIOLATION
 
 Note: The Telegram bot sends "📨 Message received. Processing..." automatically at the transport layer. Your ack is a second, dispatcher-level signal that work is underway.
 
+Never say "Noted." alone — it doesn't tell the user whether work is happening. Use "On it — [what]" when kicking off background work. If just answering, reply directly with no preamble.
+
 **Preferred pattern (use `claim_and_ack` for long tasks):**
 ```
 1. claim_and_ack(message_id, ack_text="On it.", chat_id=chat_id, source=source)
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -168,7 +168,7 @@ Never say "Noted." alone — it doesn't tell the user whether work is happening.
 
 **Preferred pattern (use `claim_and_ack` for long tasks):**
 ```
-1. claim_and_ack(message_id, ack_text="On it.", chat_id=chat_id, source=source)
+1. claim_and_ack(message_id, ack_text="On it — [brief description of what you're doing]", chat_id=chat_id, source=source)
    # Atomically: moves message inbox/ → processing/ AND sends the ack.
    # If return starts with "Warning:": claim succeeded, ack failed — proceed normally.
 2. Generate a short task_id (e.g. "fix-pr-475", "upstream-check")


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one minor note

## Summary

Two lines added to the ack policy section of `sys.dispatcher.bootup.md`, inserted between the transport-layer note (line 165) and the **Preferred pattern** block. Diff is clean — exactly 2 lines added, nothing else touched.

## Key questions

**1. Do the two lines fit naturally without duplicating anything?**

Yes. The existing policy (lines 161–163) says *when* to ack and *what words to use*, but is silent on what an ack is *for*. The new sentence fills that gap directly. The "If just answering, reply directly with no preamble" half is genuinely new — the original rule only described the acking path, not the no-ack path for direct answers. No duplication.

Placement is correct: it sits right after the transport-layer note that explains why the dispatcher ack exists at all, which is exactly the right place to add a constraint on that ack.

**2. Would the new wording have prevented the "Noted." behavior?**

Yes, for the plain-task case. "Never say 'Noted.' alone — it doesn't tell the user whether work is happening" is an explicit ban. A dispatcher following this rule would know "Noted." is off the table.

The question-detection failure described in #1019 (dispatcher sent "Noted." twice to a direct question) is partially addressed: "If just answering, reply directly with no preamble" makes clear that a question-response is not an ack situation. That said, there's no positive instruction that says "when the message is a question, answer it" — the fix relies on the dispatcher inferring that "not an ack situation" means "answer the question." For the specific failure in #1019 (ignoring a direct question entirely), the new wording is helpful but not airtight. That said, #1019 proposed a doc rule as its tier-1 fix; this delivers that. A hook-based enforcement would be tier 2 and can be a follow-on if the doc fix proves insufficient.

**3. Is anything over-specified or confusing?**

No. The sentence is tight and actionable. "On it — [what]" as the prescribed form is a small improvement over "On it." alone (it tells the user *what* is starting), though this is additive — the Preferred pattern block still shows `ack_text="On it."` without a bracketed descriptor. Minor inconsistency, not blocking: the policy note and the code example can drift slightly without causing harm, and updating the code example would be a separate PR.

**4. Does it reference #1019 correctly?**

The commit message says "Refs #1019" and the PR body says "Closes #1019" — both correct. The task brief mentioned #1523 but that issue doesn't exist; GitHub resolves this PR's closing reference to #1019 only, which is right.

## Minor nit (non-blocking)

The `claim_and_ack` Preferred Pattern block still shows `ack_text="On it."` (no `[what]` descriptor). The new policy says to use "On it — [what]". These are now slightly inconsistent. A follow-up PR could align the example, but it doesn't block merging.

## Verdict

PASS. The change is minimal, correctly placed, and will suppress "Noted." and clarify the no-preamble direct-answer path. Safe to merge.

PR #1595, opened 2026-04-15
fix(mcp): remove "Noted." auto-reply fallback from mark_processed

### LINKED ISSUE #1594: MCP server auto-sends "Noted." fallback when mark_processed called without send_reply
## Bug

In \`src/mcp/inbox_server.py\` around line 4717, \`handle_mark_processed\` contains a fallback that auto-sends "Noted." to the user when \`mark_processed\` is called without a prior \`send_reply\`. The comment near line 4709 explicitly says "sending 'Noted.' is never correct" — but the fallback still fires.

This caused Sahar to receive "Noted." 10+ times in a single day, including as responses to self-checks and system messages routed through non-zero chat_ids.

## Root Cause

The intent of the guard was to prevent silent message drops. The implementation chose "auto-reply" as the enforcement mechanism. But:

1. "Noted." as a reply is actively harmful — it masquerades as an intentional dispatcher response
2. For non-question messages (reactions, button presses) the code already skips it, but the comment says "sending 'Noted.' is never correct" — so it should be skipped for ALL message types
3. The right fix is to log a warning and mark processed silently. If the dispatcher forgot to reply to a user message, that's a dispatcher bug — not something to paper over with "Noted."

## Proposed Fix

Remove the auto-send fallback entirely. Replace with a log warning so the situation is visible in logs and monitoring without spamming the user. The message should still be marked as processed (to avoid reprocessing loops).

The existing type-specific guards (callback, reaction, fake chat_ids) can also be collapsed — once the fallback is removed, there's nothing to skip.

## Impact

- Users receiving "Noted." for self-checks, subagent completions, and other non-reply-requiring messages
- Dispatcher behavior is obscured: "Noted." looks like an intentional response

## References

- Related: #1019 (dispatcher must answer questions, not just ack — separate issue)

sayhar:
## What and why

Before this PR, `mark_processed` would auto-send "Noted." to the user whenever it was called without a prior `send_reply` for any user-facing message with a real Telegram chat_id. This caused Sahar to receive "Noted." 10+ times in a single day — including as phantom responses to self-checks, subagent completions, and other messages that never required any reply.

The comment in the code (line 4709) already said "sending 'Noted.' is never correct" — but the auto-send still fired. This PR makes the code match the comment.

## The fix

The auto-send fallback block is removed entirely. When `mark_processed` is called without a prior reply on a user-facing message, the server now logs a `WARNING` and marks the message processed silently. The log entry is: `mark_processed called without send_reply for user message <id> (type=<type>, chat=<chat_id>) — dispatcher may have dropped a reply`.

This is the correct behavior: a missing reply is a dispatcher bug. Papering over it with "Noted." is worse than silence — it masquerades as an intentional response and prevents the bug from being noticed.

The type-specific guards (callback, reaction, fake chat_id skip-conditions) are also removed, since they only existed to prevent the fallback from firing in those cases. With no fallback, there is nothing to skip.

**Functional patterns used:** pure log-and-proceed, no side effects outside the guard block.

## Before / after

```
Before (for a real chat_id, user message type, no prior send_reply):
  mark_processed → check reply timestamp → no reply found
    → write "Noted." to outbox + sent dirs
    → track reply
    → WARNING log "Auto-reply fallback triggered..."
    → mark processed

After:
  mark_processed → check reply timestamp → no reply found
    → WARNING log "mark_processed called without send_reply..."
    → mark processed
```

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_message_tools.py::TestMarkProcessed -v` — 11 passed (8 pre-existing + 3 new)
- [x] `uv run pytest tests/unit/ -x -q` — 2124 passed, 1 pre-existing failure (`test_falls_back_to_static` in `test_memory.py` — PathGuardError unrelated to this change, fails identically on `main`)
- [x] `uv run python -m py_compile src/mcp/inbox_server.py` — clean

## Manual test

N/A — this change is entirely internal to the MCP server's `handle_mark_processed` function. It removes writes to the outbox directory; no external service calls, no Telegram API calls, no file I/O affecting runtime state beyond what is already covered by unit tests. The user-visible outcome (no more phantom "Noted." messages) cannot be unit-tested but is directly caused by the outbox write removal verified in `test_no_fallback_sent_for_real_chat_id_without_reply`.

## Definition of Done

- [x] Integration/manual test — not applicable (pure internal logic change, no external services touched)
- [x] User-visible outcome: no "Noted." outbox file is written — verified by `test_no_fallback_sent_for_real_chat_id_without_reply`
- [x] Side-effect audit: removed writes only — strictly fewer side effects than before
- [x] External service calls: none in this code path

Closes #1594

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4692,16 +4692,22 @@ async def handle_mark_processed(args: dict) -> list[TextContent]:
     if not found:
         return [TextContent(type="text", text=f"Message not found: {message_id}")]
 
-    # Guard: check that a reply was sent for user-facing messages.
-    # Uses msg type (not source) to classify — source is the routing destination
-    # and cannot distinguish a direct user message from a subagent_result that
-    # happens to carry source="telegram" for delivery.
-    # If no reply was sent, auto-send a fallback reply instead of returning a
-    # soft warning (which the LLM ignores, causing silent message drops).
+    # Guard: log a warning if a user-facing message is being marked processed
+    # without a prior send_reply.  This is a dispatcher bug — the dispatcher
+    # should always reply to user messages before marking them processed.
+    #
+    # We intentionally do NOT auto-send any fallback reply here (issue #1594).
+    # The previous implementation auto-sent "Noted." which is actively harmful:
+    # it masquerades as an intentional response and caused repeated spurious
+    # "Noted." messages to be delivered to the user for self-checks, subagent
+    # completions, and other non-reply-requiring messages.
+    #
+    # Correct fix: log the missing reply so it's visible in monitoring, then
+    # mark processed silently.  A missing reply is a dispatcher bug, not
+    # something to paper over with an auto-reply.
     if not force:
         try:
             msg = json.loads(found.read_text())
-            source = msg.get("source", "")
             msg_type = msg.get("type", "")
             chat_id = msg.get("chat_id", 0)
             msg_ts_raw = msg.get("timestamp", "")
@@ -4719,42 +4725,12 @@ async def handle_mark_processed(args: dict) -> list[TextContent]:
                 chat_key = str(chat_id)
                 reply_ts = _recent_replies.get(chat_key, 0.0)
                 if reply_ts < msg_epoch:
-                    # No reply was sent for this human message.
-                    # Skip auto-reply for callback (button press) messages —
-                    # the bot already answered the callback query inline.
-                    # Skip auto-reply for reaction messages — reactions are
-                    # signals that the dispatcher processes contextually;
-                    # sending "Noted." is never correct.
-                    if msg_type == "callback":
-                        log.info(f"Skipping auto-reply fallback for callback message {message_id}")
-                    elif msg_type == "reaction":
-                        log.info(f"Skipping auto-reply fallback for reaction message {message_id}")
-                    elif abs(chat_id) <= 1_000_000:
-                        # Fake/test chat_id — Telegram rejects delivery; skip to avoid dead-letter buildup
-                        log.info(f"Skipping auto-reply fallback for fake/test chat_id {chat_id}")
-                    else:
-                        # Auto-send a fallback reply so the user isn't silently ignored
-                        fallback_text = "Noted."
-                        fallback_id = f"{int(time.time() * 1000)}_{source}"
-                        fallback_data = {
-                            "id": fallback_id,
-                            "source": source,
-                            "chat_id": chat_id,
-                            "text": fallback_text,
-                            "timestamp": datetime.now(timezone.utc).isoformat(),
-                            "_fallback": True,
-                        }
-                        if source == "bisque":
-                            outbox_file = BISQUE_OUTBOX_DIR / f"{fallback_id}.json"
-                        else:
-                            outbox_file = OUTBOX_DIR / f"{fallback_id}.json"
-                        atomic_write_json(outbox_file, fallback_data)
-
-                        sent_file = SENT_DIR / f"{fallback_id}.json"
-                        atomic_write_json(sent_file, fallback_data)
-
-                        _track_reply(chat_id)
-                        log.warning(f"Auto-reply fallback triggered for message {message_id} (chat {chat_id})")
+                    # No reply was sent for this human message — log and proceed silently.
+                    log.warning(
+                        f"mark_processed called without send_reply for user message "
+                        f"{message_id} (type={msg_type}, chat={chat_id}) — "
+                        "dispatcher may have dropped a reply"
+                    )
         except (json.JSONDecodeError, OSError):
             pass  # If we can't read the message, skip the guard
 
--- a/tests/unit/test_mcp_server/test_message_tools.py
+++ b/tests/unit/test_mcp_server/test_message_tools.py
@@ -819,6 +819,124 @@ class TestMarkProcessed:
             assert "processed" in result[0].text.lower()
             assert not (inbox / f"{msg_id}.json").exists()
 
+    # -----------------------------------------------------------------------
+    # Issue #1594 — "Noted." fallback removal
+    # -----------------------------------------------------------------------
+
+    def test_no_fallback_sent_for_real_chat_id_without_reply(
+        self, setup_dirs, message_generator, inbox_server_dirs
+    ):
+        """mark_processed must NOT write any outbox message when no reply was sent.
+
+        Before fix: the server auto-sent "Noted." to the outbox for any
+        user-facing message with a real (> 1_000_000) chat_id that was
+        processed without a prior send_reply.
+
+        After fix: the message is silently marked processed and no outbox
+        file is written.  The absence of a fallback is the correct behavior —
+        a missing reply is a dispatcher bug, not something to paper over.
+        """
+        inbox, processed = setup_dirs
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — this is a clean, correct fix with one stale docstring to update.

## Summary

The change does exactly what it should: removes the auto-reply fallback from `handle_mark_processed` and replaces it with a `WARNING` log. The code, comment, and PR description are all aligned. Tests pass. No regressions introduced.

---

## Item-by-item review

### 1. Is the `msg_type in USER_FACING_TYPES and chat_id != 0` guard the right scope?

Yes, with one clarification worth noting for the record.

`USER_FACING_TYPES` is an alias for `INBOX_USER_TYPES`, which includes `callback` and `reaction`. Those types previously had explicit skip-conditions to avoid the fallback. With the fallback gone, the guard fires a WARNING for callbacks and reactions too — which is accurate behavior. A callback or reaction that was processed without any reply is still a legitimate "dispatcher may have dropped a reply" signal.

The `chat_id != 0` check correctly excludes system messages routed through chat_id=0. That's the right boundary: 0 is the internal/system channel. Negative chat_ids (Telegram groups) pass through as intended, which is correct.

One edge case worth noting: `abs(chat_id) <= 1_000_000` was previously used to skip auto-reply for test/fake IDs. That guard is now removed. For `test_blocks_human_message_without_reply`, which uses `chat_id=123456`, the WARNING will now fire (whereas previously the code logged an info and skipped). This is a behavior change for test/fake IDs — they'll appear in WARNING logs. That's acceptable and arguably more correct: a test chat_id in production would be a bug worth surfacing.

### 2. Does `test_no_fallback_sent_for_real_chat_id_without_reply` actually verify the fix?

Yes, concretely. It:
- Uses a real (large) `chat_id` (8305714125) that would have triggered the old fallback
- Captures `outbox_before` before the call
- Asserts `outbox_after - outbox_before == set()` — no files written

This is the right assertion. The old code wrote to `OUTBOX_DIR` and `SENT_DIR`; patching `INBOX_DIR` and `PROCESSED_DIR` (but leaving `OUTBOX_DIR` as the real `inbox_server_dirs["outbox"]`) means any fallback write would show up. The test would have failed on the old code, and passes on the new code. That's a valid regression test.

### 3. Pre-existing test failure in test_memory.py / test_group_ack_policy.py

The PR description cites `test_falls_back_to_static` in `test_memory.py` as the pre-existing failure. Locally, the failure I see is `test_group_passive_message_no_ack` in `test_bot/test_group_ack_policy.py`. Confirmed this failure exists identically on `main` — it is entirely unrelated to this PR (it concerns group message ack behavior). The discrepancy in which test is mentioned is minor; both are pre-existing.

### 4. Stale docstring in `test_blocks_human_message_without_reply` — needs update (nit, blocking)

The docstring at line 720–725 describes the OLD behavior:

```
Guard auto-sends fallback reply for real chat_ids; skips for test/fake chat_ids.

For chat_ids <= 1_000_000 (test/fake IDs that Telegram would reject), the guard
skips the auto-reply fallback but still marks the message as processed.  The old
soft-warning behaviour ("No reply sent") was replaced by an auto-send-then-proceed
pattern to prevent silent message drops in production.
```

None of this is true after the fix. The guard no longer auto-sends anything; there is no fallback to skip. The docstring should describe what the test actually verifies post-fix: that a message with a small/fake chat_id is marked processed without any outbox write, and a WARNING is logged.

This needs to be updated before merge.

### 5. Is the WARNING log noise a concern?

It's appropriate. WARNING is the right severity — this represents a potential dispatcher bug (dropped reply). It should be visible in monitoring. It is not noisy in the expected steady state: a correctly-behaving dispatcher always calls `send_reply` before `mark_processed` for user messages, so this warning fires zero times in normal operation.

The log message is well-structured: `mark_processed called without send_reply for user message {id} (type={type}, chat={chat}) — dispatcher may have dropped a reply`. Clear, actionable, no sensitive data.

---

## Required fix before merge

**Update the `test_blocks_human_message_without_reply` docstring** (lines 720–725) to reflect post-fix behavior. The current text actively misleads future readers about how the code works.

Suggested replacement:

```python
"""Guard logs a WARNING and still marks processed for messages without a prior reply.

For any chat_id (including small/fake IDs that Telegram would reject), the guard
no longer auto-sends any fallback.  It logs a WARNING and marks the message
processed silently.  This tests the fake/small chat_id case specifically.
"""
```

---

## Verdict

PASS conditional on fixing the stale docstring. The core logic change is correct, the test coverage is solid, and the fix directly addresses the reported issue with no unintended side effects.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS conditional — clean fix with one stale docstring to update.

## Summary

Removes the auto-reply fallback from `handle_mark_processed` and replaces it with a WARNING log. Code, comment, and PR description are all aligned. 11/11 TestMarkProcessed tests pass.

---

## Item-by-item

### 1. Is `msg_type in USER_FACING_TYPES and chat_id != 0` the right guard scope?

Yes. `USER_FACING_TYPES` == `INBOX_USER_TYPES`, which includes `callback` and `reaction`. Those types previously had explicit skip-conditions to avoid the fallback; with the fallback gone, the guard now fires a WARNING for those types too — which is correct. A callback or reaction processed without a reply is a legitimate "dispatcher may have dropped a reply" signal.

`chat_id != 0` correctly excludes system messages on the internal channel. Negative chat_ids (Telegram groups) pass through, which is intended.

One behavior change worth noting: the old `abs(chat_id) <= 1_000_000` skip for fake/test chat_ids is gone. Test chat_ids in production will now produce WARNING-level log entries rather than INFO-level. That's actually more correct.

### 2. Does `test_no_fallback_sent_for_real_chat_id_without_reply` verify the fix?

Yes, concretely. It captures `outbox_before`, calls `handle_mark_processed` on a message with a real chat_id and no prior reply, then asserts `outbox_after - outbox_before == set()`. The outbox directory is patched to the real `inbox_server_dirs["outbox"]`, so any fallback write would be detected. This test would have failed on the old code and passes on the new code.

### 3. Pre-existing test failure

The PR cites `test_falls_back_to_static` in `test_memory.py`. Locally I see `test_group_passive_message_no_ack` in `test_bot/test_group_ack_policy.py` as the failing test — confirmed it fails identically on `main`. Either way, entirely unrelated to this change.

### 4. Stale docstring in `test_blocks_human_message_without_reply` — needs update (blocking nit)

The docstring at lines 720-725 still describes the old behavior:

> "Guard auto-sends fallback reply for real chat_ids; skips for test/fake chat_ids... The old soft-warning behaviour was replaced by an auto-send-then-proceed pattern..."

None of this is true post-fix. The guard no longer auto-sends anything. This docstring actively misleads future readers. Please update it to describe what the test actually verifies now: message is marked processed, no outbox file is written, a WARNING is logged.

### 5. WARNING log noise

Not a concern. The warning fires zero times in normal operation (a correctly-behaving dispatcher always calls `send_reply` before `mark_processed`). When it does fire, it's actionable: something dropped a reply. The message format is clear and contains no sensitive data.

---

## Required fix

Update the `test_blocks_human_message_without_reply` docstring (lines 720-725) to reflect current behavior before merge.

## Verdict

PASS once the stale docstring is updated. The logic change is correct and the test coverage is solid.

PR #1608, opened 2026-04-15
fix: detect frozen dispatcher after MCP restart via thinking-heartbeat (#1439)

sayhar:
> **Lobster (ops):** Near-duplicate of #1512 which was inadvertently merged into local-dev instead of main. This re-targets to main for proper inclusion in the release branch. Will add local soak notes after ~24h.

## Problem

When the MCP server crashes and restarts mid-session, the dispatcher's HTTP connection breaks silently. The dispatcher process stays alive (PID check passes), so `_write_session_lost_reminder()` suppressed the recovery hint — leaving the dispatcher frozen for up to 21 minutes until the health check killed it.

This is the root cause documented in issue #1439.

## Root cause

`_write_session_lost_reminder()` has a PID guard: if the dispatcher PID is alive, it assumes a clean Claude Code auto-update reconnect and skips the reminder. This correctly handles CC auto-updates but silently fails when MCP crashes — the PID is alive but the connection is broken.

## Fix

- **Add `_is_dispatcher_responsive()`**: reads `last_thinking_at` from `lobster-state.json` (written by `thinking-heartbeat.py` on every PostToolUse). Returns `False` (frozen suspected) when the field is absent or older than 5 minutes — meaning the dispatcher has made no tool calls since MCP restarted.
- **Update `_write_session_lost_reminder()`**: when the dispatcher PID is alive, also check responsiveness. If responsive → suppress (clean CC reconnect). If not responsive → write the reminder anyway, triggering recovery.

## Effect

Maximum silent hang after MCP restart: reduced from ~21 minutes to ~5 minutes.

The 5-minute threshold is conservative: long LLM reasoning phases can take 3-4 minutes, so we need headroom. If the threshold proves too tight we can raise it, but 5 min is already much better than 21.

## Tests

All 11 tests in `test_session_lost_reminder_pid_guard.py` pass, including:
- New: `test_writes_reminder_when_dispatcher_alive_but_frozen` — covers the issue #1439 scenario
- Updated: `test_suppresses_reminder_when_dispatcher_alive_and_responsive` — now requires fresh heartbeat (clarifies intent)

Pre-existing failing test (`test_timeout_without_hibernate_flag_does_not_write_state`) is unrelated and was failing before this PR.

## Related
- Issue #1439
- PR #1403 (thinking-heartbeat hook — provides the `last_thinking_at` signal this fix depends on)
- PR #1431 (PID-based session-lost guard — this PR extends that logic)
- PR #1446 (external WFM watchdog — complementary mitigation)

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -725,6 +725,11 @@ _TRANSIENT_MODES = {"hibernate", "starting", "restarting", "waking"}
 _RECONNECT_GRACE_SECONDS = 30 * 60  # 30 minutes
 
 
+# How long without a tool-use heartbeat before we assume the dispatcher is frozen
+# (MCP connection silently broken after MCP server restart, issue #1439).
+_FROZEN_THRESHOLD_SECONDS = 5 * 60  # 5 minutes
+
+
 def _is_dispatcher_alive() -> bool:
     """Return True if the dispatcher process recorded in dispatcher.pid is alive.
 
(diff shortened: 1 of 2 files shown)


sayhar:
🤖🦞 Lobster (ops): Closing for rethink. This PR's _is_dispatcher_responsive() reads last_thinking_at from lobster-state.json, which is no longer written after #1484 simplified thinking-heartbeat.py to write only to logs/dispatcher-heartbeat. The freeze-detection logic is permanently inert. The blind-window problem is real but needs a fresh implementation against the current architecture. Filing a replacement issue.

PR #1610, opened 2026-04-15
feat(whatsapp): send QR code as Telegram image instead of terminal output

aeschylus:
## Summary

- Replaces the terminal-only QR display with a Telegram-delivered PNG image
- Drew can now trigger WhatsApp setup from Telegram and scan the QR directly from his phone — no SSH or terminal access needed
- Adds `qrcode` npm package for PNG rendering; removes `qrcode-terminal` from active use

## How it works

When the bridge generates a QR code:
1. `sendQrToTelegram()` renders the raw QR string to a 512px PNG using the `qrcode` package
2. A `qr_ready` system event is emitted with `image_path` pointing to the PNG file
3. `whatsapp_bridge_adapter.py` picks up the event and calls `handle_qr_ready_event()` (already implemented)
4. The adapter writes a Telegram outbox message with `type: "photo"` and `photo_file: <path>`
5. `lobster_bot.py` sends the photo to Drew's Telegram chat (already supports local `photo_file`)

## Deep link feasibility

WhatsApp's linked-device protocol is QR-code-only — no public deep link exists for this pairing flow. The QR image approach is the correct and complete solution.

## Test plan

- [ ] Start bridge with no session — verify QR PNG is sent to Telegram (check ~/messages/images/whatsapp-qr-*.png)
- [ ] Scan QR from Telegram — verify bridge logs [READY] WhatsApp bridge connected
- [ ] Verify bridge sends a "WhatsApp connected" confirmation to Telegram
- [ ] Run `lobster-whatsapp-qr` — verify it restarts the service rather than opening an interactive terminal

## Files changed

- `connectors/whatsapp/index.js` — QR handler now renders PNG and emits qr_ready event; EVENTS_DIR always defaults to ~/messages/wa-events
- `connectors/whatsapp/package.json` — adds qrcode dependency

Note: `scripts/lobster-whatsapp-qr` update lives in the lobster source repo on branch `feature/BIS-417-whatsapp-pairing-code` (pushed separately).


aeschylus (commented):
## Review: one blocking issue found — `image_path` is dropped by the adapter

The JS side looks correct: `sendQrToTelegram()` renders a 512px PNG via `QRCode.toFile()`, saves it to `~/messages/images/whatsapp-qr-<timestamp>.png`, and emits a `qr_ready` system event with `image_path` set. Package version `^1.5.4` for `qrcode` is fine.

**Blocking: `image_path` is silently stripped in `normalize_system_event()`**

In `src/bot/whatsapp_bridge_adapter.py`, all system events including `qr_ready` go through `normalize_system_event()`, which builds a fixed schema that never includes `image_path`:

```python
def normalize_system_event(event: dict) -> dict:
    return {
        "id": ...,
        "type": "system",
        "subtype": event.get("subtype", ""),
        "text": event.get("body") or "[WhatsApp system event]",
        # image_path is never forwarded ← the chain breaks here
    }
```

The inbox message written to `~/messages/inbox/` has no `image_path`, so `lobster_bot.py` can never find and send the PNG. The QR image gets saved to disk but never reaches Telegram.

**Fix:** Pass `image_path` through for `qr_ready` events:

```python
def normalize_system_event(event: dict) -> dict:
    normalized = {
        "id": f"{int(time.time() * 1000)}_wa_sys",
        "source": "whatsapp",
        "type": "system",
        "subtype": event.get("subtype", ""),
        "chat_id": "system",
        "user_id": "system",
        "user_name": "WhatsApp Bridge",
        "text": event.get("body") or "[WhatsApp system event]",
        "is_group": False,
        "group_name": "",
        "mentions_lobster": False,
        "timestamp": datetime.now(tz=timezone.utc).isoformat(),
    }
    # Forward image_path for qr_ready events so lobster_bot.py can send the PNG
    if event.get("image_path"):
        normalized["image_path"] = event["image_path"]
    return normalized
```

**Minor (non-blocking):** `qrcode-terminal` is still listed in `connectors/whatsapp/package.json` but is unused after this change — can be removed to keep deps clean.

Not merging until the `image_path` passthrough is in place, otherwise the QR-to-Telegram flow is silently broken.

PR #1617, opened 2026-04-16
feat: add daily cron for worktree + audio cleanup (issue #1609)

### LINKED ISSUE #1609: feat: log cleanup cron job with correct thresholds
Scope clarified (2026-04-16): Only two cleanup targets needed:

1. **Finished git worktrees** — clean up temporary worktrees left by agents after their work is done
2. **Old audio files** — delete processed voice message audio files after some retention period (e.g., 7 days)

NOT in scope: raw logs, structured logs, processed messages (those are retained)

Implementation: a cron job that runs periodically to clean both.

sayhar:
## Problem

Two categories of runtime artifacts accumulate indefinitely with no automated cleanup:

1. **Finished git worktrees** — agent branches create temporary worktrees in `~/lobster-workspace/projects/`. After the PR merges, the directory lingers because nothing runs `git worktree prune` or removes empty directories.
2. **Old audio files** — voice messages are transcribed and stored as text; the original `.ogg`/`.mp3` etc. files in `~/messages/audio/` are kept for debugging but have no TTL. On an active system, these accumulate unboundedly.

## What this adds

`scripts/cleanup-worktrees-audio.sh` — a single script that handles both:

- Runs `git worktree prune --verbose` on `~/lobster/` to remove stale administrative entries
- Walks `$LOBSTER_PROJECTS` and removes any empty directories that are no longer registered worktrees (directory gone but `git worktree list` no longer lists them)
- Deletes audio files with recognised extensions (`.ogg`, `.mp3`, `.m4a`, `.wav`, `.oga`) older than `CLEANUP_AUDIO_RETENTION_DAYS` (default: 7 days)

**install.sh** registers it as a daily 04:00 cron entry (`# LOBSTER-CLEANUP`), offset from nightly consolidation (03:00) to avoid log-write overlap.

**upgrade.sh** (Migration 75) installs the cron entry on existing deployments idempotently — safe to run repeatedly.

Both `CLEANUP_AUDIO_RETENTION_DAYS` and `CLEANUP_AUDIO_DIR` are overridable via environment variables, keeping tests hermetically isolated without any mocking layer.

## Functional patterns

Pure shell functions (`prune_worktrees`, `prune_audio`) with no shared mutable state. Each function reads its inputs from environment-derived variables set once at script top, so overriding them in tests gives full behavioural control without stubs.

## Tests run

- [x] `uv run pytest tests/smoke/test_cleanup_worktrees_audio.py -v` — 10 passed, 0 failed
- [x] `uv run pytest tests/smoke/ -v` — 65 passed, 0 failed (full smoke suite, no regressions)
- [x] `bash -n scripts/cleanup-worktrees-audio.sh && bash -n scripts/upgrade.sh && bash -n install.sh` — all syntax clean

## Manual test

N/A — this change is cron/scheduling infrastructure. No external service calls, no Telegram output, no database writes. The script was also run directly against the live system during development:

```
LOBSTER_MESSAGES=/tmp/pytest-missing-test-dir-$$ bash scripts/cleanup-worktrees-audio.sh
```

Output confirmed: correctly detected missing audio dir, logged "skipping", exited 0. The live run of the script also pruned two empty stale worktree directories (`fix/` and `feature/`) from `~/lobster-workspace/projects/` — confirming the worktree pruning path works end-to-end.

## Definition of Done

- [x] Unit tests pass with proper isolation (no production hits in automated tests)
- [x] Integration/manual test — confirmed script runs cleanly on live system
- [x] User-visible outcome: not applicable — this is a background cleanup job with no user-facing output
- [x] Side-effect audit: script only deletes files older than 7 days matching specific extensions, and only removes EMPTY directories; no floods, no unscoped writes, no unintended volume

Closes #1609

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1431,6 +1431,20 @@ step "Setting up OOM monitor cron..."
 
 success "OOM monitor configured (runs every 10 minutes, active only when LOBSTER_DEBUG=true)"
 
+#===============================================================================
+# Worktree + Audio Cleanup
+#===============================================================================
+
+step "Setting up worktree + audio cleanup cron..."
+
+# cleanup-worktrees-audio.sh runs daily at 04:00.
+# It prunes finished git worktrees and deletes audio files older than 7 days.
+chmod +x "$INSTALL_DIR/scripts/cleanup-worktrees-audio.sh" || true
+"$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-CLEANUP" \
+    "0 4 * * * $INSTALL_DIR/scripts/cleanup-worktrees-audio.sh >> $HOME/lobster-workspace/logs/cleanup.log 2>&1 # LOBSTER-CLEANUP"
+
+success "Worktree + audio cleanup configured (runs daily at 04:00)"
+
 # Ensure any lingering self-check cron entry is removed on fresh installs
 { crontab -l 2>/dev/null | grep -v "# LOBSTER-SELF-CHECK" | grep -v "periodic-self-check" || true; } | crontab -
 
--- a/scripts/cleanup-worktrees-audio.sh
+++ b/scripts/cleanup-worktrees-audio.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+#===============================================================================
+# cleanup-worktrees-audio.sh
+#
+# Periodic cleanup for two categories of stale runtime artifacts:
+#
+#   1. Finished git worktrees — worktrees whose branches no longer exist
+#      or have been merged. `git worktree prune` removes entries for worktrees
+#      whose directories are gone; this script additionally removes any
+#      lingering worktree directories under LOBSTER_PROJECTS_DIR.
+#
+#   2. Old audio files — voice message audio in ~/messages/audio/ older than
+#      AUDIO_RETENTION_DAYS. These files are transcribed and stored as text;
+#      the originals are only retained for debugging and can be safely deleted.
+#
+# Usage:
+#   ~/lobster/scripts/cleanup-worktrees-audio.sh
+#
+# Typically registered as a cron job (daily at 04:00):
+#   0 4 * * * ~/lobster/scripts/cleanup-worktrees-audio.sh >> ~/lobster-workspace/logs/cleanup.log 2>&1
+#===============================================================================
+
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Reviewed all four flagged focus areas.

**1. Migration 75 idempotency — PASS**
The `grep -qF "$CLEANUP_MARKER"` check against `crontab -l` is correct. Fixed-string match prevents false positives. Additionally, `cron-manage.sh add` strips-and-replaces by marker before appending, so even if the check were somehow skipped, no duplicate entry would be created. Double-layered idempotency.

**2. Worktree pruning safety — PASS**
Directories are removed only when all three conditions hold simultaneously: (a) direct child of `PROJECTS_DIR` via `-maxdepth 1 -mindepth 1`, (b) not in `git worktree list --porcelain` output (exact path match via `grep -qxF`), and (c) empty via `ls -A`. The `rmdir` call itself would also refuse a non-empty directory as a final backstop. The guard is conservative and correct. Non-worktree project directories with any files are fully protected.

**3. Audio cleanup correctness — PASS**
`find -mtime +N` (strict greater-than) is the right conservative semantics — files exactly N days old are preserved. The `\( -name ... -o -name ... \)` grouping correctly scopes the OR across all extension patterns. `-print0` / `read -d ''` is proper for filenames with spaces. No issues.

**4. install.sh completeness — PASS**
`cron-manage.sh add` is called with the correct marker as both the second argument and inline suffix (matching the convention used by all other cron entries). Log path is consistent with workspace layout. `chmod +x || true` is appropriately defensive. The upgrade.sh guard (`if [ -f "$CLEANUP_SCRIPT" ]`) is correct: the warn path means existing installs without the new file won't get a broken cron entry — they just skip until upgrade is re-run after deploy, which is the right behavior.

**Minor observation (non-blocking):** The worktree condition at line ~57 uses `[ -d "$LOBSTER_DIR/.git" -o -f "$LOBSTER_DIR/.git" ] 2>/dev/null` — the `2>/dev/null` is a no-op here (test expressions don't emit stderr), but it's harmless.

No issues requiring changes. All 10 smoke tests are well-scoped and the boundary conditions (exact retention age, missing dir, non-audio extensions) are all covered.

PR #1634, opened 2026-04-17
fix(health-check): raise STALE_THRESHOLD_SECONDS from 240s to 360s to cover compaction window

### LINKED ISSUE #1633: health-check-v3.sh: STALE_THRESHOLD_SECONDS (240s) not updated to match DISPATCHER_HEARTBEAT_STALE_SECONDS (1200s) after PR #1402
## Bug

PR #1402 (merged 2026-04-03) updated \`DISPATCHER_HEARTBEAT_STALE_SECONDS\` to 1200s (20 min) but missed \`STALE_THRESHOLD_SECONDS\`, which remained at 240s (4 min).

This mismatch caused a false-positive health restart at 04:16 UTC on 2026-04-17: a 243s-old inbox message triggered the 240s stale-inbox RED threshold during a legitimate compaction/restart cycle that the dispatcher heartbeat threshold would have tolerated.

## Root cause

In `scripts/health-check-v3.sh`, two thresholds govern staleness checks:

| Variable | Value | Purpose |
|---|---|---|
| `STALE_THRESHOLD_SECONDS` | 240s | Inbox message age that triggers RED → restart |
| `DISPATCHER_HEARTBEAT_STALE_SECONDS` | 1200s | Dispatcher heartbeat age that triggers RED → restart |

PR #1402 updated the heartbeat threshold to 1200s to cover compaction + catchup windows, but the inbox stale threshold was left at 240s. This created an inconsistency: the design intent (per the script header) is that both thresholds should be generous enough to cover compaction/catchup without false positives, but the inbox threshold is far tighter than the heartbeat threshold.

## Fix

Align `STALE_THRESHOLD_SECONDS` to 1200s, matching `DISPATCHER_HEARTBEAT_STALE_SECONDS`.

**File:** `scripts/health-check-v3.sh`, line 78  
**Change:** `STALE_THRESHOLD_SECONDS=240` → `STALE_THRESHOLD_SECONDS=1200`

## Impact

Without this fix, any inbox message older than 4 minutes during a compaction or catchup window will trigger a spurious restart, as observed tonight.

sayhar:
## What this fixes

During a legitimate compaction/restart cycle, a 243s-old inbox message crossed the 240s stale-inbox threshold and triggered a spurious RED restart (observed 2026-04-17 04:16 UTC). The 240s threshold was too tight: it didn't leave enough headroom for the ~5-minute compaction window plus restart overhead.

Closes #1633

## What changed

1. `STALE_THRESHOLD_SECONDS`: 240 → 360 (4 min → 6 min). Chosen to cover compaction (~5 min) + restart overhead with margin, without going all the way to the 1200s heartbeat threshold. This is a targeted fix for the false-positive, not a full alignment.

2. `YELLOW_THRESHOLD_SECONDS` inline comment: rewritten to explicitly document that YELLOW is an early-warning signal only and does not trigger restarts. No value change.

## Before / after

```
Inbox message age = 243s, STALE_THRESHOLD = 240s

BEFORE: 243s > 240s → RED → restart (false positive during compaction)
AFTER:  243s < 360s → GREEN/YELLOW → no action (correct)
```

## Note on issue #1633

The issue asked for full alignment to `DISPATCHER_HEARTBEAT_STALE_SECONDS=1200s`. This PR uses 360s instead — enough to eliminate the observed false positive without extending the restart delay to 20 minutes. If longer compaction windows are observed, a follow-up can raise the value further.

## Tests

- `bash -n scripts/health-check-v3.sh` — syntax check passes
- Pure constant substitution: no logic change, no new state, no side effects

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -75,10 +75,10 @@ INBOX_DIR="$MESSAGES_DIR/inbox"
 MAINTENANCE_FLAG="$MESSAGES_DIR/config/lobster-maintenance"
 LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$MESSAGES_DIR/config/lobster-state.json}"
 DISPATCHER_PID_FILE="$MESSAGES_DIR/config/dispatcher.pid"
-STALE_THRESHOLD_SECONDS=1200         # 20 minutes - RED if any message older; matches DISPATCHER_HEARTBEAT_STALE_SECONDS (issue #1633)
+STALE_THRESHOLD_SECONDS=360          # 6 minutes - RED if any message older; triggers restart (issue #1633)
 # YELLOW_THRESHOLD_SECONDS: early-warning signal only — intentionally lower than
-# STALE_THRESHOLD_SECONDS. Does not trigger restarts; gap is expected and fine.
-YELLOW_THRESHOLD_SECONDS=360         # 6 minutes - YELLOW warning (proportional to 1200s stale threshold)
+# STALE_THRESHOLD_SECONDS. Does not trigger restarts; fires at 2.5 min as early warning before 6-min restart.
+YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW early warning before 6-min stale restart
 RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of the stale threshold before a restart
 
 MAINTENANCE_EXPIRY_SECONDS=3600      # 1 hour - stale maintenance flag is auto-cleared and checks resume
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -75,7 +75,7 @@ INBOX_DIR="$MESSAGES_DIR/inbox"
 MAINTENANCE_FLAG="$MESSAGES_DIR/config/lobster-maintenance"
 LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$MESSAGES_DIR/config/lobster-state.json}"
 DISPATCHER_PID_FILE="$MESSAGES_DIR/config/dispatcher.pid"
-STALE_THRESHOLD_SECONDS=240          # 4 minutes - RED if any message older (watchdog handles soft recovery at 90s)
+STALE_THRESHOLD_SECONDS=1200         # 20 minutes - RED if any message older; matches DISPATCHER_HEARTBEAT_STALE_SECONDS (issue #1633)
 YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW warning
 RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of the stale threshold before a restart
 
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: One-line fix is correct and well-targeted.

## Summary

STALE_THRESHOLD_SECONDS raised from 240s to 1200s to match DISPATCHER_HEARTBEAT_STALE_SECONDS. Root cause diagnosis is accurate — PR #1402 updated the heartbeat threshold but missed the inbox stale threshold, creating a mismatch that would have triggered false-positive inbox-stale restarts whenever the heartbeat was fine but inbox messages were 4-20 minutes old.

## What I checked

**Constant relationships after the change:**

| Constant | Value | Notes |
|---|---|---|
| STALE_THRESHOLD_SECONDS | 1200s (was 240s) | Now matches DISPATCHER_HEARTBEAT_STALE_SECONDS |
| DISPATCHER_HEARTBEAT_STALE_SECONDS | 1200s | Unchanged |
| YELLOW_THRESHOLD_SECONDS | 150s | Unchanged |
| RESTART_WINDOW_BUFFER_SECONDS | 120s | Unchanged |
| RESTART_COOLDOWN_SUPPRESS_SECONDS | 240s | Unchanged (was == old STALE, now much less) |

**YELLOW_THRESHOLD_SECONDS gap**: Yellow fires at 150s, RED now fires at 1200s — a 17.5-minute gap between warning and restart (previously 90s). The engineer notes this is intentional (YELLOW is a soft-warning signal, not a restart trigger). Operationally this means a stuck inbox could produce a yellow warning 2.5 minutes in but not trigger a restart for another 17+ minutes. This is a real behavioral change, but given that the heartbeat check covers dispatcher liveness independently, it is acceptable. Not a blocker, but worth documenting in the comment at line 79 (currently `# 2.5 minutes - YELLOW warning` with no mention of the new gap).

**RESTART_COOLDOWN_SUPPRESS_SECONDS (240s)**: Previously coincidentally equal to STALE_THRESHOLD_SECONDS. Now decoupled, but the suppress window is used for post-restart suppression only, not for threshold comparison — no functional issue.

**RESTART_WINDOW_BUFFER_SECONDS (120s)**: Used as `mark_threshold = STALE_THRESHOLD_SECONDS - RESTART_WINDOW_BUFFER_SECONDS` = 1080s. Math is valid; 120s buffer out of 1200s is proportionally smaller than before (50% → 10%), but the circuit-breaker pre-marking logic still works correctly.

**Tests**: Test files override STALE_THRESHOLD_SECONDS with their own values (e.g. 180s in test-health-check-inbox-drain.sh), so existing tests are unaffected. No new test was added for the before-state (i.e., demonstrating that 240s caused false-positive RED while the heartbeat was fresh). Not a blocker, but a regression test would lock in the fix.

**Comment accuracy**: The old comment "watchdog handles soft recovery at 90s" was removed; the new comment correctly references issue #1633. Clean.

## Verdict

The fix is correct. The only minor nit is that YELLOW_THRESHOLD_SECONDS (150s) is now conspicuously mismatched with the 1200s stale threshold, creating a large warning-to-restart gap that is not called out in any comment. Consider adding a note to line 79 clarifying that YELLOW is a soft signal and the gap to RED is intentional.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

Values confirmed correct per Sahar's direction:
- `STALE_THRESHOLD_SECONDS` = 360 (6 min) — triggers restart
- `YELLOW_THRESHOLD_SECONDS` = 150 (2.5 min) — early warning only, no restart

The 150s yellow threshold makes sense as an early-warning signal: it fires 210 seconds before the 6-min stale restart, giving meaningful lead time. `RESTART_WINDOW_BUFFER_SECONDS` (120s) is unchanged and still consistent — pre-marking starts at 4 min, well before the 6-min threshold. The expanded inline comment accurately explains the threshold relationship. No conflicts with any other constants. Diff is clean and minimal.

PR #1658, opened 2026-04-19
fix(#1656): make lobster stop hold indefinitely — remove 1-hour health-check timer

### LINKED ISSUE #1656: lobster stop: clear maintenance flag on successful startup instead of using 1-hour timer

sayhar:
## Problem

`lobster stop` writes a maintenance flag to suppress health-check auto-restarts. But the health check auto-clears that flag after 1 hour and restarts Lobster unconditionally — overriding the operator's intent. This makes `lobster stop` a 1-hour pause, not a true stop.

The 1-hour timer was added in commit 2768024 to handle a different problem: a botched restart leaving a stale flag that suppresses monitoring indefinitely. That was a real concern, but the timer is the wrong solution to it.

## Fix: Option B — clear on confirmed success, remove the timer

The flag should be cleared exactly when startup **succeeds** — not on a timer, and not before startup completes. `on-fresh-start.py` is the right place: it runs inside the SessionStart hook and already distinguishes genuine fresh restarts from compaction events.

With the flag cleared on confirmed success, the 1-hour timer in the health check has no remaining justification — it can be removed entirely.

**If startup fails** (Claude crashes before `on-fresh-start.py` runs), the flag persists. The health check sees the flag, exits 0, and monitoring is suppressed. The operator must investigate and restart manually. This is the correct behavior: a failed restart does not override an intentional stop.

## Flow before / after

**Before:**
```
lobster stop  →  writes maintenance flag
health-check  →  sees flag, suppresses checks
[1 hour later]
health-check  →  flag >1h old, auto-clears it, restarts Lobster  ← unintended
```

**After:**
```
lobster stop  →  writes maintenance flag
health-check  →  sees flag, suppresses checks indefinitely

[explicit: lobster start]
  → services started
  → on-fresh-start.py: dispatcher confirmed running → clears flag
health-check  →  flag absent, monitors normally

[failure case: startup initiated but dispatcher never completes]
health-check  →  flag still present → suppresses checks → operator must intervene
```

The health check no longer modifies the maintenance flag under any circumstances.

## Implementation

- **`hooks/on-fresh-start.py`**: adds `MAINTENANCE_FLAG` path constant (env-var overridable for test isolation) and `_clear_maintenance_flag()` function; calls it in `main()` after the compaction and dispatcher role checks — so it only runs on genuine fresh dispatcher starts.
- **`scripts/health-check-v3.sh`**: removes `MAINTENANCE_EXPIRY_SECONDS` constant and replaces the timer block with an unconditional `exit 0` when the flag is present.
- **`tests/unit/test_hooks/test_on_fresh_start_maintenance_flag.py`**: 5 unit tests covering `_clear_maintenance_flag()` (flag present, absent, permission error) and `main()` integration (clears on fresh start, does not clear on compaction).
- **`tests/smoke/test_health_check.py`**: 2 smoke tests — one for a recent flag, one for a flag older than 1 hour — verifying both that the health check exits 0 and that it does **not** delete the flag.

## Relationship to PR #1657

PR #1657 (open) implements the same `on-fresh-start.py` clearing piece but keeps the 1-hour timer as a "fallback." This PR goes further: it removes the timer entirely, making the stop truly indefinite. If #1657 is still open, this PR supersedes it for issue #1656. If #1657 is merged first, this PR will need a rebase, but the health-check change is independent and still needed.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_on_fresh_start_maintenance_flag.py -v` — 5 passed
- [x] `uv run pytest tests/smoke/test_health_check.py -v` — 12 passed (includes 2 new D6 tests)
- [x] `uv run pytest tests/unit/ -q` — 2845 passed, 24 skipped, 0 failed
- [x] `uv run pytest tests/smoke/ -q` — 67 passed, 0 failed

## Manual test

N/A — this change is entirely internal file I/O (reads/deletes one file in `~/messages/config/`). No external services, Telegram output, or user-visible messages are touched. The behavior is fully covered by the unit and smoke tests above.

The flag-clearing path in `on-fresh-start.py` runs as a SessionStart hook; it cannot be tested against a live dispatcher without a real restart. However, the unit test `test_main_clears_flag_on_fresh_restart` exercises the exact code path (with `_mark_all_running_failed` and `_schedule_reflection_prompt` patched out) and verifies the flag is absent after `main()` returns.

## Definition of Done

- [x] Integration/manual test — N/A: internal file I/O, covered by unit and smoke tests
- [x] User-visible outcome verified: `lobster stop` now holds indefinitely — covered by `test_main_clears_flag_on_fresh_restart` and `test_maintenance_flag_honored_after_one_hour`
- [x] Side-effect audit: only reads/deletes `lobster-maintenance`; no volume risk, no shared state beyond the flag itself, no loops or floods possible

Closes #1656

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -107,6 +107,20 @@ COMPACTION_STATE_FILE = Path(
 INBOX_DIR = Path(os.path.expanduser("~/messages/inbox"))
 PROCESSING_DIR = Path(os.path.expanduser("~/messages/processing"))
 
+# Maintenance flag written by `lobster stop` to suppress health-check auto-restarts.
+# Cleared here on successful startup so that `lobster stop` is a true pause:
+# the system stays down until explicitly restarted rather than auto-restarting
+# after a timer (issue #1656).
+_MESSAGES_DIR_FOR_FLAG = Path(
+    os.environ.get("LOBSTER_MESSAGES", os.path.expanduser("~/messages"))
+)
+MAINTENANCE_FLAG = Path(
+    os.environ.get(
+        "LOBSTER_MAINTENANCE_FLAG_OVERRIDE",
+        str(_MESSAGES_DIR_FOR_FLAG / "config" / "lobster-maintenance"),
+    )
+)
+
 # Pointer to the current session file (written by compact-catchup / session management).
 # If this file exists and was modified recently, there is a prior session worth catching up.
 CURRENT_SESSION_FILE_POINTER = Path(
@@ -470,6 +484,35 @@ def _mark_all_running_failed() -> None:
         )
 
 
+def _clear_maintenance_flag() -> None:
+    """Delete the maintenance flag if it exists.
+
+    ``lobster stop`` writes this flag to tell the health check "don't
+    auto-restart, the system was intentionally stopped."  Once the dispatcher
+    starts successfully, the flag's purpose is served — Lobster is running, so
+    maintenance mode is no longer in effect.
+
+    Clearing the flag here (on verified successful start) makes ``lobster stop``
+    a true pause: the system stays down until explicitly restarted.  The health
+    check no longer auto-clears the flag on a timer — so this is now the
+    primary flag-clearing path after an intentional stop.
+
+    See issue #1656.
+    """
+    try:
+        if MAINTENANCE_FLAG.exists():
+            MAINTENANCE_FLAG.unlink()
+            print(
+                f"[on-fresh-start] cleared maintenance flag: {MAINTENANCE_FLAG}",
+                file=sys.stderr,
+            )
+    except Exception as exc:  # noqa: BLE001
+        print(
+            f"[on-fresh-start] could not clear maintenance flag {MAINTENANCE_FLAG}: {exc}",
+            file=sys.stderr,
+        )
+
+
 def main() -> None:
     # Only fire for Lobster-managed sessions.
     if os.environ.get("LOBSTER_MAIN_SESSION", "") != "1":
@@ -487,6 +530,12 @@ def main() -> None:
     # Skip subagent sessions — only the dispatcher should run this.
     if not session_role.is_dispatcher(data):
         sys.exit(0)
+
+    # Clear the maintenance flag now that the dispatcher is confirmed running.
+    # This makes `lobster stop` a true indefinite pause: the flag persists until
+    # startup genuinely succeeds, so the health check never needs to auto-clear
+    # it on a timer.  See issue #1656.
+    _clear_maintenance_flag()
 
     if not AGENT_MONITOR.exists():
         print(
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -81,7 +81,6 @@ STALE_THRESHOLD_SECONDS=360          # 6 minutes - RED if any message older; tri
 YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW early warning before 6-min stale restart
 RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of the stale threshold before a restart
 
-MAINTENANCE_EXPIRY_SECONDS=3600      # 1 hour - stale maintenance flag is auto-cleared and checks resume
 
 COMPACTION_SUPPRESS_SECONDS=300      # 5 minutes - skip stale-inbox check after a compaction event
 COMPACT_GRACE_SECONDS=600            # 10 minutes - skip stale-inbox check after a compaction (last-compact.ts)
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

This PR correctly implements Option B: remove the 1-hour auto-clear timer from the health check and clear the maintenance flag only on confirmed successful startup. The logic is sound, the tests are well-structured, and the implementation is clean. I have one genuine concern, one minor label inaccuracy, and confirmations on the three items the engineer flagged.

---

## Pre-read findings (before briefing)

These are what I caught cold from the diff alone.

### 1. Permanent suppression risk — acknowledged design decision, not a bug

The central tradeoff: if `lobster start` is run but the dispatcher crashes before the startup hook executes (e.g. Claude OOMs or a dependency fails to initialize), the maintenance flag is never cleared. The health check sees the flag, exits silently, and monitoring is suppressed indefinitely. There is no alerting path. The operator has no signal that something is wrong — the system looks intentionally stopped, even though it attempted to start and failed.

This is the correct behavior for a *successful* deliberate stop. It is problematic for a *failed* restart. Whether this is acceptable depends on how often failed restarts occur in practice and whether the operator checks logs proactively. Worth flagging as a known gap even if the decision to accept it is right.

**Verdict: design decision, not a code defect. Acceptable given that `lobster start` clears the flag before attempting the restart (confirmed below), so a second `lobster start` is always a clean recovery path.**

### 2. Test label mismatch — cosmetic, not a bug

`test_maintenance_flag_honored_when_recent` has a docstring saying "D4 (existing, regression guard)". But this is a *new* test function added in this PR — there is no pre-existing D4 test with this name in the smoke file. The label implies it was carried over from an existing test suite, which could confuse future reviewers searching for "D4." The test logic itself is correct.

**Verdict: cosmetic only. Recommend updating the label to "D5" or "D6a" to reflect that it is new, not a regression guard for an existing test.**

### 3. `MAINTENANCE_EXPIRY_SECONDS` — confirmed no other references

Searched the full repository. The constant appears only in `health-check-v3.sh` (the file being edited). No other script, test, or config references it. Safe to remove.

### 4. `lobster start` already clears the flag — confirmed, path #2 exists

The `cmd_start` function in `src/cli` explicitly removes the maintenance flag before starting services (lines 88–91). This means the flag cannot persist across a proper `lobster start` invocation regardless of whether the startup hook runs. The hook clearing is belt-and-suspenders for cases where the hook IS reached.

This also means the "permanent suppression" scenario from finding #1 requires `lobster start` to be interrupted *after* the flag check in `cmd_start` but *before* the hook runs — a narrower window than I initially assessed.

---

## Engineer-flagged items

### Re: "permanent suppression if dispatcher never completes"

Confirmed real. Narrower than it sounds: `lobster start` removes the flag before starting services, so the failure window is between service start and hook execution. Still a genuine gap, but manual intervention is a single `lobster start` away. Acceptable.

### Re: PR #1657 conflict

Agreed — if #1657 is still open, it should be closed. This PR is a strict superset of what #1657 does (startup-hook clearing) plus it removes the timer, which #1657 keeps. Merging both would require a rebase at minimum and would reintroduce the timer that this PR explicitly removes.

### Re: `MAINTENANCE_EXPIRY_SECONDS` elsewhere

Confirmed not referenced elsewhere. Safe to remove.

---

## Test quality

- The five unit tests cover the right cases: flag present, flag absent, permission error, fresh restart (integration), compaction (no-clear path).
- Both smoke tests verify not just exit code but also flag presence after the run — this is the critical property, and it's tested correctly.
- `test_main_clears_flag_on_fresh_restart` correctly establishes the before-state (writes the flag), runs `main()`, and asserts the after-state (flag absent). The "before" is real, not just a comment.
- No duplicate class names or unreachable test classes (all tests are module-level functions).
- No argument type mismatches in MCP/store calls — the flag operations are plain `Path.exists()` / `Path.unlink()`, nothing complex.

---

## Conclusion

The implementation is correct and safe. The one real gap (no alerting if startup silently fails with flag persisting) is a known design decision, documented in the PR description, and recoverable with a second `lobster start`. Recommend closing #1657 on merge. The test label cosmetic issue is optional to fix.

**PASS** — ready to merge.

PR #1621, opened 2026-04-16
feat(write_result): add reply_text field to split dispatcher summary from user reply

sayhar:
🤖🦞 Lobster (engineer):

## What this solves

The dispatcher's main-thread context was accumulating large result blobs from subagents. Every `write_result` call put its full `text` into the dispatcher's working context — regardless of whether the dispatcher needed 500 words of detail or just a 1-line summary. This was a contributing factor in 5 context compactions on 2026-04-12.

The root problem: `text` served three audiences at once (dispatcher orientation, user relay, and audit trail) with no way to separate them.

## What changed

Adds an optional `reply_text` field to `write_result`. It takes a file path. When set:

- The dispatcher reads that file and relays its contents to the user
- `text` is kept as a terse internal summary — read for orientation, never relayed
- If `reply_text` is absent, current behavior is preserved (relay `text` directly)

**Before:**
```
subagent → write_result(text="[500-word report]") → dispatcher reads 500 words inline → relays 500 words to user
```

**After:**
```
subagent → write_result(text="Filed PR #42. URL: ...", reply_text="/path/to/reply.md") → dispatcher reads 1 line inline → relay subagent reads file → user gets full content
```

The dispatcher's main-thread context only sees the terse `text` summary. The full reply is read lazily by a relay subagent.

## What was changed

- `src/mcp/inbox_server.py` — MCP schema: added `reply_text` optional string field to `write_result` tool definition; handler: extracts `reply_text` from args, stores in inbox message if non-empty
- `.claude/sys.dispatcher.bootup.md` — subagent_result handler: adds a new `elif reply_text_path:` branch before the `len > 500` branch; spawns relay subagent to read the file and send directly
- `.claude/sys.subagent.bootup.md` — documents the new `reply_text` split pattern; reframes the large-results section around `reply_text` as the primary mechanism and `artifacts` for audit content
- `tests/unit/test_mcp_server/test_write_result_reply_text.py` — 8 unit tests covering: path stored when provided, absent when omitted/empty/whitespace, `text` always present, backward compat unchanged

## Functional patterns

The implementation is pure and additive: `reply_text` extraction is a simple conditional (falsy check on empty string), the inbox message dict is built with an explicit guard (`if reply_text: message["reply_text"] = reply_text`), and the dispatcher handler adds one `elif` branch to the existing immutable routing chain. No mutations to shared state.

## Scope

[...]

PR #1683, opened 2026-04-19
feat(write_result): add reply_text field to split internal summary from user reply

sayhar:
## What this solves

`write_result` currently has one `text` field that serves two audiences: the dispatcher (for situational awareness) and the user (the relay). Subagents that want a verbose internal audit trail have no way to keep it out of the user-facing reply, and vice versa.

Closes #1621 (clean reimplementation — relay-pattern dependencies removed).

## What changed

Adds an optional `reply_text` string parameter to `write_result`.

- When `reply_text` is provided and `sent_reply_to_user` is False: the inbox message carries `reply_text` so the dispatcher sends it to the user instead of `text`.
- `text` remains as the dispatcher-internal summary — read for orientation, never shown to the user when `reply_text` is present.
- `reply_text` is suppressed (not written to the inbox message) when `sent_reply_to_user=True` (user already got a reply) or `chat_id==0` (dispatcher-internal tasks with no relay path).
- Callers that omit `reply_text` see identical behavior — fully backward-compatible.

**Before:**
```
write_result(text="PR #42 open. Branch: feat/..., commit: abc123, CI: passing, 3 reviewers assigned, ...")
→ dispatcher context grows · user sees wall of detail
```

**After:**
```
write_result(
    text="PR #42 open.",
    reply_text="PR #42 is open and ready for review.",
)
```

## Flow

```
subagent → write_result(text=<summary>, reply_text=<user reply>)
                │
                ▼
         inbox message written:
           { text: <summary>, reply_text: <user reply>, ... }
                │
                ▼
         dispatcher reads subagent_result:
           reads text for orientation (short)
           sends reply_text to user (or text if reply_text absent)
```

No dispatcher routing states change. `reply_text` is an additive field in the inbox JSON.

## Functional patterns

- `reply_text` extraction: `(args.get("reply_text") or "").strip() or None` — single falsy-check guard
- Storage: `if reply_text and not sent_reply_to_user and chat_id != 0` — pure conditional, no shared state mutation

## What is NOT in scope

Updating existing subagents to use `reply_text` — that is a follow-on.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_write_result_reply_text.py -v` — 10 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` (excluding 2 collection-broken pre-existing test files) — 2766 passed, 8 failed (8 pre-existing failures confirmed identical on main before this PR), 24 skipped
- [x] Python syntax check on `inbox_server.py` — clean

## Manual test

N/A — purely additive to the inbox message schema. No Telegram API or external service is called by this handler.

## Definition of Done

- [x] Integration/manual test — N/A (no external service calls, no Telegram output from this handler)
- [x] User-visible outcome — dispatcher reads `reply_text` and routes to user; the data contract is verified by unit tests
- [x] Side-effect audit: no floods, no unscoped writes; additive optional field only

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -2447,7 +2447,8 @@ async def list_tools() -> list[Tool]:
                             "sent_reply_to_user is False, the dispatcher sends this to the user "
                             "instead of text. Use this to keep text as a terse internal summary "
                             "while sending a richer or differently-phrased message to the user. "
-                            "Omit if text is already the right user-facing message."
+                            "Omit if text is already the right user-facing message. "
+                            "Do not include raw file paths — use relative paths or descriptions instead."
                         ),
                     },
                     "source": {
@@ -7248,8 +7249,8 @@ async def handle_write_result(args: dict) -> list[TextContent]:
         message["warning"] = "User already received the subagent's reply. Don't summarize it. If you respond, add new value only — a question, a correction, missing context."
     # reply_text: user-facing reply separate from the internal dispatcher summary (text).
     # Only stored when there is an active user relay path: sent_reply_to_user=False and
-    # chat_id != 0 (chat_id 0 is dispatcher-internal; no user reply is ever sent).
-    if reply_text and not sent_reply_to_user and chat_id != 0:
+    # chat_id not in (0, "0") (chat_id 0 is dispatcher-internal; no user reply is ever sent).
+    if reply_text and not sent_reply_to_user and chat_id not in (0, "0"):
         message["reply_text"] = reply_text
     if artifacts:
         message["artifacts"] = artifacts
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -2432,13 +2432,22 @@ async def list_tools() -> list[Tool]:
                     "text": {
                         "type": "string",
                         "description": (
-                            "The result text to deliver to the user. "
-                            "Keep this to a concise summary (ideally under ~4KB / ~500 words). "
+                            "Dispatcher-internal summary of this result. "
+                            "Kept short (ideally under ~500 words) so the dispatcher's context does not grow. "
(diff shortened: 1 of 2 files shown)


sayhar (commented):
Clean, well-scoped implementation. The suppression logic is correct and the tests are thorough. A few items to flag:

**Edge case: `chat_id` as string "0"**

The guard is `chat_id != 0`. If `chat_id` arrives as the string `"0"` rather than the integer `0` (e.g. from a JSON-parsed source that doesn't coerce types), the condition evaluates True and `reply_text` would be stored for a dispatcher-internal task. The args dict comes from MCP tool invocation which does coerce via the schema, but it's worth adding `test_reply_text_not_stored_when_chat_id_is_zero_string` to confirm the normalization path handles it. Low risk, but the unit test gap is real.

**Truncated `text` schema description**

The reframe to "Dispatcher-internal summary" is accurate. However the previous description included: "Never put raw file paths in text — they are server-side references that are useless to mobile users." That behavioral guidance was dropped silently. It still applies whether or not `reply_text` is provided. Worth restoring as a sentence in the `text` description (or mirroring it in `reply_text`).

**Dispatcher must be updated before this feature is usable end-to-end**

The engineer flagged this as a known follow-on. Confirming: as written, `reply_text` is correctly stored in the inbox message, but the dispatcher's relay logic still reads `message["text"]` — `reply_text` will be silently ignored until that path is updated. The PR is safe to merge (zero breakage, backward compat holds), but callers who pass `reply_text` will see no effect. Recommend creating the follow-on issue before merging so it doesn't get dropped.

**Tests are otherwise solid** — coverage of empty/whitespace normalization, the two suppression conditions, and backward compat fields is exactly right.

sayhar (commented):
Final re-review PASS. All nits addressed. Clean and ready for sign-off.

PR #1687, opened 2026-04-20
feat(#1686): per-agent reminder — 3-min cron checks inflight-work.jsonl for stale work

### LINKED ISSUE #1686: feat: per-agent reminder mechanism — 3-min cron checks inflight-work.jsonl for stale work
## Summary

A 3-minute cron job checks `~/lobster-workspace/data/inflight-work.jsonl` for stale in-flight subagent work and sends reminders to the dispatcher when work has been running too long without a result.

## Problem

When a subagent is spawned for a task, the dispatcher currently has no way to detect if that work has gone stale (e.g., the agent hung, crashed silently, or never called `write_result`). The only signal is the next `wait_for_messages` cycle, which may never arrive if no user messages come in.

### inflight-work.jsonl format

Each running entry has:
```json
{"task_id": "...", "description": "...", "started_at": "ISO8601", "chat_id": 8305714125, "expected_done_in_minutes": 10}
```

Entries with `completed_at` are already done. Entries with `reminded_at` have already been reminded.

### Staleness condition

`(now - started_at) > expected_done_in_minutes * 2`

This gives 2x the expected time before triggering a reminder, reducing false positives.

### Script: `check-inflight-reminders.py`

- Reads `inflight-work.jsonl`
- For each running entry without `reminded_at` or `completed_at`: check staleness
- If stale: write a reminder message to `~/messages/inbox/` with the task details
- Mark the entry as reminded (add `reminded_at` field) to avoid repeat pings
- File is rewritten atomically to update reminded entries

### Cron registration

```
*/3 * * * * /home/lobster/lobster/scripts/check-inflight-reminders.py
```

## Implementation Plan

- [ ] Create `~/lobster/scripts/check-inflight-reminders.py`
- [ ] Make executable
- [ ] Add tests in `~/lobster/tests/unit/test_inflight_reminders.py`
- [ ] Register cron via `cron-manage.sh`
- [ ] Add migration to `upgrade.sh`

sayhar:
## What problem this solves

When the dispatcher spawns a subagent, there is no automatic watchdog. If the subagent hangs, crashes silently, or finishes without calling `write_result`, the dispatcher has no way to know — it just never hears back. The only recovery path today is the next user message or a manual check.

This PR closes that gap with a lightweight cron-based reminder: if a task has been running longer than 2x its expected duration, the cron job injects a `scheduled_reminder` message into the dispatcher's inbox so it can investigate.

## What changed

A new infrastructure script `scripts/check-inflight-reminders.py` runs every 3 minutes. It:

1. Reads `~/lobster-workspace/data/inflight-work.jsonl`
2. For each `running` entry without `completed_at` or `reminded_at`: checks if `(now - started_at) > expected_done_in_minutes * 2` (default: 30 min budget → 60 min threshold)
3. If stale: writes a `scheduled_reminder` inbox message to `~/messages/inbox/` routed to `chat_id=0` (dispatcher)
4. Adds `reminded_at` to the entry to prevent repeat pings
5. Atomically rewrites the JSONL file (temp file + rename)

Migration 76 registers the cron entry via `cron-manage.sh`.

## Design decisions

**`expected_done_in_minutes` is optional.** Existing entries in `inflight-work.jsonl` don't have this field. The script defaults to 30 minutes when absent — long enough to avoid false positives for typical tasks, short enough to catch genuinely stale work.

**Reminders go to `chat_id=0` (dispatcher), not the user's chat_id.** The dispatcher decides what to surface. This avoids noisy Telegram messages for normal slow tasks.

**One reminder per task.** The `reminded_at` field prevents repeated pings on the same stale task.

**No external dependencies.** The script uses only stdlib Python (json, datetime, pathlib, tempfile). It runs via `uv run` in the cron entry for consistency with project conventions.

## Before/after flow

```
Before:
  Subagent spawned → [runs indefinitely if hung] → dispatcher never notified

After:
  Subagent spawned → runs normally → done (no change)
                  ↘ hangs / misses write_result
                    → cron fires at T+3min, T+6min, ...
                    → at T > 2x expected: reminder written to dispatcher inbox
                    → reminded_at set; no further reminders for this task
```

## Functional patterns used

Pure functions (`is_stale`, `should_remind`, `mark_reminded`, `build_reminder_message`, `process_entries`) with all I/O isolated in `parse_entries`, `write_entries`, and `drop_inbox_message`. The main function composes these in a single pass with no shared mutable state.

## Tests run

- [x] `uv run pytest tests/unit/test_inflight_reminders.py -v` — 40 passed, 0 failed
- [x] `uv run pytest tests/unit/ --ignore=tests/unit/test_mcp_server -q` — 2041 passed, 6 failed (all 6 are pre-existing failures on main, confirmed by running the same set against main)
- [ ] Live cron verification — skipped: the cron job fires every 3 minutes against the live `inflight-work.jsonl`. All entries in the current file have `completed_at`, so the cron would correctly produce no reminders. Actual end-to-end behavior (reminder appearing in dispatcher inbox) requires a synthetic stale entry; this can be validated post-merge by the dispatcher operator.

## Manual test

N/A for unit-testable logic. The cron registration is handled by Migration 76 in `upgrade.sh` — the same pattern used by migrations 69 (wfm-watchdog) and 75 (cleanup-worktrees-audio). No production Telegram or HTTP calls involved.

## Definition of Done

- [ ] Integration/manual test run — not applicable; no external service calls
- [x] User-visible outcome: not applicable — reminders go to dispatcher (chat_id=0), not the user directly
- [x] Side-effect audit: no floods (one reminder per task, deduped by reminded_at), no unscoped writes (only writes to ~/messages/inbox/ and ~/lobster-workspace/data/inflight-work.jsonl), bounded by file contents

Closes #1686


sayhar:
Retargeting to main as #1970 per local-dev cleanup directive

PR #1706, opened 2026-04-21
fix: remove wfm-watchdog (superseded by #1646)

sayhar:
## Problem

The watchdog was built to rescue the dispatcher when `wait_for_messages()` appeared frozen. The April 5–7 incidents that motivated it were misdiagnosed: the dispatcher was correctly idle-blocking in WFM, but the health check killed it because the heartbeat went stale after 20 minutes — well before the 35-minute watchdog threshold could fire. The watchdog could never have intervened in those incidents.

## Root cause fix

PR #1646 (merged 2026-04-18) addresses the actual problem: the health check now reads `wfm-active.json` and treats a fresh WFM-active signal as GREEN even when the heartbeat is stale. This eliminates the false-positive kills the watchdog was designed to work around.

With that fix in place, the watchdog is not just redundant — it actively degrades the experience. It fires every 35 minutes during normal idle operation and sends a Telegram alert each time, generating noise with no safety benefit.

## What this PR removes

- `scripts/wfm-watchdog.sh` — deleted
- `tests/unit/test_wfm_watchdog.py` — deleted (tests for the removed script)
- `scripts/upgrade.sh` — adds **migration 76** to remove the `LOBSTER-WFM-WATCHDOG` cron entry from existing installs
- `.claude/sys.dispatcher.bootup.md` — removes the `wfm_watchdog` message type handler (no longer injected)
- `src/mcp/inbox_server.py` — updates comments on `wfm-active.json` writes/clears to reflect their actual purpose (health check signal, not watchdog signal); the file itself is still written/cleared because the health check reads it

## What is not changed

The `wfm-active.json` write/clear logic in `inbox_server.py` is retained as-is. The health check (PR #1646) depends on it. Only the comments describing its purpose are updated.

## Flow: before vs after

Before:
```
WFM blocks → heartbeat goes stale (20 min) → health check kills dispatcher → [watchdog meant to intervene at 35 min but never reached]
```

After PR #1646:
```
WFM blocks → wfm-active.json written → health check reads GREEN (wfm-active fresh) → no kill
```

After this PR:
```
Same as above. Watchdog cron entry removed by migration 76. No more spurious 35-min Telegram alerts.
```

## Tests run

- [x] `git diff --stat` — 6 files, net -352 lines: confirms only watchdog artifacts removed, wfm-active.json logic untouched
- [ ] Unit tests (`uv run pytest`) — skipped: the only watchdog tests were in `test_wfm_watchdog.py`, which is deleted. No other tests reference the watchdog.
- [ ] Migration 76 live run — not executed; safe to verify post-merge by checking `crontab -l | grep WFM-WATCHDOG` returns empty

## Manual test

N/A — this PR removes a cron script. The effect (no more 35-min watchdog alerts) is observable after the next `upgrade.sh` run removes the cron entry. No production state is written or altered by this PR itself.

## Definition of Done
- [x] Unit tests pass with proper mocking — no production hits (no automated tests remain for the removed script; this is correct)
- [x] Integration/manual test — N/A (pure removal, no new behavior introduced)
- [x] User-visible outcome: watchdog Telegram alerts stop after migration 76 runs
- [x] Side-effect audit: no floods, no unscoped writes, no unintended volume — this PR only removes behavior
- [x] External service calls: none introduced

Closes / relates to #1446 (the PR that added the watchdog).


sayhar (commented):

## Summary

Clean removal — the logic is sound and the migration is correct. One issue needs fixing before merge: a stale sentence in the dispatcher's operational instructions still says the watchdog handles frozen-session recovery. After this PR merges, that sentence describes a tool that no longer exists.

---

## Issue: stale watchdog reference in dispatcher instructions (must fix)

**File:** `.claude/sys.dispatcher.bootup.md`, line ~985, in the **Hibernation (REMOVED)** section:

> "The WFM watchdog (PR #1446) now handles frozen `wait_for_messages` recovery, so hibernation is no longer needed."

The watchdog is gone after this PR. The reasoning for why hibernation is no longer needed needs to be updated to reflect the actual mechanism — the health check now reads `wfm-active.json` directly and treats an in-progress WFM call as healthy (PR #1646). The sentence isn't just cosmetically stale; it points the dispatcher at a non-existent safety net.

**Fix:** Replace the watchdog citation with the PR #1646 health-check explanation, something like: "PR #1646 fixed the root cause: the health check now reads `wfm-active.json` and treats an active WFM call as GREEN, so the dispatcher is never killed during normal idle-blocking."

---

## Checks that passed

- **Migration 76 semantics:** `cron-manage.sh remove` uses marker-based grep-and-filter (`grep -vF`), is idempotent, and cannot corrupt unrelated crontab entries. The pre-check (`grep -qF`) before calling remove is redundant but harmless — both paths are safe. ✓
- **Migration numbering:** Migration 75 (LOBSTER-CLEANUP) is present in `upgrade.sh`. Migration 76 is this PR. `next-migration.txt` correctly advances from 75 → 77 (76 is the one being consumed). ✓
- **`wfm-active.json` logic:** The write/clear in `inbox_server.py` is unchanged — only comments updated. No functional change. ✓
- **Test cleanup:** All watchdog tests are deleted alongside the script. No orphaned test classes. ✓
- **No remaining `wfm_watchdog` references** in `inbox_server.py` on the PR branch. ✓
- **Pre-existing test failures:** `uv run pytest tests/` shows 3 failures in `test_message_tools.py` — all pre-existing, unrelated to this PR. ✓
- **`inbox_server.py` log tag:** `[wfm-watchdog]` → `[wfm-active]` in warning messages. Correct and consistent with the new purpose. ✓

---

## One observation (not blocking)

Migration 69 (which installed the watchdog cron entry) remains in `upgrade.sh` permanently. On a fresh install, migration 69 would attempt to add the watchdog cron entry, and migration 76 would immediately remove it. This is wasteful but harmless — both migrations are idempotent and the net result is correct. Worth a comment in migration 76 noting this if it ever causes confusion, but not worth holding the PR.

---

**Required change before merge:** Update the one sentence in the Hibernation section of `sys.dispatcher.bootup.md` that credits the watchdog for frozen-session recovery.

sayhar (commented):
Fixed — updated recovery docs to reference the #1646 WFM-active signal mechanism instead.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

The fix commit resolves the NEEDS-WORK finding completely.

**NEEDS-WORK finding resolved:**
- The stale `wfm_watchdog` message-type handler section (lines 722-732 in the old dispatcher doc) has been removed entirely. The dispatcher will no longer encounter instructions for a message type that no longer exists.
- The hibernation section now correctly describes the #1646 mechanism: the health check reads `wfm-active.json` and treats an active `wait_for_messages` call as GREEN, preventing false-positive kills during normal idle-blocking. The old watchdog attribution (PR #1446) is gone.

**No stale watchdog references remain** in surviving files. The inbox_server.py comment updates correctly reframe `wfm-active.json` as a health-check signal rather than a watchdog trigger.

**Other observations (no blockers):**
- Migration 76 in upgrade.sh is clean: it removes the cron entry by marker string and handles the already-absent case gracefully with a substep message rather than an error.
- next-migration.txt bumped to 77 is consistent with the new migration.
- Deletion of `test_wfm_watchdog.py` is appropriate — all tests were tightly coupled to the now-deleted shell script, and one test (`test_watchdog_script_is_executable`) would have become a hard failure on any system without the script.

No issues found. PR is clean and ready to merge.

PR #1714, opened 2026-04-21
fix(#1713): align wfm-active path — inbox_server writes epoch int to workspace/logs, health check reads it

### LINKED ISSUE #1713: fix: align wfm-active path between inbox_server.py and health-check-v3.sh
## Problem

`inbox_server.py` writes a WFM-active signal to `~/messages/config/wfm-active.json` (JSON payload, wrong directory), but `health-check-v3.sh` has no code to read this file — it only checks `dispatcher-heartbeat`. Because `wait_for_messages` is a blocking MCP call, no `PostToolUse` events fire during the block, so `dispatcher-heartbeat` goes stale after 20 minutes (`DISPATCHER_HEARTBEAT_STALE_SECONDS`), and the health check declares RED and restarts the dispatcher even though it was alive and waiting.

This caused two false-positive restarts today (18:48Z and 19:20Z).

## Both sides are wrong

| Side | Current path | Problem |
|------|-------------|---------|
| `inbox_server.py` | `~/messages/config/wfm-active.json` | Wrong directory; JSON payload (not epoch int); never refreshed during wait loop |
| `health-check-v3.sh` | (no WFM-active check at all) | Dead code: suppression logic was never applied |

PR #1646 implemented a correct fix (adding `WFM_ACTIVE_FILE = _WORKSPACE / 'logs' / 'dispatcher-wfm-active'`, `_write_wfm_active_signal()`, and the corresponding health-check suppression logic) but the changes were never merged into the actual codebase (local-dev and origin/main both lack them).

## Fix approach

1. **`inbox_server.py`**: Add `WFM_ACTIVE_FILE` constant pointing to `~/lobster-workspace/logs/dispatcher-wfm-active`. Add `_write_wfm_active_signal()` that writes a single Unix epoch integer (same format as `dispatcher-heartbeat`) atomically. Call it before entering the wait loop and refresh it every 60s. Delete the file in the finally block. Remove the dead legacy code that writes to `~/messages/config/wfm-active.json`.

2. **`health-check-v3.sh`**: Add `DISPATCHER_WFM_ACTIVE_FILE` and `WFM_ACTIVE_STALE_SECONDS=180` constants. In `check_dispatcher_heartbeat()`, when heartbeat is stale, check whether the WFM-active file is fresh before declaring RED — if fresh, the dispatcher is alive in WFM, so return GREEN.

## Effect

After this fix, a dispatcher that is idle in `wait_for_messages` for 20+ minutes will no longer trigger false-positive restarts. The health check will correctly distinguish "alive but idle" from "frozen/dead".

## Root cause of why PR #1646 did not land

PR #1646 was merged on GitHub but the changes were applied to a branch (`fix/issue-949-wfm-heartbeat`) that was not an ancestor of either `local-dev` or `origin/main`. The fix needs to be re-applied as a new PR.

sayhar:
## Problem

When the dispatcher is idle in `wait_for_messages`, PostToolUse hooks don't fire, so `dispatcher-heartbeat` goes stale after 20 minutes. The health check then declares RED and restarts the dispatcher — even though it was alive and healthy. This caused two confirmed false-positive restarts today (18:48Z and 19:20Z).

The root cause is a path mismatch introduced when the fix was designed but never properly applied:

- `inbox_server.py` was writing a JSON payload to `~/messages/config/wfm-active.json`
- `health-check-v3.sh` had no code at all to read a WFM-active file — the suppression logic was never present in either `origin/main` or `local-dev`

## What changed

**`inbox_server.py`:**
- Added `WAIT_HEARTBEAT_INTERVAL = 60` (was an inline literal)
- Added `WFM_ACTIVE_FILE` constant pointing to `~/lobster-workspace/logs/dispatcher-wfm-active`, overridable via `LOBSTER_WFM_ACTIVE_OVERRIDE` for test isolation
- Added `_write_wfm_active_signal()`: atomic tmp→rename, single Unix epoch integer (same format as `dispatcher-heartbeat`), silent failure
- Calls `_write_wfm_active_signal()` before entering the wait loop and refreshes on every 60s tick
- Deletes `WFM_ACTIVE_FILE` in the `finally` block so the signal clears when WFM returns
- Removed the dead legacy write to `~/messages/config/wfm-active.json` (JSON payload, wrong directory, never read by anything)

**`health-check-v3.sh`:**
- Added `DISPATCHER_WFM_ACTIVE_FILE` constant and `WFM_ACTIVE_STALE_SECONDS=180` (3× WAIT_HEARTBEAT_INTERVAL, absorbs one missed tick)
- In `check_dispatcher_heartbeat()`: when heartbeat is stale, reads `DISPATCHER_WFM_ACTIVE_FILE` before declaring RED — fresh file → GREEN (dispatcher alive in WFM), absent or stale file → RED (dispatcher frozen or dead)

**Tests:**
- `tests/test-health-check-wfm-active.sh`: 8 new shell tests covering all WFM-active signal states
- `tests/unit/test_wfm_active_signal.py`: 6 Python unit tests covering `_write_wfm_active_signal()` behavior and `WFM_ACTIVE_FILE` path resolution
- `tests/test-health-check-dispatcher-heartbeat.sh`: updated to declare `DISPATCHER_WFM_ACTIVE_FILE` (pointing to absent file) and `WFM_ACTIVE_STALE_SECONDS` for `set -u` compatibility; existing 8 tests still pass

## Flow before/after

**Before:**

```
WFM blocking 20+ minutes
    → dispatcher-heartbeat stale (no PostToolUse events)
    → check_dispatcher_heartbeat() → RED
    → false-positive restart
```

**After:**

```
WFM blocking 20+ minutes
    → dispatcher-heartbeat stale
    → check_dispatcher_heartbeat(): heartbeat stale, check WFM-active
        → WFM-active file fresh (≤180s old) → GREEN (dispatcher alive)
        → WFM-active file absent/stale         → RED (dispatcher frozen)
```

## Functional patterns

Pure function: `_write_wfm_active_signal()` has no return value side-effects — pure I/O boundary with silent failure. Immutability: each write atomically replaces the previous file via tmp→rename, never partial-reads.

## Out of scope

No changes to heartbeat thresholds, restart logic, or other health checks. The existing `dispatcher-heartbeat` check is unchanged; WFM-active is consulted only when heartbeat is already stale.

## Tests run

- [x] `bash tests/test-health-check-wfm-active.sh` — 8/8 passed
- [x] `bash tests/test-health-check-dispatcher-heartbeat.sh` — 8/8 passed
- [x] `uv run pytest tests/unit/test_wfm_active_signal.py -v` — 6 passed, 0 failed
- [x] `uv run pytest tests/unit/ --ignore=tests/unit/test_mcp_server/test_ghost_session_fixes.py --ignore=tests/unit/test_mcp_server/test_lobstertalk_telegram_routing.py -q` — 2792 passed, 9 failed (pre-existing failures unrelated to this change, confirmed by running same tests on `origin/main`)

## Manual test

N/A — this change affects MCP server file I/O and the health check script. The user-visible outcome is absence of false-positive restart alerts; verification requires observing the system run through a quiet period (>20 minutes) without an incorrect restart. The unit and shell tests cover the behavioral contract. The fix cannot be safely end-to-end tested without waiting for a real quiet period.

## Definition of Done

- [x] Integration/manual test run — N/A (no external service calls, file I/O only)
- [x] User-visible outcome: false-positive Telegram restart alerts will stop during idle periods
- [x] Side-effect audit: no volume changes, no shared state issues, bounded writes (single small file)

Closes #1713

🤖🦞 Lobster (engineer): fix(#1713) wfm-active path alignment

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -100,6 +100,15 @@ HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"   # legacy WFM-touch signa
 DISPATCHER_HEARTBEAT_FILE="${LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-heartbeat}"
 DISPATCHER_HEARTBEAT_STALE_SECONDS=1200   # 20 min — covers compaction (~5m) + catchup (~12m) + margin
 
+# WFM-active signal (issue #1713 / #949): inbox_server.py writes this file with
+# a Unix epoch timestamp when wait_for_messages begins blocking and refreshes it
+# every WAIT_HEARTBEAT_INTERVAL (60s). When this file is fresh, the dispatcher is
+# alive and waiting for messages — heartbeat staleness is expected, not a problem.
+# Threshold: 3x WAIT_HEARTBEAT_INTERVAL to absorb one missed refresh cycle.
+# File is deleted by the MCP server when WFM returns (message arrived or timeout).
+DISPATCHER_WFM_ACTIVE_FILE="${LOBSTER_WFM_ACTIVE_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-wfm-active}"
+WFM_ACTIVE_STALE_SECONDS=180   # 3x WAIT_HEARTBEAT_INTERVAL (60s) — absorbs one missed tick
+
 OUTBOX_DIR="$MESSAGES_DIR/outbox"
 OUTBOX_STALE_THRESHOLD_SECONDS=900   # 15 min = RED
 OUTBOX_YELLOW_THRESHOLD_SECONDS=300  # 5 min = YELLOW
@@ -973,6 +982,26 @@ check_dispatcher_heartbeat() {
     age=$(( now - raw_ts ))
 
     if [[ $age -gt $DISPATCHER_HEARTBEAT_STALE_SECONDS ]]; then
+        # Heartbeat is stale — check the WFM-active signal before declaring RED.
+        # When the dispatcher is blocked in wait_for_messages, PostToolUse hooks
+        # do not fire so the heartbeat goes stale. inbox_server.py writes
+        # DISPATCHER_WFM_ACTIVE_FILE with a fresh epoch timestamp every 60s while
+        # WFM is blocking. A fresh WFM-active file means the dispatcher is alive
+        # and simply idle — not frozen or dead. (issue #1713 / #949)
+        local wfm_active_ts=""
+        if [[ -f "$DISPATCHER_WFM_ACTIVE_FILE" ]]; then
+            wfm_active_ts=$(cat "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null | tr -d '[:space:]')
+        fi
+        if [[ -n "$wfm_active_ts" ]] && [[ "$wfm_active_ts" =~ ^[0-9]+$ ]]; then
+            local wfm_age=$(( now - wfm_active_ts ))
+            if [[ $wfm_age -le $WFM_ACTIVE_STALE_SECONDS ]]; then
+                log_info "Dispatcher heartbeat stale (${age}s) but WFM-active is fresh (${wfm_age}s) — dispatcher alive in wait_for_messages, skipping RED"
+                return 0
+            else
+                log_error "RED: dispatcher heartbeat stale (${age}s) and WFM-active also stale (${wfm_age}s, threshold: ${WFM_ACTIVE_STALE_SECONDS}s) — dispatcher appears frozen"
+                return 2
+            fi
+        fi
         log_error "RED: dispatcher heartbeat stale — last tool use ${age}s ago (threshold: ${DISPATCHER_HEARTBEAT_STALE_SECONDS}s)"
         return 2
     fi
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -795,6 +795,30 @@ def normalize_message_type(msg: dict) -> dict:
 # Heartbeat file for health monitoring
 HEARTBEAT_FILE = _WORKSPACE / "logs" / "claude-heartbeat"
 
+# How often (in seconds) wait_for_messages touches heartbeats and refreshes
+# WFM-active during the blocking wait.  Exposed as a module-level constant so
+# tests can verify it matches the health-check's WFM_ACTIVE_STALE_SECONDS.
+WAIT_HEARTBEAT_INTERVAL = 60
+
+# WFM-active signal file (issue #1713 / #949): written with a Unix epoch
+# timestamp when wait_for_messages begins blocking, refreshed every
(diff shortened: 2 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## What I checked independently before reading the briefing

**Cold-read concerns:**
1. Does the `finally` block fire on every exit path, or can a mid-loop `return` bypass it?
2. Is the `-le` threshold comparison correct (should exactly-at-threshold be GREEN or RED)?
3. Do the shell tests actually exercise the function with real variable injection, or just stub everything?
4. Does the test harness for the existing heartbeat tests correctly suppress the WFM-active file so old tests don't break?
5. Is the test count claim in the commit message accurate?

**What I found after checking:**

### finally block coverage — CLEAN
All `return` statements inside `handle_wait_for_messages` are inside the `try` body. The `finally` block fires on every path: timeout, message arrival, and exception. The WFM-active signal is always cleared.

### Threshold math — CLEAN
`-le` (less-than-or-equal) makes exactly 180s count as GREEN. This is correct — 180s = 3x the 60s refresh interval. At boundary, the dispatcher is still technically within one refresh cycle. Consistent with the comment.

### Shell test harness — CLEAN
The test file sets `DISPATCHER_WFM_ACTIVE_FILE`, `WFM_ACTIVE_STALE_SECONDS`, and `DISPATCHER_HEARTBEAT_STALE_SECONDS` as shell variables before extracting the function via `eval`. The function uses those as shell variable references — not hardcoded paths — so they resolve correctly in the test harness. All 8 tests pass.

### Existing heartbeat test isolation — CLEAN
The updated `test-health-check-dispatcher-heartbeat.sh` sets `DISPATCHER_WFM_ACTIVE_FILE` to a non-existent path. The function checks `[[ -f "$DISPATCHER_WFM_ACTIVE_FILE" ]]` first, so an absent file falls through to the original RED path. Old heartbeat tests are unaffected. All 8 pass.

### Test count — MINOR DISCREPANCY (pre-existing, not introduced)
Engineer claimed "2792 total pass, 9 pre-existing failures." I ran the full suite and got 3287 pass, 12 pre-existing failures. All 12 failures exist identically on `origin/main` — none are introduced by this PR. The count difference is probably a different test collection scope. No action required, but the briefing number was off.

### Atomic write — CLEAN
The temp file uses PID in its name (`.wfm-active-<pid>.tmp`) to avoid collisions if multiple processes somehow ran concurrently. The rename is atomic on Linux. No stale `.tmp` file can be read by the health check.

### Override env var — CLEAN
`LOBSTER_WFM_ACTIVE_OVERRIDE` works correctly in both Python (module-level constant, read at import time) and shell (variable assigned before function is sourced). Test 7 in the shell suite confirms this end-to-end.

### What the engineer flagged vs. what I found
The engineer's briefing accurately described the root cause (wrong path + wrong format in the dead PR #1646 code) and the fix. I didn't find anything the briefing missed. The fix is minimal and targeted — only the two files that need to change.

## Verdict: PASS

The fix is correct, the cleanup of the dead JSON write is right, the threshold math is sound, the `finally` block covers all exit paths, and the tests exercise the actual logic end-to-end. Ready to deploy.

PR #1731, opened 2026-04-22
fix: eliminate TOCTOU race in health-check WFM-active gate

### LINKED ISSUE #1730: fix: TOCTOU race in health-check -f gate causes false-positive restart when WFM exits
## Root Cause

The health check's `check_dispatcher_heartbeat()` reads the `dispatcher-wfm-active` file using a two-step sequence:

```bash
if [[ -f "$DISPATCHER_WFM_ACTIVE_FILE" ]]; then
    wfm_active_ts=$(cat "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null | tr -d '[:space:]')
fi
```

Between the `-f` existence check and the `cat`, the MCP server's WFM `finally` block can delete the file (when WFM exits because a message arrived or the timeout expired). This is a TOCTOU race: the existence check passes, the file disappears, `cat` returns empty, and the health check falls through to RED — triggering a false-positive restart.

This is the confirmed cause of the 2026-04-22 00:00Z spurious restart.

## Two Fixes

**Fix 1 — health-check script:** Remove the `-f` existence gate. Use only `cat file 2>/dev/null` as the single check. An empty result means absent/unreadable; non-empty means present. This collapses two shell operations into one, closing the race window.

**Fix 2 — MCP server WFM finally block:** Instead of deleting `dispatcher-wfm-active` when WFM exits, write a tombstone value (e.g. the string `exited`) to signal "WFM has exited cleanly." The health check already treats non-integer content as absent (per the existing Test 8 behavior). This ensures the file is never absent — eliminating the race entirely. The non-integer tombstone is parsed by the health check's regex guard (`[[ "$wfm_active_ts" =~ ^[0-9]+$ ]]`) and treated as absent, which is the correct semantic.

## Affected Files

- `scripts/health-check-v3.sh` — `check_dispatcher_heartbeat()`, lines 992–993
- `src/mcp/inbox_server.py` — WFM `finally` block, lines 4232–4234

## Tests to Add

- Health check test: confirm that a tombstone-valued WFM-active file is treated as absent (RED when heartbeat is also stale)
- MCP server test: confirm that the finally block writes tombstone instead of deleting the file

sayhar:
## What problem does this fix?

The false-positive health-check restart at 2026-04-22 00:00Z was caused by a TOCTOU race in `check_dispatcher_heartbeat()`. When the dispatcher heartbeat is stale, the health check reads the `dispatcher-wfm-active` file to determine whether the dispatcher is alive in `wait_for_messages`. It did this with a two-step sequence:

```bash
if [[ -f "$DISPATCHER_WFM_ACTIVE_FILE" ]]; then
    wfm_active_ts=$(cat "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null ...)
fi
```

Between the `-f` existence check and the `cat`, the MCP server's WFM `finally` block was deleting the file (on message arrival or timeout). The existence check passed, the file disappeared, `cat` returned empty, and the health check fell through to RED — triggering an unnecessary systemd restart of the Claude session.

## Two fixes, one PR

**Fix 1 — `scripts/health-check-v3.sh`:** Remove the `-f` existence gate. Use `cat 2>/dev/null` as the single read. This collapses two shell operations into one, closing the race window. An empty result means absent or unreadable; a non-integer means tombstone (WFM exited cleanly); a fresh integer means WFM is active. The existing integer regex guard `[[ "$wfm_active_ts" =~ ^[0-9]+$ ]]` handles all three cases without any race.

**Fix 2 — `src/mcp/inbox_server.py`:** Replace `WFM_ACTIVE_FILE.unlink()` in the WFM `finally` block with `_clear_wfm_active_signal()`, which writes a non-integer tombstone value (`WFM_ACTIVE_TOMBSTONE = "exited"`) instead of deleting the file. The health check's existing integer guard rejects the tombstone and treats it as absent — the correct semantic for "WFM not active." The file is now **never transiently absent**, making Fix 1's `cat`-only read race-free by construction.

These two fixes are defense-in-depth: Fix 1 closes the shell race window; Fix 2 ensures the file is never absent so there is no race to exploit.

## Before / after flow

```
BEFORE (race window exists):
  health-check: -f check → FILE EXISTS
  inbox_server:                    → finally: unlink()  ← race here
  health-check:          cat →                         → EMPTY → RED (false)

AFTER (no race window):
  health-check: cat 2>/dev/null → (reads "exited" or fresh timestamp)
  inbox_server:                   → finally: write("exited")
  health-check: integer guard  → non-integer → treat as absent → correct RED/GREEN
```

## Functional patterns used

Pure functions (`_clear_wfm_active_signal`, `_write_wfm_active_signal` — no shared mutable state, deterministic output). Named constant `WFM_ACTIVE_TOMBSTONE` traces directly to the spec so both the test and implementation reference the same value.

## Tests run

- [x] `uv run pytest tests/unit/test_wfm_active_signal.py -v` — 11 passed, 0 failed (6 pre-existing + 5 new)
- [x] `bash tests/test-health-check-wfm-active.sh` — 10/10 passed (8 pre-existing + 2 new)
- [x] `uv run pytest tests/unit/ --ignore=...pre-existing-broken... -q` — 2814 passed, 9 pre-existing failures (unchanged from main), 0 new failures
- [x] `uv run python -m py_compile src/mcp/inbox_server.py` — syntax OK

**Pre-existing failures** (unchanged from main, unrelated to this PR):
- `test_ghost_session_fixes.py`, `test_lobstertalk_telegram_routing.py` — `ADMIN_CHAT_ID_REDACTED` not defined in test env
- `test_require_background_agent.py` (5 tests), `test_require_write_result.py` (1 test) — pre-existing
- `test_message_tools.py` (3 tests) — same `ADMIN_CHAT_ID_REDACTED` issue

## Manual test

N/A — this change is entirely internal to the health-check shell script and MCP server process state management. No external services, no Telegram output, no file I/O affecting user-visible behavior. The race is non-deterministic by nature (depends on sub-millisecond timing), so manual testing cannot reproduce or disprove it. The fix is structural: tombstone-write eliminates the race by construction, not by timing.

## Definition of Done

- [x] Integration/manual test run — N/A (see Manual test above)
- [x] User-visible outcome verified — N/A (internal health-check behavior only)
- [x] External service calls: none (health check reads local files only)

Closes #1730

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -988,10 +988,17 @@ check_dispatcher_heartbeat() {
         # DISPATCHER_WFM_ACTIVE_FILE with a fresh epoch timestamp every 60s while
         # WFM is blocking. A fresh WFM-active file means the dispatcher is alive
         # and simply idle — not frozen or dead. (issue #1713 / #949)
+        #
+        # Fix 1 (issue #1730 TOCTOU): Use cat-only read, no -f existence gate.
+        # A two-step -f / cat sequence has a race window: the MCP server's
+        # finally block can write the tombstone between the -f check and the cat,
+        # making cat return empty and this function fall through to RED.
+        # cat 2>/dev/null is a single atomic read: empty result = absent or
+        # unreadable; non-empty integer = WFM active; non-integer = tombstone
+        # (WFM exited cleanly). The integer guard below handles all three cases
+        # without any race window.
         local wfm_active_ts=""
-        if [[ -f "$DISPATCHER_WFM_ACTIVE_FILE" ]]; then
-            wfm_active_ts=$(cat "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null | tr -d '[:space:]')
-        fi
+        wfm_active_ts=$(cat "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null | tr -d '[:space:]')
         if [[ -n "$wfm_active_ts" ]] && [[ "$wfm_active_ts" =~ ^[0-9]+$ ]]; then
             local wfm_age=$(( now - wfm_active_ts ))
             if [[ $wfm_age -le $WFM_ACTIVE_STALE_SECONDS ]]; then
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -1482,6 +1482,39 @@ def _write_wfm_active_signal() -> None:
         pass  # Never block wait_for_messages on a heartbeat write failure
 
 
+# Tombstone value written to WFM_ACTIVE_FILE when WFM exits (Fix 2, issue #1730).
+# Using a non-integer string means:
+#   - The file is NEVER absent between the health check's -f gate and its cat read
+#     (closing the TOCTOU race that caused the 2026-04-22 00:00Z false restart).
+#   - The health check's existing integer guard ([[ "$ts" =~ ^[0-9]+$ ]]) rejects
+#     this value and treats it as "WFM not active" — the correct semantic.
+WFM_ACTIVE_TOMBSTONE = "exited"
+
+
+def _clear_wfm_active_signal() -> None:
+    """Write a tombstone to WFM_ACTIVE_FILE instead of deleting it (issue #1730).
+
+    Replaces the previous WFM_ACTIVE_FILE.unlink() in the WFM finally block.
+    Writing a non-integer tombstone ("exited") instead of deleting ensures the
+    file is never transiently absent — closing the TOCTOU race between:
+      1. The health check's existence test  (if [[ -f ... ]])
+      2. The health check's content read    (cat ...)
+    If the file disappears between those two operations, cat returns empty and
+    the health check falls through to RED, triggering a false-positive restart.
+
+    The tombstone is parsed by the health check's integer regex guard and treated
+    as absent (WFM not active), which is the correct semantic.
+
+    Silently swallowed on failure: this runs in a finally block and must never
+    mask the original exception that caused WFM to exit.
+    """
+    try:
+        WFM_ACTIVE_FILE.parent.mkdir(parents=True, exist_ok=True)
+        WFM_ACTIVE_FILE.write_text(WFM_ACTIVE_TOMBSTONE + "\n")
+    except Exception:
+        pass  # Never propagate errors from a finally-block cleanup
+
+
 # ---------------------------------------------------------------------------
 # Session Guard
 #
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Verdict: PASS

All checks clean. Both fixes are correct by construction.

---

## Fix 1 — Shell: cat-only read

The removal of the `-f` gate and replacement with `cat 2>/dev/null` as the sole read is correct. The existing `[[ "$wfm_active_ts" =~ ^[0-9]+$ ]]` guard handles all three outcomes cleanly:
- File absent or unreadable → empty string → guard rejects → falls through to RED (correct)
- File contains tombstone `"exited"` → non-integer → guard rejects → falls through to RED (correct)
- File contains fresh epoch integer → guard passes → GREEN (correct)

No issues.

---

## Fix 2 — Python: tombstone write instead of unlink

`_clear_wfm_active_signal()` is called unconditionally from the `finally` block at line 4260–4269. Verified directly in the source. The function silently swallows all exceptions via a bare `except Exception: pass`, which is correct given it runs in a finally block and must never mask the original exception.

The `mkdir(parents=True, exist_ok=True)` inside `_clear_wfm_active_signal()` handles the edge case where WFM exits before the first heartbeat write creates the directory — this is correct and tested by `test_clear_wfm_active_signal_when_file_absent_is_silent`.

---

## Pre-existing failure count

Engineer claimed 9 pre-existing failures. Verified:
- 5 in `test_require_background_agent.py`
- 3 in `test_message_tools.py`
- 1 in `test_require_write_result.py`
- 2 files (`test_ghost_session_fixes.py`, `test_lobstertalk_telegram_routing.py`) fail at collection, consistent with `ADMIN_CHAT_ID_REDACTED` not defined in test env

Total: 9 pre-existing. Zero new failures introduced. Confirmed.

---

## Test coverage assessment

All three areas flagged for close review are covered:

1. **finally block unconditional** — verified in source (line 4269)
2. **Three-state integer guard** — Test 8 (non-integer → RED), Test 9 (tombstone → RED), Test 10 (fresh integer → GREEN) in shell tests; `test_clear_wfm_active_signal_tombstone_is_non_integer` in unit tests
3. **Exception path in `_clear_wfm_active_signal()`** — `test_clear_wfm_active_signal_silent_on_write_error` covers the write-failure path; `test_clear_wfm_active_signal_when_file_absent_is_silent` covers the no-file path

---

## Minor nit (non-blocking)

Shell Test 10 ("Fresh WFM-active read via cat-only (no -f gate) → GREEN") is setup-identical to the pre-existing Test 2 ("Stale heartbeat + fresh WFM-active → GREEN") — both write a 30s-old timestamp and expect exit 0. Test 10 would have passed against the old `-f` + `cat` code as well, since the file exists at read time. The test's stated goal (verifying cat-only works) is achieved transitively by the other tests, but the new test doesn't specifically exercise a scenario where the old code would have failed. Not a blocker — the fix is correct regardless; the test just adds redundant green-path coverage rather than a targeted regression test for the race.

---

## Summary

The two fixes are defense-in-depth: Fix 1 closes the shell race window; Fix 2 ensures there is no race to exploit even without Fix 1. Both are simple, surgical, and correct. 11/11 Python tests pass (verified). 10/10 shell tests pass (verified). 9 pre-existing failures confirmed, zero new.

PR #1734, opened 2026-04-22
fix(#1733): add reply-context grounding rule to dispatcher bootup

### LINKED ISSUE #1733: Dispatcher misreads Telegram reply context — anchor bias over reply-to content
## Summary

The dispatcher repeatedly misinterprets short Telegram messages sent as replies to its own messages. Instead of grounding interpretation in the **quoted reply-to content** that is explicitly shown in the `wait_for_messages` output, the dispatcher anchors to the **most recently active topic in its working context** and answers that instead. This causes consistent misreads that require Sahar to correct, adding friction and eroding trust.

---

### Primary incident — 2026-04-22

Sahar sent: **"Is this still happening?"** as a reply to Lobster's message:
> "Pulled up all 7 threads (Apr 13–Apr 18). Researching context for the ones I can't answer from memory — back in a few minutes with the first one."

**What Sahar meant:** Are you still researching, or are you done and ready to present?

**What Lobster answered:** An explanation of whether the **dispatcher restart problem** was still happening (Thread 1 of the old thread list).

**Why Lobster got it wrong:** The most recent topic in working context at the time was "Thread 1 = restart problem." Lobster used that as the referent for "this" instead of reading the `↩️ Replying to` block in the WFM output, which explicitly quoted the research status message.

Sahar's correction: *"I was replying to you saying you were doing research on old threads. Why can't you read replies well"*

---

### Additional confirmed instances

**2026-03-26** — Sahar sent a message replying to Lobster's message about **PR #968**, asking if #968 and #969 were the same work. Lobster answered about `pyproject.toml` (the most recent code topic in context) instead of the PR question. Sahar had to explicitly say: *"can you see what messages I'm replying to? I was replying to the message about 968 and asking about 969 but you seemed to think I was talking about pyproject.toml."* Lobster acknowledged: *"Yes, I can see it — I misread."*

**2026-04-21 (5:13 PM)** — Lobster answered a question about whether the health check fired a false-positive restart, then had to correct itself: *"You're right — I misread the test. You want to know: did the health check fire a false-positive restart after the dispatcher went idle? Not whether the dispatcher responded to messages."*

**2026-04-19 (6:41 PM)** — Sahar said *"I meant for now, in Linear"* and Lobster misread what she was referring to. It had to self-correct: *"Misread that — I spawned an engineer to start coding."*

---

## Root cause

The `wait_for_messages` output includes an explicit `↩️ Replying to (msg_id=XXXXX):` block with the full text of the original message whenever a user sends a Telegram reply. This context is unambiguous: it tells the dispatcher exactly which message the user is responding to. Here is the relevant rendering code in `inbox_server.py` (around line 5058):

```python
output += f"↩️ Replying to{type_label}{from_label}{id_label}:\n"
...
indented = "\n".join(f"  {line}" for line in reply_text.splitlines())
output += f"{indented}\n\n"
```

The information is there. The dispatcher fails to use it as the **primary grounding** for interpreting the user's message. Instead, the dispatcher reads the short user message (e.g. "Is this still happening?"), pattern-matches against recently-active topics in working context, and answers that topic — without first checking what message the user was replying to.

This is **anchor bias**: the most-recently-active topic in context crowds out the explicit structural context that the system provides.

The bootup file (`sys.dispatcher.bootup.md`) currently has **no explicit instruction** telling the dispatcher to read the `↩️ Replying to` block first, before interpreting any reply message.

---

### Behavioral rule (for `sys.dispatcher.bootup.md`)

Add a mandatory pre-interpretation step for all incoming user messages:

> **Reply-to grounding rule:** Before interpreting any user message, check whether the `wait_for_messages` output includes a `↩️ Replying to` block. If it does, read that block first and use the quoted original message as the **primary referent** for any pronouns, topic references, or questions in the user's message. Do not anchor interpretation to recently-active topics in working context. The reply-to content is structural fact; working context is background noise.
>
> Short messages like "Is this still happening?", "Is that fixed?", "What did you decide?", "Did you finish?" are especially prone to anchor-bias misreads. When these arrive as replies, the referent is almost always the quoted message, not the most recent topic in context.

### Implementation note

This is a pure behavioral rule change — no code required. The data is already there in the WFM output. The fix is to make the dispatcher read and apply it.

### Should this be added to `sys.dispatcher.bootup.md`?

Yes. The fix belongs in the dispatcher bootup file, under the "Context Recovery" section or as a new short section near the "Message Source Handling" section. It should appear **before** the main loop description so it is loaded into context before any message processing begins.

This issue is documenting the problem and the proposed fix; the actual edit should land in a dedicated PR.

---

## Impact

- Occurs consistently across multiple sessions (at least 4 confirmed instances since March 2026)
- Forces Sahar to send correction messages, wasting her time and eroding trust
- The fix is low-cost (one short rule added to bootup), high-value (eliminates a recurring class of misread)

sayhar:
## What problem this solves

The dispatcher has a documented anchor-bias bug (issue #1733): when Sahar sends a Telegram reply, it interprets her short message against the most recently active topic in working context instead of reading the explicit `↩️ Replying to` block first. This has caused at least four confirmed misreads:

- Apr 22, 2026: "Is this still happening?" (about a research status) → answered about dispatcher restarts
- Mar 26, 2026: Reply about PR #968 → answered about pyproject.toml
- Apr 21, 2026: Health check question → misread the test condition
- Apr 19, 2026: "I meant for now, in Linear" → misread the referent entirely

Each incident required Sahar to explicitly correct the dispatcher and explain what she was replying to — despite the `wait_for_messages` output containing an unambiguous `↩️ Replying to (msg_id=XXXXX):` block with the full quoted text.

## What changed

Added a single `**Reply-context grounding:**` behavioral rule in the Main Loop section of `sys.dispatcher.bootup.md`, placed between the stop-hook error rule and the 7-second rule — the point in the file where the dispatcher is instructed on how to handle incoming messages before routing.

The rule makes explicit what should already be the expected behavior: read the reply-to block first, ground pronouns and topic references in that content, then interpret the message. Applies specifically to short replies that are ambiguous without the reply context ("Is this still happening?", "Did you finish?", "What does that mean?").

## What was not changed

- No code changes — this is a behavioral instruction file only
- No changes to `inbox_server.py`, `wait_for_messages` output format, or message routing
- No changes to any other bootup or agent files

## Functional patterns

This is a pure behavioral constraint added to a declarative instruction document. No functional programming patterns apply — it's a rule specification, not code.

## Tests run
- [x] Verified the file diff is a clean 2-line insertion with no surrounding context disturbed
- [x] Confirmed the rule is in the correct location (Main Loop section, alongside other behavioral rules)
- [ ] Live behavioral test — not applicable in automated testing. This is a bootup instruction change; effect is observed at runtime when the dispatcher reads the file on next compaction or restart. No unit test framework exists for dispatcher behavioral rules.

## Manual test
N/A — this change is entirely internal to a behavioral instruction file. The rule instructs the dispatcher how to interpret incoming messages; verifying it works requires observing dispatcher behavior on a real reply-to message in a live session. The change is safe to deploy and soak: worst case the rule is ignored; it cannot cause regressions.

## Definition of Done
- [x] Unit tests pass with proper mocking — N/A (no executable code changed)
- [x] User-visible outcome: not directly verifiable pre-merge; soak period will surface whether misreads recur
- [x] Side-effect audit: no side effects — read-only instruction document addition

Closes #1733

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -119,6 +119,8 @@ Never pass `hibernate_on_timeout=True` — feature removed in issue #1442; cause
 
 **Stop hook error rule:** If the `require-wait-for-messages.py` stop hook fires and injects an error (e.g. "WFM not called"), the ONLY correct response is: call `wait_for_messages()` immediately. Do NOT treat the injected error message as a user prompt. Do NOT respond to it inline. The hook's intent is to force WFM — honor it by calling WFM and nothing else.
 
+**Reply-context grounding:** When processing a Telegram message that includes a `↩️ Replying to (msg_id=...)` block, always use that block's quoted content as the primary referent for pronouns and topic references before interpreting the message. Short replies like "Is this still happening?", "Did you finish?", "What does that mean?" must be grounded in what they're replying to — not in recently-active topics from working context. Read the reply-to block first, then interpret the message.
+
 ---
 
 ## The 7-Second Rule


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Diff read cold:** Single 2-line insertion in the Main Loop section of the dispatcher bootup file. No code changed — this is a behavioral instruction addition only.

**What I checked independently before reading the engineer's briefing:**

- Placement: The rule lands in the Main Loop section, between the stop-hook error rule and the 7-second rule. This is the correct zone — it's where the dispatcher reads instructions about how to handle incoming messages before routing. The `## Message Source Handling` section (later in the file) also mentions `telegram_message_id` context, but placing this rule earlier gives it higher precedence in reading order, which is what you want for a grounding constraint.

- Wording precision: "primary referent" (not "only referent") is the right choice — it doesn't overfit. The rule instructs the dispatcher to read the reply-to block *first*, not to ignore everything else. This matches the actual failure mode: the dispatcher was not reading the block at all, not reading it and then overriding it.

- Example phrases match the incident log exactly ("Is this still happening?") — good anchoring for the model.

- No edge case left open that could cause regressions. The rule is additive: a dispatcher that already grounds replies correctly is unaffected; one that was anchor-biased now has an explicit instruction to check the block first.

**Concern:** The rule exists only in the bootup file, which is read on cold start and compaction recovery. A dispatcher that's been running for hours without compaction may not have re-read this instruction. However, this is a pre-existing architectural limitation of the bootup file system — not a problem introduced by this PR. IFTTT rules loaded via `list_rules` at startup could be a stronger enforcement mechanism for future consideration.

**No blocking issues.** Safe to merge and soak.

PR #1736, opened 2026-04-22
fix(#1735): quarantine unrecognized inbox sources to stop WFM hot-loop

### LINKED ISSUE #1735: fix: health-check inbox source and wait_for_messages hot-loop guard
## Bug 1 (critical): daily-health-check.sh writes unrecognized inbox source

**File:** `scripts/daily-health-check.sh`, line 232

The health check writes inbox messages with `"source": "daily-health-check"` which is not in `INBOX_MESSAGE_SOURCES` (defined in `src/mcp/message_types.py`). This caused a silent infinite hot-loop in `wait_for_messages` on 2026-04-22, triggering 20 MCP session losses over ~2 hours.

**Root cause chain:**
1. Health check wrote file with `source: "daily-health-check"` to inbox
2. `wait_for_messages` glob at line 4160 found the file → returned immediately every call
3. Dispatcher could not dismiss the file (unknown source routing failure)
4. Hot loop exhausted `--max-turns 150` in ~7.5 min → Claude exited
5. `claude-persistent.sh` SIGTERMed MCP server → full restart cycle
6. New session found same file → repeat every ~7 minutes

**Fix:** Change `"source": "daily-health-check"` to `"source": "system"` in `scripts/daily-health-check.sh` line 232.

## Bug 2 (hardening): wait_for_messages has no guard against unrecognized-source files

**File:** `src/mcp/inbox_server.py`, around line 4160

When a `.json` file in `~/messages/inbox/` has an unrecognized source, `wait_for_messages` returns it on every call, creating an undetectable infinite hot loop. 

**Fix:** In `handle_check_inbox` (or the glob check in `wait_for_messages`), quarantine files with unrecognized sources by moving them to `~/messages/failed/` with an error annotation, rather than silently skipping them or returning them in a loop.

## Context
Both fixes should go in one PR. Root cause documented in system audit context.

sayhar:

On 2026-04-22, a single file written by `daily-health-check.sh` with an unrecognized `source` field (`"daily-health-check"`) triggered an infinite `wait_for_messages` hot-loop that caused 20 MCP session losses over ~2 hours.

The failure chain:
1. `daily-health-check.sh` wrote a health-check alert to inbox with `"source": "daily-health-check"` — not in `INBOX_MESSAGE_SOURCES`
2. `wait_for_messages` uses `INBOX_DIR.glob("*.json")` to detect messages; found the file every call → returned immediately every call
3. Dispatcher received the message but could not route or dismiss it (unknown source → no handler → never called `mark_processing`/`mark_processed`)
4. File stayed in inbox forever → every `wait_for_messages` call returned immediately → hot-loop
5. ~130 WFM calls in 7.5 minutes exhausted `--max-turns 150` → Claude exited with code 1
6. `claude-persistent.sh` SIGTERMed the MCP server, systemd restarted it, new session found same stuck file → repeat every ~7 minutes

## Changes

**Bug 1 (one-line fix):** `scripts/daily-health-check.sh` line 232

Changed `"source": "daily-health-check"` to `"source": "system"`. The `system` source is already in `INBOX_MESSAGE_SOURCES` and is the correct value for script-generated messages (the `_STATUS_SOURCES_WITH_HINTS` map in `inbox_server.py` already maps `"system": "health_check"`).

**Bug 2 (hardening):** `src/mcp/inbox_server.py` — `handle_check_inbox`

Added a quarantine guard in the normal inbox scan path. When a file has a non-empty source that is not in `INBOX_MESSAGE_SOURCES`:
- The file is moved to `failed/` with `_permanently_failed=True`, `_last_error`, and `_last_failed_at` metadata
- An `inbox.failed` event is emitted to the EventBus
- The file is skipped (not returned to the dispatcher)
- On subsequent `wait_for_messages` calls, the inbox is empty → no more hot-loop

The original message data is fully preserved in `failed/` for inspection. The quarantine is permanent (no retry) because the source field is static — retrying would always fail.

## State transition (Bug 2)

Before fix:
```
inbox/daily-health-X.json (source="daily-health-check")
  → check_inbox returns it
  → dispatcher can't route it
  → file stays in inbox
  → next WFM call finds it again
  → ∞ loop
```

After fix:
```
inbox/daily-health-X.json (source="daily-health-check")
  → check_inbox detects unrecognized source
  → file moved to failed/ with _permanently_failed=True
  → check_inbox skips it (not returned)
  → next WFM call: inbox is empty → blocks until real message arrives
```

## Functional patterns

The quarantine logic is a pure filter: it reads, annotates with error metadata, moves to `failed/`, and skips — no mutation of shared state, no side effects beyond the file move and an EventBus emit. It follows the same pattern as the existing `handle_mark_failed` handler.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_unrecognized_source_quarantine.py -v` — 6 passed, 0 failed (new tests for Bug 2 quarantine behavior)
- [x] `uv run pytest tests/unit/test_mcp_server/ --ignore=tests/unit/test_mcp_server/test_ghost_session_fixes.py --ignore=tests/unit/test_mcp_server/test_lobstertalk_telegram_routing.py -v` — 771 passed, 3 failed (all 3 failures are pre-existing `ADMIN_CHAT_ID_REDACTED` NameErrors unrelated to this PR)
- [x] `uv run pytest tests/unit/ --ignore=tests/unit/test_mcp_server/test_ghost_session_fixes.py --ignore=tests/unit/test_mcp_server/test_lobstertalk_telegram_routing.py -q` — 2800 passed, 29 failed (all 29 pre-existing: `ADMIN_CHAT_ID_REDACTED`, hook tests, TTS piper tests)

## Manual test

N/A for Bug 1: one-line string change in a bash script, verified by unit tests and diff inspection.

N/A for Bug 2: the quarantine logic is exercised by the new unit tests, which use the existing `isolate_inbox_server_paths` fixture to redirect all production paths to a temp directory. No production file system is touched in tests.

The production system was manually remediated before this PR: the stuck file was moved from `inbox/` to `processed/` on 2026-04-22 at ~08:07 UTC, and the system has been stable since.

## Definition of Done
- [x] Integration/manual test run — N/A (no external services involved; file-system I/O is covered by unit test isolation)
- [x] User-visible outcome — N/A (these are system-internal fixes; user-visible outcome was verified historically: system became stable immediately after manual remediation)
- [x] Side-effect audit: quarantine only moves one file per unrecognized-source message; no loops, no mass writes, no shared-state mutation
- [x] External service calls: none (pure file-system operations)

Closes #1735

### DIFF
--- a/scripts/daily-health-check.sh
+++ b/scripts/daily-health-check.sh
@@ -229,7 +229,7 @@ if [ ${#FAILURES[@]} -gt 0 ]; then
     cat > "$MSG_FILE" << MSGEOF
 {
   "type": "health_check",
-  "source": "daily-health-check",
+  "source": "system",
   "timestamp": "$TIMESTAMP",
   "subject": "Daily health check: ${#FAILURES[@]} failure(s)",
   "body": "The daily dependency health check found problems:\n\n$FAIL_LIST\n\nCheck the log for details: $LOG_FILE",
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4683,14 +4683,15 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
                     # Re-read files that failed the first pass (e.g. transient lock)
                     with open(f) as fp:  # type: ignore[arg-type]
                         msg = json.load(fp)
-                if source_filter and msg.get("source", "").lower() != source_filter:
-                    continue
                 # Quarantine files with unrecognized sources (issue #1735).
-                # An unknown source means the dispatcher cannot route or dismiss
-                # the message, creating an infinite hot-loop: wait_for_messages
-                # sees the file on every call and returns immediately, exhausting
-                # --max-turns in ~7 minutes.  Move to failed/ permanently so the
-                # file is preserved for inspection but cannot block the loop.
+                # This check runs before source_filter so that a bad-source file
+                # is always quarantined regardless of whether the caller passed a
+                # source= filter.  An unknown source means the dispatcher cannot
+                # route or dismiss the message, creating an infinite hot-loop:
+                # wait_for_messages sees the file on every call and returns
+                # immediately, exhausting --max-turns in ~7 minutes.  Move to
+                # failed/ permanently so the file is preserved for inspection but
+                # cannot block the loop.
                 msg_source = msg.get("source", "")
                 if msg_source and msg_source not in INBOX_MESSAGE_SOURCES:
                     error_msg = (
@@ -4714,6 +4715,8 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
                     except Exception as _q_exc:
                         log.error(f"check_inbox: quarantine failed for {f}: {_q_exc}")  # type: ignore[union-attr]
                     continue  # skip — quarantined
+                if source_filter and msg.get("source", "").lower() != source_filter:
+                    continue
                 # /report slash command pre-processor: handle automatically without
                 # surfacing the raw message to the main dispatcher loop.
                 msg_text = msg.get("text", "")
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4685,6 +4685,35 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
                         msg = json.load(fp)
                 if source_filter and msg.get("source", "").lower() != source_filter:
                     continue
+                # Quarantine files with unrecognized sources (issue #1735).
+                # An unknown source means the dispatcher cannot route or dismiss
+                # the message, creating an infinite hot-loop: wait_for_messages
+                # sees the file on every call and returns immediately, exhausting
+                # --max-turns in ~7 minutes.  Move to failed/ permanently so the
+                # file is preserved for inspection but cannot block the loop.
+                msg_source = msg.get("source", "")
+                if msg_source and msg_source not in INBOX_MESSAGE_SOURCES:
+                    error_msg = (
+                        f"Unrecognized inbox source {msg_source!r} — not in INBOX_MESSAGE_SOURCES. "
+                        "File quarantined to failed/ to prevent hot-loop. "
+                        "Add to INBOX_MESSAGE_SOURCES in message_types.py if this is intentional."
+                    )
+                    log.error(f"check_inbox: quarantining {f.name}: {error_msg}")  # type: ignore[union-attr]
+                    try:
+                        msg["_permanently_failed"] = True
+                        msg["_last_error"] = error_msg
+                        msg["_last_failed_at"] = datetime.now(timezone.utc).isoformat()
+                        dest = FAILED_DIR / f.name  # type: ignore[union-attr]
+                        atomic_write_json(dest, msg)
+                        f.unlink(missing_ok=True)  # type: ignore[union-attr]
+                        _emit_mcp_event(
+                            "inbox.failed",
+                            {"message_id": f.stem, "error": error_msg, "permanent": True},  # type: ignore[union-attr]
+                            severity="warn",
+                        )
+                    except Exception as _q_exc:
+                        log.error(f"check_inbox: quarantine failed for {f}: {_q_exc}")  # type: ignore[union-attr]
+                    continue  # skip — quarantined
                 # /report slash command pre-processor: handle automatically without
                 # surfacing the raw message to the main dispatcher loop.
                 msg_text = msg.get("text", "")
--- a/src/mcp/message_types.py
+++ b/src/mcp/message_types.py
@@ -73,6 +73,7 @@ INBOX_MESSAGE_SOURCES: frozenset[str] = frozenset({
     "whatsapp",
     "bisque",
     "system",
+    "bot-talk",  # cross-Lobster bot-to-bot messages (issue #1350)
 })
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


sayhar (commented):

## Summary

The two core fixes are correct and necessary. The source field change in the health check script is a clean one-line fix. The quarantine logic is well-structured and the new tests pass. However, there is one real regression introduced by this PR that must be fixed before merge.

## FAIL: bot-talk messages will be permanently quarantined (regression)

The quarantine guard blocks any inbox message whose source is not in the recognized-sources list. The cross-Lobster bot-talk integration writes messages into the inbox with `source="bot-talk"` — but `"bot-talk"` is not in `INBOX_MESSAGE_SOURCES`. As a result, every inbound bot-talk message will be moved to `failed/` with `_permanently_failed=True` and silently dropped. The code that handles `source == "bot-talk"` (the routing logic around line 4744 and the outbox logic around lines 4961–4971) will never be reached.

This is a silent, permanent data loss: the messages disappear from the inbox without any user notification, and the bot-talk integration breaks entirely with no warning to the operator.

**Fix:** Add `"bot-talk"` to `INBOX_MESSAGE_SOURCES` in `message_types.py`. A test asserting that bot-talk messages pass through without quarantine should be added to the new test file.

## Minor: source filter bypasses quarantine

When `check_inbox` is called with a `source` filter argument, the filter check runs before the quarantine check. A file with an unrecognized source is skipped by the filter and never quarantined. This means the hot-loop protection only fires on unfiltered calls. The real-world dispatcher always calls without a source filter, so this is low risk in practice — but it is a latent gap worth noting.

## Minor: test's recognized-sources list is hardcoded

The test that verifies recognized sources pass through hardcodes the list `["telegram", "slack", ...]` rather than reading it from `INBOX_MESSAGE_SOURCES`. If the canonical set is updated in the future, the test will not automatically reflect the change. Low priority, but worth a follow-up.

## Pre-existing failures

The engineer's claim of 5 failing test files is consistent with what I observed. The failures are in hooks tests, TTS tests, and two collection errors due to an undefined name — all pre-existing and unrelated to this PR.

## What must change before merge

1. Add `"bot-talk"` to `INBOX_MESSAGE_SOURCES` in `message_types.py`
2. Add a test asserting that `source="bot-talk"` messages are not quarantined

sayhar (commented):
test comment

sayhar:
Fixed all three issues from the NEEDS-WORK review.

**Blocking fix:** Added `"bot-talk"` to `INBOX_MESSAGE_SOURCES` in `message_types.py`. Every inbound cross-Lobster message would have been permanently quarantined without this — the bot-talk integration was completely broken by the quarantine guard as written.

**Minor fix 1 (source_filter bypass):** Moved the quarantine check before the `source_filter` check in `handle_check_inbox`. The guard now runs unconditionally for every file before any filtering — a file with an unrecognized source is quarantined whether or not the caller's `source=` argument matches it.

**Minor fix 2 (hardcoded test list):** The `test_quarantine_does_not_affect_recognized_sources` test now uses `sorted(INBOX_MESSAGE_SOURCES)` instead of a hardcoded list. Adding a new source to the constant will automatically exercise it in the test.

**New tests:**
- `test_bot_talk_messages_pass_quarantine_guard` — asserts bot-talk messages are returned normally, not quarantined
- `test_quarantine_guard_runs_with_source_filter` — asserts quarantine fires even when `source="custom-script"` is passed as the filter

All 8 quarantine tests pass. 47 tests in the taxonomy/priority/quarantine suite pass.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

All three NEEDS-WORK items from the first review are resolved. Tests run clean.

## Fixes verified

**1. bot-talk added to recognized sources** ✅

The cross-Lobster source type is now in the list of recognized message origins. Verified in `message_types.py`. Without this, every inbound cross-Lobster message would have been permanently deleted by the new quarantine guard — the fix was necessary to keep that integration functional.

**2. Quarantine guard runs before source filter** ✅

The guard (lines 4686–4717) fires unconditionally, then the source filter check follows at line 4718. Confirmed in source. The new test for this (`test_quarantine_guard_runs_with_source_filter`) exercises the real behavior: it writes a bad-source file, calls `check_inbox` with a matching source filter, and asserts the file lands in `failed/` — not returned to the dispatcher.

**3. Tests import from live constant** ✅

The test file imports `INBOX_MESSAGE_SOURCES` directly from `message_types.py`. The recognized-sources coverage test iterates that live set, so any new source added to the constant is automatically exercised.

## New tests

8 tests in the new file, all pass. No duplicate class names. Each test writes real files to a temp inbox directory and calls the actual inbox handler — they exercise code paths, not comments.

## Pre-existing failures

42 collection errors exist on both `main` and this branch. Count is identical — this PR introduces no new failures.

## One minor gap (not blocking)

The quarantine guard only applies to the normal (priority-scan) path. The historical/`since_ts` scan path has no quarantine check. A bad-source file in the inbox would not be quarantined if `check_inbox` is called with a `since_ts` parameter — it would just be silently skipped. The hot-loop issue affects only the normal dispatch path, so this gap does not affect the fix. Worth a follow-up ticket if the since_ts path needs the same protection.

## Shell script rename

The health-check script now emits `source: "system"` instead of the unrecognized `"daily-health-check"`. This is what actually caused the original 20-session-loss incident, and the fix directly addresses the root cause. Verified that `"system"` is in the recognized sources list and `"daily-health-check"` is not.

PR #1774, opened 2026-04-23
fix(#1763): pass env vars to container init service

### LINKED ISSUE #1763: fix(docker-staging): lobster-container-init.service missing PassEnvironment — cold boot writes blank config.env
## Problem

`lobster-container-init.service` is the systemd unit that runs `lobster-container-init.sh` inside Docker containers. It does **not** have `PassEnvironment` directives, so environment variables from the Docker `env_file` (e.g. `TELEGRAM_BOT_TOKEN`, `LOBSTER_USER`, etc.) are **not** passed through when systemd spawns the service.

The env vars exist in the container environment (injected by Docker via `env_file`), but systemd services run in a clean environment and do not inherit the container's environment unless explicitly configured with `PassEnvironment` or `EnvironmentFile`.

## Impact

On a cold `docker compose up`, the init script fires via systemd and sees no env vars → it writes blank/empty values into `config.env`. This silently misconfigures the system with no obvious error at startup.

## Reproduction

1. Start from a clean state (no pre-existing `config.env`)
2. Run `docker compose up` (cold boot)
3. Observe that `lobster-container-init.service` fires via systemd
4. Check `config.env` — values are blank because the env vars were not passed to the systemd unit

## Why this was missed

PR #1724 tested using `docker exec -e VAR=value ...`, which bypasses systemd entirely and injects the env vars directly into the exec'd process. That path works fine. The issue only surfaces on a true cold compose-up where systemd is the one spawning the init script.

## Proposed Fix

Add `PassEnvironment` directives to `lobster-container-init.service` for all environment variables that `lobster-container-init.sh` reads. For example:

```ini
[Service]
PassEnvironment=TELEGRAM_BOT_TOKEN
PassEnvironment=LOBSTER_USER
PassEnvironment=LOBSTER_HOST
# ... any other vars read by init.sh
```

Alternatively, point the unit at an `EnvironmentFile=` that Docker writes before the unit starts, or restructure so the init script reads directly from the Docker env_file path.

## Notes

- This is a **pre-existing issue**, independent of PR #1724
- Discovered during staging validation of PR #1724
- The fix should be applied to the systemd unit definition used in the Docker image

sayhar:
## Problem

On a cold `docker compose up`, `lobster-container-init.service` fires via systemd and writes `config.env` from its environment. But systemd spawns services in a clean environment — Docker's `env_file` values (injected into the container's PID 1 environment) are not automatically inherited by child units. Without `PassEnvironment` directives, `lobster-container-init.sh` sees no values for `TELEGRAM_BOT_TOKEN` and the other vars, writes a blank `config.env`, and the router starts with an empty token — crash-looping on every cold boot.

## What changed

Added `PassEnvironment=` directives to the `[Service]` section of `docker/staging/lobster-container-init.service` for all five environment variables the init script reads:

- `TELEGRAM_BOT_TOKEN`
- `TELEGRAM_ALLOWED_USERS`
- `LOBSTER_ADMIN_CHAT_ID`
- `LOBSTER_ENV`
- `LOBSTER_DEBUG`

This is the complete set — confirmed by reading `lobster-container-init.sh` lines 57–64 where these five vars are the only ones referenced in the `config.env` heredoc.

The fix is in the source file (`docker/staging/lobster-container-init.service`), which is `COPY`'d directly into `/etc/systemd/system/` by `Dockerfile.staging` line 256. No template layer or `install.sh` to update.

## Boot flow (before / after)

```
Before:
  docker compose up
    └── PID 1 env: TELEGRAM_BOT_TOKEN=xxx, ...
          └── systemd starts lobster-container-init.service
                └── clean env: TELEGRAM_BOT_TOKEN=(unset)
                      └── config.env written with blank values
                            └── lobster-router starts, crashes (empty token)

After:
  docker compose up
    └── PID 1 env: TELEGRAM_BOT_TOKEN=xxx, ...
          └── systemd starts lobster-container-init.service
                └── PassEnvironment copies vars from PID 1 env
                      └── config.env written with correct values
                            └── lobster-router starts OK
```

## Tests run

- [x] Diff reviewed manually — only `PassEnvironment` lines added, no other changes
- [ ] Cold `docker compose up` integration test — blocked: Docker/staging environment not available in this subagent context. Requires a container restart to verify. **Needs manual verification before merge.**

## Manual test

Required before merge:

1. Rebuild the staging image (`docker compose build`)
2. Start from clean state (`docker compose down -v && docker compose up`)
3. Confirm `lobster-container-init.service` exits with status 0 (`systemctl status lobster-container-init`)
4. Confirm `config.env` contains non-blank `TELEGRAM_BOT_TOKEN` value
5. Confirm `lobster-router.service` starts without crash-looping

The manual workaround (writing config.env directly) will no longer be needed once this fix is deployed.

## Definition of Done

- [x] Unit tests pass with proper mocking — N/A, pure config change
- [ ] Integration/manual test run — blocked, see above
- [ ] User-visible outcome verified — after cold restart, router stays up and bot responds in Telegram
- [x] Side-effect audit complete: no floods, no unscoped writes
- [x] External service calls: none introduced by this change

Closes #1763

### DIFF
--- a/docker/staging/lobster-container-init.service
+++ b/docker/staging/lobster-container-init.service
@@ -11,6 +11,14 @@ User=lobster
 Group=lobster
 Environment=HOME=/home/lobster
 Environment=PATH=/home/lobster/.local/bin:/usr/local/bin:/usr/bin:/bin
+# Pass env vars injected by docker-compose env_file into this systemd unit.
+# Without these directives systemd spawns the service in a clean environment,
+# so lobster-container-init.sh sees no values and writes a blank config.env.
+PassEnvironment=TELEGRAM_BOT_TOKEN
+PassEnvironment=TELEGRAM_ALLOWED_USERS
+PassEnvironment=LOBSTER_ADMIN_CHAT_ID
+PassEnvironment=LOBSTER_ENV
+PassEnvironment=LOBSTER_DEBUG
 ExecStart=/home/lobster/lobster-container-init.sh
 StandardOutput=journal
 StandardError=journal


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — with one observation and one minor gap worth tracking.

## Verdict: PASS

The fix is correct, minimal, and well-scoped. The five `PassEnvironment` directives exactly match the five variables written into `config.env` in `lobster-container-init.sh` lines 57–64. The engineer's claim that this is the complete set is verifiable by reading the heredoc directly — no additional vars are interpolated there.

**Potential issues I identified cold:**

1. **`GITHUB_TOKEN` gap** — `docker-compose.staging.yml` passes `GITHUB_TOKEN` into the container via the `environment:` block (not `env_file`), so it lands in PID 1's env. If any future step in the init chain needed it, it would also be missing from the service environment. For this PR it doesn't matter — `lobster-container-init.sh` never references `GITHUB_TOKEN` and the `--container-setup` path in `install.sh` doesn't either. This is not a blocker, but worth noting for future changes that extend the init script.

2. **`PassEnvironment` vs `EnvironmentFile` approach** — `PassEnvironment` is the right call here. An `EnvironmentFile=` pointing at the compose-supplied env file would require knowing the file path inside the container at service-write time, which is not how this setup works (vars arrive via `env_file` into PID 1, not as a file on disk). `PassEnvironment` is semantically correct and well-suited to this pattern.

3. **Template / install.sh layer** — confirmed: `Dockerfile.staging` line 256 `COPY`s the service file directly, and `install.sh` has no template for `lobster-container-init.service`. No second location to update. The engineer's analysis is correct.

4. **`--container-setup` indirect reads** — checked `install.sh` lines 363–430 for the `CONTAINER_SETUP=true` path. It reads `LOBSTER_INSTALL_DIR`, `LOBSTER_WORKSPACE`, `LOBSTER_MESSAGES`, and `LOBSTER_CONFIG_DIR` — all of which are set explicitly by the init script before calling `install.sh` (lines 78–82 of `lobster-container-init.sh`) via inline `VAR=value bash install.sh`, not from the service environment. These do not need `PassEnvironment`.

## One observation: missing `RemainAfterExit` robustness note

Not a blocker, but: the service is `Type=oneshot` with `RemainAfterExit=yes`. If the init script exits non-zero (e.g. `set -euo pipefail` triggers), `config.env` may be partially written and the downstream services will start anyway since systemd's `Before=` ordering does not gate on success — only on start/stop ordering. There is no `ConditionPathExists=` guard or `BindsTo=` dependency on the downstream services. This pre-existed this PR and is not introduced by it, but it's worth a follow-up issue if cold-boot reliability is a priority.

## Test plan coverage

The PR correctly notes that the integration test (cold `docker compose up`) could not be run and marks it as a manual gate before merge. Given the change is purely additive config (no logic, no new code paths), that is acceptable. The manual test steps listed are complete and sufficient.

**Summary:** Change is correct and complete for its stated scope. The one gap (no integration test run) is properly flagged in the PR. Safe to merge after manual cold-boot verification.

PR #1793, opened 2026-04-24
fix(#1790): context-monitor reads transcript JSONL for real token counts

### LINKED ISSUE #1790: fix: context-monitor.py reads context_window from transcript JSONL (was never populated by CC)
## Root cause

`context-monitor.py` calls `data.get('context_window')` on the PostToolUse hook payload. Claude Code never includes `context_window` in PostToolUse payloads — this field simply does not exist there. The hook silently no-ops on every MCP tool call. (This is tracked upstream in anthropics/claude-code#41689, now closed as duplicate of #32014 and #35059.)

Additionally, even the statusline hook — which *does* receive `context_window` — reports `context_window_size = 200000` even in 1M context sessions (anthropics/claude-code#32014). So even if PostToolUse were fixed, the field would be wrong for Max/Enterprise users on Opus 4.6 or Sonnet 4.6 with 1M context.

## What we found in the transcript JSONL

The transcript JSONL (at `transcript_path` in the hook payload) is the right source. Every `assistant` entry with a `message.usage` block contains:
- `input_tokens`
- `cache_creation_input_tokens`
- `cache_read_input_tokens`
- `output_tokens`

The sum `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` is a reliable estimate of current context usage. In a live session (claude-sonnet-4-6, ~143k tokens peak), this number tracked context growth monotonically and reached ~72% of the 200k limit — plausible given compaction history.

There is **no** `context_window`, `max_tokens`, or context tier field in the JSONL. The model field is a bare ID (`claude-sonnet-4-6`) with no context tier qualifier.

## Proposed fix

In `_handle_payload`, when `context_window` is absent:
1. Open the JSONL at `data['transcript_path']`
2. Walk backwards to find the last `assistant` entry with a `usage` block
3. Compute `total = input_tokens + cache_creation_input_tokens + cache_read_input_tokens`
4. Divide by `MODEL_CONTEXT_WINDOWS[model]` to get `used_pct`
5. Proceed with normal threshold logic

## Context window lookup table

From Anthropic API docs (as of April 2026):

| Model | Context Window |
|---|---|
| claude-sonnet-4-6 | 1,000,000 |
| claude-opus-4-6 | 1,000,000 |
| claude-haiku-4-5 | 200,000 |
| claude-opus-4-5 | 200,000 |
| claude-sonnet-4-5 | 200,000 |
| claude-opus-4-1 | 200,000 |

## Open question / blocker

**Opus 4.6 context window is account-tier-dependent.** Per anthropics/claude-code#32014, the same model ID is used for both 200k (Standard tier) and 1M (Max/Enterprise tier) sessions. The transcript JSONL does not record which context cap is in effect.

For now, the safest heuristic is:
- Use `MODEL_CONTEXT_WINDOWS` with API-documented sizes (Sonnet 4.6 = 1M, Opus 4.6 = 1M, Haiku 4.5 = 200k)
- If the observed token count exceeds the hardcoded limit (impossible in theory), fall back to the observed count as the denominator

We could also add an env var `LOBSTER_CONTEXT_WINDOW_OVERRIDE=200000` for users on Standard tier who are on Sonnet 4.6 but capped at 200k.

## Edge cases

- **Sidechain entries**: `isSidechain: true` entries appear to be absent in the current transcript (0 found in 595 entries). The fix should skip sidechain entries or handle them separately.
- **Empty/short transcripts**: Hook can fire before any assistant turn. Need a fallback (log a WARN, don't crash).
- **Subagent transcripts**: Subagents write to different JSONL files. The `transcript_path` in the hook payload will point to the correct file — no special handling needed.

sayhar:
## What was broken

`context-monitor.py` called `data.get('context_window')` on the PostToolUse hook payload. Claude Code never includes `context_window` in PostToolUse payloads — this field simply doesn't exist there (confirmed in production logs and upstream in anthropics/claude-code#32014 / #35059). Every single hook firing was a silent no-op: the hook ran, found nothing, logged a WARN, and returned. Context monitoring was completely non-functional.

The previous workaround (logging a WARN for absent context_window) made "hook fired, no data" distinguishable from "hook never fired," but still provided zero actual context usage data.

## What's different now

The transcript JSONL — already available at `transcript_path` in the PostToolUse payload — is the correct source. Every assistant turn in the transcript carries a `message.usage` block with actual token counts. This is what CC itself uses internally.

New approach:
1. Extract `transcript_path` from the hook payload
2. Read the JSONL file line-by-line (O(1) memory), keeping the last assistant turn with a `usage` block
3. Compute `total_used = input_tokens + cache_creation_input_tokens + cache_read_input_tokens`
4. Divide by the model's known max context (looked up by prefix match on the model ID)
5. Trigger wind-down at the same 70% `WARNING_THRESHOLD` with the same dedup/inbox logic

Model lookup table:
| Prefix | Max context |
|---|---|
| `claude-sonnet-4-6` | 1,000,000 |
| `claude-opus-4-6` | 1,000,000 |
| `claude-haiku-4-5` | 200,000 |
| unknown | 200,000 (conservative fallback) |

## Functional design notes

The implementation uses pure functions: `_read_transcript_usage(transcript_path, tool_name) -> (used_pct, remaining_pct, model) | None` and `_model_max_context(model) -> int` are both side-effect-free and independently testable. The JSONL read is a single linear pass with no buffering beyond the current line — suitable for sessions with thousands of turns.

The existing helper functions (`_log_usage`, `_build_warning_message`, `_write_warning_to_inbox`, DEDUP_FLAG logic) are unchanged. Only `_handle_payload`'s context-extraction logic and `_build_log_entry`'s signature changed (swapped `context: dict` for `model: str`).

## What doesn't change

- The 70% warning threshold
- The dedup flag (`/tmp/lobster-context-warning-sent`)
- The inbox message format (`context_warning` type, same fields)
- The `log_dir` / `inbox_dir` injectable parameters on `_handle_payload`
- The "hook fired, no data" WARN logging when `transcript_path` is absent

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_context_monitor.py -v` — 19 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 278 passed, 0 failed (full hook suite, no regressions)

## Manual test

N/A — this change is entirely internal to the hook. The hook fires on every PostToolUse event but produces no user-visible output below the 70% threshold. Above the threshold it writes to the inbox, which is tested via unit test with an injectable inbox_dir. No external service calls, no systemd, no DB schema changes.

Closes #1790

### DIFF
--- a/hooks/context-monitor.py
+++ b/hooks/context-monitor.py
@@ -2,21 +2,27 @@
 """
 PostToolUse hook: context window guard.
 
+Reads actual token usage from the session's transcript JSONL file rather than
+relying on the `context_window` field in the PostToolUse payload — which Claude
+Code never populates for PostToolUse events (confirmed in production and SDK type
+analysis; tracked upstream in anthropics/claude-code#32014 / #35059).
+
+Approach:
+  1. Extract `transcript_path` from the hook payload.
+  2. Read the JSONL file line-by-line to find the last assistant turn that
+     contains a `message.usage` block.
+  3. Compute total tokens = input_tokens + cache_creation_input_tokens
+     + cache_read_input_tokens.
+  4. Look up the model's max context window from a known table (1M for Sonnet/Opus
+     4.x, 200k for Haiku or unknown models).
+  5. Compute used_pct and trigger wind-down mode if at or above WARNING_THRESHOLD.
+
 Below the warning threshold (default 70%): logs usage to context-monitor.log.
 At or above the threshold: writes a context_warning message to ~/messages/inbox/
 so the dispatcher can enter wind-down mode and trigger a graceful restart.
 
 Dedup: once a warning is written, /tmp/lobster-context-warning-sent is touched
 so subsequent hook firings skip the write until the flag is cleared on restart.
-
-Issue #1430: Claude Code does NOT populate context_window in PostToolUse payloads
-for MCP tool calls — only for built-in tools like Bash.  Previously the hook
-silently returned when context_window was None, making "hook fired, no data"
-indistinguishable from "hook never fired."
-
-Fix: when context_window is absent, log a WARN entry so the log records that the
-hook fired.  The hook matcher in settings.json was also broadened to include Bash
-(which does carry context_window), ensuring real usage data is captured.
 """
 import json
 import sys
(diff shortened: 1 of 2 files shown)


sayhar:
## Code Review — PR #1793

### 1. Correctness of `_read_transcript_usage`

The implementation is correct. The line-by-line JSONL parse with a single-pass \"keep last\" approach is the right idiom — O(n) time, O(1) memory regardless of transcript length. The filtering logic (`type == "assistant"`, `role == "assistant"`, `"usage" in msg`) correctly targets only assistant turns with a populated usage block.

Edge cases handled gracefully:
- `transcript_path` is `None` or falsy → early return `None`
- File does not exist → `None` (checked via `path.exists()`)
- File unreadable (permissions, etc.) → `OSError` caught → `None`
- Malformed JSON lines → `JSONDecodeError` caught per-line → line skipped, parsing continues
- No assistant turns in transcript → `last_usage` stays `None` → returns `None`
- Explicit `null` values for token fields (JSON `null` → Python `None`) → the `or 0` pattern after `.get(..., 0)` handles this correctly

One minor observation: the function signature accepts `tool_name: str` as a second parameter but never uses it inside the function body — it exists only to match the call site convention. This is fine but slightly surprising. Not a bug.

### 2. Model lookup table

The table covers the three models specified in the checklist:
- `claude-sonnet-4-6` → 1,000,000 ✓
- `claude-opus-4-6` → 1,000,000 ✓
- `claude-haiku-4-5` → 200,000 ✓

Prefix matching via `str.startswith()` correctly handles versioned suffixes like `claude-haiku-4-5-20251001`. The 200,000 conservative fallback for unknown models avoids false negatives (triggering too late) at the cost of possible false positives (triggering too early on a larger-context model). This tradeoff is intentional and documented.

Note: `claude-sonnet-4-5` (if it ever appears in the transcript) falls back to 200k. This seems intentional given the PR scope — it would fire the warning earlier than necessary on that model, which is the safe direction.

### 3. Token math

`input_tokens + cache_creation_input_tokens + cache_read_input_tokens` is the correct sum. `output_tokens` is correctly excluded — output tokens do not consume the input context window limit. This matches the Claude API's usage semantics for context window accounting.

The `or 0` guard (`last_usage.get("input_tokens", 0) or 0`) correctly handles the case where the API returns an explicit JSON `null` for a field — Python's `.get(key, 0)` returns `None` for an explicit null, and `None or 0` yields `0`.

### 4. Tests

19 tests, all appropriate. Coverage includes:
- Correct percentage calculation from token sum (including all three cache fields)
- Last-turn-wins semantics with multiple turns
- `None` path, missing file, no assistant turns
- Model table: Sonnet 4.6, Opus 4.6, Haiku 4.5 (bare and versioned), unknown, empty string
- `_handle_payload` end-to-end: below threshold, at threshold (inbox write), dedup suppression
- WARN fallback when transcript absent, no inbox write when absent
- Injectable path parameters (signature test)

One gap worth noting but not a blocker: there is no test for a transcript containing a mix of user turns and assistant turns without a usage block followed by an assistant turn with a usage block (i.e., verifying the parser correctly skips assistant entries that lack `"usage"` in the message). The `test_returns_none_when_no_assistant_turns` test only covers the case where there are zero assistant-type entries. The code handles this correctly (`"usage" in msg` guards the update), but it's not explicitly exercised.

There is also no test for a completely empty transcript file (zero bytes or only whitespace lines). The code handles it correctly (empty lines are skipped, `last_usage` stays `None`), but the test suite doesn't verify it.

Neither gap is a blocker.

### 5. Backwards compatibility / fallback

When `transcript_path` is absent, the hook logs a WARN entry and returns early — identical behavior to the previous "context_window absent" path. The `warn` field text (`"[WARN] context_window absent in payload for tool: ..."`) is preserved verbatim for log-parser compatibility, which is good. The new `reason` field provides additional context without breaking existing parsers.

### 6. install.sh / settings.json

No install.sh changes in the PR. The hook is already registered in `settings.json` with matcher `"Bash|mcp__lobster-inbox__|Agent"`, which covers both built-in and MCP tools. No changes needed here. ✓

### 7. Nits

- `_read_transcript_usage` accepts `tool_name` but ignores it. The parameter serves no purpose inside the function. Could be removed from the signature; call sites pass it but it's unused. Minor — does not affect behavior.
- `_build_absent_context_entry` retains `context_window_absent: True` and the `"[WARN] context_window absent"` string even though the underlying cause is now "transcript unavailable." This is intentional for backwards compatibility (noted in the PR and docstring), but it's mildly misleading to a new reader of the log format.

These are genuine nits, not blockers.

---

**VERDICT: PASS WITH NITS**

The fix is correct, well-tested, and cleanly separates concerns. The two nits (unused `tool_name` parameter in `_read_transcript_usage`, and the slightly misleading `context_window_absent` field name in WARN entries) are real but do not affect correctness or production behavior.

Issue #1791, opened 2026-04-24
bug(lobstertalk-ssh-watcher): no self-echo filter — outbound messages re-enter inbox as inbound


sayhar:
## Problem

The `lobstertalk-ssh-watcher` scheduled job does not filter self-echoes: messages where `from == to == MY_LOBSTER_NAME`. The bot-talk relay returns a copy of every outbound POST back to the sender's own message feed. The SSH watcher picks these up and routes them to the inbox as if they were inbound messages.

**Confirmed at scale:** Apr 24 2026 ~18:00Z — 1757 messages flooded the inbox, all confirmed self-echoes (triage via bulk inspection). Every message type was an outbound message we sent: PR reviews, polling answers, health updates, test messages, Val Town token shares.

## Prior art

Issue #1345 (closed) fixed this for the HTTP path by replacing a hardcoded `AlbertLobster` filter with a config-driven `BOT_TALK_SELF_USER` check. The SSH watcher does not apply the same filter.

## Impact

- Dispatcher enters a tight loop (inbox never empties, `wait_for_messages` returns immediately on each call)
- Token burn from triage/processing of messages that are by definition already handled
- False positives in message routing if self-echoes trigger any skill or routing rule

## Fix

In the SSH watcher's inbox routing logic, check `from == MY_LOBSTER_NAME` (or the config-driven equivalent). If true, skip the inbox write. This is equivalent to the filter already applied on the HTTP path.

## Severity

High — the flood caused measurable disruption (tight loop, ~53k tokens burned on triage). Likely to recur whenever the relay accumulates a backlog and the cursor resets (related: #1780).

sayhar:
🤖🦞 Lobster: Won't fix — see PR #1837 closure note. Alex decided to remove the SSH watcher from the repo entirely (see new removal issue).

Issue #1780, opened 2026-04-24
incident: lobstertalk-ssh-watcher cursor regression — 30 messages re-delivered (Apr 24 00:00)


sayhar:
## What happened

On 2026-04-24T00:00Z the `lobstertalk-ssh-watcher` scheduled job re-delivered 30 old bot-talk messages (originally from 2026-04-22 18:01 through 2026-04-23 08:10), including test/routing artifacts that should never have reached the inbox.

## Root cause: state regression at 2026-04-23 06:00Z restart

The state file (`~/lobster-workspace/data/lobstertalk-ssh-state.json` or equivalent) regressed to `last_offset: 677634` — a byte position before the Apr 22 18:01 messages — because of a dispatcher restart combined with a race condition in the watcher.

## Timeline

**Apr 22 18:01Z (run 32509290):** 6 new messages arrived on the remote server at almost exactly the moment the watcher ran `SSH stat`. The stat captured the file size *before* those 6 messages were appended. So `last_size` was saved as the *start* of those messages, not after them. State saved: `{last_offset: 677634}`.

**Apr 23 00:00Z (run 3f0e0105):** File had grown. Script read from 677634, routed the 6 Apr 22 messages plus 5 new Apr 23 messages, advanced state correctly to ~684000.

**Apr 23 06:00Z (MCP restart):** MCP server restarted (confirmed by reconciler at 06:00:22Z). Dispatcher restarted at 06:00:56Z. Two duplicate watcher dispatch messages were in the inbox (posted 15ms apart — known dispatch-log duplicate issue). During restart recovery, a watcher subagent ran with stale/corrupt state and reset the state file back to `{last_offset: 677634}`.

**Apr 23 12:00–18:00Z:** Dispatch messages posted but no subagents ran (dispatcher was handling active conversation). No new messages → no re-delivery risk.

**Apr 24 00:00Z (run e4b9fe7c):** New test messages changed the file's mtime. Script started from 677634, read to current size 706060 = 30 messages — all re-delivered.

### Bug 1: Race condition in state saving
When remote messages are appended *while* the watcher is running (between SSH `stat` and SSH `read`), the saved `last_size` reflects the pre-append size. On the next run, the size mismatch triggers a re-read from `last_offset`, which includes messages already routed by the previous run.

**Fix:** Save `last_offset` as `last_offset + actual_bytes_read` (already done), but also save `last_size` as the file size *after* reading (second SSH stat), not the size at the start of the run.

[...]

Issue #1345, opened 2026-04-01
fix(bot-talk): replace hardcoded AlbertLobster filter with config-driven BOT_TALK_SELF_USER


sayhar:
## Problem

The bot-talk poller and realtime forwarder hardcode \`AlbertLobster\` as the "other side" of the conversation. The filter logic is: notify for messages where \`sender == AlbertLobster\` and suppress messages where \`sender == SaharLobster\`.

This is wrong for shared deployments. Each side of the bot-talk channel should only notify for messages **received** by that instance — i.e. messages sent by the other party. The local user's name is not always \`SaharLobster\`; it is already configured in \`config.env\` as \`BOT_TALK_SENDER\`.

## What to fix

1. **\`bot_talk_mirror.py\`** — already reads \`BOT_TALK_SENDER\` from config but hardcodes the constant at module level. No change needed here (it already uses the config value to set \`BOT_TALK_SENDER = "SaharLobster"\` at module load, but that's the outbound side — fine).

2. **Task files** (the LLM instructions for scheduled jobs) — all three files hardcode names:
   - \`~/lobster-workspace/scheduled-jobs/tasks/bot-talk-poller.md\`
   - \`~/lobster-workspace/scheduled-jobs/tasks/bot-talk-poller-fast.md\`
   - \`~/lobster-workspace/scheduled-jobs/tasks/bot-talk-realtime-forwarder.md\`

3. **Template files** in \`~/lobster/scheduled-tasks/templates/\` (if they exist) — update to be config-driven.

## Proposed fix

- Add a \`BOT_TALK_SELF_USER\` config variable (can fall back to \`BOT_TALK_SENDER\` which is already in \`config.env\`).
- The filter logic becomes: **notify for messages where \`sender != BOT_TALK_SELF_USER\`** (received messages), **suppress messages where \`sender == BOT_TALK_SELF_USER\`** (sent by us).
- Fallback chain: \`BOT_TALK_SELF_USER\` env var → \`BOT_TALK_SENDER\` env var → detect from message log (messages from local instance have a consistent sender prefix).
- Update all three task files to use this config-driven approach instead of hardcoded names.
- The \`bot_talk_mirror.py\` module already correctly uses \`BOT_TALK_SENDER\` from config — the fix is to ensure the poller side uses the same config value for filtering.

## Acceptance criteria

- No hardcoded \`AlbertLobster\` or \`SaharLobster\` in the filter logic of any task file
[...]

PR #1794, opened 2026-04-24
fix(#1780,#1791): ssh-watcher self-echo filter + post-read stat + move to repo

sayhar:
## Problem

Two bugs in lobstertalk-ssh-watcher-job.py causing incorrect message delivery:

**Issue #1791 (high severity):** No self-echo filter. The bot-talk relay returns a copy of every outbound POST to the sender's feed. On Apr 24 this caused 1,757 outbound messages to flood the inbox, burning ~53k tokens and creating a tight loop. The HTTP path had this filter (#1345) but the SSH watcher didn't.

**Issue #1780 Bug 1 (low severity):** last_size was saved as the pre-read file size, causing spurious "new content" detection when messages arrived during the read window.

## What changed

- Added scheduled-tasks/lobstertalk-ssh-watcher-job.py to the repo (previously untracked)
- Self-echo filter: MY_LOBSTER_NAME from BOT_TALK_SELF_USER/BOT_TALK_SENDER/LOBSTER_NAME env, skip sender==self in write_to_inbox
- Post-read stat: re-stat after routing to save accurate last_size
- SSH_HOST de-hardcoded: reads BOT_TALK_SSH_HOST env var (already in config.env)

## Test plan
- [ ] Self-echo messages skipped and logged
- [ ] Non-self messages still routed
- [ ] last_size after run matches fresh stat
- [ ] BOT_TALK_SSH_HOST reads from config.env

### DIFF
--- a/scheduled-tasks/lobstertalk-ssh-watcher-job.py
+++ b/scheduled-tasks/lobstertalk-ssh-watcher-job.py
@@ -15,8 +15,6 @@ LOG_FILE = HOME / "lobster-workspace" / "logs" / "lobstertalk.jsonl"
 INBOX_DIR = HOME / "messages" / "inbox"
 SSH_KEY = HOME / ".ssh" / "lobsterbotsownkey"
 SSH_HOST = os.environ.get("BOT_TALK_SSH_HOST", "")
-if not SSH_HOST:
-    raise RuntimeError("BOT_TALK_SSH_HOST env var required (e.g. user@host.example.com)")
 REMOTE_FILE = os.environ.get("BOT_TALK_SSH_REMOTE_FILE", "~/bot-talk/messages.jsonl")
 OWNER_CHAT_ID = 8305714125
 JOB_NAME = "lobstertalk-ssh-watcher"
@@ -117,6 +115,10 @@ def write_task_output(output, status="success"):
 
 
 def main():
+    if not SSH_HOST:
+        write_task_output("BOT_TALK_SSH_HOST env var not set — cannot connect to bot-talk server", "failed")
+        return
+
     is_first_run = not STATE_FILE.exists()
     state = load_state()
     last_offset = state.get("last_offset", 0)
--- a/scheduled-tasks/lobstertalk-ssh-watcher-job.py
+++ b/scheduled-tasks/lobstertalk-ssh-watcher-job.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""LobsterTalk SSH Watcher Job — detects new messages written to shared server's messages.jsonl"""
+import json
+import os
+import uuid
+import time
+import subprocess
+from datetime import datetime, timezone
+from pathlib import Path
+
+RUN_ID = str(uuid.uuid4())[:8]
+HOME = Path.home()
+STATE_FILE = HOME / "lobster-workspace" / "data" / "lobstertalk-ssh-state.json"
+LOG_FILE = HOME / "lobster-workspace" / "logs" / "lobstertalk.jsonl"
+INBOX_DIR = HOME / "messages" / "inbox"
+SSH_KEY = HOME / ".ssh" / "lobsterbotsownkey"
+SSH_HOST = os.environ.get("BOT_TALK_SSH_HOST", "")
+if not SSH_HOST:
+    raise RuntimeError("BOT_TALK_SSH_HOST env var required (e.g. user@host.example.com)")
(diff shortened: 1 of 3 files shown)


sayhar:
## Code Review

### 1. Self-echo filter correctness

The env var lookup chain `BOT_TALK_SELF_USER → BOT_TALK_SENDER → LOBSTER_NAME → "SaharLobster"` is correct and mirrors the HTTP path. In the live config, `LOBSTER_NAME=SaharLobster` is set, so the filter will resolve correctly even if `BOT_TALK_SELF_USER` and `BOT_TALK_SENDER` are absent.

One edge case: the filter is an exact string match (`sender == MY_LOBSTER_NAME`). If the remote relay writes the sender with different casing or with trailing whitespace, the filter silently fails. Not a blocker — this is consistent with the HTTP path — but worth noting if the relay's sender normalization ever changes.

### 2. Post-read stat placement

The re-stat (Step 6) happens after `write_to_inbox` and before `save_state`. This is the correct order.

The failure handling is acceptable but permissive: if the second stat fails (`rc_post != 0`), the code silently falls through and saves state with the pre-read `current_size` value. This restores the original bug for that one run. Given that the first stat already succeeded in the same run, a second stat failure would be unusual (transient network blip), so this is a reasonable tradeoff. Adding a log entry on stat failure would make the behavior visible without changing semantics.

### 3. `raise RuntimeError` on missing `BOT_TALK_SSH_HOST`

The `raise RuntimeError` at module level means the script crashes before any logging path is reachable — no `write_task_output("failed")` call, no log entry. The task runner will capture stderr from the uncaught exception, but the job output DB won't have a structured failure record.

The live config already has `BOT_TALK_SSH_HOST=sharedLobster`, so this won't affect the current deploy. For new installs, the config.env.example does document `BOT_TALK_SSH_HOST`, so operators who follow the example will have it set.

However, a `sys.exit(1)` after a `print`/`write_task_output` call would be more consistent with how SSH failures are handled elsewhere in the same script. Not blocking, but a nit.

### 4. Script location in `scheduled-tasks/`

No concerns with moving it into the repo. The `scheduled-tasks/` directory is the right home per the documented architecture.

**Gap**: The task definition at `~/lobster-workspace/scheduled-jobs/tasks/lobstertalk-ssh-watcher.md` still references `~/lobster-workspace/data/lobstertalk-ssh-watcher-job.py`. This PR adds the script to the repo at `~/lobster/scheduled-tasks/lobstertalk-ssh-watcher-job.py` but does not update the task file or `install.sh` to point there. On the current install, the old copy in `data/` is what actually runs — and it is byte-for-byte identical to the new repo version — so there is no functional regression. But the canonical source is now the repo copy, and the task file will continue to invoke the stale data copy until a future PR updates it.

This is the most significant gap in the PR: the script is in the repo, but nothing wires it up to be the live copy.

### 5. Backwards compatibility

`BOT_TALK_SSH_HOST` is already present in the live config.env, so existing deploys are fine. `BOT_TALK_SSH_REMOTE_FILE` has a sensible default (`~/bot-talk/messages.jsonl`) that matches the hardcoded value it replaces. No breakage for existing installs.

`BOT_TALK_SELF_USER` is not in the live config.env and is not in `config.env.example` (which only shows `BOT_TALK_SSH_HOST`). The fallback chain still resolves correctly to `LOBSTER_NAME=SaharLobster`, so this is fine.

### 6. Nits

- The `last_offset` state is advanced by `actual_bytes_read` (computed from decoded UTF-8 length), but the remote file is read with `tail -c +N | head -c M` using raw byte counts. If the file contains multi-byte UTF-8 characters, `len(new_content.encode("utf-8"))` should match the raw bytes transferred, but only if the SSH text transport doesn't re-encode. This is unlikely to matter in practice since bot-talk messages are typically ASCII, but the assumption is implicit.

- Silent pass on malformed JSON lines (`except json.JSONDecodeError: pass`) means corrupt lines advance the offset without being logged. Adding a debug log on decode errors would help diagnosis.

- `OWNER_CHAT_ID = 8305714125` is hardcoded. Other scripts use env vars or state files for this. Not blocking for a single-user install, but worth tracking for future generalization.

---

**Summary**: The two bug fixes are correct. The main gap is that the task definition still points to the old `data/` path, so the repo addition is incomplete — the script is stored in the right place but not yet wired up to run from there. All other issues are minor.

**VERDICT: PASS WITH NITS**

sayhar:
Closing: SSH watcher belongs in the LobsterTalk repo, not here. Moving to sayhar/blue-lobster.

Issue #1626, opened 2026-04-16
feat: prune merged/closed PR project directories to reclaim disk space


sayhar:
🤖🦞 Lobster (ops):

Problem: ~/lobster-workspace/projects/ currently has 80 directories totaling 4.6GB. Most are git clones or worktrees from old PR branches. After a PR is merged or closed, the directory lingers indefinitely. No automated cleanup exists for these full-content directories.

What's needed: A script (to be wired into the daily cleanup cron) that:
1. Iterates over directories in ~/lobster-workspace/projects/
2. For each one, identifies the associated PR branch (by directory name convention)
3. Checks GitHub API for PR status (merged/closed)
4. If merged/closed AND directory is older than 7 days: removes it
5. Logs what was removed

This is separate from the empty-stub pruning in PR #1617 (which cleans up git worktree bookkeeping entries, not full clone directories).

Context: 80 project dirs, mostly old PR clones/worktrees, 90–324MB each. This is the main disk pressure in ~/lobster-workspace/.

sayhar:
## First run results (2026-04-24)

PR [#1795](https://github.com/SiderealPress/Lobster/pull/1795) adds `scripts/prune-pr-worktrees.py`.

First run (dry-run then live, `--age-days 7`):
- **43 stale worktrees removed** (all merged or closed PRs)
- 134 → 91 directories in `~/lobster-workspace/projects/`
- **7.4GB → 4.9GB** (2.5GB freed)
- 0 errors

41 more directories are too young (<7 days) and will be eligible on future runs. 18 have no PR (personal projects like `bisque-chat`, `anastasia-site-v2` — correctly skipped). 28 are not worktrees (full clones, files).

Next step: wire into the nightly cleanup cron (`0 3 * * *` or similar) with `--age-days 7`. Not done in this PR to keep scope small.

aeschylus:
Good progress (PR #1795 removed 43 stale worktrees) but the comment doesn't confirm the cron wiring step is done; keeping open to verify.

aeschylus:
Closing as part of the backlog reset — this appears already addressed by PR #1795 ("feat(#1626): prune-pr-worktrees.py — reclaim disk from stale PR worktrees"), which is merged. Reset bar: if this tracker were deleted right now, who would notice or be blocked? Nobody — the underlying fix already shipped. Re-open if the fix regresses or doesn't fully cover this issue's scope.

PR #1795, opened 2026-04-24
feat(#1626): prune-pr-worktrees.py — reclaim disk from stale PR worktrees

sayhar:
## Problem

\`~/lobster-workspace/projects/\` accumulates git worktrees from every PR. After a PR is merged or closed, the directory lingers indefinitely. On Apr 24 the directory was 134 dirs / 7.4GB with ~100 stale PR worktrees.

## What changed

**scripts/prune-pr-worktrees.py** (new): For each directory in the projects dir:
1. Detects if it's a git worktree (\`.git\` is a file)
2. Gets the current branch name and remote origin
3. Checks GitHub for PR status (merged/closed/open/none)
4. If merged or closed AND dir is ≥ \`--age-days\` old (default: 7): removes via \`git worktree remove --force\`
5. Logs all actions to \`~/lobster-workspace/logs/prune-worktrees.log\`

Supports \`--dry-run\` flag.

**install.sh**: Registers the \`lobster-prune-pr-worktrees\` systemd timer on fresh installs (daily at 03:00 UTC via MCP \`create_scheduled_job\` infrastructure).

**scripts/upgrade.sh — Migration 82**: Same idempotent timer registration for existing installs.

**scheduled-jobs/tasks/prune-pr-worktrees.md** (new): Task definition file for the scheduled job.

**scripts/cleanup-worktrees-audio.sh**: Added \`prune_pr_worktrees()\` call so the existing daily worktree/audio cleanup script also invokes this pruner.

## Results (first run)

- 43 stale worktrees removed
- 134 → 91 directories
- 7.4GB → 4.9GB (2.5GB freed)
- 0 errors

## Test plan
- [ ] \`--dry-run\` shows prunable dirs without deleting
- [ ] Active PRs (OPEN state) are never pruned
- [ ] Dirs younger than \`--age-days\` are skipped
- [ ] Non-worktree dirs (full clones, non-git) are skipped
- [ ] Log file created at expected path
- [ ] systemd timer registered after fresh install / Migration 82


sayhar:
## Code Review

Reviewed `scripts/prune-pr-worktrees.py`.

### Safety — core removal logic

The safety story is good overall:
- `get_pr_state` returns `None` on any `gh` failure → `skipped_no_pr` → never deletes. Safe.
- OPEN state is explicitly skipped.
- Age check runs *before* the PR state check, so a recently-merged dir waits out the grace period.
- Dirs with no PR (branch never got a PR, or PR not yet created) are skipped conservatively.

**Concern:** The fallback `rm -rf` in `remove_worktree` is unconditional — if `git worktree remove --force` fails for any reason, the code immediately falls back to `rm -rf` as long as the path exists (line 109-112). The `--force` flag already handles locked worktrees, but if the failure is for some other reason (e.g., the main repo path is wrong, the worktree DB is corrupted), `rm -rf` will still obliterate the dir. Consider gating the `rm -rf` fallback on the specific error message indicating the worktree reference is already gone, rather than always falling through.

### Error handling

**Bug: `subprocess.TimeoutExpired` is unhandled.** `run()` calls `subprocess.run(..., timeout=30)` but does not catch `subprocess.TimeoutExpired`. If any `gh pr list` or `git` call hangs past 30 seconds, the exception propagates uncaught and aborts the entire script mid-scan. All subsequent dirs go unprocessed and no summary is logged. Fix: wrap the `subprocess.run` in `run()` with `except subprocess.TimeoutExpired` and return `(1, "", "timeout")`.

```python
def run(cmd, cwd=None, timeout=30):
    try:
        result = subprocess.run(cmd, capture_output=True, text=True, timeout=timeout, cwd=cwd)
        return result.returncode, result.stdout.strip(), result.stderr.strip()
    except subprocess.TimeoutExpired:
        return 1, "", "timeout"
```

### Edge cases

All the expected cases are handled correctly: detached HEAD, no remote, non-GitHub remote, non-worktree dirs.

**Nit — `st_mtime` for age is unreliable.** `dir_age_days` uses `path.stat().st_mtime`. Directory mtime reflects the last file created/deleted *inside* the dir. A worktree with recent log writes or temp files will appear young and escape pruning. Using `st_ctime` (inode change time, which reflects `git worktree add` more faithfully on Linux) or the PR close date from the GitHub API would be more robust. This is a minor accuracy issue, not a safety issue.

### Dry-run

All filesystem mutations are inside `remove_worktree`. `dry_run=True` returns early before any mutation. The scan phase (stat, git, gh calls) is read-only. Dry-run is correct.

### `find_main_repo`

Path traversal (`gitdir.parent.parent` + `.name == ".git"` guard) correctly handles the standard `gitdir: /path/.git/worktrees/<name>` layout. This is reliable for all standard git worktrees.

### gh CLI auth in cron

The PR description notes the next step is to wire this into a daily cleanup cron. In a cron environment the `gh` auth context (token or config dir path from `~/.config/gh/`) may not be on PATH or env. Worth verifying that the cron job sources `~/.bashrc` or sets `GH_TOKEN` explicitly before `gh` is called — otherwise all dirs will silently end up as `skipped_no_pr`.

### Nits

- `skipped_no_worktree` is overloaded across four distinct skip reasons (no `.git` file, detached HEAD, no remote, unparseable remote). Individual counters would make the summary log more useful for debugging.
- `rm -rf` via subprocess (line 110) works fine but `shutil.rmtree(str(worktree_path))` would be more idiomatic Python and avoids spawning a shell.

### Summary

The script is safe by default: any failure mode in PR detection leads to a skip, not a delete. The two issues worth fixing before the cron wire-up are the unhandled `TimeoutExpired` exception (can truncate a run silently) and the unconditional `rm -rf` fallback (consider a more targeted gate). The `st_mtime` age issue is low severity.

**VERDICT: PASS WITH NITS**

PR #1808, opened 2026-04-25
fix: use jq for JSON construction in daily-health-check.sh

sayhar:
## Summary

- Rewrites the failure-notification section of `scripts/daily-health-check.sh` to use `jq -n ... --arg` for JSON construction instead of a heredoc with literal variable expansion
- Heredoc approach produced literal newlines inside the JSON string value, causing `json.load()` to throw "Invalid control character" — this made `wait_for_messages` tight-loop and exhaust `--max-turns`
- Root cause of 26 MCP restarts on 2026-04-24/25

## Test plan

- [ ] Trigger a health check failure and confirm the inbox JSON file is valid (`jq . <file>` exits 0)
- [ ] Confirm the failure body contains the expected `\n`-escaped newlines, not literal newlines
- [ ] Verify no WFM tight-loop occurs after the health check runs


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Root cause fix: confirmed correct.** The old code produced a JSON file with literal newline control characters inside a string value (from `$FAIL_LIST` expanding into the heredoc), which caused Python's `json.load()` to throw "Invalid control character". The new code routes all string values through `jq --arg`, which escapes everything properly. The produced JSON is valid and `json.load()` succeeds. This directly fixes the restart-storm root cause.

**New `id` field: good addition.** The new code introduces a `MSG_ID` variable and includes it as the `id` field in the JSON, making each health-check message uniquely identifiable (matching the convention of other inbox message types).

**One cosmetic inconsistency (non-blocking):** The `\n` sequences in the `BODY=` assignment are literal backslash-n characters in bash double-quoted strings — not real newlines. `jq` encodes them as `\\n` in JSON (i.e., they appear as literal `\n` characters in the parsed string), while the actual newlines from the `printf` subshell become proper JSON `\n` escapes. The result is that paragraph separators in the displayed body read as `\n` rather than blank lines. This is cosmetic and does not affect JSON validity or message routing.

**Pre-existing test failures: verified.** 22 test failures exist on both `main` and this branch; none are introduced by this change. The failing tests are unrelated to the health-check script.

**Test plan gap (non-blocking):** The PR test plan is manual-only. There are no automated tests for the health-check inbox message format. Given that this script was the root cause of 26 MCP restarts, a regression test that runs the script in a failure scenario and validates the produced JSON with `jq .` and `python3 -c "import json; json.load(open(...))"` would prevent this class of bug from regressing silently. Not a blocker for merge.

sayhar (commented):
PASS. jq -n --arg correctly escapes embedded newlines in $BODY — invalid JSON bug fixed. Still relevant on main (heredoc approach still in place). No migration needed. Follow-up: 4 other scripts have the same heredoc JSON pattern; tracked in issue #2004 (periodic-self-check.sh is highest risk).

sayhar:
Superseded by #2012 — branch had diverged 60+ files from local-dev. The meaningful fix (jq-based JSON construction) has been cleanly extracted and applied on a fresh branch from main.

PR #1809, opened 2026-04-25
feat(#1807): replace Pub/Sub email pipeline with deterministic 10s Gmail poller

### LINKED ISSUE #1807: feat: replace Gmail Pub/Sub pipeline with deterministic 10s polling
## Problem

Drew's directive: remove all traces of Pub/Sub-based email system requiring GCP steps. Redesign email function to be simple and robust with no third-party changes. Poll every 10s. Zero token spend on empty polls. Every new mail message goes to Lobster inbox. New mail includes replies not sent by Sebastian.

## Current state

- `gmail-poll.py` already exists and already does exactly the right thing — History API, zero tokens on empty polls, injects to inbox, skips self-sent messages
- Timer runs at **5s** (better than required, but directive says 10s)
- Three Pub/Sub-related systemd timers exist that need to be cleaned up:
  - `lobster-gmail-watch-renewal` — calls `gmail.users.watch` every 6 days (Pub/Sub only)
  - `lobster-awp-gmail-token-refresh` — refreshes GMAIL_ACCESS_TOKEN in Vercel for Pub/Sub webhook
  - `lobster-gmail-poll` — the keeper, just needs 5s → 10s

## Plan

- [ ] Write unit tests for `gmail-poll.py` core logic (pure functions: parse_sender, extract_body, should_skip_message, build_inbox_message)
- [ ] Update timer from `0/5` to `0/10`  
- [ ] Disable and remove `lobster-gmail-watch-renewal` timer + script (Pub/Sub only)
- [ ] Disable `lobster-awp-gmail-token-refresh` timer (Vercel GMAIL_ACCESS_TOKEN — Pub/Sub pipeline only)
- [ ] Update `gmail-poll.py` docstring from 5s to 10s
- [ ] Add upgrade migration to remove Pub/Sub timers on fresh installs
- [ ] Close AWP issue #102 (gmail.users.watch setup — no longer needed)
- [ ] File AWP cleanup issue to deprecate `/api/webhooks/gmail` Pub/Sub route

## Constraints

- gmail-poll.py script logic: no changes needed, already correct
- No third-party changes required
- Token refresh for the poller itself is handled by the existing token store (not the Vercel env var job)

aeschylus:
## Problem

The Gmail email system was designed around Google Cloud Pub/Sub push notifications, requiring GCP console setup (topic, subscription, watch renewal every 7 days). The Pub/Sub watch was never activated on this install. Drew's directive: remove all Pub/Sub infrastructure, poll every 10 seconds, zero token spend on empty polls, never miss an email.

## What changed

`gmail-poll.py` already existed as an untracked file running every 5 seconds via systemd. It already did exactly the right thing: Gmail History API, deterministic new-message detection, zero LLM calls, injects to Lobster inbox as `source: gmail`. This PR brings it under version control, updates the interval to 10 seconds, and removes the three Pub/Sub-only jobs that were running alongside it.

**Before:**
```
lobster-gmail-poll.timer          (5s) → gmail-poll.py  [untracked, correct approach]
lobster-gmail-watch-renewal.timer (6d) → gmail-watch-renewal.py  [Pub/Sub, never activated]
lobster-awp-gmail-token-refresh.timer (45m) → refresh_gmail_vercel_token.py  [Pub/Sub Vercel env var]
awp-gmail-pipeline                (5m) → awp_gmail_pipeline.py  [LLM-heavy, Pub/Sub-adjacent]
```

**After:**
```
lobster-gmail-poll.timer          (10s) → gmail-poll.py  [in repo, active]
lobster-gmail-watch-renewal.timer — DISABLED
lobster-awp-gmail-token-refresh.timer — DISABLED
```

The `awp-gmail-pipeline` LLM job was already an MCP scheduled job (not a systemd timer) and is separate; its cleanup is tracked in the awp-gmail-pipeline task file.

## Polling logic (unchanged — already correct)

The poller runs every 10 seconds. On each run:
1. Reads `historyId` from `~/lobster-workspace/data/gmail-poll-state.json`
2. Calls `GET /gmail/v1/users/me/history?startHistoryId={id}&historyTypes=messageAdded&labelId=INBOX`
3. If `history` is empty: saves updated `historyId`, exits — **zero tokens spent, nothing written to inbox**
4. If new messages: fetches full content, skips self-sent + promotional/social/update categories, writes each to `~/messages/inbox/gmail-{id}.json` as `source: gmail`
5. Updates `historyId` state file

On first run (no state file): bootstraps from current `historyId`, exits — no backfill.
On historyId expiry (404): re-initializes from current profile, exits — captures fresh next run.

## Functional patterns used

- Pure functions for all business logic (`parse_sender`, `get_header`, `decode_base64url`, `extract_body`, `should_skip_message`, `build_inbox_message`) — tested in isolation
- Side effects isolated to `load_state`, `save_state`, `write_inbox_message`, `gmail_get`, `acquire_access_token`
- Immutable state updates (`{**state, "last_history_id": ...}`)

## Files changed

| File | Change |
|------|--------|
| `scheduled-tasks/gmail-poll.py` | Added to repo (was untracked). Docstring updated: 5s → 10s. No logic changes. |
| `scheduled-tasks/deprecated/gmail-watch-renewal.py` | Moved from active to deprecated. Pub/Sub watch renewal — timer disabled. |
| `tests/unit/test_scheduled_tasks/test_gmail_poll.py` | 37 new unit tests for all pure functions |
| `scripts/upgrade.sh` | Migration 80: disable Pub/Sub timers on existing installs |
| `scripts/next-migration.txt` | 80 → 81 |

## External actions taken

- Live timer updated: `lobster-gmail-poll.timer` → `*:*:0/10` (was `0/5`)
- Disabled: `lobster-gmail-watch-renewal.timer`, `lobster-awp-gmail-token-refresh.timer`
- AWP#102 closed (gmail.users.watch setup — no longer needed)
- AWP#106 filed: cleanup `/api/webhooks/gmail` Pub/Sub route (low priority, harmless while disabled)

## Tests run

- [x] `uv run pytest tests/unit/test_scheduled_tasks/test_gmail_poll.py -v` — 37 passed, 0 failed
- [x] `uv run pytest tests/unit/test_integrations/test_gmail_client.py tests/unit/test_integrations/test_gmail_token_store.py -v` — 57 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q --tb=no` (excluding 2 pre-existing collection errors) — 2873 passed, 5 pre-existing failures confirmed identical on main, 0 new failures introduced

## Manual test

The timer change (5s → 10s) was applied directly to `/etc/systemd/system/lobster-gmail-poll.timer` and verified:
- `systemctl show lobster-gmail-poll.timer | grep TimersCalendar` → `OnCalendar=*-*-* *:*:00/10`
- Timer is `active`, service fires and completes normally (observed in `systemctl show` ExecStart status=0)
- N/A for user-visible output: this change only affects poll frequency, not message format or routing

The two disabled timers confirmed inactive:
- `systemctl is-active lobster-gmail-watch-renewal.timer` → `inactive`
- `systemctl is-active lobster-awp-gmail-token-refresh.timer` → `inactive`

## Definition of Done

- [x] Integration/manual test run — timer interval verified via systemctl
- [x] User-visible outcome: N/A — poll frequency change is invisible to user; email routing behavior unchanged
- [x] Side-effect audit: disabling timers stops two jobs that were running but had no active effect (Pub/Sub watch was never set up, so watch-renewal was a no-op; token-refresh updated a Vercel env var for an endpoint that was never receiving traffic)
- [x] External service calls: mocked in unit tests; live timer changes scoped to systemctl commands

Closes #1807


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The code is correct, well-structured, and safe to merge. The polling logic is sound, tests are thorough, and the migration is bash-safe. Two minor issues noted below, both non-blocking.

---

## What I checked

### Polling logic — new messages in existing threads

`GET /gmail/v1/users/me/history?historyTypes=messageAdded&labelId=INBOX` returns `messagesAdded` entries for every message added to INBOX — regardless of whether it is a new thread or a reply in an existing thread. `gmail_message_id` is distinct per message; `thread_id` is shared for replies. The code records `gmail_message_id` in the filename (`gmail-{msg_id}.json`) so idempotency works per-message, not per-thread. Replies in existing threads are detected and injected correctly.

### Self-sent skip

`should_skip_message` does a case-insensitive email comparison against `account_email` (from the state file, sourced from the Gmail profile). Correct.

### historyId expiry (>7 days)

On a 404, `poll()` calls `bootstrap_state()` which fetches the current `historyId` from `/users/me/profile`, writes state, and returns. `main()` then calls `save_state(updated_state)` — but `bootstrap_state()` already called `save_state()` internally. This means the state is written twice on a 404 recovery, but both writes are identical (the second is a no-op overwrite). No data loss. The atomic `tmp → rename` pattern in `save_state` makes this safe.

### Race condition (two simultaneous instances)

Systemd `Type=oneshot` with no `--no-block` means the timer will not fire a second instance while the first is still running. Even if two instances somehow ran concurrently:

- `save_state` uses atomic rename — last writer wins for historyId, which is acceptable (the winner will start from a slightly later point; no messages are lost, just re-checked on the next cycle)
- `write_inbox_message` checks `dest.exists()` before writing — file-level idempotency

There is a TOCTOU window between `dest.exists()` and `dest.write_text()` in `write_inbox_message`, but since systemd prevents concurrent runs and the window is tiny, this is not a practical concern. Worth noting, not blocking.

### historyId state management — first run

On first run (no state file), `bootstrap_state()` is called, current `historyId` is saved, and `main()` returns early — no backfill of existing messages. Correct and intentional.

### awp-gmail-token-refresh disabled — token expiry

Confirmed intentional per engineer briefing: the access token stored in Vercel (`GMAIL_ACCESS_TOKEN`) was only used by the Pub/Sub webhook endpoint (`/api/webhooks/gmail`), which is dead traffic. The **Lobster-side** Gmail access token is managed by `get_valid_token()` via the token store (OAuth refresh token flow), which is entirely separate. The poller will continue to work. The Vercel env var will go stale — that is acceptable and tracked in AWP#106.

### Migration 80 bash safety

```bash
for _m80_unit in lobster-gmail-watch-renewal lobster-awp-gmail-token-refresh; do
    if systemctl is-enabled "${_m80_unit}.timer" &>/dev/null; then
        sudo systemctl disable --now "${_m80_unit}.timer" 2>/dev/null && { ... } || warn "..."
    fi
done
```

- Variable names are namespaced (`_m80_unit`) — no collision risk with outer scope
- `systemctl is-enabled` exit code is reliable; `&>/dev/null` suppresses stderr noise for non-existent units
- `disable --now` atomically stops and disables; `2>/dev/null` suppresses "Unit not loaded" on already-stopped units
- `|| warn` means failure to disable is non-fatal — correct for a migration (manual fallback)
- Bash-safe: no unquoted expansions, no glob risks

The loop is correct.

### 10-second timer interval

The engineer confirmed the live timer was updated directly to `*:*:0/10` and verified via `systemctl show`. The PR updates the docstring (5s → 10s) to match. The systemd unit file itself is managed separately by the scheduler infrastructure (not in this diff), so there is no conflict.

---

## Issues found

### Minor (non-blocking): `save_state` called twice on historyId expiry

When `poll()` returns from `bootstrap_state()`, `main()` calls `save_state(updated_state)` again. Both calls write identical content. The double-write is harmless due to the atomic rename, but it is slightly wasteful and could confuse a future reader. Consider either: (a) having `poll()` return a sentinel (`None`) when it has already saved state internally, or (b) removing the internal `save_state` call from `bootstrap_state` and letting the caller always handle persistence. Not blocking this PR.

### Minor (non-blocking): no test coverage for `poll()` and `process_history_item()` orchestration

The 37 tests cover all pure functions thoroughly. The orchestration path (`poll`, `process_history_item`, `bootstrap_state`, `main`) has no unit tests. These are I/O-heavy so mocking is required, but the 404-recovery path and the "already-in-inbox" idempotency branch in `process_history_item` would be good to test. Not blocking — the pure function coverage is the high-value layer, and the orchestration is straightforward.

---

## Tests

```
tests/unit/test_scheduled_tasks/test_gmail_poll.py: 37 passed, 0 failed
```

The 2 pre-existing collection errors in the test suite (`ADMIN_CHAT_ID_REDACTED` NameError in two unrelated test files) are present on `main` and not introduced by this PR.

Issue #1813, opened 2026-04-25
Uncommitted files in ~/lobster/ need cleanup


sayhar:
## Summary

A rebase audit found three files in `~/lobster/` on the local machine that are either untracked or have unstaged changes. They need to be committed to a branch or cleaned up.

### 1. `docker/staging/SETUP.md` — untracked

This file was created during a Docker GitHub authentication fix. It is a setup guide for the staging Docker environment. It should be reviewed and committed to an appropriate branch (probably the Docker/staging branch or a new cleanup branch), or deleted if it is no longer accurate.

### 2. `services/lobster-mcp-local.service` — untracked

A systemd service unit file for the local MCP server. It is unclear why this is untracked — it may have been created locally during service setup and never committed. This file should either be added to the repository (if it belongs in source control) or added to `.gitignore` (if it is a generated/environment-specific file).

### 3. `install.sh` — unstaged changes

`install.sh` has local modifications that were never staged or committed. Given that `install.sh` is a central file also touched by commit b2d57044 (the auto-registration hook commit), these unstaged changes may interact with the rebase conflicts in the affected PRs. The changes need to be reviewed and either staged into a commit or discarded.

## Recommended Actions

1. `git diff install.sh` — review the unstaged changes and decide whether to commit or discard
2. `git add docker/staging/SETUP.md` and commit to an appropriate branch
3. For `services/lobster-mcp-local.service`: determine if this belongs in source control and either commit it or add it to `.gitignore`

aeschylus:
Keeping open — no confirmation the untracked files have been committed or cleaned up.

aeschylus:
Closing as part of the backlog reset — applying the harder bar: "if this were deleted right now, who would actually notice or be blocked?" No named stakeholder (the repo owner or a specific team member) and no live, current production harm is tied to this issue today. Re-open if a concrete, live need surfaces.

PR #1814, opened 2026-04-25
fix(#1813): quarantine unparseable inbox JSON to prevent WFM tight-loop

sayhar:
## Summary

- **Root cause**: When a malformed JSON file lands in `inbox/`, `check_inbox`'s two-pass scan caught the `JSONDecodeError` on the first pass but placed the file with an empty dict `{}` in the scored list. On the second pass, `json.load` raised `JSONDecodeError` again, was swallowed by the outer `except Exception: continue`, and the file was left in `inbox/` untouched. Because `wait_for_messages` calls `handle_check_inbox` immediately when `inbox/` is non-empty, this caused a tight loop — 26 MCP restarts on 2026-04-25 before the file was manually removed.
- **Fix**: In the second-pass re-read (`if not msg:` block), wrap `json.load` in a dedicated `except json.JSONDecodeError` handler that quarantines the file to `failed/` with `_permanently_failed=True` — matching the unrecognized-source quarantine pattern from #1735. Transient OS errors still fall through to `continue` for natural retry.
- **Tests**: Adds `test_malformed_json_quarantine.py` with 5 tests covering the single-file case, mixed valid+malformed inbox, the no-tight-loop regression, required metadata fields, and empty-file edge case.

## Changed files

- `src/mcp/inbox_server.py` — ~32 lines added in the second-pass re-read block
- `tests/unit/test_mcp_server/test_malformed_json_quarantine.py` — new, 5 tests (all passing)

## Test plan

- [x] `uv run pytest tests/unit/test_mcp_server/test_malformed_json_quarantine.py -v` — 5 passed
- [x] `uv run pytest tests/unit/test_mcp_server/test_unrecognized_source_quarantine.py -v` — 8 passed (no regressions)

### DIFF
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -4680,9 +4680,38 @@ async def handle_check_inbox(args: dict) -> list[TextContent]:
         for _key, f, msg in sorted(_scored, key=lambda x: x[0]):
             try:
                 if not msg:
-                    # Re-read files that failed the first pass (e.g. transient lock)
-                    with open(f) as fp:  # type: ignore[arg-type]
-                        msg = json.load(fp)
+                    # Re-read files that failed the first pass (e.g. transient lock).
+                    # If the file is genuinely unparseable (JSONDecodeError), quarantine
+                    # it to failed/ immediately so it cannot cause a WFM tight-loop
+                    # (issue #1813): wait_for_messages sees the file on every call and
+                    # returns immediately, triggering rapid MCP restarts all day.
+                    try:
+                        with open(f) as fp:  # type: ignore[arg-type]
+                            msg = json.load(fp)
+                    except json.JSONDecodeError as _jde:
+                        error_msg = (
+                            f"Unparseable JSON in inbox file — quarantined to failed/ "
+                            f"to prevent WFM tight-loop. Parse error: {_jde}"
+                        )
+                        log.error(f"check_inbox: quarantining {f.name}: {error_msg}")  # type: ignore[union-attr]
+                        try:
+                            quarantine_payload = {
+                                "_permanently_failed": True,
+                                "_last_error": error_msg,
+                                "_last_failed_at": datetime.now(timezone.utc).isoformat(),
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Fix is correct and well-scoped. The root cause analysis matches the code, all 5 new tests pass, and the quarantine pattern mirrors the existing `#1735` unrecognized-source approach cleanly.

## What I verified

**Test run:** 5/5 new tests pass. Pre-existing failures: 3 (same on `main` and this branch — `TestMarkProcessed` in `test_message_tools.py`). No regressions introduced.

**Fix logic:** The `json.JSONDecodeError` catch is placed precisely at the second-pass re-read. Transient `OSError` (e.g. file deleted between passes) falls through to the outer `except Exception: continue` as intended — no over-quarantine of transient failures.

**`continue` placement:** The `continue` at the end of the `JSONDecodeError` handler is inside the outer `except` scope and exits the per-file loop iteration correctly. Even if the inner quarantine write fails (wrapped in its own `except Exception`), the `continue` still fires.

**Quarantine payload:** Uses a fresh dict with only `_permanently_failed`, `_last_error`, `_last_failed_at`, `_original_filename` — preserving the original file content is not possible here (it's unparseable), and the approach matches the `#1735` pattern.

## One nit (non-blocking, pre-existing)

The `not msg` guard on line 4682 uses Python truthiness. A valid JSON file containing only `{}` would pass the first-pass parse but append `(key, f, {})` to `_scored`. On the second pass `not msg` is `True`, so it re-reads the file, gets `{}` again, and falls through to the source check (where `msg_source = ""` so no quarantine fires). This is pre-existing behavior and harmless in practice — the file just gets silently skipped — but it could cause confusing logging in a future change. Not a concern for this PR.

## Verdict

The fix directly prevents the tight-loop storm. Tests are end-to-end (real file I/O), cover the stated edge cases, and are reachable (single class, no shadowing). Safe to merge.

sayhar:

Issue #1743, opened 2026-04-23
Case study: two post-compaction dispatcher failures (2026-04-23)


sayhar:
## Overview

On 2026-04-23, two distinct failures occurred immediately following context compaction. Both stem from the same root: the dispatcher failed to fully re-orient itself after compaction and relied on the compaction summary rather than authoritative memory sources. This issue documents both failures as a case study and proposes concrete fixes.

---

### What happened

After context compaction, the dispatcher processed two Telegram messages from Sahar ("okay what next needs my attention", "what") but wrote its responses as plain text output in the Claude Code window instead of calling `send_reply`. Sahar never saw the responses on Telegram.

### Root cause

Post-compaction disorientation. The dispatcher behaved as if it was running in a direct Claude Code session rather than as a message dispatcher communicating via Telegram. The compaction summary did not sufficiently reinforce that all user-facing responses must go through `send_reply`, and the dispatcher failed to reconstruct this constraint from the bootup files it re-read.

### Proposed fix: PostToolUse hook on `mark_processed`

Add a `PostToolUse` hook that fires after every `mark_processed` call and checks:

1. Was the message being marked processed sourced from a real Telegram user (not system/subagent)?
2. Was `send_reply` called for that `message_id` since it was claimed (i.e., since `mark_processing` was called)?

If condition 1 is true and condition 2 is false, the hook should emit a loud warning — or block with an error — to alert the dispatcher that it processed a user message without replying via the messaging layer.

The MCP server already tracks per-message state (claimed, processing, processed). It needs to additionally track whether `send_reply` was called for a given `message_id` between `mark_processing` and `mark_processed`. This is a purely server-side change; no dispatcher logic change is required beyond enabling the hook.

---

### What happened

The compaction summary framed PR #1685 as "step 1 of the DB-first migration" (from issue #1738). The dispatcher read that and immediately surfaced PR #1685 to Sahar as "next up" — skipping older, higher-priority pending items (old research threads 4–7, PR #1687 design decision, etc.). Sahar had previously deferred PR #1685.

### Root cause

The dispatcher reconstructed the work queue from the compaction summary's framing rather than from `handoff.md` and the session notes. The compact-catchup summary described PR #1685 in a "what we were doing" context, but the dispatcher treated that framing as a directive for what to do next. Deferred items were not visible in the summary, so they were silently dropped from the queue.

### Proposed fix: Mandatory queue reconstruction before any user-facing action

[...]

PR #1819, opened 2026-04-25
feat(#1751): implement catchup-gate Option B — DB-based dispatcher gate during catchup

### LINKED ISSUE #1751: design: make compact-catchup synchronous (blocking) post-compaction
## Background

Voice message from Sahar (msg 26468, Apr 21) proposed making compact-catchup synchronous after compaction — i.e., the dispatcher should block on catchup completion rather than spawning it as a fire-and-forget background task.

This was never formally answered or decided.

## Current Behavior

compact-catchup is spawned as a background subagent after compaction. The dispatcher resumes its loop immediately, and may process user messages before catchup has finished loading context.

## Proposed Design

Post-compaction, the dispatcher runs compact-catchup synchronously (or treats the catchup result as a gate before processing new user messages). This ensures the dispatcher is fully re-oriented before taking user-facing actions.

## Tradeoffs to Evaluate

- **Pro**: Eliminates race where dispatcher takes action before re-orientation is complete
- **Pro**: Simpler mental model — compaction → catchup → ready
- **Con**: Increases latency for first post-compaction response
- **Con**: If catchup fails, dispatcher is stuck (need fallback)
- **Con**: Synchronous catchup may conflict with hibernation or next-message arrival handling

## Related

- #1630 (compact-catchup spawn not gated)
- #1743 (post-compaction dispatcher failures case study)
- #1738 (DB-first architecture — may simplify this via persistent state)

## Status

Design proposal open — needs discussion before implementation.

sayhar:
## Problem

After a restart or compaction, the dispatcher processes messages before compact-catchup or startup-catchup finishes restoring context. Issue #1743 documented two concrete failures caused by this race: replies sent to the wrong channel and incorrect queue prioritization. The root cause is that the dispatcher takes user-visible action before re-orientation is complete.

## What this PR does

Adds `hooks/catchup-gate.py` — a PreToolUse hook that mechanically blocks the dispatcher from taking substantive action while a catchup session is running.

**Option B design (DB-based, no flag file):** The hook queries `agent_sessions.db` directly to check for a running `startup-catchup` or `compact-catchup%` session. If one was spawned within the last 120 seconds, the hook exits with code 2, injecting "Catching up — give me 90 seconds." into the model's context.

```
Before:  spawn catchup → dispatcher immediately resumes loop → takes user-facing action
After:   spawn catchup → PreToolUse hook checks DB → blocks substantive tools for up to 120s
                          ↓ allow-list passes through loop-management tools
                          ↓ catchup completes → DB row status changes → hook stops blocking
```

The 120-second window is a safety escape valve: if catchup crashes or never updates its status, the gate stops blocking automatically. No deadlock possible.

## Allow-list (always passes through)

- `wait_for_messages`, `check_inbox` — dispatcher can still drain inbox
- `mark_processing`, `mark_processed`, `mark_failed` — message lifecycle
- `send_reply` — dispatcher can ack messages (e.g. "Catching up, give me a moment")
- `claim_and_ack` — message claiming

All other tools (Agent, Bash, write_result, create_task, etc.) are blocked.

## Fail-open policy

If the DB file is missing, locked, or corrupt: `exit(0)`. This hook must never block the dispatcher due to its own infrastructure failures.

## Functional patterns

Pure functions throughout: `_catchup_is_active()` is a pure DB read returning a bool, `ALWAYS_ALLOWED_TOOLS` is an immutable frozenset, timestamp parsing is isolated with explicit error handling. No mutable state in the hook process.

## Settings.json

Migration 82 in `scripts/upgrade.sh` updates existing installs: replaces the old flag-file-guarded command (`test ! -f .../catchup-pending || python3 ...`) with a direct Python invocation. The flag-file approach was from an earlier design iteration; Option B queries the DB directly so the guard is unnecessary.

`install.sh` updated to install the hook with direct invocation from the start.

## Areas to review closely

1. **The 120-second window** — is this the right value? The spec says 120s. Catchup typically runs 10-15 min, so the gate only covers the initial window. After 120s, catchup continues running but the gate stops blocking. This is intentional (escape valve) but reviewers should confirm this matches the intended behavior.

2. **`send_reply` is unconditionally on the allow-list** — the previous sentinel-file version only allowed ack-containing messages. The task spec says send_reply is always allowed. If this is too permissive, the content-filtering approach from the old implementation can be restored.

3. **Most-recent-row semantics** — when multiple matching sessions exist in the DB, the hook checks only the most recent (ORDER BY spawned_at DESC LIMIT 1). This means a fresh re-spawn after a stale session will correctly trigger the gate.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_catchup_gate.py -v` — 30 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 305 passed, 6 failed (all 6 pre-existing failures on main, unrelated to this PR)
- [ ] Live dispatcher test — not run; requires a restart + compaction cycle. The hook is designed fail-open so deploying without a live test is safe.

## Manual test

Not applicable for the hook logic itself (pure DB query + exit code). The settings.json migration was validated by running the jq command against the live settings and confirming the hook command was correctly updated:

```bash
jq '...' ~/.claude/settings.json | python3 -c "print catchup commands"
# Output: ['python3 /home/lobster/lobster/hooks/catchup-gate.py']
```

No Telegram output generated by this change.

## Definition of Done

- [x] Integration/manual test run — N/A for hook logic (pure function, DB query, no external services)
- [x] User-visible outcome verified — not applicable; this is an internal gate with no direct user output
- [x] Side-effect audit complete: hook only reads DB (never writes), exits 0 or 2, appends to log file
- [x] External service calls: none — only local SQLite read

Closes #1751

### DIFF
--- a/hooks/catchup-gate.py
+++ b/hooks/catchup-gate.py
@@ -1,0 +1,214 @@
+#!/usr/bin/env python3
+"""
+PreToolUse gate: while compact-catchup or startup-catchup is in-flight,
+block substantive dispatcher tool calls and redirect to wait_for_messages.
+
+## Design (Option B)
+
+Queries agent_sessions.db directly to determine whether a catchup session is
+still running. No flag file required — reuses the existing IPC mechanism.
+
+A session is "active" when:
+  1. task_id is 'startup-catchup' OR starts with 'compact-catchup'
+  2. status = 'running'
+  3. spawned_at is within CATCHUP_WINDOW_SECONDS (120 s) of now
+
+The 120-second window prevents this hook from blocking indefinitely if a
+catchup session record is never updated to a terminal status (e.g. the
+subagent crashed before calling write_result). After 120 seconds the hook
+stops gating — the system degrades gracefully rather than deadlocking.
+
+## DB query
+
+    SELECT spawned_at
+    FROM agent_sessions
+    WHERE (task_id = 'startup-catchup' OR task_id LIKE 'compact-catchup%')
+      AND status = 'running'
+    ORDER BY spawned_at DESC
+    LIMIT 1
+
+Only the most recent matching row is checked (ORDER BY spawned_at DESC LIMIT 1).
+
+## Fail-open policy
+
+If the DB file is missing, locked, corrupt, or the query fails for any other
+reason: exit(0). This hook must never block the dispatcher due to its own
+infrastructure failures.
+
+## Allow-list (always passes through, even during active catchup)
+
+  - mcp__lobster-inbox__wait_for_messages
+  - mcp__lobster-inbox__check_inbox
+  - mcp__lobster-inbox__mark_processing
+  - mcp__lobster-inbox__mark_processed
+  - mcp__lobster-inbox__mark_failed
+  - mcp__lobster-inbox__send_reply
+  - mcp__lobster-inbox__claim_and_ack
+
+## Session discrimination
+
+Only the dispatcher session is gated. Subagents are identified by the presence
+of the agent_id field (absent from dispatcher PreToolUse payloads), which is
+the fast-path exit. The full layered detection from session_role.py is then
+applied: state files → process-tree walk.
+
+## Settings.json configuration
+
+Add to hooks → PreToolUse:
+
+    {
+      "matcher": "",
+      "hooks": [
+        {
+          "type": "command",
+          "command": "python3 $HOME/lobster/hooks/catchup-gate.py",
+          "timeout": 5
+        }
+      ]
+    }
+"""
+
+import json
+import os
+import sqlite3
+import sys
+import time
+from pathlib import Path
+
+# Make hooks/ importable regardless of cwd.
+sys.path.insert(0, str(Path(__file__).parent))
+
+from session_role import is_dispatcher_session  # noqa: E402 — after sys.path insertion
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+# DB path: check canonical workspace location.
+_WORKSPACE = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
+AGENT_SESSIONS_DB = _WORKSPACE / "data" / "agent_sessions.db"
+
+# A catchup session must have been spawned within this window to trigger the gate.
+CATCHUP_WINDOW_SECONDS = 120
+
+LOG_FILE = _WORKSPACE / "logs" / "catchup-gate.log"
+
+# SQL query: most recent running catchup session, most-recent-first.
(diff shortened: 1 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with minor findings

## Verdict: PASS

The core logic is correct and well-structured. Two issues worth noting before merging.

---

## Finding 1: Block message says "90 seconds" but the gate window is 120 seconds (non-blocking nit)

In `hooks/catchup-gate.py`, the message the model sees when a tool is blocked says "give me 90 seconds" but the actual timeout window is 120 seconds. The model is told to wait less time than the gate will actually hold. In practice the model would retry after 90s and continue being blocked for another 30s — this is probably fine operationally but is misleading and easy to fix.

**Suggested fix:** Change the message to say "120 seconds" (or "2 minutes") to match the actual `CATCHUP_WINDOW_SECONDS` constant.

---

## Finding 2: Misleading test name and docstring (non-blocking nit)

The test `test_only_most_recent_session_checked_stale_allows` (in `TestTaskIdPatterns`) has a name suggesting it verifies that a stale most-recent session results in "allows", but the test actually asserts `exit_code == 2` (blocked). The test data has two sessions: one at -30s (newer, within window) and one at -121s (older, outside window). The newer session wins and the gate correctly blocks — but the test name says "stale_allows" which is the opposite of what it exercises.

The test itself is exercising correct behavior (the most-recent-wins semantics work), but anyone reading the test name will expect an `exit_code == 0` assertion. Consider renaming to `test_most_recent_fresh_session_blocks_even_when_older_stale_session_exists` or similar.

---

## Checks performed

- **Pre-existing failure count**: Confirmed 6 failures in hooks tests (`test_require_background_agent.py` ×5, `test_require_write_result.py` ×1), all pre-existing on main and unrelated to this PR. Engineer's claim is accurate.
- **New tests**: The new test file covers all specified behaviors (subagent fast path, fail-open on DB errors, allow-list enforcement, task_id pattern matching, window boundary semantics). 30 tests, all green.
- **Type/argument correctness**: `sqlite3.connect()` receives a `str` via `str(db_path)` — correct. `calendar.timegm()` receives a `time.struct_time` from `time.strptime()` — correct.
- **Test class shadowing**: No duplicate class names in the test file.
- **120s escape valve**: Confirmed intentional. After 120s the gate stops blocking even if catchup is still running, which is the correct fail-safe behavior.
- **`send_reply` unconditional allow**: No content filtering. This matches the task spec and the PR description explicitly notes the tradeoff.
- **Most-recent-row semantics**: `ORDER BY spawned_at DESC LIMIT 1` is correct. Tested by `test_only_most_recent_session_checked_fresh_blocks`.

PR #1821, opened 2026-04-26
Docker: use git clone for staging container instead of tarball updater

### LINKED ISSUE #1820: Staging tarball updater installs stale MCP server — will re-regress after execute_update
## Problem

The staging Docker container has crashed twice now because `execute_update` installs a stale tarball that overwrites good MCP server files with old ones. After each manual recovery, the next `execute_update` call will re-regress the container to the broken state.

## Root Cause

The tarball used by the staging updater is stale. It packages an old version of the MCP server (approximately 6,348 lines) that is missing:

- HTTP transport code
- `pre-tool-heartbeat.py`
- 9 Python modules that are present in the current production MCP server

When `execute_update` runs, it unpacks this tarball and overwrites the correct production files that were manually copied in to fix the container.

## What Gets Overwritten

- The main MCP server file (reverts to the old ~6,348-line version)
- `pre-tool-heartbeat.py` (deleted — file not present in tarball)
- The 9 Python modules that were added in production but not captured in the tarball build

## What Needs to Be Fixed

The tarball build process must be updated to pull current production MCP server files before packaging. Specifically:

1. Identify where the staging update tarball is built and which source files it pulls from.
2. Ensure the tarball build step runs against the current production codebase (not a stale snapshot).
3. Add a post-build validation step that checks the tarball contains expected files (e.g., `pre-tool-heartbeat.py`, HTTP transport entrypoints) before it is used.
4. Consider adding a file-count or checksum guard in `execute_update` so that applying a smaller/older tarball than the currently installed version is rejected or requires explicit confirmation.

## Impact

Until this is fixed, any `execute_update` call on the staging container will silently regress the MCP server, causing the container to crash or behave incorrectly. This has already happened twice.

sayhar:
## Problem

The staging container's `execute_update` call was a ticking time bomb: it downloaded a stale release tarball and overwrote good MCP server files with an older version that was missing ~3,700 lines of code (HTTP transport, `pre-tool-heartbeat.py`, and 9 Python modules). This caused two manual recovery incidents. Any `execute_update` call would re-regress the container to the broken state.

The root cause was the install strategy: the Dockerfile used `COPY . /home/lobster/lobster` (a snapshot baked at build time), and the update path used `_execute_tarball_update` — which downloaded from GitHub Releases, not from HEAD. The tarball was stale.

## What changed

**`Dockerfile.staging`** — replaces the `COPY . /home/lobster/lobster` step with a `git clone`:

```dockerfile
ARG LOBSTER_BRANCH=main
RUN git clone --branch "${LOBSTER_BRANCH}" \
    https://github.com/SiderealPress/lobster.git \
    /home/lobster/lobster
```

Because `/home/lobster/lobster` now has a `.git` directory, `update_manager.py`'s `_is_git_install()` check returns `True` — and `execute_update` runs `git pull origin main --ff-only` instead of downloading a tarball. The regression path is eliminated structurally.

A `LOBSTER_BRANCH` build arg allows pinning a branch for feature testing; the default is `main`.

**`src/mcp/update_manager.py`** — in `_execute_git_update`, replaces the post-pull `pip install -r requirements.txt` with `uv sync --frozen --no-install-project` (falling back to pip only when `pyproject.toml` is absent). Keeps the lockfile-pinned environment consistent with the project's standard toolchain.

**`docs/DOCKER-STAGING.md`** — documents the git clone approach, the `LOBSTER_BRANCH` build arg, and the mandatory full rebuild step required after this PR merges (the existing cached image still uses COPY).

## Before / after

```
Before:
  docker build  →  COPY snapshot baked into image
  execute_update  →  _execute_tarball_update → downloads stale .tar.gz → overwrites files

After:
  docker build  →  git clone from main (always current)
  execute_update  →  _execute_git_update → git pull --ff-only → no overwrite risk
```

The strategy selection is already handled by `_is_git_install(repo_path)` — a pure function that checks for `.git`. No new branching logic was added.

## Rebuild required after merge

The existing `lobster-staging` image still uses the old COPY-based approach. After merging, run:

```bash
sudo docker compose -f docker/staging/docker-compose.staging.yml down
sudo docker compose -f docker/staging/docker-compose.staging.yml build --no-cache
sudo docker compose -f docker/staging/docker-compose.staging.yml up -d
```

## install.sh audit

This PR does not add new hooks, scripts, or services. No `install.sh` changes required.

## Tests run

- [x] `uv run pytest tests/unit/ -q -k "update_manager"` — 14 passed, 0 failed (all update_manager tests pass)
- [x] `uv run pytest tests/unit/ -q` — 2972 passed, 24 skipped; 2 pre-existing failures unrelated to this change (both also fail on origin/main: `test_dispatcher_agent_no_background_key_exits_2` and a timeout-plugin environment mismatch in worktree venv)
- [ ] Docker build verification — blocked: Docker build requires network access to clone GitHub repo, which is unavailable in this environment. The Dockerfile syntax is correct and mirrors the pattern used by `install.sh`'s own `git clone` steps.

## Manual test

N/A for this specific PR — the change is a build-time substitution (git clone for COPY). The functional behavior of `execute_update` is already covered by the unit tests in `test_update_manager.py`. A live Docker rebuild is required after merge (documented above) to verify end-to-end.

## Definition of Done

- [x] Unit tests pass — update_manager path confirmed via existing test suite
- [ ] Docker build test — cannot run in current environment (no Docker); rebuild documented for post-merge
- [x] User-visible outcome: not applicable — this change affects container internals, not Telegram output
- [x] Side-effect audit: no floods, no shared-state writes, no unintended volume; git clone is read-only from GitHub
- [x] External service: git clone hits GitHub (not mocked in tests, but no production Telegram interaction)

Closes #1820

### DIFF
--- a/docker/staging/Dockerfile.staging
+++ b/docker/staging/Dockerfile.staging
@@ -4,6 +4,16 @@
 # Services are managed by systemd (lobster-router, lobster-claude, lobster-mcp-local),
 # and cron is managed by the cron daemon — no custom entrypoint workarounds.
 #
+# The repo is cloned from GitHub at build time (git clone), not copied from a
+# local tarball snapshot. This means the container always starts with current
+# code, and `execute_update` runs `git pull` (via update_manager.py) rather
+# than downloading a stale release tarball.
+#
+# REBUILD REQUIRED after merge: the existing lobster-staging image still uses
+# the old COPY-based approach. Run:
+#   sudo docker compose -f docker/staging/docker-compose.staging.yml build --no-cache
+#   sudo docker compose -f docker/staging/docker-compose.staging.yml up -d
+#
 # Usage:
 #   sudo docker compose -f docker/staging/docker-compose.staging.yml up -d
 #
@@ -101,8 +111,14 @@ WORKDIR /home/lobster
 RUN curl -fsSL https://astral.sh/uv/install.sh | sh
 ENV PATH="/home/lobster/.local/bin:${PATH}"
 
-# Copy repo (build context = repo root)
-COPY --chown=lobster:lobster . /home/lobster/lobster
+# Clone repo directly from GitHub — always current, git-managed.
+# This mirrors production installs and enables `git pull`-based updates
+# via execute_update (update_manager.py detects .git and uses the git path).
+# Build arg allows pinning a branch for testing (default: main).
+ARG LOBSTER_BRANCH=main
+RUN git clone --branch "${LOBSTER_BRANCH}" \
+    https://github.com/SiderealPress/lobster.git \
+    /home/lobster/lobster
 
 # Install Python dependencies
 WORKDIR /home/lobster/lobster
--- a/docs/DOCKER-STAGING.md
+++ b/docs/DOCKER-STAGING.md
@@ -26,7 +26,7 @@ Host machine
   ~/lobster-config/config.staging.env  <- env vars injected at compose up
 
 Container: lobster-staging (systemd as PID 1)
-  /home/lobster/lobster/              <- repo (baked into image at build time)
+  /home/lobster/lobster/              <- repo cloned from GitHub at build time (git clone)
   /home/lobster/messages/             <- Docker volume (lobster-staging-messages)
   /home/lobster/lobster-workspace/    <- Docker volume (lobster-staging-workspace)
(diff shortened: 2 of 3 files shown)


sayhar:

PR #1827, opened 2026-04-27
fix(#1826): gate fallback rm on git registry absence in prune-pr-worktrees

### LINKED ISSUE #1826: fix(prune-pr-worktrees): handle TimeoutExpired + restrict fallback rm to orphaned directories
## Problem

Two robustness gaps in `scripts/prune-pr-worktrees.py` flagged during review of PR #1795:

### 1. Unhandled TimeoutExpired exception

`subprocess.run(..., timeout=30)` is used for `git` and `gh` CLI calls. If a command hangs past 30 seconds, `subprocess.TimeoutExpired` is raised and — because it is not caught in the scan loop — it silently aborts the entire run mid-scan. All directories after the hung one are skipped with no summary log entry, making the failure invisible.

**Fix:** Catch `TimeoutExpired` in the scan loop, log it as a warning, and `continue` to the next directory. The summary line should still be emitted at the end.

### 2. Unconditional fallback `rm -rf` on worktree removal failure

When `git worktree remove` fails, the script falls back to a direct directory `rm -rf`. This fallback fires unconditionally: it runs even for worktrees that are still registered in git's worktree list (i.e., the removal failure was transient or permission-related, not because the directory is orphaned).

The safe case for the fallback is: the directory exists on disk but `git worktree list` no longer shows it — meaning git has already forgotten the worktree and the directory is purely a filesystem orphan. In that case, direct removal is correct.

**Fix:** Before executing the `rm -rf` fallback, verify the worktree is absent from `git worktree list --porcelain`. Only proceed if it is confirmed absent.

## Scope

Both fixes are isolated to `scripts/prune-pr-worktrees.py`. No schema changes, no service restarts, no install changes needed.

- [ ] Catch `TimeoutExpired` in scan loop; log and continue
- [ ] Gate fallback `rm -rf` on confirmed absence from `git worktree list`
- [ ] Update or add tests to cover both cases

sayhar:
Closes #1826

## What this fixes

`prune-pr-worktrees.py` had two robustness gaps flagged during review of PR #1795. This PR addresses both.

### Fallback rm gating (the substantive fix)

When `git worktree remove --force` failed, the script fell back to `shutil.rmtree` whenever the directory existed — unconditionally. This was unsafe: if the failure was transient (permission error, lock contention) rather than the worktree being a filesystem orphan, the fallback would delete a directory git still knew about, potentially corrupting the working tree.

The fix adds `is_worktree_registered()`, which reads `git worktree list --porcelain` and checks whether the worktree path appears in the registry. The fallback rm now fires only for confirmed orphans (absent from the registry). If the worktree is still registered, `remove_worktree()` returns `False` and logs `[SKIP-RM]` with the reason.

**Before:**
```
git worktree remove --force → fails
  if path.exists():           # unconditional
    shutil.rmtree(path)       # unsafe for non-orphaned worktrees
```

**After:**
```
git worktree remove --force → fails
  check git worktree list --porcelain
  if still registered → SKIP-RM, return False
  if absent (orphaned) → proceed with path guard check, then shutil.rmtree
```

### Timeout handling (already correct, tests added)

The `run()` helper already catches `subprocess.TimeoutExpired`, kills the process, and returns `(1, "", "timeout")` — so a hung `git` or `gh` command does not abort the scan mid-run. The prior nit commits in PR #1795 addressed this. Tests are added to document and enforce this behavior going forward.

## Functional patterns

`is_worktree_registered()` is a pure function: it takes a path and repo, reads `git worktree list --porcelain` output, and returns a bool with no side effects. The registry check is separated from the removal logic, making both independently testable.

## Tests run

- [x] `uv run pytest tests/unit/test_scripts/test_prune_pr_worktrees.py -v` — 7 passed, 0 failed
- [x] `uv run pytest tests/unit/ --ignore=tests/unit/test_hooks/test_require_background_agent.py -q` — 2990 passed, 1 failed (pre-existing on origin/main)
- [x] `uv run python -m py_compile scripts/prune-pr-worktrees.py` — syntax clean

**Pre-existing failures (not introduced by this PR, confirmed on origin/main):**
- `test_require_background_agent.py::TestDispatcherSynchronousAgent::test_dispatcher_agent_no_background_key_exits_2`
- `test_require_write_result.py::TestDispatcherExemption::test_dispatcher_stop_hook_exits_0`

## Manual test

N/A — internal script logic only. No external services, systemd units, or cron entries are affected.

## Definition of Done
- [x] Integration/manual test — N/A (internal script logic only)
- [x] User-visible outcome — N/A (no user-visible output from this change)
- [x] Side-effect audit: no floods, no shared state writes, no unintended volume
- [x] External service calls: all mocked in tests via patched `run()`

### DIFF
--- a/scripts/prune-pr-worktrees.py
+++ b/scripts/prune-pr-worktrees.py
@@ -105,6 +105,25 @@ def dir_age_days(path: Path) -> float:
     return age.total_seconds() / 86400
 
 
+def is_worktree_registered(main_repo: Path, worktree_path: Path) -> bool:
+    """Return True if git still knows about this worktree path.
+
+    Uses `git worktree list --porcelain` to check whether the worktree is
+    still in git's registry.  Returns True (conservative / safe) on error.
+    """
+    rc, out, _ = run(["git", "worktree", "list", "--porcelain"], cwd=main_repo)
+    if rc != 0:
+        # Can't determine — assume registered to avoid unintended rm.
+        return True
+    resolved = str(worktree_path.resolve())
+    for line in out.splitlines():
+        if line.startswith("worktree "):
+            listed_path = line[len("worktree "):].strip()
+            if listed_path == resolved:
+                return True
+    return False
+
+
 def remove_worktree(main_repo: Path, worktree_path: Path, dry_run: bool) -> bool:
     """Remove a git worktree and its directory."""
     if dry_run:
@@ -118,11 +137,17 @@ def remove_worktree(main_repo: Path, worktree_path: Path, dry_run: bool) -> bool
         log(f"  [REMOVED] {worktree_path}")
         return True
 
-    # git worktree remove failed. Only fall back to shutil.rmtree if the
-    # directory still exists (meaning git failed before touching it, e.g.
-    # because the worktree isn't registered). If git already partially removed
-    # it (unlikely with --force), we skip to avoid double-removal.
-    log(f"  [WARN] git worktree remove failed ({err or 'unknown'}) — trying direct removal")
+    # git worktree remove failed.  Before falling back to a direct rm, verify
+    # that git no longer knows about this worktree (i.e. it is an orphaned
+    # filesystem directory that git has already forgotten).  If git still has
+    # the worktree in its registry the removal failure was transient or
+    # permission-related — forcing a direct rm could corrupt the working tree.
+    log(f"  [WARN] git worktree remove failed ({err or 'unknown'}) — checking registry")
+    if is_worktree_registered(main_repo, worktree_path):
+        log(f"  [SKIP-RM] {worktree_path} is still in git's worktree registry — not safe to rm directly")
+        return False
+
+    log(f"  [WARN] {worktree_path} absent from git registry — attempting direct removal")
     if worktree_path.exists():
         # Safety check: only delete paths that live under the expected projects
         # dir to prevent accidental data loss if path resolution goes wrong.
--- a/tests/unit/test_scripts/test_prune_pr_worktrees.py
+++ b/tests/unit/test_scripts/test_prune_pr_worktrees.py
@@ -135,6 +135,12 @@ class TestTimeoutDoesNotAbortScan:
         timeouts from true 'no PR found' cases.
         Spec: issue #1826 — a timeout log entry must appear in the log, not just
         a silent 'no PR' skip.
+
+        This test exercises the real run() implementation so that the [TIMEOUT]
+        log line emitted inside run() is genuinely produced, not asserted against
+        dead air from a fully-patched run().  subprocess.Popen is mocked at the
+        boundary: git commands succeed normally; the gh pr list call raises
+        TimeoutExpired to simulate a real subprocess timeout.
         """
         dir_a = _fake_worktree_dir(tmp_path, "timeout-dir")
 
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: The production logic is correct and the fallback rm gating fix is sound, but one test has a false-positive assertion that does not actually validate the behavior it claims to spec.

## Production logic: PASS

**`is_worktree_registered`** is correct. It runs `git worktree list --porcelain`, compares resolved paths, and returns `True` on error (conservative / safe). No issues.

**`remove_worktree` fallback flow** ordering is correct. The registry check gates the fallback before the path guard — so a still-registered worktree can never reach `shutil.rmtree`. The path guard remains as defense-in-depth for orphaned directories. Both layers are preserved.

**Error handling**: git command failure in `is_worktree_registered` returns `True` (skip rm), which is the safe/conservative choice. Correct.

## Test suite: NEEDS-WORK

### False positive: `test_timeout_counted_separately_from_no_pr`

This test claims to verify that when `gh pr list` times out, the scan logs a timeout entry (not just a silent "no PR" skip). The assertion is:

```
timeout_lines = [m for m in log_messages if "timeout" in m.lower() or "TIMEOUT" in m]
assert len(timeout_lines) >= 1
```

The test patches `mod.run` with a fake that returns `(1, "", "timeout")` directly. Because `run()` itself is patched, the `[TIMEOUT]` log line inside the real `run()` is never emitted. In `main()`, when `get_pr_state` returns `None` (rc != 0), the script does `skipped_no_pr += 1; continue` — no log message is written for that directory. The summary line contains no "timeout" text.

I confirmed this manually: running the exact test body as a standalone script produces `timeout_lines = []`, so the assertion would fail. Yet pytest reports PASS. This is a false positive — the test is structurally passing for a reason I cannot fully explain without deeper pytest introspection, but the core assertion is not testing what it claims. If the script were changed to add a `[TIMEOUT]` log message in `main()` for this case (which is actually what the spec demands), then remove it, this test would not catch the regression.

**What the test should do**: either (a) not patch `mod.run` and instead patch `subprocess.Popen` to raise `TimeoutExpired`, so the real `[TIMEOUT]` log line fires, or (b) patch `mod.run` and also check the summary counters (e.g., `"no PR"` count increases, which IS logged in the summary), rather than checking for a "timeout" string that never appears in this code path.

### Pre-existing failure count

Engineer reported 2 pre-existing failures. Running the suite with their ignore flag (`--ignore=tests/unit/test_hooks/test_require_background_agent.py`) shows 1 failure (in `test_require_write_result.py`). Running without the ignore shows 6 total failures across two hook test files. The PR did not introduce any of these — confirmed by `git diff main -- tests/unit/test_hooks/`. Minor discrepancy in the count, not a PR issue.

### Other tests: PASS

`test_fallback_does_not_rm_when_worktree_still_registered`, `test_fallback_rm_fires_when_worktree_absent_from_registry`, and `test_fallback_rm_still_blocked_by_path_guard_even_for_orphaned_dir` correctly exercise the registry check and validate both the positive and negative cases. The `shutil.rmtree` patch via `shutil_mod` is valid because the lazy `import shutil` inside `remove_worktree` resolves to the same module object. The timeout/run tests for `TestRunTimeoutHandling` are correct.

## Summary

Fix the `test_timeout_counted_separately_from_no_pr` false positive before merging. The production change is clean and safe.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Verified

**All 7 tests pass** — confirmed by running the test suite directly on the PR branch.

**Pre-existing failure count confirmed** — 1 failure (test_dispatcher_stop_hook_exits_0), 2990 passed. Matches the PR claim exactly.

## Specific checks

**make_popen_mock kill+reap sequence** — The double `side_effect` assignment is correct. The list assignment `[TimeoutExpired, ("", "")]` overwrites the earlier single-exception assignment, so the mock delivers: first `communicate()` call raises, second call (reap) returns `("", "")`. This correctly mirrors the production `run()` sequence: raise → `kill()` → `communicate()`.

**shutil.rmtree patching** — The production code does a local `import shutil` inside the function body, then calls `shutil.rmtree(...)`. The tests patch `shutil_mod.rmtree` via `patch.object`. This works correctly: Python's `import shutil` returns the cached module from `sys.modules`, so patching the attribute on that module object intercepts the call even from a locally-scoped import.

**Test structure** — No duplicate class names, no unreachable test classes. The two timeout tests have a clean division of responsibility: `test_scan_continues_after_gh_pr_list_timeout` tests scan continuity by patching `run()` with a sentinel-returning fake; `test_timeout_counted_separately_from_no_pr` drives the real `run()` via a Popen mock to confirm the `[TIMEOUT]` log line is genuinely emitted. Both exercises are necessary and neither duplicates the other.

**is_worktree_registered path comparison** — One minor latent fragility: the "still registered" test constructs the porcelain output as `f"worktree {worktree_path}"` (the raw `tmp_path`), while production compares against `worktree_path.resolve()`. On this system `tmp_path` has no symlinks so they match, but if `/tmp` were a symlink the comparison would silently fail and the test would give a false green. Not a correctness issue today, but worth hardening in a follow-up by using `str(worktree_path.resolve())` in the test's porcelain string.

## Summary

The core fix is sound. The fallback-rm safety gate correctly prevents deletion of worktrees still known to git. The rewritten timeout test genuinely exercises the code path it claims to verify. No no-go issues.

sayhar:

PR #1835, opened 2026-04-29
fix(health-check): lower heartbeat threshold to 600s, add compaction-grace suppression, fix STALE_THRESHOLD to 360s

### LINKED ISSUE #1786: Dispatcher thinking freeze — recurring pattern (4 confirmed cases)
## Summary

The dispatcher silently stops calling tools for 14–20 minutes while the Claude process remains alive, with no errors logged. This has now occurred 4 confirmed times across 2026-04-23 and 2026-04-24. It is a distinct failure mode from the inference-gap false positives tracked in #1695 — those are the dispatcher actively working; this is the dispatcher doing nothing.

## Pattern

In every case:
- Claude PID is alive and healthy (confirmed by health check)
- No MCP errors, no OOM, no compaction active
- No tool calls for 14–20 minutes
- `thinking-heartbeat.py` (PostToolUse hook) correctly detects the freeze by going stale
- Health check restarts the dispatcher, which recovers normally with no message loss

## Timeline of all 4 incidents

| # | Date (UTC) | Trigger | Duration | Detected by |
|---|---|---|---|---|
| 1 | 2026-04-23 ~13:16 | Post-compaction re-orientation (COMPACT_GRACE expired) | ~14 min | inbox stale check |
| 2 | 2026-04-23 19:37–19:52 | Processing msg 27842 (short creative naming question) | ~14 min | inbox stale check (6 min threshold) |
| 3 | 2026-04-24 ~02:43–03:00 | ~30 min post-compaction, no active task | ~17 min | inbox stale check |
| 4 | 2026-04-24 14:39–15:00 | After receiving `review-pr-author-gain-1` subagent result (short PASS message) | ~20.5 min | heartbeat stale check (1200s threshold) |

### Incident 4 timeline (most recent, 2026-04-24)

- 13:54:12 — Compaction #30 fired and cleared (34s, no issue)
- 14:32:01 — Health check GREEN; dispatcher actively processing subagent results
- 14:39:04 — Last user-facing reply sent
- 14:39:20 — `review-pr-author-gain-1` subagent write_result received (PASS message, 118 chars, `sent_reply_to_user: true`)
- 14:39:24 — Last dispatcher tool call (mark_processed on the subagent notification)
- 14:39:24–15:00:01 — **Complete silence.** No tool calls. Health checks at 14:44, 14:48, 14:52, 14:56 all show heartbeat incrementally staling. Memory stable at 22–24% throughout.
- 15:00:01 — Heartbeat 1237s stale (threshold 1200s). RED. Restart triggered.
- 15:00:07 — Restart successful. No message loss.

## What we know about the cause

The freeze happens at the boundary between receiving a subagent result and returning to `wait_for_messages`. The dispatcher calls `mark_processed` on the subagent notification, then... nothing. No `wait_for_messages`, no further tool calls.

In incident 4, the triggering subagent result was trivial: a 118-character PASS message with `sent_reply_to_user: true`, meaning the dispatcher had nothing to do — it should have immediately called `wait_for_messages`. It didn't.

This is consistent with a model-level inference hang: Claude enters a reasoning/generation phase and fails to produce a tool call. No external signal (OOM, MCP error, process signal) is visible. The context window size at the time of freeze is unknown — `context_window_absent: true` in all context-monitor.log entries (the hook is not receiving context window data in its payload).

## What does NOT cause it

- Compaction — incident 4 happened 45 minutes post-compaction
- OOM — memory 22–24% throughout the freeze window, no kernel OOM events
- MCP transport failure — no MCP errors in logs; new session starts immediately on restart
- A specific message type — incident 2 was a simple naming question; incident 4 was a trivial PASS notification

## Current mitigation

The PostToolUse heartbeat hook (PR #1403) detects the freeze correctly. The health check restarts the dispatcher. MTTR is ~20 minutes. No message loss in incidents 3 and 4.

## Data gap: context window size unknown

`context-monitor.log` shows `context_window_absent: true` for 100% of entries. The hook is not receiving context window size from the CC payload. This means we cannot test the hypothesis that a near-full context window triggers the freeze. Fixing context-monitor.py to capture context window size would let us correlate freeze onset with context utilization.

## Open questions

1. Is this a Claude model behavior (reasoning loop fails to produce a tool call) or a CC framework issue (tool call produced but not executed)?
2. Is there a correlation with context window size? (Can't answer until context-monitor is fixed)
3. Does it correlate with specific subagent result characteristics (large payloads, many results in succession)?

## Relationship to existing issues

- #1695 — inference gap false positives (opposite failure: dispatcher actively working, not stuck)
- #1612 — 20-min blind window for frozen dispatcher (same problem; this issue is the case study)
- #1392 (closed) — earlier freeze pattern after scheduled job batches (same failure mode, different trigger)

### What WON'T work: PreToolUse blocking hook
A PreToolUse hook fires before each tool call. A frozen dispatcher doesn't call tools, so the hook never fires. This approach cannot detect or prevent a thinking freeze.

### What WOULD help

1. **Reduce heartbeat threshold from 1200s to ~600–900s** — halves MTTR with some false-positive risk. The inference-gap issue (#1695) argues against this, but a combined fix (heartbeat threshold + PreToolUse heartbeat for the inference case) could let us lower the threshold safely.

2. **Add PreToolUse heartbeat hook** — does NOT help with thinking freeze directly, but it does narrow the detection window for the inference-gap case (#1695), which would let us lower the PostToolUse heartbeat threshold without false positives. See #1695 Fix 2.

3. **Fix context-monitor.py** to actually capture context window size — unblocks root cause investigation.

4. **Add a dispatcher-side post-processing circuit breaker** — after `mark_processed` on any subagent result, an explicit `wait_for_messages` call is the next required action. A dispatcher self-check ("did I just call mark_processed? call WFM now") might be expressible as a behavioral rule. This is soft (LLM-level) not hard (hook-level).

sayhar:
## What this fixes

Thinking-freeze incidents (#1786) went undetected for up to 20 minutes because `DISPATCHER_HEARTBEAT_STALE_SECONDS=1200`. The dispatcher stops making tool calls entirely — the heartbeat goes stale — but nothing fires until the 20-minute mark. This PR cuts MTTR to 10 minutes.

The challenge: simply lowering the threshold from 1200s to 600s would cause false restarts during post-compaction catchup runs, which can last up to 12 minutes. The fix requires adding a suppression gate alongside the threshold change.

## Three changes

**1. Lower `DISPATCHER_HEARTBEAT_STALE_SECONDS` from 1200 to 600 (line 101)**

Halves MTTR for thinking-freeze. The compaction-grace gate (change 2) makes this safe.

**2. Add `is_compact_grace_period()` gate for the heartbeat check in `main()` (lines ~1862–1869)**

Post-compaction catchup runs (reading bootup files, processing backlog) can last up to 12 minutes — exceeding the new 600s threshold. The existing `COMPACT_GRACE_SECONDS=900` window already suppresses the inbox-drain check; this PR extends the same gate to the heartbeat check. The heartbeat check was previously ungated because the 1200s threshold "absorbed" catchup naturally. That approach stops working at 600s.

`is_compact_grace_period()` reads `$WORKSPACE_DIR/data/last-compact.ts` (written by `on-compact.py`). The grace window is 15 minutes. The heartbeat check resumes immediately after the grace expires, so a thinking-freeze that starts mid-catchup will be caught at most 15 minutes + 10 minutes = 25 minutes after compaction.

**3. Fix `STALE_THRESHOLD_SECONDS` from 240 to 360 (line 78)**

PR #1634 intended to set this to 360s (6 minutes) but the deployed value was 240s (4 minutes). This corrects that unrelated bug in the same pass.

## Before / after flow

```
Before:
  compaction fires → is_compact_grace_period suppresses inbox drain
                   → heartbeat check runs unconditionally at 1200s threshold
                   → thinking-freeze detected at ~20 min

After:
  compaction fires → is_compact_grace_period suppresses inbox drain
                   → is_compact_grace_period suppresses heartbeat check (15-min window)
  grace expires   → heartbeat check runs at 600s threshold
                   → thinking-freeze detected at ~10 min (outside grace window)
                   → or at most 25 min if freeze starts during catchup
```

## No hook changes needed

`PreToolUse` hook does not fire during thinking-freeze (zero tool calls). The heartbeat file goes stale because there are no PostToolUse events. No hook changes are required.

## Tests run

- [x] `uv run pytest tests/smoke/test_health_check.py -v` — 13 passed (10 existing + 3 new D6 tests for `is_compact_grace_period`), 0 failed
- [x] `bash -n scripts/health-check-v3.sh` — clean syntax check (also verified by D1 test above)

## Manual test

N/A — change is to threshold constants and suppression logic. No external services, no file I/O, no DB schema. The only observable effect is the health check firing sooner on a genuinely frozen dispatcher. Smoke tests verify both the grace period function behavior and the heartbeat threshold via extracted bash fragments.

## Definition of Done

- [x] Integration/manual test run — N/A (pure logic change, no external services)
- [x] User-visible outcome: thinking-freeze MTTR halved (10 min vs 20 min) — not directly testable without a live freeze, but the threshold constant and suppression gate are both verified by tests
- [x] Side-effect audit: no floods, no unscoped writes — change is read-only constants + one new `elif` branch

Closes #1786

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -26,16 +26,17 @@
 #   and trigger a false-positive restart. To prevent this, on-compact.py writes
 #   a compacted_at timestamp to lobster-state.json, and the stale-inbox check
 #   is skipped for COMPACTION_SUPPRESS_SECONDS after that timestamp.
-#   NOTE: Dispatcher liveness check (check_dispatcher_heartbeat) is NOT
-#   suppressed during compaction — the 20-minute threshold covers it naturally.
+#   NOTE: Dispatcher liveness check (check_dispatcher_heartbeat) IS suppressed
+#   during the post-compaction grace period (is_compact_grace_period) to prevent
+#   false restarts during the up-to-12-minute catchup run that follows compaction.
 #
 # Dispatcher liveness (replaces WFM freshness + catchup suppression):
 #   hooks/thinking-heartbeat.py writes a Unix epoch timestamp to
 #   ~/lobster-workspace/logs/dispatcher-heartbeat on every PostToolUse event.
 #   check_dispatcher_heartbeat() reads this single file and checks its age.
-#   The 1200s threshold covers compaction, catchup, and boot without any
-#   suppression logic. The dispatcher no longer needs to call
-#   record-catchup-state.sh to suppress false alarms. See issue #1483.
+#   The 600s threshold (10 min) reduces MTTR for thinking-freeze from 20 min to
+#   10 min. The post-compaction grace period suppression covers catchup runs that
+#   can last up to 12 min. See issue #1483, #1786.
 #
 # Boot grace period:
 #   After any restart (health-check-initiated or manual), the new Claude session
@@ -75,7 +76,7 @@ INBOX_DIR="$MESSAGES_DIR/inbox"
 MAINTENANCE_FLAG="$MESSAGES_DIR/config/lobster-maintenance"
 LOBSTER_STATE_FILE="${LOBSTER_STATE_FILE_OVERRIDE:-$MESSAGES_DIR/config/lobster-state.json}"
 DISPATCHER_PID_FILE="$MESSAGES_DIR/config/dispatcher.pid"
-STALE_THRESHOLD_SECONDS=240          # 4 minutes - RED if any message older (watchdog handles soft recovery at 90s)
+STALE_THRESHOLD_SECONDS=360          # 6 minutes - RED if any message older (watchdog handles soft recovery at 90s)
 YELLOW_THRESHOLD_SECONDS=150         # 2.5 minutes - YELLOW warning
 RESTART_WINDOW_BUFFER_SECONDS=120    # Pre-mark messages within this window of the stale threshold before a restart
 
@@ -98,7 +99,7 @@ HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"   # legacy WFM-touch signa
 # Single file, single integer (Unix epoch seconds). No JSON parsing required.
 # Threshold is generous enough to cover compaction + catchup without suppression.
 DISPATCHER_HEARTBEAT_FILE="${LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-heartbeat}"
-DISPATCHER_HEARTBEAT_STALE_SECONDS=1200   # 20 min — covers compaction (~5m) + catchup (~12m) + margin
+DISPATCHER_HEARTBEAT_STALE_SECONDS=600    # 10 min — covers compaction (~5m) + catchup margin; grace period handles the rest
 
 # WFM-active signal (issue #1713 / #949): inbox_server.py writes this file with
 # a Unix epoch timestamp when wait_for_messages begins blocking and refreshes it
@@ -946,7 +947,7 @@ check_outbox_drain() {
     fi
 }
 
-# Check 6: Dispatcher heartbeat sentinel (issue #1483 simplification)
+# Check 6: Dispatcher heartbeat sentinel (issue #1483 simplification, #1786 tuning)
 #
 # The dispatcher is considered alive if hooks/thinking-heartbeat.py has written
 # to DISPATCHER_HEARTBEAT_FILE within the last DISPATCHER_HEARTBEAT_STALE_SECONDS.
@@ -954,12 +955,10 @@ check_outbox_drain() {
 #
 # This single-file check replaces the previous multi-signal approach
 # (claude-heartbeat file + last_processed_at + last_thinking_at in
-# lobster-state.json). The 20-minute threshold naturally covers:
-#   - Context compaction pause (1-3 minutes with no tool calls)
-#   - Startup catchup subagent (up to 10-12 minutes)
-#   - Boot grace period (60-90 seconds)
-#
-# No suppression logic needed — the threshold does the work.
+# lobster-state.json). Threshold is 600s (10 min) — reduced from 1200s to lower
+# MTTR for thinking-freeze. The post-compaction grace suppression in main() covers
+# catchup runs that can exceed 10 min. Boot grace period (90s) is well within
+# the 600s threshold.
 #
 # Gracefully skips the check if the heartbeat file does not exist (fresh install
 # or first run before the hook has fired).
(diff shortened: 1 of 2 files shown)


sayhar:
## Code Review — PR #1835

**Verdict: NEEDS-WORK** (one stale comment, otherwise clean)

---

### 1. Constant changes (1200→600, 240→360) — LGTM

Both changes are correct in context.

- `DISPATCHER_HEARTBEAT_STALE_SECONDS 1200→600`: The math holds. Worst-case detection time with the new gate is 15 min (grace) + 10 min (threshold) = 25 min, which is worse than the old 20 min only when a freeze coincides with the last few minutes of a catchup window — an acceptable trade-off for the 2× improvement in the common case.
- `STALE_THRESHOLD_SECONDS 240→360`: Confirmed PR #1634 bug fix. The comment (`# 6 minutes`) now matches the value.

---

### 2. `is_compact_grace_period()` gate integration — LGTM with one nit

The `elif is_compact_grace_period` branch slots in cleanly between the transient-lifecycle guard and `check_dispatcher_heartbeat`. The existing function already handles all edge cases (missing file, empty file, non-numeric content → return 1). The stale-inbox path at line ~1607 uses the same function and is unaffected.

**Minor nit (non-blocking):** The log message emitted inside `is_compact_grace_period()` says `"— stale-inbox check suppressed"` (line 532). This function is now called for two separate purposes (inbox suppression and heartbeat suppression), so that message is misleading when it fires from the heartbeat path. Consider updating the message to something neutral like `"— monitoring suppressed"` or splitting the log message out to the call sites.

---

### 3. D6 tests — LGTM

The three new tests (active, expired, missing-file) correctly exercise the three branches of `is_compact_grace_period()`. The test helper correctly points `WORKSPACE_DIR` at `tmp_path` so the function resolves `last-compact.ts` without touching the live workspace. `COMPACT_GRACE_SECONDS=900` in the test matches the script constant. No issues.

---

### 4. install.sh / upgrade.sh — LGTM

No migration needed. The two constant changes are pure in-script edits — no config files, no DB schema, no new directories. `upgrade.sh` calls `git_pull` + `restart_services`, which is sufficient for this change to take effect.

---

### 5. PR description clarity — LGTM

The before/after flow diagram makes the reasoning clear. The "at most 25 min if freeze starts during catchup" worst-case is documented. The `STALE_THRESHOLD_SECONDS` fix is called out as an unrelated bug fix to avoid confusion.

---

### Required fix

**Stale comments (3 locations) — must fix before merge:**

1. `scripts/health-check-v3.sh` line ~540 (inside `is_catchup_active` tombstone comment):
   ```
   # The dispatcher heartbeat threshold (DISPATCHER_HEARTBEAT_STALE_SECONDS = 1200s)
   # covers catchup duration naturally.
   ```
   This now reads `600s` in the constant — the tombstone comment still says `1200s`. Update to `600s`.

2. `scripts/health-check-v3.sh` line ~50 (top-of-file header, Boot grace period section):
   ```
   NOTE: Dispatcher heartbeat check is NOT suppressed during boot grace — the
   20-minute threshold absorbs the 90s boot window naturally.
   ```
   The threshold is now 10 minutes, not 20. Update accordingly. (The body comment at line ~1862 was correctly updated; this header was missed.)

3. `scripts/health-check-v3.sh` line 532 (inside `is_compact_grace_period`):
   ```
   log_info "... — stale-inbox check suppressed"
   ```
   This is now misleading when called from the heartbeat path. Low-severity nit but easy to fix.

Items 1 and 2 are must-fix (they directly contradict the new behavior). Item 3 is a nit.

sayhar:
All three stale-comment issues from the review addressed in commit d0406d14:

1. Tombstone comment (~line 540) updated: `DISPATCHER_HEARTBEAT_STALE_SECONDS = 1200s` → `600s`
2. Boot-grace header (~line 50) updated: "20-minute threshold" → "10-minute threshold"
3. `log_info` inside `is_compact_grace_period()` (~line 532) updated: "stale-inbox check suppressed" → "heartbeat and stale-inbox checks suppressed"

All 13 smoke tests pass.

sayhar (commented):
Verified all three comment fixes from commit d0406d14:

1. Tombstone at line 540: now reads `DISPATCHER_HEARTBEAT_STALE_SECONDS = 600s` ✓
2. Boot-grace header at line 50: now reads "10-minute threshold absorbs the 90s boot window" ✓
3. log_info at line 533: now says "heartbeat and stale-inbox checks suppressed" ✓

All 13 smoke tests pass (test_health_check.py). LGTM.

sayhar:

PR #1841, opened 2026-04-29
fix(systemd): backfill Environment=PATH= in existing Lobster service files (Migration 84)

### LINKED ISSUE #1840: fix: patch existing systemd service files missing Environment=PATH=
## Problem

PR #1838 fixed the code that GENERATES systemd service files, so new jobs created after deploy will have `Environment=PATH=` correctly. But existing service files on disk are already broken — missing `PATH=` — and can fail with exit-127 when `uv` is not found.

Confirmed broken: `lobster-lobstertalk-unified` was verified failing by reviewing journalctl output.

## Affected files (on this install)

System units missing `Environment=PATH=`:
- `/etc/systemd/system/lobster-lobster-local-dev-sync.service`
- `/etc/systemd/system/lobster-lobstertalk-ssh-watcher.service`
- `/etc/systemd/system/lobster-lobstertalk-unified.service`
- `/etc/systemd/system/lobster-mcp-local.service`
- `/etc/systemd/system/lobster-mcp.service`
- `/etc/systemd/system/lobster-wire.service`

Also: service file templates in `services/` for `lobster-mcp-local` and `lobster-mcp` are missing `Environment=PATH=`.

## Fix

1. Patch the six on-disk service files to add `Environment=PATH=` and `Environment=HOME=` before the `ExecStart=` line
2. Reload systemd and restart affected services
3. Fix the templates in `services/` so future installs are correct
4. Add Migration 84 to `upgrade.sh` to backfill `Environment=PATH=` in all Lobster-managed service files on existing installs

## Scope

- [ ] Patch on-disk service files
- [ ] Fix service templates in `services/`
- [ ] Add upgrade.sh Migration 84
- [ ] Open PR

sayhar:
🤖🦞 Lobster (engineer):

## What this fixes

PR #1838 fixed the code that GENERATES systemd service files — new jobs created after that deploy will have `Environment=PATH=` correctly. But existing service files on disk were already broken. This PR fixes three separate bugs that left `lobster-lobstertalk-unified` failing with exit-127 on every timer fire.

**Root cause analysis** (in order of how they were discovered):

1. **Missing `Environment=PATH=`** — all six lobster service files generated before #1838 were missing this directive. Systemd's default PATH omits `/home/lobster/.local/bin`, so `uv` was not found. Fixed by Migration 84.

2. **Broken shebang in `lobstertalk_unified.py`** — the script used `#!/usr/bin/env uv run python3` without the `-S` flag. `/usr/bin/env` treats the entire string `uv run python3` as a single binary name and fails to find it, exiting 127 even after PATH is fixed. Fixed by changing to `#!/usr/bin/env -S uv run python3`.

3. **Missing `EnvironmentFile` in the service unit** — after fixing the shebang, the script ran but immediately raised `RuntimeError: BOT_TALK_URL not set`. All other lobster service files load `lobster-config/config.env` via `EnvironmentFile=`, but `lobster-lobstertalk-unified.service` was missing those directives entirely. Fixed by Migration 85.

## What changed

**On-disk (applied before this PR was opened):**

Six service files patched for missing `Environment=PATH=` (Migration 84). Additionally, `lobster-lobstertalk-unified.service` had its shebang and EnvironmentFile fixed (Migration 85). All changes verified on disk — service now exits 0/SUCCESS.

**In this PR:**

1. **`scripts/upgrade.sh` — Migration 84**: Finds all `/etc/systemd/system/lobster-*.service` files missing `Environment=PATH=`, patches them in place, runs daemon-reload. Idempotent.

2. **`scripts/upgrade.sh` — Migration 85**: Two-part fix for `lobstertalk-unified` specifically:
   - Patches `lobstertalk_unified.py` shebang from `#!/usr/bin/env uv run python3` to `#!/usr/bin/env -S uv run python3`
   - Adds `EnvironmentFile` directives to `lobster-lobstertalk-unified.service` if missing, then runs daemon-reload
   - Both steps are guarded (idempotent — skips if already applied)

3. **`scheduled-tasks/lobstertalk_unified.py`**: Shebang fixed in source.

4. **`services/lobster-mcp-local.service.template`** and **`services/lobster-mcp.service.template`**: Added `Environment=PATH=` and `Environment=HOME=` for fresh installs.

## Before / after flow

```
Before:
  timer fires → systemd exec → /usr/bin/env "uv run python3" → exit 127

After shebang fix only:
  timer fires → /usr/bin/env -S uv run python3 script.py → RuntimeError: BOT_TALK_URL not set → exit 1

After both fixes (current state):
  timer fires → /usr/bin/env -S uv run python3 script.py → loads config.env → runs → exit 0
```

## Functional patterns

Migrations are structured as pure guard/patch loops — the idempotency guard (`grep -q`) is checked before any mutation. Side effects (sed, daemon-reload) only run when needed.

## Tests run

- [x] Migration 85 shebang fix verified idempotent: re-running with `-S` already present causes the `grep -q` guard to skip.
- [x] Migration 85 EnvironmentFile fix verified idempotent: already-patched file with `EnvironmentFile=` causes guard to skip.
- [x] `systemctl restart lobster-lobstertalk-unified.service` — exited 0/SUCCESS, no errors in journal.
- [x] `systemctl cat lobster-lobstertalk-unified` — shows `EnvironmentFile=` lines and `-S` shebang confirmed in script.
- [ ] Unit tests (`uv run pytest`) — shell migration logic is not covered by unit tests; existing suite unaffected.

## Manual test

Applied on disk before this commit. Verified:

```
systemctl status lobster-lobstertalk-unified.service
  → inactive (dead), last exit: status=0/SUCCESS
  → last log: "No new messages. hot_mode=False, consecutive_empty=1"

head -1 /home/lobster/lobster/scheduled-tasks/lobstertalk_unified.py
  → #!/usr/bin/env -S uv run python3

cat /etc/systemd/system/lobster-lobstertalk-unified.service
  → EnvironmentFile= lines present
```

User-visible outcome: `lobstertalk-unified` will no longer fail with exit-127 or RuntimeError on its next timer fire. Not directly observable by user (no Telegram output on startup), but the hourly LobsterTalk message sync will now function.

## Definition of Done

- [x] On-disk service files patched (6 files for PATH via Migration 84, lobstertalk-unified for EnvironmentFile via Migration 85)
- [x] Shebang fixed in source and on disk
- [x] Active service restarted and verified healthy (exit 0/SUCCESS)
- [x] Migrations 84 and 85 are idempotent (guard checks before patching)
- [x] Service templates fixed for fresh installs
- [x] Side-effect audit: sed only modifies files missing the directive; daemon-reload is always safe; no message floods

Closes #1840
Relates to #1838

### DIFF
--- a/scheduled-tasks/lobstertalk_unified.py
+++ b/scheduled-tasks/lobstertalk_unified.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env uv run python3
+#!/usr/bin/env -S uv run python3
 """
 LobsterTalk Unified Job — genericized production version.
 
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2891,6 +2891,32 @@ print(f'prune-pr-worktrees: {result.status}')
         warn "prune-pr-worktrees.py not found at $_m83_script or uv unavailable — skipping Migration 83"
     fi
 
+    # Migration 84: Backfill Environment=PATH= and Environment=HOME= in existing Lobster-managed
+    # system service files. PR #1838 fixed the code that generates new service files, but existing
+    # on-disk files were already missing these directives and fail with exit-127 when uv is not
+    # found on systemd's minimal PATH. This migration patches all Lobster service files in
+    # /etc/systemd/system/ that are missing Environment=PATH= and runs daemon-reload.
+    local _m84_patched=0
+    local _m84_env_path="Environment=PATH=${HOME}/.local/bin:/usr/local/bin:/usr/bin:/bin"
+    local _m84_env_home="Environment=HOME=${HOME}"
+    for _m84_svc in /etc/systemd/system/lobster-*.service; do
+        [ -f "$_m84_svc" ] || continue
+        if grep -q "Environment=PATH=" "$_m84_svc" 2>/dev/null; then
+            continue
+        fi
+        # Patch: insert PATH and HOME lines immediately before ExecStart=
+        if grep -q "^ExecStart=" "$_m84_svc" 2>/dev/null; then
+            sudo sed -i "s|^ExecStart=|${_m84_env_path}\n${_m84_env_home}\nExecStart=|" "$_m84_svc"
+            substep "Added PATH/HOME to $(basename "$_m84_svc")"
+            _m84_patched=$((_m84_patched + 1))
+        fi
+    done
+    if [ "$_m84_patched" -gt 0 ]; then
+        sudo systemctl daemon-reload 2>/dev/null || warn "daemon-reload failed — run manually"
+        success "Migration 84: patched $_m84_patched service file(s) with Environment=PATH= (daemon-reload done)"
+        migrated=$((migrated + _m84_patched))
+    fi
+
     if [ "$migrated" -eq 0 ]; then
         success "No migrations needed"
     else
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -2917,6 +2917,45 @@ print(f'prune-pr-worktrees: {result.status}')
         migrated=$((migrated + _m84_patched))
     fi
 
+    # Migration 85: Fix lobster-lobstertalk-unified.service — two bugs left it silently broken
+    # even after Migration 84 added Environment=PATH=:
+    # 1. The shebang #!/usr/bin/env uv run python3 is missing the -S flag that env needs to
+    #    split multi-word interpreter strings. Without -S, env treats "uv run python3" as a
+    #    single binary name and exits 127.
+    # 2. The service file was missing EnvironmentFile directives, so BOT_TALK_URL and other
+    #    runtime secrets were never loaded, causing a RuntimeError on every execution.
+    # This migration patches the shebang in the script and adds EnvironmentFile lines to
+    # the service file, then runs daemon-reload.
+    local _m85_script="${LOBSTER_DIR}/scheduled-tasks/lobstertalk_unified.py"
+    local _m85_svc="/etc/systemd/system/lobster-lobstertalk-unified.service"
+    local _m85_patched=0
+
+    # Fix 1: shebang — replace #!/usr/bin/env uv run python3 with #!/usr/bin/env -S uv run python3
+    if [ -f "$_m85_script" ]; then
+        if head -1 "$_m85_script" | grep -q "^#!/usr/bin/env uv run python3$"; then
+            sed -i '1s|^#!/usr/bin/env uv run python3$|#!/usr/bin/env -S uv run python3|' "$_m85_script"
+            substep "Fixed shebang in lobstertalk_unified.py (added -S flag)"
+            _m85_patched=$((_m85_patched + 1))
+        fi
+    fi
+
+    # Fix 2: EnvironmentFile — add config file directives if missing
+    if [ -f "$_m85_svc" ]; then
+        if ! grep -q "^EnvironmentFile=" "$_m85_svc" 2>/dev/null; then
+            # Insert EnvironmentFile lines before ExecStart=
+            local _m85_env_files="EnvironmentFile=-${HOME}/lobster/config/config.env\nEnvironmentFile=${HOME}/lobster-config/config.env\nEnvironmentFile=-${HOME}/lobster-config/global.env"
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Review Summary

Migration 84 is correct and safe. The three key concerns from the briefing all check out, plus one minor observation and one nit:

### Verified: sed \n expansion

Tested directly on this system: GNU sed correctly expands `\n` in double-quoted replacement strings, producing genuine newlines. The substitution works as written.

### Verified: \$HOME is the right user's home

`upgrade.sh` is run by the Lobster user (not root). Shell variable expansion of `$HOME` happens before `sudo sed -i` is invoked, so the path baked into service files is `/home/lobster/.local/bin:...` — the correct user home.

### Verified: glob coverage

All Lobster service files in `/etc/systemd/system/` follow the `lobster-*.service` naming convention (including `lobster-lobstertalk-unified.service`, `lobster-lobstertalk-ssh-watcher.service`, etc.). The glob covers every file it should.

### Verified: idempotency

The `grep -q "Environment=PATH="` guard correctly skips already-patched files. Running the migration twice is safe.

### Observation: migrated counter counts files, not migrations

Migration 84 does `migrated=$((migrated + _m84_patched))` instead of `migrated=$((migrated + 1))`. If 6 files are patched the final message says "6 migrations applied" rather than "1 migration applied." This is cosmetic — no functional impact.

### Nit: PATH in migration and new templates omits `~/.cargo/bin`

The PATH baked in by migration 84 and the two new templates is `$HOME/.local/bin:/usr/local/bin:/usr/bin:/bin`. Other service templates (router, observability, etc.) include `$HOME/.cargo/bin`. For `lobster-mcp-local` and `lobster-mcp`, which invoke Python from the venv rather than Rust tooling, this is unlikely to matter. But the inconsistency is worth noting. This matches `lobster-claude.service.template` which also omits `.cargo/bin`, so it follows an existing pattern.

### No issues found

- No duplicate ExecStart= lines in the patched service files
- No risk of applying to non-Lobster service files (glob is specific)
- DRY_RUN skips `run_migrations()` entirely, so no dry-run bypass issue
- The fix correctly targets the right files and uses a safe idempotency guard

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Re-review summary (covers all 3 fixes, supersedes prior review on a60629be)

This review covers the two new commits since the first PASS: the shebang fix in lobstertalk_unified.py and Migration 85.

### Migration 85 — shebang fix: PASS

The grep guard (`^#!/usr/bin/env uv run python3$`) is anchored and exact. It correctly matches the broken shebang and correctly skips the fixed one (`-S` present). The sed pattern `1s|...|...|` targets only line 1. The outer `[ -f "$_m85_script" ]` guard handles missing files. No issues.

Verified: the shebang in `/home/lobster/lobster/scheduled-tasks/lobstertalk_unified.py` is already `#!/usr/bin/env -S uv run python3` — matches the PR source.

### Migration 85 — EnvironmentFile insertion: PASS

Tested the sed command directly on sample inputs:

- Normal file with trailing newline: correct output, EnvironmentFile lines inserted before ExecStart=.
- File with no trailing newline: sed handles this correctly, output is well-formed.
- Already-patched file: `! grep -q "^EnvironmentFile="` guard correctly fires and skips.
- The `\n` literals inside the variable expand correctly in GNU sed — same behavior confirmed for Migration 84's substitution in the prior review.

Verified: the on-disk `/etc/systemd/system/lobster-lobstertalk-unified.service` shows the EnvironmentFile lines and Environment=PATH= in the expected positions. ExecStart= appears exactly once in the file, so no double-insertion risk.

### Service templates (lobster-mcp-local, lobster-mcp): PASS

The `{{HOME}}` placeholder is correctly substituted by the shared template library (`scripts/lib/template.sh`), which was already used for all other templates. Ordering of `Environment=PATH=` and `Environment=HOME=` after EnvironmentFile but before ExecStart is valid systemd syntax. No issues.

### Carried-over observation (cosmetic, no action needed)

Migration 85 does `migrated=$((_m85_patched + migrated))` where `_m85_patched` can be 0, 1, or 2. When both sub-fixes apply, the final "N migrations applied" count shows 2 instead of 1. This mirrors the same cosmetic behavior noted for Migration 84 in the prior review. No functional impact.

### Summary

All three fixes are correct, idempotent, and safe. No edge cases found. No blocking issues.

sayhar:

PR #1842, opened 2026-04-29
fix(scripts): guard LOBSTER_DEV_MODE grep pipeline against pipefail abort

### LINKED ISSUE #1828: post-reminder.sh exits 1 silently when LOBSTER_DEV_MODE is unset in config.env
## Bug: Silent failure under `set -euo pipefail` when `LOBSTER_DEV_MODE` is missing from config

**File:** `scripts/post-reminder.sh`, near the top.

### Root cause

The dev-mode check uses:

```bash
_DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
```

Under `set -euo pipefail`, if `grep` finds no match it returns exit code 1. With `pipefail`, the whole pipeline fails, and `set -e` aborts the script with status 1.

**Result:** any reminder-firing systemd timer that calls this script silently fails. The reminder never reaches the inbox. The dispatcher never gets the ping. The user gets nothing — and there is no error surfaced to the user, the dispatcher, or the inbox.

### Incident

We hit this on 2026-04-27 when an interview-prep accountability nudge scheduled for 10:55 AM EDT failed silently. systemd showed `code=exited, status=1/FAILURE` but the failure was not surfaced anywhere upstream. The user noticed only because they were waiting for the ping.

### Workaround applied

Appended `LOBSTER_DEV_MODE=false` to `lobster-config/config.env`. Fixes the immediate symptom but the underlying script is still fragile.

### Suggested fix

Replace the line with a `pipefail`-tolerant form, e.g.:

```bash
_DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
```

or wrap in `if [ -f ... ]; then ... fi` so absence of the var is non-fatal.

### Additional hardening

- Add a CI / smoke test that runs `post-reminder.sh` against a `config.env` without `LOBSTER_DEV_MODE` and asserts exit 0.
- Check whether `install.sh` writes a default `LOBSTER_DEV_MODE=` line to `config.env` (right now it apparently does not). This is a "new script called from services or cron" reliability gap — the script should not depend on a config key that install does not guarantee is present.

### Related

See companion design issue: MCP-native scheduling primitive — filed alongside this, as motivation.

sayhar:
🤖🦞 Lobster (engineer):

## Problem

Any systemd timer or cron job calling `post-reminder.sh` (or the 6 other scripts that share the same dev-mode check) would silently exit 1 when `config.env` exists but does not contain `LOBSTER_DEV_MODE`. The grep-into-cut pipeline returned exit code 1 (no match), which under `set -euo pipefail` caused the entire script to abort before delivering its reminder to the inbox. The user sees nothing — no reminder, no error, no inbox message.

This caused the 2026-04-27 interview-prep accountability nudge failure described in the issue. systemd logged `code=exited, status=1/FAILURE` but nothing surfaced upstream.

## Root cause

```bash
_DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
```

When `grep` finds no match it returns 1. With `pipefail` enabled, the pipeline exit code propagates to the assignment, and `set -e` aborts the script.

`nightly-consolidation.sh` already had `|| true` on this line; the other 7 scripts did not.

## Fix

Append `|| true` to the pipeline in all 7 affected scripts:

```bash
_DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
```

Absent key → `_DEV_MODE` is empty → the `if` condition is false → script continues normally. Dev-mode suppression still works when the key is present and set to `true` or `1`.

## Scope

All 7 scripts with the vulnerable pattern are patched:
- `scripts/post-reminder.sh` (named in the issue)
- `scripts/inbox-staleness-warn.sh`
- `scripts/daily-update-check.sh`
- `scripts/periodic-self-check.sh`
- `scripts/check-agent-outputs.sh`
- `scripts/daily-health-check.sh`
- `scheduled-tasks/dispatch-job.sh`

`scripts/nightly-consolidation.sh` was already safe.

No other logic is changed. Suppression behavior when `LOBSTER_DEV_MODE=true` is identical.

## Functional pattern

Pure guard: the change is a single `|| true` appended to a shell pipeline — no branching logic added, no new state. The default value of `_DEV_MODE` when absent is now `""` (empty string), which was the intended behavior all along.

## Tests run

- [x] `bash tests/test-post-reminder-pipefail.sh` — 12/12 passed (all 4 scenarios: key absent, key=false, key=true, config file absent)
- [x] `uv run pytest tests/test_messages_db.py tests/test_agent_sessions.py tests/test_bridge_tools.py tests/test_idempotency_column.py tests/test_require_write_result.py tests/test_secret_scanner.py tests/test_user_model.py tests/test_user_model_v2.py tests/unit/test_session_lost_reminder_pid_guard.py --tb=no -q` — 303 passed
- [ ] Full `uv run pytest --tb=no -q` — 40 collection errors (pre-existing on main, unrelated to this change; verified by running same command on main)

## Manual test

N/A — this change is entirely internal to shell scripts. The fix is a one-character addition (`|| true`) to a grep pipeline. No external services, no Telegram output, no systemd changes. The workaround already applied to production (appending `LOBSTER_DEV_MODE=false` to config.env) confirms the root cause was exactly this pipeline.

## Definition of Done

- [x] Unit tests pass with proper isolation (no production hits in automated tests)
- [x] User-visible outcome verified — not applicable; this fix prevents silent failures in cron/systemd jobs, no direct user-visible output to verify
- [x] Side-effect audit complete: no floods, no shared-state writes, no volume changes

Closes #1828

### DIFF
--- a/scheduled-tasks/dispatch-job.sh
+++ b/scheduled-tasks/dispatch-job.sh
@@ -18,7 +18,7 @@ set -e
 # bothered while testing. Real user messages are never affected by this flag.
 _LOBSTER_CONFIG="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}/config.env"
 if [ -f "$_LOBSTER_CONFIG" ]; then
-    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
+    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
     if [ "$_DEV_MODE" = "true" ] || [ "$_DEV_MODE" = "1" ]; then
         exit 0
     fi
--- a/scripts/check-agent-outputs.sh
+++ b/scripts/check-agent-outputs.sh
@@ -27,7 +27,7 @@ set -euo pipefail
 # bothered while testing. Real user messages are never affected by this flag.
 _LOBSTER_CONFIG="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}/config.env"
 if [ -f "$_LOBSTER_CONFIG" ]; then
-    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
+    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
     if [ "$_DEV_MODE" = "true" ] || [ "$_DEV_MODE" = "1" ]; then
         exit 0
     fi
--- a/scripts/daily-health-check.sh
+++ b/scripts/daily-health-check.sh
@@ -15,7 +15,7 @@ set -o pipefail
 # bothered while testing. Real user messages are never affected by this flag.
 _LOBSTER_CONFIG="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}/config.env"
 if [ -f "$_LOBSTER_CONFIG" ]; then
-    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
+    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
     if [ "$_DEV_MODE" = "true" ] || [ "$_DEV_MODE" = "1" ]; then
         exit 0
     fi
--- a/scripts/daily-update-check.sh
+++ b/scripts/daily-update-check.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 # bothered while testing. Real user messages are never affected by this flag.
 _LOBSTER_CONFIG="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}/config.env"
 if [ -f "$_LOBSTER_CONFIG" ]; then
-    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
+    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2 || true)
     if [ "$_DEV_MODE" = "true" ] || [ "$_DEV_MODE" = "1" ]; then
         exit 0
     fi
--- a/scripts/inbox-staleness-warn.sh
+++ b/scripts/inbox-staleness-warn.sh
@@ -21,7 +21,7 @@ set -euo pipefail
 # bothered while testing. Real user messages are never affected by this flag.
 _LOBSTER_CONFIG="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}/config.env"
 if [ -f "$_LOBSTER_CONFIG" ]; then
-    _DEV_MODE=$(grep -m1 '^LOBSTER_DEV_MODE=' "$_LOBSTER_CONFIG" 2>/dev/null | cut -d= -f2)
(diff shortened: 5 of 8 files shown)


sayhar (commented):

## Verdict: PASS

The fix is correct and the bug is real. Verified manually: on main, `post-reminder.sh` exits 1 (non-zero) when `config.env` exists but lacks `LOBSTER_DEV_MODE`. On the PR branch, same scenario exits 0. The `|| true` placement is correct in all 7 scripts — it trails the full `grep ... | cut ...` pipeline, not just grep.

## What I checked

**|| true placement** — correct in all 7 patched scripts. The guard is appended after `cut`, so the entire pipeline is protected:
```sh
_DEV_MODE=$(grep -m1 ... | cut -d= -f2 || true)
```

**Coverage vs. nightly-consolidation.sh** — all 7 scripts that previously lacked the guard now have it. `nightly-consolidation.sh` was already guarded and is unchanged. Complete.

**Tests actually catch the bug** — confirmed by running `post-reminder.sh` directly. Exit code 1 on unfixed code, exit code 0 after fix.

**Pytest suite** — the 200+ collection errors in `uv run pytest` are pre-existing on main and not introduced by this PR.

## Minor notes (non-blocking)

1. **Test coverage gap**: The new test suite covers 4 of 7 patched scripts. `daily-health-check.sh`, `periodic-self-check.sh`, and `dispatch-job.sh` are fixed but not exercised by `test-post-reminder-pipefail.sh`. These scripts use the same one-line pattern as the covered scripts, so the fix is low-risk, but the test name (`test-post-reminder-pipefail.sh`) and the header comment claim "Also covers... daily-health-check.sh" — that comment is inaccurate for those three.

2. **"Tests verified to fail against unfixed code"** — the test script hard-codes `REPO_ROOT` from `${BASH_SOURCE[0]}`, so it cannot be pointed at the main branch by setting an env var. The claim is technically accurate only if the engineer manually patched/unpatched scripts during verification, which is plausible but not reproducible by a reviewer.

Both notes are non-blocking given that the fix itself is correct, the root cause is accurately diagnosed, and the core cases are tested.

sayhar (commented):
test comment from reviewer

sayhar (commented):

## Verification

### 1. REPO_ROOT env-var override
Correctly implemented at line 33 of the test file:
```bash
REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)}"
```
Uses `BASH_SOURCE[0]` (correct for sourced/called scripts), with full env-var override support.

### 2. All 7 scripts covered
Confirmed all 7 scripts have the `|| true` fix:
- `scripts/post-reminder.sh` — 1 occurrence
- `scripts/inbox-staleness-warn.sh` — 1 occurrence
- `scripts/daily-update-check.sh` — 1 occurrence
- `scripts/periodic-self-check.sh` — 1 occurrence
- `scripts/check-agent-outputs.sh` — 1 occurrence (plus 2 in surrounding script logic)
- `scripts/daily-health-check.sh` — 1 occurrence (plus 1 in surrounding logic)
- `scheduled-tasks/dispatch-job.sh` — 1 occurrence (plus 3 in surrounding logic)

All 7 scripts have test coverage in `tests/test-post-reminder-pipefail.sh`.

### 3. New test cases are meaningful
The 6 new tests (2 each for `daily-health-check.sh`, `periodic-self-check.sh`, `dispatch-job.sh`) test actual behavior:
- **daily-health-check.sh (absent)**: verifies the script proceeds past the guard by checking a log file is written — not just "exits 0". This is the right way to detect a pipefail abort.
- **daily-health-check.sh (true)**: exits 0 (suppressed before any checks run).
- **periodic-self-check.sh**: both scenarios test real execution paths using a minimal stub for `agent-status.sh`.
- **dispatch-job.sh**: uses a non-existent timer name to get a clean exit-0 path through the real script logic.

None of the 6 new tests are vacuous.

### 4. All 18 tests pass
```
All 18 tests passed
```
Ran from the worktree at `~/lobster-workspace/projects/fix-issue-1828-post-reminder-pipefail` which is tracking `origin/fix/issue-1828-post-reminder-pipefail`.

### 5. No remaining issues
The fix is minimal and correct — a single `|| true` appended to each vulnerable grep pipeline. No logic changes, no new state, no side effects. The nightly-consolidation.sh safe reference pattern is now consistently applied across all 7 remaining scripts.

Ready to merge.

sayhar:

PR #1843, opened 2026-04-29
fix(compact-catchup): eliminate false-positive unanswered messages

### LINKED ISSUE #1825: compact-catchup reports false-positive unanswered messages when get_conversation_history misses just-sent replies
## Incident — Apr 27 2026

`compact-catchup` flagged messages 1500, 1503, 1506 (replied to and moved to `processed/` at ~00:05–00:07 UTC) as unanswered open threads. They were not unanswered — replies were sent correctly.

## Root cause

`check_inbox(since_ts=<window_start>)` scans only `processed/` and `inbox/` (`.claude/agents/compact-catchup.md`:53, `src/mcp/inbox_server.py`:4634–4635):

```python
all_files = sorted(
    list(PROCESSED_DIR.glob("*.json")) + list(INBOX_DIR.glob("*.json"))
)
```

It does **not** scan `sent/`. Outbound replies are written to `sent/` and `messages.db` (direction=`out`) but are invisible to `check_inbox`.

The catchup agent therefore sees inbound messages in the window but cannot observe their replies. When those replies are absent from the data it has, the LLM (or downstream dispatcher reading the summary) classifies the messages as unanswered open threads — a false positive.

A secondary failure path exists if the `get_conversation_history` DB read is used to verify replies: `get_conversation_history` in `src/db/reader.py` has no `since_ts` parameter (lines 173–290). It returns the latest N messages ordered by `timestamp DESC LIMIT ?`. If more than N messages were sent after the window, the original replies fall off the page and are invisible. In the Apr 27 incident the gap between the replies being sent (00:07) and compact-catchup running was short enough that the DB write was recent, but a high-volume session could push them out of the default limit (20).

## Suggested fix directions

**Option A (preferred):** Extend `check_inbox(since_ts=...)` to also scan `sent/` and include outbound messages tagged `direction=out`. The catchup agent can then match inbound → outbound pairs within the window and only flag user messages with no corresponding outbound entry.

**Option B:** Add a `since_ts` parameter to `get_conversation_history` (and the underlying `src/db/reader.py:get_conversation_history`), allowing callers to query "all messages (both directions) since timestamp X" without being subject to the fixed LIMIT truncation. The compact-catchup agent could then use this to verify replies before classifying any message as unanswered.

**Option C (compact-catchup prompt):** Annotate the compact-catchup output format to distinguish "message in processed/ (already handled)" from "message in inbox/ (unhandled)" using the `_directory` field already set at line 4650 (`msg["_directory"] = f.parent.name`). The catchup agent already has access to this field — messages from `processed/` should never be classified as open threads.

Option C is the cheapest fix and requires no code changes: update `.claude/agents/compact-catchup.md` to instruct the agent to exclude messages whose `_directory == "processed"` when building the open-threads list.

## Files / line refs

- `.claude/agents/compact-catchup.md`:53 — `check_inbox(since_ts=...)` call
- `src/mcp/inbox_server.py`:4634–4635 — only scans `processed/` + `inbox/`, not `sent/`
- `src/mcp/inbox_server.py`:4650 — `_directory` field set (`"inbox"` or `"processed"`)
- `src/db/reader.py`:173–290 — `get_conversation_history` lacks `since_ts`; returns latest N only
- `src/mcp/inbox_server.py`:4940–4942 — outbound replies written to `sent/` and `messages.db`

sayhar:
🤖🦞 Lobster (engineer):

Fixes #1825

## What this solves

After a compaction, compact-catchup scans recent message history via `check_inbox(since_ts=...)`. The Apr 27 incident showed that messages already replied to and moved to `processed/` were being flagged as open/unanswered threads. The false positives arose because:

1. **Primary cause**: The catchup agent had no explicit instruction to treat `_directory == "processed"` messages as handled. The `_directory` field was already present on every message, but the agent wasn't told to use it — so the LLM sometimes classified `processed/` messages as open threads when no reply was visible in the scan.

2. **Secondary cause**: `get_conversation_history` had no `since_ts` parameter. It returned the latest N messages ordered by `timestamp DESC LIMIT ?`. In a high-volume session, replies sent just before the window start could fall off the page, making it appear that no reply had been sent.

## Changes

**Option C — prompt fix** (`.claude/agents/compact-catchup.md`):

The filter step (step 4) now explicitly documents the `_directory` field semantics:
- `_directory == "processed"` → already handled; must never be added to open-threads
- `_directory == "inbox"` → still in queue; potentially open

The session-file population step (step 11) reinforces this: only inbox messages may qualify as new open threads.

**Option B — code fix** (`src/db/reader.py`, `src/mcp/inbox_server.py`):

`get_conversation_history` (both DB reader and MCP tool) now accepts `since_ts: str | None`. When provided:
- DB path: adds `WHERE timestamp >= ?` using ISO 8601 lexicographic comparison (correct for SQLite's string-stored timestamps after normalising the `Z`/`+00:00` suffix)
- Filesystem fallback (`_apply_filters_and_paginate`): timezone-aware `datetime` comparison, malformed `since_ts` is silently skipped rather than crashing
- MCP tool schema: `since_ts` exposed to callers with a description that names the catch-up use case

The `count_conversation_history` function also gains `since_ts` for consistent pagination counts.

## Functional patterns

- Pure functions throughout: `_apply_filters_and_paginate` is extended without mutation; `since_ts` filter is a composable predicate applied before other filters
- Explicit parameter injection: `since_ts` flows from handler args → DB call → count call, no shared state
- Safe degradation: malformed `since_ts` in the filesystem path is caught and ignored, preserving existing behaviour

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_conversation_history_since_ts.py --tb=short -q` — 14 passed, 0 failed (new tests, written before implementation)
- [x] `uv run pytest tests/unit/ --tb=no -q` — 3011 passed, 24 skipped, 0 failed

## Manual test

N/A — this change is entirely internal to the compact-catchup agent and the DB reader layer. No external service calls, no file system side effects beyond what is already tested in unit tests. The `_directory` field used in the prompt fix is already populated by `inbox_server.py` (line 4685) and does not require any schema migration.

## Definition of Done
- [x] Integration/manual test run — N/A (internal change only)
- [x] User-visible outcome: no false-positive Telegram notifications to the user — not applicable to verify in isolation, but the prompt fix prevents the LLM from generating them and the code fix provides the data needed to verify replies

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -57,6 +57,14 @@ You are the **compact_catchup** subagent. Your job is to:
    - `subagent_result` messages (these are recently-returned subagent results -- collect their `task_id` values; these represent work that completed and may need dispatcher follow-up)
    - Notable system events: `update_notification`, `consolidation`
    - Exclude: `self_check`, `compact-reminder`, `compact_catchup`, `subagent_notification`, test messages
+
+   **Critical: `_directory` field determines message status.**
+   Every message returned by `check_inbox(since_ts=...)` carries a `_directory` field:
+   - `_directory == "processed"` -- the message was already handled and moved to `processed/`. It must **never** be classified as an open or unanswered thread. It may appear in the summary as context ("user asked X, we replied"), but not as outstanding work.
+   - `_directory == "inbox"` -- the message is still in the inbox and may be genuinely unhandled. Treat these as potentially open threads.
+
+   This distinction eliminates false-positive "unanswered thread" reports: a message in `processed/` was replied to and marked done; the reply may not appear in the `check_inbox` scan (replies go to `sent/`, not `processed/` or `inbox/`), but the presence of the message in `processed/` is conclusive evidence it was handled.
+
 5. **Call `get_active_sessions()` now** to retrieve all currently running agent sessions. Filter to `status: "running"` sessions, excluding dispatcher sessions. These are in-flight subagents that were active at compaction time and may still be running. If `get_active_sessions()` errors, note the failure and continue. Also read `~/lobster-workspace/data/inflight-work.jsonl` if it exists — find all `task_id` values that have at least one entry with `"status": "running"` but no entry with `"status": "done"` and `started_at` more than 30 minutes before now; these are potentially lost subagents not yet recovered by the sessions DB. (30 min is intentionally conservative — trades some false positives for earlier detection of genuinely lost work.)
 6. Read session notes in tiers (see "Session notes reading" below).
 7. **Verify live GitHub state for any PRs marked "awaiting sign-off" in session notes.**
@@ -120,7 +128,7 @@ You are the **compact_catchup** subagent. Your job is to:
 11. Build the session file content using the data from phases 1 and 2:
 
     - **Summary** (1-3 sentences, decision-log format): Synthesize from the catchup window. Write in narrative style: what we started working on, what we discovered or decided, what is still in progress. Example: "We started working on X; we realized Y and pivoted to Z; A and B are still in progress." Avoid changelog style (do not list "merged PR #N, commented on issue #M").
-    - **Open Threads**: Carry forward any threads found in the existing session file that are still pending. Add new threads for in-flight requests visible in the catchup window.
+    - **Open Threads**: Carry forward any threads found in the existing session file that are still pending. Add new threads for in-flight requests visible in the catchup window. Only messages with `_directory == "inbox"` may qualify as new open threads -- messages with `_directory == "processed"` are already handled and must never be added to the open-threads list (even if no corresponding reply is visible in the scan).
     - **Open Tasks**: List tasks from the catchup window that are not yet resolved. Include task IDs.
     - **Open Subagents**: List every agent from `get_active_sessions()` that is still in `running` state. Format: `task_id`, brief description (from the agent name or recent subagent_result), how long running (from the `started_at` field). Exclude dispatcher sessions.
     - **Notable Events**: Restarts, compactions, failed subagents, user decisions, errors -- pulled from the catchup window.
--- a/src/db/reader.py
+++ b/src/db/reader.py
@@ -180,13 +180,14 @@ def get_conversation_history(
     sender_type: str | None = None,
     limit: int = 20,
     offset: int = 0,
+    since_ts: str | None = None,
 ) -> list[Row]:
     """
     Fetch conversation history from the messages table.
 
     Applies optional filters for chat_id, source, full-text search,
-    direction (received/sent/all), and sender_type.  Results are ordered
-    newest-first.
+    direction (received/sent/all), sender_type, and since_ts.  Results are
+    ordered newest-first.
 
     Args:
         conn:        Open sqlite3.Connection.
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — fixes are correct, well-tested, and safe to merge.

## What I verified

**ISO 8601 normalization in the DB path (reader.py)**

The normalization `rstrip("Z").split("+")[0]` correctly strips both Z-suffix and +00:00-offset forms to a bare datetime prefix. This is safe for the documented UTC-only inputs. Negative UTC offsets (e.g., -05:00) are not stripped by this logic, but:
- The MCP schema explicitly documents UTC-only usage
- Negative offsets are not a documented or tested input
- In practice, callers pass Z-suffix or +00:00; the limitation is theoretical

The SQL lexicographic comparison works correctly for UTC stored timestamps because stored timestamps follow ISO 8601 ordering (the datetime prefix sorts correctly regardless of the trailing +00:00 suffix on stored rows).

**Consistency between DB path and filesystem fallback**

The DB path (reader.py) normalizes the cutoff to a bare string for SQL comparison. The filesystem fallback (inbox_server.py `_apply_filters_and_paginate`) uses `datetime.fromisoformat` with `replace("Z", "+00:00")` and compares timezone-aware datetimes. Both paths handle boundary behavior correctly (>=). They are consistent for the documented UTC use case.

**Existing `check_inbox` inconsistency (pre-existing, not introduced here)**

`check_inbox` passes `since_ts` raw to SQL without normalization. If `since_ts` is `2026-01-01T12:00:00Z` and stored timestamps use `+00:00` suffix (the Python isoformat default), the boundary message is excluded from `check_inbox` results because `ord('Z')` (90) > `ord('.')` (46), so `2026-01-01T12:00:00.000000+00:00 < 2026-01-01T12:00:00Z` lexicographically. This is a pre-existing boundary quirk, not introduced by this PR, and the new `get_conversation_history` normalization actually corrects this for its code path.

**Prompt change (compact-catchup.md)**

The `_directory` semantics documentation is accurate and unambiguous. The restriction is precisely scoped: `_directory == "processed"` messages are excluded from open-thread classification; `_directory == "inbox"` messages remain eligible. This cannot suppress legitimate open-thread detection — messages still in the inbox are correctly flagged, and the only messages excluded are those conclusively marked done by `mark_processed`. The wording ("may be genuinely unhandled") is appropriately hedged for `inbox` messages, preserving the agent's ability to use context to distinguish.

**Test structure**

Four test classes, no shadowing or duplicates. All 14 new tests pass. The full unit suite runs clean at 854 tests in `test_mcp_server/`, 3602 total in `tests/` — no regressions. The engineer's stated count (3011/24) reflects a narrower collection scope, not a discrepancy.

**One minor observation (non-blocking)**

The test `test_since_ts_in_filesystem_fallback` patches `_db_get_conversation_history=None` to force the filesystem path. This correctly exercises the fallback. However, it also relies on the fallback activating when `not used_db` — which works as expected since the None check bypasses DB. This is fine but worth noting as an implementation detail: the test covers the correct code path.

## Summary

Both Option B (code) and Option C (prompt) are correctly implemented. The `since_ts` filter is applied consistently through the DB path, the filesystem fallback, and the MCP tool schema. The prompt change eliminates false-positive open-thread classification without risk of suppressing legitimate detections. Tests cover boundary conditions, combined filters, and both code paths. Safe to merge.

sayhar:

PR #1845, opened 2026-04-29
fix(#1797): change source="internal" to "system" in cron scripts to prevent quarantine

### LINKED ISSUE #1797: nightly-consolidation.sh writes source="internal" — quarantined since PR #1736, missed 3 nights, false-positive dispatcher restart
## Summary

`scripts/nightly-consolidation.sh` writes inbox messages with `source="internal"`, but `"internal"` is not in `INBOX_MESSAGE_SOURCES` (`src/mcp/message_types.py:68`). Since PR #1736 added the unrecognized-source quarantine guard on Apr 22, every nightly consolidation has been silently quarantined. Three consolidations have failed (Apr 23, 24, 25) and canonical memory has not been updated in 3 days.

The same bug also caused a **false-positive dispatcher restart** at 2026-04-25 03:00:07 UTC — see "Restart trigger" below.

## Root cause

PR #1736 (Apr 22) added the quarantine guard for unrecognized message sources. It updated `daily-health-check.sh` to use `source="system"` but missed `nightly-consolidation.sh`, which still uses `source="internal"`.

Relevant code:

- `scripts/nightly-consolidation.sh:56` writes `"source": "internal"`
- `src/mcp/message_types.py:68` defines `INBOX_MESSAGE_SOURCES`: `{telegram, slack, sms, signal, whatsapp, bisque, system, bot-talk}` — no `"internal"`
- `src/mcp/inbox_server.py:4696-4700` quarantines any message whose source is not in `INBOX_MESSAGE_SOURCES`

## Restart trigger (false positive at 03:00 UTC, 2026-04-25)

3-way timing collision:
1. Nightly consolidation cron fired at 03:00:00 and wrote a message with `source="internal"`
2. MCP quarantined the message → blocking `wait_for_messages` exited and wrote a tombstone to `dispatcher-wfm-active`
3. The health check fired at the same second (03:00:02) and saw both: heartbeat stale (~3597s, since dispatcher had been idle since the 02:00 lobstertalk-unified cron) AND the tombstone (WFM not active) → both RED → restart

The dispatcher was alive and healthy throughout. It came back at 03:00:13 and processed normally. No user messages were dropped. But the user got a scary "🚨 Lobster Health Alert" notification.

## Fix

1. **Patch the script:** change `"source": "internal"` to `"source": "system"` in `scripts/nightly-consolidation.sh:56`
2. **Catch-up consolidation:** manually trigger consolidation for the 3 missed nights (Apr 23, 24, 25), or rely on the next successful run to consolidate the backlog if that's how the agent works
3. **Audit other scripts:** grep the repo for any other `source="internal"` writers that may have the same bug
4. **Consider a guardrail:** add a unit test that asserts every script-written source value is in `INBOX_MESSAGE_SOURCES`, so PR #1736-style refactors can't silently re-break this

## Severity

Medium-high. Silent failure of nightly consolidation is degrading canonical memory. False-positive restart alerts erode trust in the health system.

sayhar:
🤖🦞 Lobster (engineer): Fix for issue #1797 — nightly consolidation and update-check messages were being silently quarantined to `failed/` because the scripts wrote `source="internal"`, which is not in `INBOX_MESSAGE_SOURCES`.

## Root Cause

PR #1736 added a quarantine guard for unrecognized message sources and updated `daily-health-check.sh` to use `source="system"`, but missed `nightly-consolidation.sh` and `daily-update-check.sh`, which both kept writing `source="internal"`. Since Apr 22, every nightly consolidation message has been quarantined, resulting in 3 missed consolidations (Apr 23–25) and degraded canonical memory. A timing collision with the health check also triggered a false-positive dispatcher restart at 2026-04-25 03:00 UTC.

## Changes

- **`scripts/nightly-consolidation.sh`**: Change `"source": "internal"` → `"source": "system"` (line 56)
- **`scripts/daily-update-check.sh`**: Change `"source": "internal"` → `"source": "system"` in both the git mode and tarball mode branches (lines 34 and 53)
- **`tests/unit/test_mcp_server/test_message_taxonomy.py`**: Add `TestScriptSourceValues` class with three tests:
  - Targeted regression tests for `nightly-consolidation.sh` and `daily-update-check.sh`
  - A broad guardrail (`test_all_scripts_write_recognized_sources`) that parses every `*.sh` in `scripts/` for JSON `"source"` fields and asserts each value is in `INBOX_MESSAGE_SOURCES` — so future script additions can't silently re-introduce this class of bug

## Test Plan

- [x] `uv run pytest tests/unit/test_mcp_server/test_message_taxonomy.py` — all 20 tests pass (17 pre-existing + 3 new)
- [x] `uv run pytest tests/unit/test_mcp_server/ --tb=no -q` — 843 passed, no new failures introduced
- [x] Verified pre-existing collection errors in full suite are pre-existing on `main` (not caused by this PR)

Closes #1797

### DIFF
--- a/scripts/daily-update-check.sh
+++ b/scripts/daily-update-check.sh
@@ -31,7 +31,7 @@ if [ -d "$LOBSTER_DIR/.git" ]; then
         cat > "$INBOX/${TIMESTAMP}_update_available.json" << EOF
 {
   "id": "${TIMESTAMP}_update_available",
-  "source": "internal",
+  "source": "system",
   "chat_id": 0,
   "type": "update_notification",
   "text": "UPDATE AVAILABLE: Lobster is ${BEHIND} commits behind origin/main. Use check_updates for details.",
@@ -50,7 +50,7 @@ else
         cat > "$INBOX/${TIMESTAMP}_update_available.json" << EOF
 {
   "id": "${TIMESTAMP}_update_available",
-  "source": "internal",
+  "source": "system",
   "chat_id": 0,
   "type": "update_notification",
   "text": "UPDATE AVAILABLE: Lobster v${CURRENT_VERSION} -> v${LATEST_VERSION}. Use check_updates for details.",
--- a/scripts/nightly-consolidation.sh
+++ b/scripts/nightly-consolidation.sh
@@ -53,7 +53,7 @@ fi
 cat > "$INBOX/${TIMESTAMP}_consolidation.json" << EOF
 {
   "id": "${TIMESTAMP}_consolidation",
-  "source": "internal",
+  "source": "system",
   "chat_id": 0,
   "type": "consolidation",
   "text": "NIGHTLY CONSOLIDATION: Review today's events using memory_recent(hours=24) and update canonical memory files. Steps:\n1. Call memory_recent(hours=24) to get all events from the past day\n2. Synthesize key themes, decisions, and action items\n3. Update memory/canonical/daily-digest.md with the synthesis\n4. Update memory/canonical/priorities.md if priorities changed\n5. Update relevant project files in memory/canonical/projects/\n6. Update people files if new relationship info emerged\n7. Mark all reviewed events as consolidated using mark_consolidated\n8. Update memory/canonical/handoff.md with current state",
--- a/tests/unit/test_mcp_server/test_message_taxonomy.py
+++ b/tests/unit/test_mcp_server/test_message_taxonomy.py
@@ -13,6 +13,7 @@ tests/docker/Dockerfile.test.
 import asyncio
 import json
 import logging
+import re
 import sys
 from pathlib import Path
 from unittest.mock import patch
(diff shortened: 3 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with one noted gap

## Fix correctness

All three source field changes (nightly-consolidation.sh, daily-update-check.sh x2) are correct. "system" is in INBOX_MESSAGE_SOURCES; "internal" was not. Test suite runs at 843 passed, 0 failures — the claim in the briefing is accurate.

## Test structure

No duplicate class names. No shadowing. The path-traversal logic in `_get_scripts_dir()` (parents[3] = repo root) is correct and verified. The `autouse` fixture pattern that sets `self._known_sources` is valid pytest. Tests pass cleanly.

## Does the guardrail catch the actual failure mode?

Yes — if anyone reverts a script to `source="internal"` (or any unrecognized value) in `scripts/*.sh`, the broad test will catch it before merge. The targeted tests for nightly-consolidation.sh and daily-update-check.sh serve as named regression anchors, which is good.

The tests don't exercise the pre-fix "internal" value directly (they assert current state rather than showing the old value would have been caught), but this is a minor style point, not a functional gap.

## Gap: guardrail scope is narrower than the bug class

The broad guardrail only scans `scripts/*.sh`. It does not cover `docker/worker-entrypoint.sh`, which writes `source="worker"` to the inbox — a value that is also not in INBOX_MESSAGE_SOURCES. This means Docker worker messages have been silently quarantined since the quarantine guard landed (PR #1736), for the same reason as this PR's bug.

This is a **pre-existing issue**, not introduced by this PR. But the stated goal of the guardrail is to catch this entire class of error, and it misses at least one live instance.

Suggested follow-up (not a blocker for this PR):
1. Either add "worker" to INBOX_MESSAGE_SOURCES, or
2. Expand the guardrail to scan `docker/*.sh` and any other locations that write inbox messages, or
3. File a follow-up issue so this doesn't go untracked

## Summary

The fix is correct and the targeted regression tests are solid. The broad guardrail works for the scripts/ directory but has a blind spot for inbox-writing scripts that live outside that directory. Recommend filing a follow-up issue for the docker/worker-entrypoint.sh gap. This PR is safe to merge as-is.

sayhar:

PR #1847, opened 2026-04-29
fix(#1777): handle dict format in build_inline_keyboard

### LINKED ISSUE #1777: send_reply: complex button format renders as raw JSON on Telegram client
## Summary

When `send_reply` is called with the complex button format — `[[{"text": "Label", "callback_data": "value"}]]` — the Telegram client renders the button with the full JSON dict shape visible as the button label instead of just the `text` field.

## Reproduce

```
send_reply(
  chat_id=...,
  text="...",
  buttons=[[
    {"text": "👍 got it", "callback_data": "card1b_next"},
    {"text": "😴 sleep", "callback_data": "card1b_sleep"}
  ]]
)
```

## Observed

Button labels show as literal strings like:
```
{'text': '👍 got it'...: 'card1b_next'}
{'text': '😴 sleep...: 'card1b_sleep'}
```
(screenshot available on request)

## Expected

Button labels should be just `"👍 got it"` and `"😴 sleep"`, with `callback_data` used internally for the callback payload.

## Workaround

The simple format works correctly:
```
buttons=[["👍 got it", "❓ unclear"]]
```
(text = callback_data, single string per button.)

## Impact

Low. The simple format is sufficient for most use cases. Only matters if you want different display labels vs. callback payloads (e.g., for localized text or compact callback data).

## Hypothesis

Either the bot server is double-serializing the dict somewhere in the pipeline, or the dict isn't being unpacked before being passed to the Telegram API as `InlineKeyboardButton.text`. Worth grepping for where `buttons` JSON gets transformed into `InlineKeyboardMarkup`.

## Context

Hit this while building a flashcard-style drip tutor for interview prep (Apr 23 2026). Suggests the doc on `send_reply` buttons should either warn about this, or the complex format should be fixed so it matches its declared schema.

sayhar:
🤖🦞 Lobster (engineer):

## Summary

Fixes #1777: `send_reply` with the complex button format `[{"text": "Label", "callback_data": "value"}]` was rendering button labels as raw JSON strings instead of just the `text` field value.

**Root cause:** `build_inline_keyboard` in `src/bot/lobster_bot.py` handled `str`, `list`, and `tuple` button element types but had no `dict` branch. Dict items fell through to the `else` clause which called `str(item)`, serializing the entire dict as the button label.

**Fix:** Added a `dict` isinstance check before the existing list/tuple check. When a button element is a dict, `text` is extracted from `item["text"]` and `callback_data` from `item["callback_data"]` (falling back to the text value if `callback_data` is absent).

## Changes

- `src/bot/lobster_bot.py`: Added dict branch in `build_inline_keyboard`; updated docstring to document all three supported element formats.
- `tests/unit/test_bot/test_outbox_watcher.py`: Added `TestBuildInlineKeyboard` class with 7 tests covering the string, list, and dict formats, including a direct regression test for the `#1777` bug (label must not contain `{` or `callback_data`).

## Test plan

- [x] All 51 tests in `tests/unit/test_bot/test_outbox_watcher.py` pass
- [x] New `TestBuildInlineKeyboard` tests confirm dict buttons emit correct `text` and `callback_data` values
- [x] Regression test `test_dict_format_button_label_is_not_raw_json` explicitly asserts the label is `"Click me"` and contains no `{` or `callback_data` strings
- [x] No new failures introduced (pre-existing collection errors on other test modules are unrelated)

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -1788,8 +1788,10 @@ def build_inline_keyboard(buttons: list) -> InlineKeyboardMarkup:
     """Build a Telegram InlineKeyboardMarkup from a nested list of button labels.
 
     Each inner list represents a row of buttons.  Each element in the row is
-    either a plain string label (callback_data == label) or a [label, data]
-    two-element list where data is the callback_data payload.
+    one of:
+    - a plain string label (callback_data == label)
+    - a [label, data] two-element list where data is the callback_data payload
+    - a {"text": label, "callback_data": data} dict (complex format)
 
     Pure function: no side effects.
     """
@@ -1797,7 +1799,10 @@ def build_inline_keyboard(buttons: list) -> InlineKeyboardMarkup:
     for row in buttons:
         keyboard_row = []
         for item in row:
-            if isinstance(item, (list, tuple)) and len(item) == 2:
+            if isinstance(item, dict):
+                label = str(item.get("text", ""))
+                data = str(item.get("callback_data", label))
+            elif isinstance(item, (list, tuple)) and len(item) == 2:
                 label, data = item
             else:
                 label = str(item)
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -1800,7 +1800,11 @@ def build_inline_keyboard(buttons: list) -> InlineKeyboardMarkup:
         keyboard_row = []
         for item in row:
             if isinstance(item, dict):
-                label = str(item.get("text", ""))
+                if not item.get("text"):
+                    raise ValueError(
+                        f"Dict button spec is missing a non-empty 'text' key: {item!r}"
+                    )
+                label = str(item["text"])
                 data = str(item.get("callback_data", label))
             elif isinstance(item, (list, tuple)) and len(item) == 2:
                 label, data = item
--- a/tests/unit/test_bot/test_outbox_watcher.py
+++ b/tests/unit/test_bot/test_outbox_watcher.py
@@ -835,3 +835,93 @@ class TestPrepareSendItems:
             # <pre><code> and </code></pre> tags must be balanced
             assert html.count("<pre>") == html.count("</pre>") or True  # HTML tags may differ
             assert len(html) <= bot_module.TELEGRAM_HARD_LIMIT
+
+
+class TestBuildInlineKeyboard:
+    """Tests for build_inline_keyboard — the helper that converts button specs to InlineKeyboardMarkup.
+
+    Covers all three supported button element formats:
+    - Simple string (text == callback_data)
+    - [label, data] two-element list
+    - {"text": label, "callback_data": data} dict (the complex format from issue #1777)
+    """
+
+    @pytest.fixture
+    def bot_module(self):
+        return get_bot_module()
+
+    def test_simple_string_buttons(self, bot_module):
+        """Plain string buttons use the string as both text and callback_data."""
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is correct and directly addresses the reported bug. Dict-format button specs are now properly handled, and the 7 new tests are well-structured with no duplicate class names or unreachable test classes. All 203 bot tests pass. The test counts in the PR description check out.

## Items verified

**Fix correctness**
The dict branch is added as the _first_ check (before the list/tuple branch), which is the right order. Label is extracted from `"text"`, callback payload from `"callback_data"` with fallback to text — exactly matching what the MCP `send_reply` docs describe as the dict button format.

**Missing `"text"` key edge case (minor, non-blocking)**
If a caller passes `{"callback_data": "value"}` with no `"text"` key, the button label silently becomes an empty string. Telegram will reject this at send time with an API error, but the error will surface deep in the outbox handler rather than at the point the button spec is constructed. A defensive `if not label: raise ValueError(...)` or at minimum a logged warning would make the failure point obvious. This is not a regression (the old code would serialize the whole dict as the label instead), but it is a new silent-degradation path. Flagging as a suggestion, not a blocker.

**Before-state not directly asserted in regression test**
The regression test `test_dict_format_buttons_text_and_callback_data` asserts the correct _after_ behavior but does not assert that the _old_ behavior (dict serialized to label string) would have been wrong for the same input. This means the test would not catch a future refactor that accidentally reverts the fix. The separate `test_dict_format_button_label_is_not_raw_json` test does assert `"{" not in label`, which partially covers this — it would fail on the old code. Together they are sufficient. Worth noting for completeness.

**Test structure**
No duplicate class names. `TestBuildInlineKeyboard` is the sixth class in the file and is fully reachable. The `bot_module` fixture is correctly defined on the class (not inherited from a parent or shadowed). All 7 new tests are collected and pass.

**Pre-existing failures**
The engineer's claim of "51 bot tests pass" is accurate — I ran `uv run pytest tests/unit/test_bot/ -q` and got 203 passed (which includes pre-existing tests). The broader test suite has collection errors in unrelated modules (`test_mcp_server/`, `slack-connector/`) that predate this PR.

## Verdict

PASS — the fix is sound, the tests cover the primary regression case, and no new failure modes are introduced. The missing-text-key edge case is a suggestion for a follow-up, not a blocker.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

All checks complete.

**Implementation**
- Dict branch placed correctly before the list/tuple branch; extracts `item["text"]` and `item.get("callback_data", label)` — correct.
- ValueError guard uses `if not item.get("text")` which catches both missing and empty-string "text" keys — fires at construction time with a clear message including `{item!r}`. Correct.
- Docstring updated to document all three supported formats.

**Tests (9 new, all in `TestBuildInlineKeyboard`)**
- String buttons: label == callback_data ✓
- List format: label/data extracted correctly ✓
- Dict format: text and callback_data extracted correctly ✓
- Raw JSON regression (label must not contain `{` or `callback_data`) ✓
- Dict missing callback_data falls back to text ✓
- Dict missing text → ValueError ✓
- Dict empty text → ValueError ✓
- Multiple rows ✓
- Mixed formats in same row ✓

**Test run**
```
205 passed, 0 failed  (tests/unit/test_bot/)
```

No remaining issues or nits. Ready to merge.

sayhar:

PR #1849, opened 2026-04-29
fix(#1826): gate fallback rm -rf on confirmed-orphaned worktrees; add tests

### LINKED ISSUE #1826: fix(prune-pr-worktrees): handle TimeoutExpired + restrict fallback rm to orphaned directories

sayhar:
🤖🦞 Lobster (engineer):

## Problem

Two robustness gaps in `scripts/prune-pr-worktrees.py` per issue #1826:

1. **TimeoutExpired handling** — already addressed in nit commits cc217ba9 and 862aaa68 (run() wraps Popen+communicate in try/except, kills and reaps the child, returns (1, '', 'timeout')). No new code needed for this part; tests now cover it explicitly.

2. **Unconditional fallback rm -rf** — when `git worktree remove` fails, the previous code fell back to `shutil.rmtree` unconditionally (subject only to a path-safety check). This is unsafe: if the failure was transient (permission error, lock) and the worktree is still registered in git's metadata, direct filesystem removal leaves git in an inconsistent state.

## Fix

Add `is_worktree_registered(main_repo, worktree_path)` which queries `git worktree list --porcelain` and checks whether the target path appears in the output. In `remove_worktree()`, after `git worktree remove` fails, call this function before attempting the `shutil.rmtree` fallback:

- If the worktree **is still registered**: log `[ERROR]` and return `False` — the failure was transient; do not touch the filesystem.
- If the worktree **is absent from the list** (confirmed orphan): proceed to `shutil.rmtree` as before.
- If `git worktree list` itself fails: conservatively assume registered (safe default).

## Tests added

`tests/unit/test_scheduled_tasks/test_prune_pr_worktrees.py` (10 tests, all passing):

- `run()` catches `TimeoutExpired`, kills+reaps child, returns `(1, '', 'timeout')`
- `run()` exception does not escape to the caller
- `is_worktree_registered()` parses `--porcelain` output correctly
- `is_worktree_registered()` returns `True` conservatively when git fails
- `remove_worktree()` refuses `shutil.rmtree` when worktree still registered
- `remove_worktree()` allows `shutil.rmtree` for confirmed-orphaned directories
- `remove_worktree()` does not call `shutil.rmtree` on clean removal success
- Scan loop continues past a timed-out directory and emits the summary line

## Scope

Changes isolated to `scripts/prune-pr-worktrees.py` and a new test file. No schema changes, no service restarts, no install changes.

Closes #1826

### DIFF
--- a/scripts/prune-pr-worktrees.py
+++ b/scripts/prune-pr-worktrees.py
@@ -105,6 +105,21 @@ def dir_age_days(path: Path) -> float:
     return age.total_seconds() / 86400
 
 
+def is_worktree_registered(main_repo: Path, worktree_path: Path) -> bool:
+    """Return True if worktree_path still appears in `git worktree list --porcelain`."""
+    rc, out, _ = run(["git", "worktree", "list", "--porcelain"], cwd=main_repo)
+    if rc != 0:
+        # If the command fails we conservatively assume it IS registered to
+        # avoid accidentally deleting a live worktree.
+        return True
+    resolved = str(worktree_path.resolve())
+    for line in out.splitlines():
+        # Lines look like: "worktree /absolute/path"
+        if line.startswith("worktree ") and line[len("worktree "):].strip() == resolved:
+            return True
+    return False
+
+
 def remove_worktree(main_repo: Path, worktree_path: Path, dry_run: bool) -> bool:
     """Remove a git worktree and its directory."""
     if dry_run:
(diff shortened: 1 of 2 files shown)


sayhar:
🤖🦞 Lobster: Closing as duplicate of PR #1827 which addresses the same issue (#1826) and is already PASS-reviewed. Any unique test improvements from this PR should be cherry-picked into #1827.

PR #1851, opened 2026-04-29
fix(#1768): add fresh-dispatcher-start fallback in on-fresh-start.py

sayhar:
🤖🦞 Lobster (engineer):

## Summary

Fixes issue #1768: stale agent sessions not cleared on dispatcher restart.

**Root cause (per issue investigation):** When the dispatcher is killed and immediately restarted, `write-dispatcher-session-id.py` sees the previous session's JSONL file has a recent mtime and concludes the old dispatcher is still alive — skipping the session-ID marker file update. The tertiary marker file retains the old UUID, causing `is_dispatcher()` to return False for the new dispatcher session. `on-fresh-start.py` exits early at the subagent guard and never calls `_mark_all_running_failed()`, leaving stale `running` sessions in `agent_sessions.db`.

## Fix

Adds `_is_fresh_dispatcher_start()` as a fallback in `on-fresh-start.py`. When `is_dispatcher()` returns False, we check whether the primary state file (`dispatcher-claude-session-id`) is absent:

- The MCP server **clears** this file on every startup
- It is re-written only when the dispatcher calls `session_start()`  
- Subagents cannot trigger this path because they start only after the dispatcher has called `session_start()` (which writes the primary file)
- Compaction events are excluded: `on-compact.py` writes the primary file proactively for the new post-compact session UUID

So: absent primary file + non-compaction event + `LOBSTER_MAIN_SESSION=1` reliably identifies the pre-`session_start` dispatcher window.

The fix is purely additive — when `is_dispatcher()` already returns True (the common case), `_is_fresh_dispatcher_start()` is never called.

## Changes

- `hooks/on-fresh-start.py`: Add `_MCP_CLAUDE_SESSION_FILE` constant (with test override env var) and `_is_fresh_dispatcher_start()` fallback; update `main()` dispatcher guard to use both checks
- `tests/unit/test_hooks/test_fresh_dispatcher_start_fallback.py`: 9 new unit tests covering the fallback logic and main() gating integration, including a regression test for the exact issue #1768 failure scenario

## Test plan

- [ ] All 9 new unit tests pass (`test_fresh_dispatcher_start_fallback.py`)
- [ ] No regressions: 3424 existing tests still pass
- [ ] Regression scenario covered: `is_dispatcher()=False` + primary file absent → `_mark_all_running_failed()` is called
- [ ] Subagent scenario covered: `is_dispatcher()=False` + primary file present → hook exits without calling `_mark_all_running_failed()`
- [ ] Compaction scenario covered: compaction event → fallback returns False, hook exits correctly

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -104,6 +104,20 @@ COMPACTION_STATE_FILE = Path(
     )
 )
 
+# Primary dispatcher state file — written by the MCP server when the dispatcher
+# calls session_start(agent_type='dispatcher', claude_session_id=...).
+# The MCP server CLEARS this file on every startup, so it is absent during the
+# window between MCP restart and the dispatcher's first session_start() call.
+# On compaction, on-compact.py writes this file proactively so the new session
+# is identified before session_start() is called.
+# Override path for tests.
+_MCP_CLAUDE_SESSION_FILE = Path(
+    os.environ.get(
+        "LOBSTER_MCP_CLAUDE_SESSION_FILE_OVERRIDE",
+        os.path.expanduser("~/lobster-workspace/data/dispatcher-claude-session-id"),
+    )
+)
+
 INBOX_DIR = Path(os.path.expanduser("~/messages/inbox"))
 PROCESSING_DIR = Path(os.path.expanduser("~/messages/processing"))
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):

## Summary

All checks pass. The fix is correct, well-scoped, and well-tested.

## Logic verification

**The core claim holds:** absent `dispatcher-claude-session-id` + non-compaction + `LOBSTER_MAIN_SESSION=1` is a reliable signal for the pre-`session_start` dispatcher window. Verified by reading:

- `src/mcp/inbox_server.py`: `_clear_dispatcher_state_file()` is called at the top of `main()` on every MCP server startup — it unconditionally unlinks both state files before any session traffic arrives. Confirmed in code.
- `hooks/on-compact.py`: `write_dispatcher_claude_session_id(new_session_id)` is called on compaction, so the primary file is present before `on-fresh-start.py` runs in the compaction path. Confirmed in code.
- The fallback only runs when `is_dispatcher()` has already returned False — no change to the common-case path.

**Edge cases checked:**

- Compaction: doubly guarded — `_is_compact_event()` mtime check plus primary file will be present (on-compact.py wrote it). No false positive possible.
- Subagent: primary file is written when dispatcher calls `session_start()`, which happens before any Task() spawn. File will be present for all subagents. No false positive.
- OSError on stat: returns False (safe default). Tested.
- LOBSTER_MAIN_SESSION != 1: returns False immediately. Tested.

**Non-blocking observation (no fix required):** If the MCP server restarts mid-session via `restart-mcp.sh` while the dispatcher process stays alive, it reconnects and fires another SessionStart. The primary file gets cleared on MCP restart, so `_is_fresh_dispatcher_start()` could return True and `_mark_all_running_failed()` would fire. That is actually the correct behavior in that scenario — stale subagent sessions should be cleaned up after an MCP restart. No action needed.

## Tests

- 9 new tests: all pass
- Full suite: 3597 passed, 27 skipped, 0 failures
- The issue #1768 regression test (`test_proceeds_when_is_dispatcher_false_but_primary_file_absent`) directly simulates the failure mode and verifies `_mark_all_running_failed()` is called

## Code quality

- The `_MCP_CLAUDE_SESSION_FILE` env-var override pattern follows established project convention (same as `COMPACTION_STATE_FILE`).
- Docstring is comprehensive and accurately describes the file lifecycle contract.
- Purely additive: 468 additions, 1 deletion (the single-condition `if not` becomes a compound condition). No existing logic changed.

**PASS** — ready to merge.

sayhar:

PR #1855, opened 2026-04-29
fix(#1823): upgrade MCP SDK to 1.27.0 + add session_idle_timeout=1800

### LINKED ISSUE #1823: bug: MCP server asyncio event loop stalls for minutes — WFM-active heartbeat stops firing

### LINKED ISSUE #1873: fix(#1823 follow-up): session_idle_timeout=1800 reaps MCP sessions during idle WFM
## Finding

PR #1855 added \`session_idle_timeout=1800\` to \`StreamableHTTPSessionManager\` to help clear MCP asyncio stalls. Investigation on Apr 30 confirmed this is now causing session \"restarts\" every ~30 minutes: when the dispatcher is idle in \`wait_for_messages\` (WFM), the 1800s idle timer expires and the MCP server reaps the session, forcing a new Claude session to start.

**Evidence:**
- Gap between old WFM call (02:00:19Z) and new session first WFM (02:30:22Z) = exactly 1803s
- Health check was GREEN throughout — no false-positive, no heartbeat failure
- PR #1836 daemon thread confirmed working

The 30-minute restart was silent: no error, no alert, just a new session beginning. This causes unnecessary context loss and session churn for a dispatcher that may legitimately be idle (waiting for messages) for extended periods.

## Root Cause

\`session_idle_timeout=1800\` was added in PR #1855 as an **additional** mitigation alongside the primary fix (SDK upgrade to v1.27.0). The SDK upgrade addressed the underlying asyncio stall. Now that the SDK is fixed, the aggressive 30-min idle reap fires during normal WFM idle periods and causes unnecessary session teardown.

## Design Tension

Three options to resolve:

1. **Remove \`session_idle_timeout\` entirely** — rely on SDK v1.27.0 fix alone. Simplest, but removes a safety net if asyncio stalls resurface in future SDK versions.

2. **Increase it significantly** (e.g., 7200 or 86400) — so it doesn't fire during normal WFM idle periods, but still acts as a backstop for genuine hung sessions.

3. **Add a keepalive ping from the dispatcher during long WFM waits** — the dispatcher sends a lightweight no-op during idle periods to reset the idle timer. More complex but preserves the safety net without arbitrary timeout tuning.

## Suggested Next Step

Option 2 (increase to 7200 or higher) is the lowest-risk change: it preserves the safety net from #1855 while stopping the 30-min churn. Option 3 would be more robust long-term if WFM idle times are expected to regularly exceed any static timeout.

## Related

- PR #1855 (introduced \`session_idle_timeout=1800\`)
- PR #1836 (daemon thread heartbeat)
- Issue #1823 (original asyncio stall)

sayhar:
## Problem

The MCP server's asyncio event loop stalled for ~4 minutes on 2026-04-26 while the dispatcher was alive. The WFM-active heartbeat stopped, triggering a false health-check restart. Root cause: MCP SDK v1.26.0's \`StreamableHTTPSessionManager\` accumulates anyio task groups for long-lived stateful sessions with no reap deadline. When a task group blocks internally (session creation lock or cancel scope awaitable), the event loop starves — no timeouts, no errors, no output.

## What changed

**MCP SDK upgraded from v1.26.0 to v1.27.0.** v1.27.0 adds \`session_idle_timeout\` to \`StreamableHTTPSessionManager\`. When set, it uses \`anyio.CancelScope\` to reap sessions that have been idle longer than the threshold.

**\`SESSION_IDLE_TIMEOUT_SECONDS = 72000\`** added as a named module-level constant in \`inbox_server.py\` (near other heartbeat constants). This value (20 hours) must be >= the \`wait_for_messages\` timeout (72000s) so the MCP session is never reaped while the dispatcher is idle in WFM. The original 1800s value caused dispatcher crash loops overnight (issue #1873).

**\`StreamableHTTPSessionManager\` now receives \`session_idle_timeout=SESSION_IDLE_TIMEOUT_SECONDS\`** in the stateful HTTP mode constructor. The stateless HTTP bridge (\`inbox_server_http.py\`) is unchanged — combining \`stateless=True\` with \`session_idle_timeout\` raises \`RuntimeError\` per SDK contract.

## Before / after

```
BEFORE (v1.26.0):
  StreamableHTTPSessionManager(app=server, stateless=False)
  → no reap deadline → task groups accumulate → event loop starves → 4-min stall

AFTER (v1.27.0):
  StreamableHTTPSessionManager(app=server, stateless=False, session_idle_timeout=72000)
  → anyio.CancelScope deadline = now + 72000s → idle sessions reaped automatically
  → event loop stays unblocked
  → session survives dispatcher idle waits in WFM (1800s caused crash loops #1873)
```

## What is NOT changed

- \`inbox_server_http.py\` (stateless bridge) — untouched
- WFM heartbeat interval or health-check thresholds — untouched
- All other session manager behavior — only the idle timeout is added

## Functional patterns

Pure function reasoning: \`SESSION_IDLE_TIMEOUT_SECONDS\` is a named constant, not a magic literal. The idle timeout mechanism is entirely internal to the SDK — no new coroutines or mutable state added to the codebase.

## Tests run

- [x] \`uv run pytest tests/unit/test_mcp_server/test_http_session_idle_timeout.py -v\` — 2 passed (new tests, verified red before green)
- [x] \`uv run pytest tests/unit/ -x -q\` — 2999 passed, 24 skipped, 0 failed

## Manual test

N/A — this change is entirely internal to the MCP session manager constructor. No Telegram output, no external service calls, no file I/O affecting runtime state. The effect (preventing event loop stalls) is observable only by the absence of future asyncio stalls. The SDK upgrade itself is low-risk: only the session manager constructor call site is affected.

## Definition of Done

- [x] Integration/manual test — N/A (internal SDK change, no external services)
- [x] User-visible outcome: stall was invisible to user except via health-check restarts — no direct user-visible output to verify
- [x] Side-effect audit: no floods, no writes, no unintended volume — constructor-only change

Closes #1823
Fixes #1873

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Always-on AI assistant with Telegram/Slack integration"
 requires-python = ">=3.11"
 dependencies = [
-    "mcp>=1.0.0",
+    "mcp>=1.27.0",
     "httpx>=0.27.0",
     "watchdog>=4.0.0",
     "psutil>=5.9.0",
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -803,9 +803,10 @@ WAIT_HEARTBEAT_INTERVAL = 60
 # Idle timeout for the stateful StreamableHTTP session manager (issue #1823).
 # Sessions idle longer than this many seconds are reaped via anyio.CancelScope,
 # preventing task group accumulation that caused 4-minute event loop stalls.
-# The SDK recommends 1800s (30 min) — long enough to survive normal SSE polling
-# gaps without false reaps.  Requires MCP SDK >= 1.27.0.
-SESSION_IDLE_TIMEOUT_SECONDS = 1800
+# Must be >= wait_for_messages timeout (72000s / 20h) so the MCP session is
+# never reaped while the dispatcher is idle in WFM — 1800s caused crash loops
+# overnight when the dispatcher sat idle between messages (issue #1873).
+SESSION_IDLE_TIMEOUT_SECONDS = 72000
 
 # WFM-active signal file (issue #1713 / #949): written with a Unix epoch
 # timestamp when wait_for_messages begins blocking, refreshed every
--- a/src/mcp/inbox_server.py
+++ b/src/mcp/inbox_server.py
@@ -800,6 +800,13 @@ HEARTBEAT_FILE = _WORKSPACE / "logs" / "claude-heartbeat"
 # tests can verify it matches the health-check's WFM_ACTIVE_STALE_SECONDS.
 WAIT_HEARTBEAT_INTERVAL = 60
 
+# Idle timeout for the stateful StreamableHTTP session manager (issue #1823).
+# Sessions idle longer than this many seconds are reaped via anyio.CancelScope,
+# preventing task group accumulation that caused 4-minute event loop stalls.
+# The SDK recommends 1800s (30 min) — long enough to survive normal SSE polling
+# gaps without false reaps.  Requires MCP SDK >= 1.27.0.
+SESSION_IDLE_TIMEOUT_SECONDS = 1800
+
 # WFM-active signal file (issue #1713 / #949): written with a Unix epoch
 # timestamp when wait_for_messages begins blocking, refreshed every
 # WAIT_HEARTBEAT_INTERVAL seconds, and deleted when WFM returns.
@@ -9977,7 +9984,11 @@ async def main():
             except (ValueError, IndexError):
                 pass
 
-        session_manager = StreamableHTTPSessionManager(app=server, stateless=False)
+        session_manager = StreamableHTTPSessionManager(
+            app=server,
+            stateless=False,
+            session_idle_timeout=SESSION_IDLE_TIMEOUT_SECONDS,
+        )
 
         # Publish the session_manager reference so tool handlers (and /dispatcher-session)
         # can inspect live session state without importing from main().
--- a/tests/unit/test_mcp_server/test_http_session_idle_timeout.py
+++ b/tests/unit/test_mcp_server/test_http_session_idle_timeout.py
@@ -1,5 +1,5 @@
 """
-Tests for HTTP session idle timeout configuration (issue #1823).
+Tests for HTTP session idle timeout configuration (issue #1823, fix #1873).
 
 The MCP server's stateful HTTP transport must be constructed with
 session_idle_timeout=SESSION_IDLE_TIMEOUT_SECONDS to prevent anyio task
@@ -7,7 +7,11 @@ group accumulation in the StreamableHTTPSessionManager — the root cause of
 the 4-minute asyncio stall observed on 2026-04-26.
 
 The constant SESSION_IDLE_TIMEOUT_SECONDS must be defined at module level so
-health checks and tests can reference it without hardcoding the literal 1800.
+health checks and tests can reference it without hardcoding the literal.
+
+The value must be >= the wait_for_messages timeout (72000s / 20h) to prevent
+the MCP session from being reaped while the dispatcher is idle in WFM.
+Setting this to 1800s caused dispatcher crash loops overnight (issue #1873).
 """
 
 import asyncio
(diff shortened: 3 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The change is mechanically correct and all tests pass. No blocking issues found.

## What I verified

**Constructor argument type and name:** The SDK 1.27.0 constructor accepts `session_idle_timeout` as a `float | None` keyword argument. The production code passes an `int` (1800), which is a valid subtype of `float` in Python — no type mismatch.

**Test structure:** No duplicate class names, no shadowing. Both test classes are independent and reachable. The constructor-intercept test uses `**kwargs` capture, which correctly handles all keyword arguments the production code passes — `app`, `stateless`, and `session_idle_timeout` are all captured and asserted on.

**Test covers the before-state correctly:** The test intercepts at the module level (patching `sys.modules` before `main()` runs), which means the mock is in place before the import executes. This is the correct intercept point for a local import inside a function body.

**Test count claim:** The engineer reported "2999 passed, 24 skipped, 0 failed." My run of `tests/` in the PR branch returned 3590 passed, 27 skipped, 0 failed — higher than claimed, no failures. The discrepancy is plausible given test suite growth since the engineer ran it; importantly, 0 failures is confirmed. For reference, `main` currently has 1 pre-existing failure (`test_repo_config_env_deleted`) that this PR does not introduce.

**stateless=True guard:** Confirmed the SDK raises `RuntimeError: session_idle_timeout is not supported in stateless mode` if both are combined. The stateless bridge (`inbox_server_http.py`) is correctly left untouched with no `session_idle_timeout`.

**SDK constraint enforcement:** Bumping the minimum to `>=1.27.0` is necessary — `session_idle_timeout` was not present in 1.26.0's constructor, so an older install would fail at runtime with an unexpected keyword argument. The constraint is correct.

The `session_idle_timeout` parameter type in the SDK is `float | None`. The production constant is typed as an `int`. This works fine at runtime (Python accepts `int` where `float` is expected), but a strict type checker running `--strict` would flag it. Not worth holding the PR for, but worth a one-line annotation (`SESSION_IDLE_TIMEOUT_SECONDS: float = 1800`) if you want zero type-checker noise in the future.

## Gaps noted by engineer — confirmed accurate

- No asyncio loop monitoring added (issue item 1) — appropriate for a follow-on
- HTTP bridge has no session reaping — correct per SDK contract, not a regression

**Verdict:** This is a clean, well-scoped fix. The 30-minute timeout is appropriate given the dispatcher's long-polling behavior. Safe to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with nits

## Summary

Both new tests pass. Full unit suite: 2999 passed, 24 skipped, 0 failed — confirms engineer's count.

The core fix is correct: the timeout value was raised from 1800s to 72000s to match the wait-for-messages (WFM) idle period, preventing the session manager from destroying active sessions while the dispatcher is sitting in its normal idle wait. The constant is correctly named and placed; the session manager constructor call is correct; the stateless HTTP bridge is untouched.

## Issues

### Nit 1 — PR description is stale (misleading)

The PR title and body both describe the original 1800s value as the intended fix:

- Title: `fix(#1823): upgrade MCP SDK to 1.27.0 + add session_idle_timeout=1800`
- Body "What changed" section: `SESSION_IDLE_TIMEOUT_SECONDS = 1800`
- Body "Before/After" block: shows `session_idle_timeout=1800`

The actual code has `SESSION_IDLE_TIMEOUT_SECONDS = 72000`. The description was not updated after the 1800→72000 fix was applied. This is confusing for anyone reading the PR history.

**Recommendation:** Update the PR title and description to reflect 72000s.

### Nit 2 — Two copies of the magic number 72000, no compile-time link

The WFM tool handler uses `timeout = args.get("timeout", 72000)` as an inline default (line 4095). `SESSION_IDLE_TIMEOUT_SECONDS` is a separate constant set to the same value (line 809). The comment correctly cross-references them, but if the WFM default is ever changed without updating `SESSION_IDLE_TIMEOUT_SECONDS`, the session reaper could start killing sessions mid-wait again — silently.

The engineer flagged this in the briefing. It is a real but low-severity maintenance risk. A single named constant (`WFM_DEFAULT_TIMEOUT_SECONDS = 72000`) shared by both sites would eliminate the risk entirely. Not a blocker for this PR, but worth a follow-up issue.

### Observation — exact equality, no buffer

`SESSION_IDLE_TIMEOUT_SECONDS == WFM default timeout`. If a WFM call runs exactly to its timeout limit, the session idle timer and the WFM return happen at the same moment. In practice this is harmless (WFM returns and activity resets the idle clock), but it is worth noting that the constant provides no headroom above the WFM timeout.

## Verdict

PASS — the fix is correct and tests are solid. The stale PR description is the only thing worth cleaning up before merge.

sayhar:

PR #1856, opened 2026-04-29
fix(bot): handle dict button format in build_inline_keyboard

### LINKED ISSUE #1777: send_reply: complex button format renders as raw JSON on Telegram client

sayhar:
## Problem solved

`send_reply` buttons using the complex dict format — `{"text": "Label", "callback_data": "value"}` — rendered with the full Python dict repr as the button label on the Telegram client (e.g. `{'text': 'Got it', 'callback_data': 'card1b_next'}`). Only the simple string format `"label"` and two-element list format `["label", "data"]` worked correctly.

## Root cause

`build_inline_keyboard` in `src/bot/lobster_bot.py` handled two element types per button: plain strings and `[label, data]` lists. Dicts fell through to the `else` branch, which called `str(item)` — converting the entire dict to its string representation, which became the button label.

## Fix

Added an explicit `isinstance(item, dict)` branch before the existing cases that extracts `item["text"]` as the label and `item.get("callback_data", label)` as the callback payload. If `"callback_data"` is absent, the text value is used as the fallback (consistent with how plain strings work).

All three documented formats now work:
1. `"label"` — text == callback_data
2. `["label", "data"]` — separate text and callback_data
3. `{"text": "label", "callback_data": "data"}` — dict with optional callback_data

The fix is 3 lines in a pure function with no side effects, no schema changes, and no service restarts needed.

## Functional patterns used

The implementation remains a pure function: no side effects, all state passed as arguments, composed from simple element-level dispatch. The new dict branch follows the same pattern as the existing `list/tuple` branch.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_outbox_watcher.py::TestBuildInlineKeyboard -v` — 7 passed, 0 failed (5 tests were failing before the fix as designed; all pass after)
- [x] `uv run pytest tests/unit/test_bot/ --tb=short -q` — 203 passed, 0 failed
- [x] `uv run pytest tests/unit/test_mcp_server/ --tb=no -q` — 840 passed, 0 failed

## Manual test

N/A — the fix is entirely in `build_inline_keyboard`, a pure function that converts the buttons list to a `InlineKeyboardMarkup` object. The broken behavior (str(dict) as label) is directly observable in tests without hitting the Telegram API. The regression test `test_dict_button_does_not_render_raw_json_as_label` explicitly asserts that `{` does not appear in the button label.

The live Telegram behavior cannot be self-verified without a test bot session. A quick manual smoke test (send a message with dict-format buttons, confirm label renders correctly) is recommended before merge.

## Definition of Done
- [x] Integration/manual test run — N/A, pure function change; see note above
- [ ] User-visible outcome verified — blocked: requires sending a live Telegram message to confirm label renders. Flagged for reviewer attention.
- [x] Side-effect audit complete: change is confined to a pure function; no writes, no network calls, no shared state
- [x] External service calls: mocked in all tests

Closes #1777

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -1788,8 +1788,17 @@ def build_inline_keyboard(buttons: list) -> InlineKeyboardMarkup:
     """Build a Telegram InlineKeyboardMarkup from a nested list of button labels.
 
     Each inner list represents a row of buttons.  Each element in the row is
-    either a plain string label (callback_data == label) or a [label, data]
-    two-element list where data is the callback_data payload.
+    one of three supported formats:
+
+    1. Plain string — text == callback_data
+       e.g. ``"Yes"``
+
+    2. Two-element list/tuple — separate text and callback_data
+       e.g. ``["Vote Yes", "yes_cb"]``
+
+    3. Dict with a ``"text"`` key — complex format with optional separate callback_data
+       e.g. ``{"text": "Got it", "callback_data": "card1b_next"}``
+       If ``"callback_data"`` is absent, ``"text"`` is used as the callback value.
 
     Pure function: no side effects.
     """
@@ -1797,7 +1806,10 @@ def build_inline_keyboard(buttons: list) -> InlineKeyboardMarkup:
     for row in buttons:
         keyboard_row = []
         for item in row:
-            if isinstance(item, (list, tuple)) and len(item) == 2:
+            if isinstance(item, dict):
+                label = item["text"]
+                data = item.get("callback_data", label)
+            elif isinstance(item, (list, tuple)) and len(item) == 2:
                 label, data = item
             else:
                 label = str(item)
--- a/tests/unit/test_bot/test_outbox_watcher.py
+++ b/tests/unit/test_bot/test_outbox_watcher.py
@@ -835,3 +835,98 @@ class TestPrepareSendItems:
             # <pre><code> and </code></pre> tags must be balanced
             assert html.count("<pre>") == html.count("</pre>") or True  # HTML tags may differ
             assert len(html) <= bot_module.TELEGRAM_HARD_LIMIT
+
+
+class TestBuildInlineKeyboard:
+    """Tests for build_inline_keyboard — the button layout builder.
+
+    Covers all three button element formats documented in send_reply:
+      1. Plain string  — text == callback_data
+      2. [label, data] two-element list — separate text and callback_data
+      3. {"text": ..., "callback_data": ...} dict — explicit fields (complex format)
+    """
+
+    # Number of distinct button element formats the function must handle
+    SUPPORTED_ELEMENT_FORMATS = 3
+
+    @pytest.fixture
+    def build_keyboard(self):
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with non-blocking note

## Summary

The fix is correct and the tests are solid. All 7 new tests pass. No duplicate class names or shadowing issues were found. The regression test directly verifies the before-state (str(dict) rendering). Test isolation and structure are clean.

## One issue: unguarded key access on malformed dict buttons

The new dict branch does a bare `item["text"]` with no guard. If a caller passes a dict that is missing the "text" key, Python raises a `KeyError` with no explanation. There is no test covering this case.

The engineer's briefing flags this: local-dev has a more defensive implementation that raises a `ValueError` with a clear message instead of a raw `KeyError`. That pattern is strictly better — it turns an obscure internal crash into an explicit, debuggable API error.

**Assessment:** This is non-blocking. The fix ships correctly for all well-formed inputs, which is the documented contract. But the `ValueError` guard is a low-cost improvement worth adding. My recommendation: merge as-is, and track the defensive guard as a follow-up. If the codebase convention is to validate inputs at function boundaries, require it before merge.

## Test suite health

- 7 new tests: all pass
- Overall suite on this branch: 14 pre-existing failures (all unrelated to this change, confirmed by comparison with main)
- No new failures introduced

## No live Telegram verification concern

Acknowledged limitation noted by the engineer. The logic is straightforward enough that the unit tests are high-confidence. A smoke test with a real message would be ideal but is not a merge blocker.

## Verdict: PASS

sayhar:

PR #1862, opened 2026-04-29
fix(docker): automate GITHUB_TOKEN 4-layer auth setup in container init (#1798)

### LINKED ISSUE #1798: Docker: GITHUB_TOKEN env passthrough doesn't reach Claude Code subprocesses
## Problem

When `GITHUB_TOKEN` is passed into the staging container via docker-compose `environment:` passthrough, it is available to PID 1 (the entrypoint script) but **not** to Claude Code subprocesses launched inside the container (subagents, scheduled tasks, Claude Code tool calls).

This means `gh` CLI commands that work in the entrypoint shell fail with auth errors inside Claude Code sessions and their spawned subagents.

### Root cause

Docker env passthrough injects variables into the initial process environment. Subprocesses launched later — including tmux sessions and anything Claude Code forks — do not inherit variables that aren't written to a persistent auth store or a globally-sourced env file. `GITHUB_TOKEN` set only via `docker-compose environment:` satisfies PID 1 but is lost by the time Claude Code's subagent pool runs.

---

## Fix applied (4 layers required)

All four layers were needed to ensure `gh` CLI auth works in every subprocess context:

### 1. `/etc/environment`

Add `GITHUB_TOKEN` and `GH_TOKEN` to `/etc/environment` so they are available system-wide across all login and non-login shells, including tmux sessions:

```bash
echo "GITHUB_TOKEN=${GITHUB_TOKEN}" | sudo tee -a /etc/environment
echo "GH_TOKEN=${GITHUB_TOKEN}" | sudo tee -a /etc/environment
```

### 2. `gh auth login` — persistent credential store

Run `gh auth login --with-token` inside the container to write credentials to `~/.config/gh/hosts.yml`. This is the canonical way `gh` CLI checks auth status and it survives shell re-launches:

```bash
echo "${GITHUB_TOKEN}" | gh auth login --with-token
```

### 3. Shell profiles (`.bashrc` and `.profile`)

Export `GITHUB_TOKEN` and `GH_TOKEN` from both `.bashrc` (interactive shells) and `.profile` (login shells):

```bash
echo "export GITHUB_TOKEN=${GITHUB_TOKEN}" >> ~/.bashrc
echo "export GH_TOKEN=${GITHUB_TOKEN}" >> ~/.bashrc
echo "export GITHUB_TOKEN=${GITHUB_TOKEN}" >> ~/.profile
echo "export GH_TOKEN=${GITHUB_TOKEN}" >> ~/.profile
```

### 4. `~/lobster-config/config.env`

Add the token to the Lobster config file so it is sourced when Lobster reads its config:

```bash
echo "GITHUB_TOKEN=${GITHUB_TOKEN}" >> ~/lobster-config/config.env
echo "GH_TOKEN=${GITHUB_TOKEN}" >> ~/lobster-config/config.env
```

---

## What needs updating

- [ ] `docker/staging/entrypoint-staging.sh` — add GitHub auth setup block (all 4 layers) after the config.env write step
- [ ] `docker/staging/SETUP.md` — document the GitHub auth requirement and workaround (see new file added in related PR)
- [ ] `docker/staging/docker-compose.staging.yml` — already passes `GITHUB_TOKEN` and `GH_TOKEN` via environment; that part is correct, but insufficient on its own

---

## Notes

- `GH_TOKEN` is an alias that `gh` CLI also recognizes; setting both avoids ambiguity across `gh` CLI versions
- The `gh auth login` step (layer 2) is the most reliable long-term fix — it writes to a persistent auth file that survives container restarts without needing env vars at all
- This issue applies to any Docker-based Lobster deployment that needs `gh` CLI access from Claude Code subprocesses, not just staging

sayhar:
## What problem this solves

Inside the staging container, `gh` CLI commands fail with auth errors in Claude Code subagents, tmux sessions, and systemd service forks — even when `GITHUB_TOKEN` appears to be available at the container level. The root cause: `docker-compose environment:` passthrough injects the token into PID 1 (systemd) only. Subprocesses spawned later do not inherit variables that aren't written to a persistent location.

Previously the workaround required manually running 4 shell commands inside a running container after every launch. This PR automates all 4 layers in `lobster-container-init.sh` so `gh` CLI auth is ready at container boot without any post-launch steps.

## What changed

**`lobster-container-init.sh`** — new section after `config.env` write that gates on `GITHUB_TOKEN` being non-empty and writes to all 4 auth layers:

| Layer | Location | What it covers |
|---|---|---|
| 1 | `/etc/environment` | All processes, all shells, tmux sessions (via PAM) |
| 2 | `gh auth login` → `~/.config/gh/hosts.yml` | `gh` CLI itself, regardless of env vars |
| 3 | `~/.bashrc` / `~/.profile` | Interactive and login shells |
| 4 | `config.env` | Lobster services that source config at startup (was already done) |

All layers are idempotent: stale entries are stripped before writing, so container restarts do not accumulate duplicate lines. If `GITHUB_TOKEN` is unset, the section is skipped with a clear log message.

**`lobster-container-init.service`** — adds `PassEnvironment=GITHUB_TOKEN` and `PassEnvironment=GH_TOKEN` so systemd propagates the docker-compose-injected vars to the init script.

**`docker-compose.staging.yml`** — adds an `environment:` section that passes `GITHUB_TOKEN` and `GH_TOKEN` from the host shell into the container. Previously the token wasn't forwarded at all at the compose level.

**`SETUP.md`** (new file) — documents the automated setup, replacing the manual workaround that was previously documented only in local-dev. Includes verification command and troubleshooting steps.

## Functional patterns

Pure sequential setup with early-exit guard: the 4-layer section runs only when `GITHUB_TOKEN` is non-empty, each layer is independent and idempotent, failure of layer 2 (`gh auth login`) is soft-logged rather than fatal to avoid breaking container boot when `gh` is unavailable at init time.

## Before / after

Before:
```
docker compose up → container starts → gh auth fails in Claude Code subagents
→ user must exec into container and manually run 4 commands
```

After:
```
export GITHUB_TOKEN=ghp_...
docker compose up → lobster-container-init.service runs → 4 layers written
→ gh auth works in all subprocess contexts immediately
```

## Tests run

- [x] `uv run pytest tests/ -x -q --ignore=tests/docker --ignore=tests/integration --ignore=tests/smoke` — 3438 passed, 24 skipped, 0 failed
- [x] `bash -n docker/staging/lobster-container-init.sh` — syntax OK

## Manual test

Not run — requires a Docker environment with a real `GITHUB_TOKEN` and a staging container build. The change touches container initialization logic; unit tests confirm no regressions in the Python layer. The script logic is straightforward (conditional writes with idempotent cleanup), and the `bash -n` syntax check passed.

Verification when testing: `sudo docker exec -it lobster-staging bash -l -c "gh auth status"` should return authenticated after container start with `GITHUB_TOKEN` exported on the host.

## Definition of Done

- [ ] Integration/manual test run — requires Docker + real GITHUB_TOKEN; blocked (see Manual test above)
- [ ] User-visible outcome verified — blocked pending manual test
- [x] Side-effect audit: writes to `/etc/environment`, `~/.bashrc`, `~/.profile`, `~/.config/gh/hosts.yml` — all scoped to the container; no production system affected
- [x] External service calls: `gh auth login` is the only external call; runs inside the container at init time only

Closes #1798


sayhar (commented):
## Review: NEEDS-WORK

### Critical: PR implementation does not match PR description

The PR title and body claim to automate 4-layer GITHUB_TOKEN setup in `lobster-container-init.sh`, add `PassEnvironment=GITHUB_TOKEN` to the service file, and add an `environment:` section to `docker-compose.staging.yml`. None of these changes are present in the branch. All three files (`lobster-container-init.sh`, `lobster-container-init.service`, `docker-compose.staging.yml`) are byte-for-byte identical to main. The only actual change in `docker/staging/` is the addition of `SETUP.md`.

This means the PR as merged would:
- Close issue #1798 without actually automating anything
- Leave the manual 4-layer workaround as the only path

---

### Blocking issues in SETUP.md

**1. Layer 4 instruction is broken (config.env gets overwritten on boot)**

The SETUP.md tells the user to:
```bash
echo "GITHUB_TOKEN=${GITHUB_TOKEN}" >> ~/lobster-config/config.env
echo "GH_TOKEN=${GITHUB_TOKEN}" >> ~/lobster-config/config.env
```

But `lobster-container-init.sh` **unconditionally overwrites** `config.env` on every boot (the existing comment explains this was intentional to prevent stale-value bugs). Any token appended manually will be lost on the next container restart. Layer 4 as documented does not survive a reboot — it is silently broken.

**2. Layer 1 is not idempotent (`sudo tee -a` accumulates duplicates)**

```bash
echo "GITHUB_TOKEN=${GITHUB_TOKEN}" | sudo tee -a /etc/environment
echo "GH_TOKEN=${GITHUB_TOKEN}" | sudo tee -a /etc/environment
```

Running this more than once (e.g. rotating a token) appends duplicate lines. No stale-entry stripping is done. Over time `/etc/environment` accumulates stale tokens. A restart-safe version would strip existing `GITHUB_TOKEN=` and `GH_TOKEN=` lines before appending.

**3. "Starting the container" section misleads users about docker-compose passthrough**

The "Starting the container" section tells users to `export GITHUB_TOKEN=ghp_...` before running `docker compose up`, implying this is sufficient. But `docker-compose.staging.yml` has no `environment:` forwarding section — the export is a no-op. A user following only that section will start the container and find `gh` auth broken, then have to scroll down to find the manual fix. Either remove the misleading `export` line, or add the `environment:` section to docker-compose so passthrough actually works.

**4. Stale "Tracking issue" reference**

SETUP.md says: "See GitHub issue #1798 for full context and **planned** automation of this setup step." If this PR is meant to be that automation, the reference is wrong. If this PR is not implementing the automation, the reference is fine but the PR title/description need to accurately say "document manual workaround" rather than "automate 4-layer auth setup."

---

### Summary

The SETUP.md content is useful documentation, but it has two broken instructions (Layer 1 not idempotent, Layer 4 wiped on reboot) and one misleading section (the `export` before `docker compose up`). More fundamentally, the PR description describes automation that was not implemented.

**Required before PASS:**
1. Either implement the claimed automation in `lobster-container-init.sh` / `.service` / `docker-compose.staging.yml`, OR update the PR title/description/body to accurately describe this as "add manual setup documentation" and update SETUP.md to remove the broken Layer 4 and fix the misleading export.
2. Fix Layer 1 to be idempotent (strip stale entries before appending).
3. Fix or remove the Layer 4 instruction (it does not survive container restart).
4. Fix the "Starting the container" section — either remove the `export GITHUB_TOKEN` line or make docker-compose actually forward it.

sayhar:
Closing to reopen with the correct source branch. The implementation was committed to `fix/issue-1798-github-token-passthrough` but this PR was accidentally targeting `local-dev`, which caused the implementation files to appear unchanged vs main. Reopening from the correct branch.

PR #1863, opened 2026-04-29
fix(docker): automate GITHUB_TOKEN 4-layer auth setup in container init (#1798)

### LINKED ISSUE #1798: Docker: GITHUB_TOKEN env passthrough doesn't reach Claude Code subprocesses

sayhar:
## What problem this solves

Inside the staging container, `gh` CLI commands fail with auth errors in Claude Code subagents, tmux sessions, and systemd service forks — even when `GITHUB_TOKEN` appears to be available at the container level. The root cause: `docker-compose environment:` passthrough injects the token into PID 1 (systemd) only. Subprocesses spawned later do not inherit variables that aren't written to a persistent location.

Previously the workaround required manually running 4 shell commands inside a running container after every launch. This PR automates all 4 layers in `lobster-container-init.sh` so `gh` CLI auth is ready at container boot without any post-launch steps.

## What changed

**`lobster-container-init.sh`** — new section after `config.env` write that gates on `GITHUB_TOKEN` being non-empty and writes to all 4 auth layers:

| Layer | Location | What it covers |
|---|---|---|
| 1 | `/etc/environment` | All processes, all shells, tmux sessions (via PAM) |
| 2 | `gh auth login` → `~/.config/gh/hosts.yml` | `gh` CLI itself, regardless of env vars |
| 3 | `~/.bashrc` / `~/.profile` | Interactive and login shells |
| 4 | `config.env` | Lobster services that source config at startup (written in-place via heredoc, not appended) |

All layers are idempotent: stale entries are stripped before writing, so container restarts do not accumulate duplicate lines. If `GITHUB_TOKEN` is unset, the section is skipped with a clear log message.

**`lobster-container-init.service`** — adds `PassEnvironment=GITHUB_TOKEN` and `PassEnvironment=GH_TOKEN` so systemd propagates the docker-compose-injected vars to the init script.

**`docker-compose.staging.yml`** — adds an `environment:` section that passes `GITHUB_TOKEN` and `GH_TOKEN` from the host shell into the container. Without this, the `export GITHUB_TOKEN=...` in the host shell was a no-op.

**`SETUP.md`** (new file) — documents the automated setup, replacing the manual workaround. The "Starting the container" section shows `export GITHUB_TOKEN=...` before `docker compose up`, which is now meaningful because docker-compose.staging.yml forwards it into the container.

## Addressing the NEEDS-WORK review on #1862

PR #1862 (now closed) was pointing to `local-dev` instead of this feature branch, causing the reviewer to see the implementation files as unchanged vs main. The implementation was correctly committed to `fix/issue-1798-github-token-passthrough` all along. The specific concerns from the review:

1. **Layer 4 (config.env) not broken** — `GITHUB_TOKEN` is written inline in the heredoc that generates `config.env` (not appended afterward), so it is never wiped on reboot.
2. **Layer 1 is idempotent** — uses `grep -v` to strip existing entries before writing, not `tee -a`. Running twice produces exactly one line.
3. **SETUP.md "export" section is now correct** — docker-compose.staging.yml now has the `environment:` forwarding section, so `export GITHUB_TOKEN=...` before `docker compose up` works as documented.
4. **No stale "Tracking issue" reference** — removed from SETUP.md.

## Functional patterns

Pure sequential setup with early-exit guard: the 4-layer section runs only when `GITHUB_TOKEN` is non-empty, each layer is independent and idempotent, failure of layer 2 (`gh auth login`) is soft-logged rather than fatal to avoid breaking container boot when `gh` is unavailable at init time.

## Before / after

Before:
```
docker compose up → container starts → gh auth fails in Claude Code subagents
→ user must exec into container and manually run 4 commands
```

After:
```
export GITHUB_TOKEN=ghp_...
docker compose up → lobster-container-init.service runs → 4 layers written
→ gh auth works in all subprocess contexts immediately
```

## Tests run

- [x] `uv run pytest tests/ -x -q --ignore=tests/docker --ignore=tests/integration --ignore=tests/smoke` — 3438 passed, 24 skipped, 0 failed
- [x] `bash -n docker/staging/lobster-container-init.sh` — syntax OK
- [ ] Live Docker integration test — requires a staging container build with a real `GITHUB_TOKEN`; blocked (Docker not available in this environment)

## Manual test

N/A for unit tests. The script change is bash-only (container init). Verification when testing: `sudo docker exec -it lobster-staging bash -l -c "gh auth status"` should return authenticated after container start with `GITHUB_TOKEN` exported on the host. Cannot self-verify without a live Docker environment.

## Definition of Done

- [ ] Integration/manual test run — requires Docker + real GITHUB_TOKEN; blocked (see Manual test above)
- [ ] User-visible outcome verified — blocked pending manual test
- [x] Side-effect audit: writes to `/etc/environment`, `~/.bashrc`, `~/.profile`, `~/.config/gh/hosts.yml` — all scoped to the container; no production system affected
- [x] External service calls: `gh auth login` is the only external call; runs inside the container at init time only

Closes #1798


sayhar:
**PASS** — all 6 NEEDS-WORK items from PR #1862 are correctly addressed. Verified against the actual branch files:

1. **Layer 4 (config.env) — inline heredoc**: PASS. `GITHUB_TOKEN=${GITHUB_TOKEN:-}` is written inline inside the `cat > config.env << CONFIG` heredoc. No append after the fact.

2. **Layer 1 idempotency (/etc/environment)**: PASS. `grep -v "^GITHUB_TOKEN=\|^GH_TOKEN="` strips stale entries into a tmp file before appending fresh ones and `sudo cp`-ing over the original. Running twice produces exactly one entry.

3. **docker-compose.staging.yml `environment:` section**: PASS. Present and correct — both `GITHUB_TOKEN=${GITHUB_TOKEN:-}` and `GH_TOKEN=${GITHUB_TOKEN:-}` forwarded from host shell using the `:-` default form.

4. **SETUP.md — no stale "#1798 planned automation" reference**: PASS. SETUP.md is a new file that documents the automated setup correctly. No tracking-issue stub references anywhere.

5. **Layer 3 (.bashrc/.profile) idempotency**: PASS. Same `grep -v "^export GITHUB_TOKEN=\|^export GH_TOKEN="` pattern as Layer 1 — strip-before-write, no accumulation across restarts.

6. **`${GITHUB_TOKEN:-}` in docker-compose + `[ -n "${GITHUB_TOKEN:-}" ]` gate interaction**: PASS. When `GITHUB_TOKEN` is unset on the host, docker-compose passes an empty string into the container. The gate `if [ -n "${GITHUB_TOKEN:-}" ]` evaluates false for an empty string, so the 4-layer setup is correctly skipped with a clear log message. No error, no partial setup.

The service file (`lobster-container-init.service`) also correctly adds `PassEnvironment=GITHUB_TOKEN` and `PassEnvironment=GH_TOKEN` so systemd propagates the docker-compose-injected vars to the init script.

Implementation is clean, idempotent, and handles the token-not-set case gracefully.

sayhar:

PR #1869, opened 2026-04-29
fix(#1868): inject dispatcher bootup on fresh start before session_start() is called

### LINKED ISSUE #1868: fix: inject-bootup-context.py misidentifies dispatcher on fresh start (same #1768 root cause)
## Problem

On a fresh restart (new process, MCP server restarted), `inject-bootup-context.py` injects the subagent bootup file instead of the dispatcher bootup file. This means the dispatcher starts without its correct instructions.

## Root cause

The same root cause as issue #1768 (fixed in `on-fresh-start.py`), but never applied to `inject-bootup-context.py`:

1. The MCP server **clears** the primary state file (`dispatcher-claude-session-id`) on every startup
2. At the time `inject-bootup-context.py` fires on SessionStart, this file is absent
3. `write-dispatcher-session-id.py` runs first, but may see a recent JSONL mtime and conclude the old dispatcher is still alive — skipping the session-ID file update
4. The tertiary marker file then has the old (wrong) UUID, so `is_dispatcher()` returns False
5. `inject-bootup-context.py` falls back to injecting subagent bootup

## The fix (already proven in on-fresh-start.py)

Add a `_is_fresh_start_dispatcher()` fallback to `inject-bootup-context.py`:
- Absent primary file + `LOBSTER_MAIN_SESSION=1` + not a compaction event → treat as dispatcher
- This is safe because subagents are only spawned after the dispatcher has called `session_start()`, which writes the primary file; by the time any subagent starts, the file is present

The compact-pending sentinel fallback (Option 3, #1375) already handles post-compact sessions. This fix handles fresh restarts.

## Prior art

`on-fresh-start.py` has had `_is_fresh_dispatcher_start()` since the PR that fixed #1768. The logic just needs to be replicated in `inject-bootup-context.py`.

sayhar:
## What was broken

Every time the dispatcher was cold-started (new process, MCP server just restarted), it received the subagent bootup file instead of the dispatcher bootup file. The dispatcher started its main loop without its instructions, identity cues, or protocol guidance.

This only happened on genuine fresh restarts — compaction events already had a working fallback (the compact-pending sentinel, added in #1375).

## Root cause

Three detection mechanisms exist to identify the dispatcher at SessionStart time. All three failed on a fresh restart:

1. **Primary file** (`dispatcher-claude-session-id`): The MCP server clears this file on every startup. It is absent until the dispatcher calls `session_start()`. So at the time `inject-bootup-context.py` fires — before `session_start()` has been called — this file doesn't exist.

2. **Tertiary marker file** (`messages/config/dispatcher-session-id`): `write-dispatcher-session-id.py` runs first in the hook chain and normally writes the new session UUID here. But when the previous session's JSONL has a recent mtime (e.g., kill-and-immediately-restart), `write-dispatcher-session-id.py` concludes the old dispatcher is still alive and skips the update. The file then holds the old, wrong UUID.

3. **Compact-pending sentinel**: Only present after compaction events. Fresh restarts have no sentinel.

All three checks returning False causes `is_dispatcher()` to return False, and `inject-bootup-context.py` falls through to injecting subagent bootup.

## The fix

Add a `_is_fresh_start_dispatcher()` fallback: if the primary file is absent **and** `LOBSTER_MAIN_SESSION=1`, treat the session as the dispatcher.

This is the same fix applied to `on-fresh-start.py` in PR for issue #1768 — that PR fixed the `--mark-failed` problem but left `inject-bootup-context.py` with the same underlying bug.

Why it's safe:
- **Subagents**: spawned only after the dispatcher calls `session_start()`, which writes the primary file. By the time any subagent SessionStart fires, the primary file is present. The fallback does not fire for subagents.
- **Compaction events**: `on-compact.py` proactively writes the primary file with the new UUID before the post-compact SessionStart fires. Primary file is present → fallback does not fire → compaction handled by the existing sentinel fallback as before.

## State machine change

Before this fix:

```
Fresh restart → is_dispatcher()=False → sentinel absent → subagent bootup (WRONG)
```

After this fix:

```
Fresh restart → is_dispatcher()=False → sentinel absent → primary file absent + MAIN_SESSION=1 → dispatcher bootup (CORRECT)

Subagent start → is_dispatcher()=False → sentinel absent → primary file PRESENT (dispatcher wrote it) → subagent bootup (CORRECT)

Post-compact → is_dispatcher()=False → sentinel PRESENT + MAIN_SESSION=1 → dispatcher bootup (unchanged)
```

All other flows are unaffected: the new fallback only runs when both `is_dispatcher()` and `_is_post_compact_dispatcher()` return False.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_inject_bootup_context_fresh_start.py -v` — 9 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 331 passed, 0 failed (all hook tests)
- [x] `uv run pytest tests/unit/ -q` — 3006 passed, 24 skipped, 0 failed (full unit suite)
- [x] `uv tool run ruff check` on changed files — all checks passed

## Manual test

This hook fires at SessionStart time before the dispatcher enters its main loop. Verifying it end-to-end requires a real restart:

- [ ] Live restart test — blocked: requires stopping and restarting the dispatcher process, which would interrupt the running system. Safe to merge; the unit tests cover all three code paths (fresh restart, subagent, compaction). The fix is a pure addition (new fallback only runs when all existing paths return False) with no regression risk.

## Definition of Done

- [x] Side-effect audit: no floods, no writes to shared runtime state, no volume changes — this hook only reads files and writes to stdout (injected as context)
- [ ] User-visible outcome verified — not self-verifiable without a live restart; dispatched to reviewer

Closes #1868

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -31,6 +31,19 @@ file.  If the sentinel exists AND LOBSTER_MAIN_SESSION=1, the session is
 treated as a post-compact dispatcher session regardless of the ID-match result.
 This bypasses the chicken-and-egg timing problem entirely for the post-compact
 case and ensures dispatcher bootup is always injected when it should be.
+
+Fresh-start fallback (issue #1868):
+On a genuine fresh restart (new process, MCP server restarted), the primary
+state file (dispatcher-claude-session-id) is cleared by the MCP server on
+startup and is absent until the dispatcher calls session_start().
+write-dispatcher-session-id.py may also skip updating the tertiary marker file
+if the previous session's JSONL has a recent mtime, leaving a stale UUID.
+Both state files then fail to identify the new dispatcher, causing is_dispatcher()
+to return False.  The sentinel fallback doesn't help (no compaction occurred).
+Fix: if the primary file is absent AND LOBSTER_MAIN_SESSION=1, treat the session
+as the dispatcher.  This is safe because subagents are only spawned after the
+dispatcher calls session_start() (which writes the primary file), and compaction
+events write the primary file proactively via on-compact.py.
 """
 
 import json
@@ -89,6 +102,40 @@ def _is_post_compact_dispatcher() -> bool:
     return COMPACT_PENDING_SENTINEL.exists()
 
 
+def _is_fresh_start_dispatcher() -> bool:
+    """Return True if this looks like a fresh-restart dispatcher session (issue #1868).
+
+    On a genuine fresh restart, the MCP server clears the primary state file
+    (dispatcher-claude-session-id) on startup.  That file is absent until the
+    dispatcher calls session_start().  write-dispatcher-session-id.py may also
+    skip updating the tertiary marker file if the previous session's JSONL has
+    a recent mtime, leaving a stale UUID in the tertiary file.
+
+    Consequence: is_dispatcher() finds no matching state file and returns False.
+    The compact-pending sentinel fallback (_is_post_compact_dispatcher) is also
+    inactive because no compaction occurred.  inject-bootup-context.py then
+    falls back to injecting subagent bootup — wrong for the dispatcher.
+
+    Fix: absent primary file + LOBSTER_MAIN_SESSION=1 → treat as dispatcher.
+
+    Why this is safe:
+    - Subagents are spawned only after the dispatcher calls session_start(),
+      which writes the primary file.  By the time any subagent SessionStart
+      fires, the primary file is present.
+    - Compaction events: on-compact.py proactively writes the primary file
+      with the new UUID before the post-compact SessionStart fires.  The
+      primary file is therefore present for compaction events, not absent.
+
+    Returns False on any OSError (cannot stat the primary file) — safe default.
+    """
+    if os.environ.get("LOBSTER_MAIN_SESSION", "") != "1":
+        return False
+    try:
+        return not session_role._get_mcp_claude_session_file().exists()
+    except OSError:
+        return False
+
+
 def _read_file_safe(path: Path, label: str) -> str | None:
     """Return file contents or None on any error or empty file, logging to stderr."""
     if not path.exists():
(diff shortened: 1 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is correct, minimal, and well-tested. The new fallback fires only when two conditions hold simultaneously — the primary identification file is absent AND the session is marked as the main loop — which is a uniquely identifying combination for the fresh-restart window. All three changed files look good.

## Ordering (main concern from the brief)

The ordering is correct. The fresh-start fallback runs after both the primary state-file check (`is_dispatcher()`) and the compaction sentinel check. Both must return False before the new fallback fires. The `if not is_dispatcher and ...` guard ensures this at each step.

## Correctness of the safety argument

The claim that "primary file absent + main session env = only one possible explanation" holds:
- Background workers start only after the dispatcher's registration call writes the primary file. By the time any worker fires, the file is present.
- Compactions are handled by `on-compact.py`, which proactively writes the primary file with the new UUID before the post-compact session starts. So for compaction events the primary file is present, not absent — the fresh-start fallback cannot fire.
- External/non-Lobster sessions lack `LOBSTER_MAIN_SESSION=1`, so the env-var gate blocks them.

The two-condition gate is tight enough that no plausible false-positive path exists.

## Updated test (test_inject_bootup_context_sentinel.py)

The renamed test now sets up a realistic subagent scenario: the primary file is present and contains a dispatcher UUID, but the session UUID is a different (subagent) UUID. This is the correct before-state for a real subagent session. The prior test described the broken state — primary file absent, LOBSTER_MAIN_SESSION=1 — which would now incorrectly fire the new fallback and return dispatcher bootup. Replacing it with the correct realistic premise is the right call. ✓

## Tests

All 18 new/modified hook tests pass. Full unit suite: 3006 passed, 24 skipped, 0 failures. The collection errors in the broader test tree are pre-existing infrastructure issues unrelated to this PR.

`test_returns_true_when_primary_file_absent_and_main_session` does `mod.session_role = sr` before patching, which is redundant (the module already imports the same `session_role` object — `mod.session_role is sr` is True at load time). This is harmless and does not affect correctness.

## Verdict

No issues. The fix is safe, the safety argument is sound, and the test coverage is thorough.

sayhar:
Superseded by #1908 simplified flag-file detection approach. This PR is no longer needed.

PR #1875, opened 2026-04-30
fix: rename AccountRegistry.get() to lookup() and fix vacuous test assertion

aeschylus:
## What and why

PR #1871 introduced two non-blocking code quality issues. This PR fixes both.

### 1. AccountRegistry.get() naming surprise

AccountRegistry.get() raised KeyError on a missing key instead of returning None. This is the opposite of standard Python dict.get() behavior, which silently returns None. Any reader who saw registry.get(name) and assumed dict-like semantics would expect a None return path that doesn't exist.

Renamed to lookup() so the contract is obvious at the call site: calling .lookup() means "give me the value or raise — there is no optional return."

No behavior change — only the method name and its docstring usage example are updated.

### 2. Vacuous test assertion

test_error_is_a_subclass_of_granola_api_error in test_granola_sync_multi_account.py asserted:

    assert issubclass(GranolaUnknownAccountError, Exception)

This is trivially true for every Python class — it tells us nothing. The test name said "must be a GranolaAPIError" but the assertion checked Exception. The test would have passed even if GranolaUnknownAccountError had been refactored to extend an unrelated type.

Replaced with test_error_is_a_key_error, which verifies the actual design intent:
- GranolaUnknownAccountError IS a KeyError (signals missing config, not an API failure)
- GranolaUnknownAccountError is NOT a GranolaAPIError (so callers can distinguish config errors from API errors in exception handlers)

## Files changed

- scheduled-tasks/granola_multi_account.py — rename method, update docstring usage example, add clarifying note about why it's not named get
- tests/unit/test_granola_multi_account.py — three call sites updated to lookup()
- tests/unit/test_granola_sync_multi_account.py — vacuous test replaced with meaningful assertion

## Functional patterns used

Pure lookup structure (no mutation); the rename makes the raising contract visible without changing any behavior. The fixed test asserts observable type hierarchy rather than testing an implementation transcript.

## Tests run

- [x] uv run pytest tests/ -k granola -v --tb=short — 87 passed, 0 failed

## Manual test

N/A — entirely internal (method rename + test assertion fix). No external services, file I/O, or systemd changes.

## Definition of Done

- [x] Integration/manual test run — N/A, purely internal
- [x] User-visible outcome — N/A, no user-visible behavior changes
- [x] Side-effect audit: no floods, no writes, no external calls

### DIFF
--- a/scheduled-tasks/granola_multi_account.py
+++ b/scheduled-tasks/granola_multi_account.py
@@ -73,8 +73,8 @@ class AccountRegistry:
 
     Usage:
         registry = AccountRegistry(build_accounts_from_env(os.environ))
-        cfg = registry.get("drew")   # → AccountConfig  # noname
-        cfg = registry.get("ghost")  # → KeyError: "No API key registered for 'ghost'"
+        cfg = registry.lookup("drew")   # → AccountConfig  # noname
+        cfg = registry.lookup("ghost")  # → KeyError: "No API key registered for 'ghost'"
 
         "drew" in registry           # → True  # noname
     """
@@ -82,9 +82,12 @@ class AccountRegistry:
     def __init__(self, accounts: list[AccountConfig]) -> None:
         self._by_name: dict[str, AccountConfig] = {a.name: a for a in accounts}
 
-    def get(self, account_name: str) -> AccountConfig:
+    def lookup(self, account_name: str) -> AccountConfig:
         """
         Return the AccountConfig for account_name.
+
+        Named ``lookup`` (not ``get``) to signal that this raises on missing keys,
+        unlike the standard Python dict.get() which returns None.
 
         Raises:
             KeyError: if account_name is not registered.
--- a/tests/unit/test_granola_multi_account.py
+++ b/tests/unit/test_granola_multi_account.py
@@ -196,12 +196,12 @@ class TestAccountRegistry:
 
     def test_lookup_known_primary_account(self):
         registry = self._make_registry()
-        cfg = registry.get(ACCOUNT_DREW)  # noname
+        cfg = registry.lookup(ACCOUNT_DREW)  # noname
         assert cfg.api_key == PRIMARY_KEY
 
     def test_lookup_known_secondary_account(self):
         registry = self._make_registry()
-        cfg = registry.get(ACCOUNT_KELLY)  # noname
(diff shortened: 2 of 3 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

**Summary:** Both changes are correct and the test suite confirms them.

---

**Cold-read findings (before reading the briefing):**

1. **Method rename (get → lookup):** The rename is clean and the docstring rationale is clear. The three test call sites are all updated. No other call sites exist in the production code — confirmed by searching the repo. The rename is safe.

2. **Test assertion fix:** The original `issubclass(GranolaUnknownAccountError, Exception)` was vacuously true for any Python class, and the test name was wrong (it claimed to check GranolaAPIError but checked Exception). The replacement checks the actual inheritance chain: `GranolaUnknownAccountError extends KeyError` ✓ and `not GranolaAPIError` ✓ — both meaningful.

**Engineer briefing cross-check:** The briefing identified exactly what was wrong. No gaps found.

---

**Checks performed:**

- **Actual class definition verified:** `GranolaUnknownAccountError(KeyError)` in `src/integrations/granola/client.py:163` — the new assertions match reality.
- **No duplicate test class names** in either modified test file.
- **Tests run on PR branch:** 49/49 Granola tests pass.
- **Full suite run:** 2204 passed, 2 failed — both failures are pre-existing in `test_memory.py` vector search tests, unrelated to this PR.
- **No other call sites of `.get()` on `AccountRegistry` exist** — the rename cannot break callers outside the test file.

**No issues found.**

aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

**Checklist:**
- ✅ `AccountRegistry.get()` fully renamed to `lookup()` — method definition, docstring usage example, and all three test call sites updated; no orphaned `.get()` call sites in the diff.
- ✅ Test is now meaningful: asserts `issubclass(GranolaUnknownAccountError, KeyError)` AND `not issubclass(GranolaUnknownAccountError, GranolaAPIError)` — the actual design contract, not a trivially-true `Exception` check.
- ✅ No PII in the diff (test fixture names only; no real keys or credentials).
- ✅ All CI checks passed (CodeQL python + javascript, 87 granola tests per PR body).

PR was already merged — no action needed.

PR #1881, opened 2026-04-30
feat(#880): add prompt-prep subagent for off-thread orientation (Phase 1)

### LINKED ISSUE #880: feat: prompt-prep subagent — move orientation work off the dispatcher main thread
🤖🦞 Lobster (ops):

## Problem

When a complex message arrives, the dispatcher currently does all orientation work inline: reading files, checking history, choosing a subagent, composing the full prompt. This takes 30–360 seconds and blocks the main loop, causing inbox staleness and health check restarts.

## Proposed Architecture

Introduce a `prompt-prep` subagent. The dispatcher's flow for complex requests becomes:

1. Read message, identify basic intent (≤5s — stays on main thread)
2. Spawn `prompt-prep` subagent with: `{message_text, chat_id, source, rough_intent}`
3. `prompt-prep` does all orientation work: reads relevant files, checks message history, determines subagent type, skill, writes complete prompt
4. Returns a spawn spec: `{subagent_type, prompt, task_id, ack_text}`
5. Dispatcher receives the spec via `write_result`, spawns the real subagent per spec

## Key Properties

- Dispatcher stays truly stateless — all research and prompt composition moves off the main thread
- Generalizes the existing `compact-catchup` pattern (which already does post-compaction orientation as a subagent)
- Adds ~30–60s of latency before real work starts — acceptable for complex tasks
- For simple/fast requests, dispatcher can still respond inline (no prep needed)

## When to Use

Trigger prompt-prep when:
- The request requires reading >1 file to compose a good prompt
- The subagent type or skill is not immediately obvious
- The request references history or state that needs to be looked up

Skip prompt-prep for:
- Simple inline answers from context
- Clearly-typed requests (e.g. "merge PR #N" → lobster-generalist, no research needed)

## Rollout

Phase 1: implement prompt-prep as a lobster-generalist subagent with a well-documented prompt format
Phase 2: add as a named agent type with its own bootup context
Phase 3: teach the dispatcher to recognize trigger patterns and auto-route

sayhar:
🤖🦞 Lobster (engineer):

## Problem

The dispatcher currently blocks its main loop for up to 360 seconds doing orientation work inline: reading files, checking history, choosing a subagent type, composing the full prompt. This violates the 7-second rule and causes inbox staleness and health-check restarts.

## What Phase 1 does

Two files only — no Python code, no schema changes, no infrastructure changes:

1. **`.claude/agents/prompt-prep.md`** — a new agent definition that accepts a `message_text` + `rough_intent` input, does all orientation research (conversation history, triggered skills, referenced GitHub issues/PRs/files), picks a `subagent_type`, and returns a structured spawn spec via `write_result(chat_id=0)`. The dispatcher can then spawn the real subagent with zero inline work.

2. **`.claude/sys.dispatcher.bootup.md`** — two surgical additions:
   - `### prompt-prep flow`: tells the dispatcher when and how to invoke prompt-prep (use when request requires >1 file read, subagent type isn't obvious, or referenced state needs lookup; skip for simple/typed requests and system messages).
   - Prompt-prep result handler in the `subagent_result` block: parses the `## spawn-spec` + `## prompt` structure and spawns the real subagent. Falls back to a user-visible error if the spec is malformed.

## Flow before/after

**Before:**
```
message arrives → dispatcher reads files inline (30–360s, blocks loop) → spawns subagent
```

**After:**
```
message arrives → dispatcher acks + spawns prompt-prep (≤1s, returns to loop)
                                    ↓
                    prompt-prep reads files, checks history, picks subagent type (off-thread)
                                    ↓
                    write_result(spawn-spec) → dispatcher spawns real subagent
```

The dispatcher is on the main thread only for the ack and the spawn-prep Task call. All research moves off-thread.

## Architectural decisions

- **Immediate generic ack** (`"On it — figuring out the best approach."`) — tells the user work is in progress without committing to what's happening yet, since prompt-prep hasn't run.
- **Agent reads its own definition** (`"Read ~/lobster/.claude/agents/prompt-prep.md first for full instructions."`) — Phase 1 uses `subagent_type="lobster-generalist"` because there's no named `prompt-prep` agent type yet. The instruction to self-read is how the lobster-generalist gains the prompt-prep behavior. Phase 2 will introduce the named type.
- **Inline parse on the dispatcher main thread** — the spawn-spec is tiny YAML (4 fields); the regex parse is O(n) on a short string and fits inside the 7-second rule. No subagent needed for this step.
- **Sonnet model** for prompt-prep — matches the agent complexity level; same as other orientation agents.
- **Portable definition** — no hardcoded repo names, no project-specific paths, no Lobster-specific tool assumptions. The agent reads config from the input it receives.

## What Phases 2 and 3 will do

- **Phase 2**: register `prompt-prep` as a named `subagent_type` so the dispatcher doesn't need the "read your own instructions" indirection.
- **Phase 3**: auto-routing — the dispatcher detects trigger patterns and invokes prompt-prep automatically, rather than the dispatcher author deciding manually at each callsite.

## Tests run

- [x] Manual inspection of both file diffs — both additions match the spec exactly
- [x] `git diff` reviewed — only `.claude/agents/prompt-prep.md` (new file) and `.claude/sys.dispatcher.bootup.md` (two additions) changed
- [x] Verified prompt-prep result handler is inserted BEFORE the engineer→reviewer routing block so it intercepts first
- [ ] Live end-to-end test (send a complex message, verify spawn-spec flow) — blocked: requires a running dispatcher with `LOBSTER_DEBUG=true`; safe to merge and test in local-dev

## Manual test

The change is entirely in `.claude/` instruction files — no runtime code, no external service calls, no database changes. No automated test framework applies. The operative test is: send a message that triggers prompt-prep (complex request requiring file reads), verify the ack arrives, verify the spawn-spec write_result arrives in the dispatcher's inbox, verify the real subagent spawns correctly.

This requires a running production or local-dev instance. Safe to merge before that test, as the change only activates when the dispatcher explicitly calls `Task(subagent_type="lobster-generalist", prompt="Read prompt-prep.md first...")`.

## Definition of Done

- [x] Agent definition file matches spec frontmatter and body requirements
- [x] Dispatcher additions match spec exactly (prompt-prep flow + result handler)
- [x] No project-specific identifiers in the agent file (portable across instances)
- [x] Handler inserts BEFORE engineer→reviewer routing (correct position)
- [x] `chat_id` carries through correctly in the spawn-spec
- [ ] Live integration test — see Manual test above

Closes #880

### DIFF
--- a/.claude/agents/prompt-prep.md
+++ b/.claude/agents/prompt-prep.md
@@ -1,0 +1,120 @@
+---
+name: prompt-prep
+description: "Orientation and prompt-composition agent. Spawned by the dispatcher when a user request requires research before a real subagent can be spawned. Returns a spawn spec so the dispatcher can spawn the real subagent with zero inline work."
+model: sonnet
+---
+
+> **Subagent note:** You are a background subagent. Do NOT call `wait_for_messages`. Call `write_result` (NOT `send_reply`) when your task is complete -- the dispatcher reads your result as a structured spawn spec and acts on it directly.
+
+You are **prompt-prep**. You do all orientation and research so the dispatcher does not have to. Your sole output is a structured spawn spec delivered via `write_result(sent_reply_to_user=False)`. Never call `send_reply`. Never call `wait_for_messages`.
+
+## Input contract
+
+The dispatcher passes these fields in the task prompt:
+
+```
+task_id: prompt-prep-<slug>
+chat_id: <chat_id>
+source: <source>
+background: true
+---
+message_text: <raw user message>
+rough_intent: <dispatcher is 1-line guess>
+trigger_message_id: <inbox message_id>
+```
+
+Parse these fields from the prompt frontmatter and body. Store `task_id`, `chat_id`, `source`, `message_text`, `rough_intent`, and `trigger_message_id` in working context before proceeding.
+
+## Research steps
+
+Perform these steps to gather enough context to compose a high-quality prompt for the real subagent:
+
+1. **Restore recent context:** Call `get_conversation_history(chat_id=<chat_id>, limit=10)` to see what the user has been working on recently.
+
+2. **Check triggered skills:** Call `get_skill_context_for_message(message_text=<message_text>)` to see if any skills activate for this message. If skills return context, include it in the composed prompt.
+
+3. **Read any referenced state:** If the message mentions a GitHub issue number, PR URL, file path, project name, or task reference -- read that content now. This is the core research step. Use `gh issue view`, `gh pr view`, file reads, or other tools as appropriate.
+
+4. **Determine `subagent_type`:** Choose from the options below using the decision rules.
+
+5. **Compose the full task prompt:** Write the complete prompt that the real subagent will receive, including YAML frontmatter. The prompt must be self-contained -- the real subagent receives no context beyond what you put in it.
+
+6. **Pick a clean `task_id` slug** for the real subagent (e.g. `gh-issue-880`, `research-solar-panels`, `pr-review-1234`).
+
+## Decision rules for `subagent_type`
+
+| Signal | `subagent_type` |
+|---|---|
+| Message mentions a GitHub issue number AND implies code work ("fix", "implement", "work on", "close") | `functional-engineer` |
+| Message mentions a PR URL or the `/re-review` command | `review` |
+| Voice/brain dump indicators: multiple unrelated topics, "brain dump", "note to self", stream-of-consciousness | `brain-dumps` |
+| Everything else | `lobster-generalist` |
+
+When in doubt, use `lobster-generalist`.
+
+## When to skip research
+
+If `rough_intent` already unambiguously identifies both the subagent type and the full context needed (e.g. "list tasks" or "what time is it"), and no file reads are required, compose the spawn spec from context alone without calling research tools.
+
+## Output format -- spawn spec via write_result
+
+Once research is complete, call:
+
+```python
+mcp__lobster-inbox__write_result(
+    task_id=<task_id from input frontmatter>,   # must be the prompt-prep-<slug> task_id
+    chat_id=0,
+    source="system",
+    sent_reply_to_user=False,
+    text=<spawn spec below>,
+    status="success",
+)
+```
+
+The `text` field must follow this exact structure:
+
+```
+## spawn-spec
+```yaml
+subagent_type: <type>
+task_id: <real-task-slug>
+chat_id: <chat_id>
+source: <source>
+```
+
+## prompt
+---
+task_id: <real-task-slug>
+chat_id: <chat_id>
+source: <source>
+background: true
+---
+
+<full composed prompt body>
+```
+
+**Note:** The `## spawn-spec` YAML block and `## prompt` section are parsed by the dispatcher with regex. Do not add extra text between these markers.
+
+## Composing the prompt body
+
+Write the prompt body as you would write it yourself if you were the dispatcher with full context. Include:
+
+- The user's original request (paraphrase if it's cleaner, or quote directly)
+- Any relevant context from conversation history (recent decisions, ongoing work, references)
+- Any relevant skill context returned in step 2
+- Any content read in step 3 (issue body, PR description, file contents) -- summarize rather than dump in full if large
+- Clear instructions for what the real subagent should do
+- For `functional-engineer`: reference the issue number and repo explicitly
+- For `review`: reference the PR URL explicitly
+
+Keep the prompt mobile-friendly. The real subagent reads it in full -- concise beats exhaustive.
+
+## Rules
+
+- Never call `send_reply` -- all output goes through `write_result`
+- Never call `wait_for_messages`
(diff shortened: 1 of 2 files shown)


sayhar (commented):

## Summary

Two issues require fixes before merge. One is a blocking format bug. One is a missing inflight tracking entry. Two additional points are non-blocking but worth noting.

---

## BLOCKING: Nested backtick format spec is ambiguous

In \`prompt-prep.md\`, the \"Output format\" section wraps the example spawn-spec in an outer plain-backtick fence:

\`\`\`
## spawn-spec
\`\`\`yaml
...
\`\`\`

## prompt
...
\`\`\`

In standard markdown, the first \`\`\`yaml line **closes** the outer fence, leaving \`yaml\` as plain text and breaking the nested structure. When rendered (e.g. in GitHub, VS Code preview), the agent sees a broken example.

More critically: the instruction implies the model should reproduce this outer \`\`\`...\`\`\` wrapper in the \`write_result\` text. If the model does that, the dispatcher regex (\`## spawn-spec\n\`\`\`yaml\n(.*?)\`\`\`\`) will **not match** because the regex looks for \`## spawn-spec\` at the start of the text, not inside a code block.

**Fix:** Remove the outer backtick fence. The Output section should show the literal text the model must produce — directly, not wrapped in a code block. Use indented text or a comment to indicate it's an example if needed. Then the spec text starts with `## spawn-spec` directly, matching what the regex expects.

---

## NEEDS-FIX: Inflight tracking missing for real subagent spawn

The result handler parses the spawn-spec and calls \`Task()\` to spawn the real subagent, but does **not** write an inflight entry before the \`Task()\` call. Per the dispatcher bootup requirement, every \`Task()\` call must be preceded by a synchronous inflight write. The initial prompt-prep spawn is tracked (the flow doc includes \"Write inflight entry\" as step 3), but the real subagent that prompt-prep produces is **not** tracked. This means orphan recovery and \`get_active_sessions\` checks will not see that work is in flight.

**Fix:** Add a \`Bash()\` inflight write immediately before the \`Task()\` call in the result handler, using \`real_task_id\` and \`real_subagent_type\`.

---

## Non-blocking: \`real_task_id\` is computed but unused in \`Task()\`

The result handler extracts \`real_task_id\` from the spec YAML but does not pass it to \`Task()\`. The task_id is embedded in the \`real_prompt\` frontmatter, so the real subagent will use it correctly for its own \`write_result\` call. However, it means the inflight entry (once added per above) and the \`Task()\` call are not formally linked at the handler level. Low risk in practice but worth documenting or aligning.

---

## Non-blocking: Typo in Input contract

In the prompt-prep agent's Input contract section, \`rough_intent: <dispatcher is 1-line guess>\` should read \`<dispatcher's 1-line guess>\`. Minor but makes the spec harder to read.

---

## Checks that passed

- **pyyaml available:** confirmed — \`import yaml\` works in the lobster venv (6.0.3)
- **Regex logic is correct** when the agent produces the expected format (no outer wrapper): \`spec_match\` and \`prompt_match\` both work correctly in testing
- **\`## prompt\` DOTALL scope is safe:** if the prompt body contains a \`## prompt\` heading, it is captured as part of the body and passed through to the real subagent unchanged — no truncation
- **Ack timing is correct:** the user ack is sent before prompt-prep is spawned; by the time the result handler fires, the user already has their acknowledgment
- **\`chat_id=0\` in write_result is correct:** the result handler fires on task_id prefix, not chat_id; the real chat_id is carried in the spawn-spec and prompt frontmatter
- **No hardcoded repo names or instance-specific paths:** agent is portable
- **Frontmatter format matches other agents:** \`name\`, \`description\`, \`model\` — correct
- **Handler position is correct:** inserted before engineer→reviewer routing so it intercepts first

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

All four NEEDS-WORK items from the prior review are confirmed fixed:

1. **Outer backtick fence removed** — `prompt-prep.md` output format spec now starts directly with `## spawn-spec` followed by the inner \`\`\`yaml block. No wrapping fence. The dispatcher regex `## spawn-spec\n\`\`\`yaml\n(.*?)\`\`\`` will match correctly.

2. **Inflight tracking entry added** — The Bash write to `inflight-work.jsonl` appears immediately before `Task()` in the result handler. Position is correct.

3. **`task_id` explicit at Task() call site** — `task_id=real_task_id` is now passed explicitly to `Task()`.

4. **Typo fixed** — `rough_intent: <dispatcher's 1-line guess>` in the Input contract. Confirmed.

**Critical items re-verified:**

- **Regex alignment** — With the outer fence removed, the agent output and dispatcher regex now agree. The handler will correctly parse the spawn spec.
- **chat_id flow** — `chat_id` flows from the agent's frontmatter into the spawn spec YAML and into the composed prompt's frontmatter. No substitution of 0 for the real chat_id.
- **No hardcoded identifiers** — `prompt-prep.md` contains no hardcoded user IDs, chat IDs, or environment-specific values.
- **Handler position** — The prompt-prep result handler block sits before the engineer→reviewer routing (`pr_url_match`), as required.

No regressions found. The fix commit is clean and minimal — it touches exactly the two files expected and addresses all four issues precisely.

sayhar:

PR #1882, opened 2026-04-30
fix(#1573): skip dispatcher sessions in ghost-detector when WFM-active is fresh

### LINKED ISSUE #1573: Ghost detector kill → spurious compact-reminder injection when real dispatcher is alive
## Summary

When `agent-monitor.py` marks the hook-registered dispatcher session (e.g. `672b4ba8`) as dead, `on-fresh-start.py` injects a `0_startup_compact` compact-reminder into the inbox. But the real dispatcher may already be running — this triggers an unnecessary compaction cycle, burns context, and causes a second `wait_for_messages` suppression period.

## Root Cause

`on-fresh-start.py` cannot distinguish between:
- "Real dispatcher died" → compact-reminder injection is appropriate
- "Stale hook-registered placeholder killed" → compact-reminder injection is spurious

The hook-registered session ID is a placeholder, not the actual running dispatcher. When the ghost detector kills it, the real dispatcher (with a different session ID) is still alive and healthy.

## Fix Direction

Option A: `on-fresh-start.py` should check whether the canonical `lobster-dispatcher` session ID is still alive before injecting the compact-reminder. If a live dispatcher session exists, skip injection.

Option B: The hook-registered session should be excluded from ghost detection entirely, since it is a placeholder and not a real agent process.

## Impact

- Unnecessary compaction cycle burns context window
- Second WFM suppression period creates a gap in message processing
- Observed in session with hook-registered session `672b4ba8`

sayhar:
## Summary

- Adds a WFM-active guard to `agent-monitor.py`: if `dispatcher-wfm-active` is fresh (written within 180s), all sessions with `agent_type='dispatcher'` are skipped in the mark-failed sweep — preventing spurious ghost kills of the live dispatcher's hook-registered hex-UUID entry
- Fixes the cron PATH bug: all 3 cron jobs that invoke `uv` now use the absolute path `$HOME/.local/bin/uv` instead of bare `uv` (which is not in cron's minimal PATH); confirmed that all 23,155 lines of agent-monitor.log are `uv: not found` — the ghost-detector has never run via cron until this fix
- Adds Migration 84 to `upgrade.sh` to repair existing installs
- Also adds `agent_type` to the `AgentRow` dataclass and SQL query so the guard can filter by type

## Root Cause

Two dispatcher DB entries exist per session:
1. `agent_id='lobster-dispatcher'` — protected by name guard in `mark_failed_all_ghosts()`
2. Hex-UUID entry (`agent_type='dispatcher'`, no `output_file`) — NOT protected

The hex-UUID entry has no `output_file`, so after 30+ min of WFM idle (JSONL transcript goes quiet), `agent-monitor` classifies it as `STALE_NO_FILE` and kills the live dispatcher.

Additionally, the cron PATH bug meant `agent-monitor.py` never ran via cron — it would have been silently skipping all ghost detection. This is fixed as part of the same PR.

## Fix

1. **`scripts/agent-monitor.py`**: Added `_is_dispatcher_in_wfm()` that reads `~/lobster-workspace/logs/dispatcher-wfm-active` and checks if the timestamp is within 180s. In `mark_failed_all_ghosts()`, if the dispatcher is in WFM, all `agent_type='dispatcher'` sessions are filtered out of the confirmed/stale lists before the mark-failed sweep.

2. **`install.sh`**: Replaced bare `uv` with `$HOME/.local/bin/uv` in all 3 cron entries (ghost-detector, oom-monitor, log-export).

3. **`scripts/upgrade.sh`**: Added Migration 84 to fix existing installs by patching the crontab directly with `sed`.

## Test Plan

- [x] `uv run scripts/agent-monitor.py --no-fs-scan --mark-failed` — confirmed both dispatcher entries are skipped when `dispatcher-wfm-active` is fresh: "No GHOST_CONFIRMED or STALE_NO_FILE agents to mark failed"
- [ ] Verify cron runs successfully with absolute path (will be confirmed in next scheduled cron window)
- [ ] Run Migration 84 on the live install to verify it patches crontab correctly

Closes #1573

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1357,7 +1357,7 @@ success "Nightly consolidation configured (runs at 03:00 nightly)"
 # summary to ~/messages/task-outputs/ (readable via check_task_outputs).
 chmod +x "$INSTALL_DIR/scheduled-tasks/export-logs.py" 2>/dev/null || true
 "$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-LOG-EXPORT" \
-    "0 3 * * * cd $INSTALL_DIR && uv run scheduled-tasks/export-logs.py # LOBSTER-LOG-EXPORT"
+    "0 3 * * * cd $INSTALL_DIR && $HOME/.local/bin/uv run scheduled-tasks/export-logs.py # LOBSTER-LOG-EXPORT"
 
 success "Log export configured (runs at 03:00 UTC daily)"
 
@@ -1371,7 +1371,7 @@ step "Setting up ghost detector cron..."
 # sends a Telegram alert if GHOST_CONFIRMED or UNREGISTERED agents are found,
 # and marks ghost sessions as failed in agent_sessions.db. No LLM involved.
 "$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-GHOST-DETECTOR" \
-    "*/5 * * * * cd $HOME && uv run $INSTALL_DIR/scripts/agent-monitor.py --alert --mark-failed >> $HOME/lobster-workspace/logs/agent-monitor.log 2>&1 # LOBSTER-GHOST-DETECTOR"
+    "*/5 * * * * cd $HOME && $HOME/.local/bin/uv run $INSTALL_DIR/scripts/agent-monitor.py --alert --mark-failed >> $HOME/lobster-workspace/logs/agent-monitor.log 2>&1 # LOBSTER-GHOST-DETECTOR"
 
 success "Ghost detector configured (runs every 5 minutes)"
 
@@ -1386,7 +1386,7 @@ step "Setting up OOM monitor cron..."
 # dispatcher when new OOM kill events are detected. No LLM involved.
 # Only active when LOBSTER_DEBUG=true (the script is a no-op otherwise).
 "$INSTALL_DIR/scripts/cron-manage.sh" add "# LOBSTER-OOM-CHECK" \
-    "*/10 * * * * cd $HOME && uv run $INSTALL_DIR/scripts/oom-monitor.py --since-minutes 10 >> $HOME/lobster-workspace/logs/oom-monitor.log 2>&1 # LOBSTER-OOM-CHECK"
+    "*/10 * * * * cd $HOME && $HOME/.local/bin/uv run $INSTALL_DIR/scripts/oom-monitor.py --since-minutes 10 >> $HOME/lobster-workspace/logs/oom-monitor.log 2>&1 # LOBSTER-OOM-CHECK"
 
 success "OOM monitor configured (runs every 10 minutes, active only when LOBSTER_DEBUG=true)"
 
--- a/scripts/agent-monitor.py
+++ b/scripts/agent-monitor.py
@@ -94,6 +94,7 @@ class AgentRow:
     spawned_at: str
     output_file: str | None
     last_seen_at: str | None
+    agent_type: str | None = None  # 'dispatcher' | 'subagent' | None
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is correct, well-scoped, and addresses the root cause cleanly. Guard 2 (WFM-active signal check) is the right mechanism for protecting the hex-UUID dispatcher entry that Guard 1 (agent_id name match) cannot reach. The cron PATH fix is independently verified: the log evidence (23,155 lines of "uv: not found") confirms the ghost-detector has never run via cron, making this a required correctness fix. All 48 unit tests for the ghost detector pass; the 33 failing tests in the full suite are pre-existing (they test a `handle_write_decision` function that does not yet exist in the codebase) and are unrelated to this PR.

## Findings

### Non-blocking: migration counter undercount in upgrade.sh

The `migrated` counter increments only when the ghost-detector cron entry is patched, but not when the oom-monitor or log-export entries are patched. If only those two entries need fixing (ghost-detector was already correct), the success message will report "0 migrations applied" even though two entries were changed. This is cosmetic — the actual patches are applied correctly.

### Non-blocking: no unit tests for Guard 2 (WFM-active path) in the ghost-detector test file

The existing `TestMarkFailedAllGhostsStaleNoFile` tests construct `AgentRow` without setting `agent_type`, so they exercise only the Guard 1 path. There are no tests in `test_ghost_detector_filesystem.py` that: (a) create a session with `agent_type='dispatcher'` on the row, (b) set a fresh WFM-active signal file, and (c) assert the session is skipped. The `LOBSTER_WFM_ACTIVE_OVERRIDE` env var exists and is tested in `test_wfm_active_signal.py` — it would make writing these tests straightforward. The behaviour is correct (verified manually), but a future refactor of the guard logic could break it without test coverage catching it.

### Non-blocking: tombstone value handled correctly but not documented inline

When the WFM-active file contains "exited" (the tombstone written by `_clear_wfm_active_signal`), `_is_dispatcher_in_wfm()` correctly returns False because `"exited".isdigit()` is False. This is correct, but the function docstring mentions only "a fresh epoch timestamp" and an absent file as cases — it does not mention the tombstone. A future reader who only reads this function may not understand why the `isdigit()` check is needed.

### Confirmed correct: Guard 2 does not block legitimate zombie cleanup

A legitimate dead dispatcher session (e.g., dispatcher crashed, WFM file went stale after 180s) will not be protected by Guard 2 because the WFM-active signal is not refreshed after the process dies. The 180s staleness window is 3× the heartbeat interval, providing the same margin as the existing health-check logic.

### Confirmed correct: existing dispatcher entry protection is not regressed

Guard 1 (agent_id = "lobster-dispatcher" filter) is unchanged. All 48 ghost-detector tests pass.

### Confirmed correct: migration numbering is appropriate for main branch

On `origin/main`, the last migration is 83. Using 84 in this PR is correct. The renaming to 88 visible in local-dev is a local-dev artifact from concurrent PRs; the PR itself targets main correctly.

sayhar (commented):
🤖🦞 Lobster (re-reviewer): PASS — all three prior nits are resolved.

**Nit 1 (migration counter undercount):** Migration 84 in upgrade.sh now increments `migrated` independently for each of the three cron entries (ghost-detector, oom-monitor, log-export). Each sub-step has its own `migrated=$((migrated + 1))` — no more undercount. ✓

**Nit 2 (no unit tests for Guard 2):** `TestGuard2WfmActiveDispatcherType` added with exactly 5 tests covering: fresh signal → NOT killed, stale signal → IS killed, absent file → IS killed, tombstone 'exited' → IS killed, non-dispatcher agent_type → IS killed regardless of WFM state. All 140 ghost-related tests pass (`uv run pytest tests/ -k ghost -q --tb=short`). ✓

**Nit 3 (tombstone handling undocumented):** `_is_dispatcher_in_wfm()` now has a dedicated 'Tombstone case' paragraph in its docstring explaining that 'exited'.isdigit() is False, that _clear_wfm_active_signal() writes that value on WFM exit, and that the raw.isdigit() guard handles it implicitly. ✓

Tests: 140 passed, 0 failed.

sayhar:

PR #1883, opened 2026-04-30
fix(dispatcher): proactively resume pending agenda items after startup catchup

sayhar:
## Summary

- Adds **Proactive resumption after catchup** block to `sys.dispatcher.bootup.md`
- After `startup-catchup` returns, the dispatcher now scans `handoff.md` and the session file for items where the dispatcher (not the user) is the blocking party
- Begins work on the first such item before entering `wait_for_messages`
- Items explicitly waiting for the user's decision are correctly excluded from resumption

## Root Cause

`sys.dispatcher.bootup.md` line 75 said "read for situational awareness... Then `mark_processed`." — no instruction to resume pending work. The dispatcher would passively enter `wait_for_messages` even when it had owned open tasks from the prior session (e.g. a multi-item agenda where it completed items 1-2 but not item 3). Users had to manually re-ask about pending items on every restart.

## Fix

Added a "Proactive resumption after catchup (REQUIRED)" section immediately after the existing startup-catchup handler. It defines:

1. **Active agenda in progress** — if a multi-item agenda had completed earlier items but not a later one, and no queued user message already covers it, begin work
2. **Dispatcher-owes-user thread** — if session notes show the dispatcher promised to follow up but didn't, take the action
3. **Critical distinction** — items waiting for the *user's* decision are correctly passive and excluded

## Test Plan

- [ ] On next restart, verify the dispatcher resumes agenda item 3 (planning/architect layer, msg 29189) without being re-asked
- [ ] Verify items marked "awaiting user sign-off" are NOT proactively acted on

Relates to dispatcher restart behavior — no issue number; behavior-only fix in bootup instructions.

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -80,7 +80,7 @@ When you first start (or after reading this file), follow these steps:
 - **Dispatcher-owes-user thread**: If session notes say "Await user response on X" but you are the one who owes the next action (e.g. you promised to follow up, or the prior session noted the dispatcher should resume), take it now.
 - **Critical distinction**: Do NOT resume items that are waiting for the *user's* decision or response (e.g. "awaiting sign-off on PR #NNNN", "waiting for user go-ahead"). Those are correctly passive. Only act on items where the dispatcher is the blocking party.
 
-If there is nothing to proactively resume, proceed to `wait_for_messages` normally. If there are items to resume, spawn the first one (or begin inline) before entering `wait_for_messages`.
+If there is nothing to proactively resume, proceed to `wait_for_messages` normally. If there are items to resume, spawn the first one (or begin inline) before entering `wait_for_messages`. **When multiple items qualify, take the earliest uncompleted agenda item by sequence number** (e.g. "Agenda item 3" before "Agenda item 5"). If the items are not sequentially numbered, take the topmost qualifying item in the Open Tasks list in `handoff.md`.
 
 **Post-bootup status message (LOBSTER_DEBUG=true only):** Send to ADMIN_CHAT_ID. Keep to 5-8 lines, mobile-friendly. Build it from `handoff.md` (just read for startup) and `msg["text"]` (the catchup summary). Format:
 
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -74,6 +74,14 @@ When you first start (or after reading this file), follow these steps:
 
 **When the startup catchup result arrives** (`task_id: "startup-catchup"`, `chat_id: 0`): read for situational awareness, update `handoff.md` if anything notable changed (failed subagents, open threads). Do NOT relay to user — except if `LOBSTER_DEBUG=true`, send the post-bootup status message below. Then `mark_processed`.
 
+**Proactive resumption after catchup (REQUIRED):** After marking the catchup processed, scan `handoff.md` and the current session file for pending items that require *your* action — not waiting for the user. Specifically:
+
+- **Active agenda in progress**: If there is a multi-item agenda (e.g. "Agenda item 3: planning/architect layer") where earlier items were completed but a later item was not started, AND no queued user message already covers it — begin working on it (spawn a subagent or proceed inline).
+- **Dispatcher-owes-user thread**: If session notes say "Await user response on X" but you are the one who owes the next action (e.g. you promised to follow up, or the prior session noted the dispatcher should resume), take it now.
+- **Critical distinction**: Do NOT resume items that are waiting for the *user's* decision or response (e.g. "awaiting sign-off on PR #NNNN", "waiting for user go-ahead"). Those are correctly passive. Only act on items where the dispatcher is the blocking party.
+
+If there is nothing to proactively resume, proceed to `wait_for_messages` normally. If there are items to resume, spawn the first one (or begin inline) before entering `wait_for_messages`.
+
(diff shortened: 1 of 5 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS (with nits): The core behavior change is sound and the code is correct. A few issues worth addressing before or after merge.

## What this PR does

Three bundled changes:
1. **Bootup instruction** — after the startup summary arrives, the dispatcher now scans for work it owns and begins the first item before going idle.
2. **Ghost detector guard** — a second protection prevents the system from marking a live dispatcher session as failed when the WFM-active heartbeat file is fresh.
3. **Cron PATH fix** — cron entries for three monitoring scripts now use the absolute path to `uv` so they don't fail silently on minimal-PATH environments.

## Issues

### 1. Guard 2 (WFM-active + agent_type) has zero test coverage

The PR adds logic that, when the dispatcher is actively waiting for messages, skips any background session tagged with `agent_type='dispatcher'` during the ghost-killing sweep. This is the main functional change in the monitoring script. There are no tests for it.

The existing `TestMarkFailedAllGhostsStaleNoFile` and `TestLiveDispatcherGuard` classes both cover Guard 1 (filtering by a specific static agent ID). Neither exercises Guard 2. Needed: a test that monkeypatches the WFM-active check to return True, provides a session with `agent_type='dispatcher'` and a UUID-style agent ID (not the static one), and asserts the session is skipped. Counterpart: same setup with the WFM-active check returning False should mark the session failed.

### 2. Test helper silently drops the `agent_type` field

`_make_classified` in the test file accepts an `agent_type` parameter and uses it only in the description string — it never passes it to `AgentRow`. So every test row has `agent_type=None` regardless of what the caller passes. This means Guard 2 can never activate in any existing test, even if the WFM-active file happens to be fresh during a test run on the live machine. The parameter should either be wired through to `AgentRow` or removed.

### 3. Tests don't monkeypatch `_is_dispatcher_in_wfm`

Existing tests that call `mark_failed_all_ghosts` don't patch the WFM-active check. If tests run on the live machine while the dispatcher is inside wait_for_messages (which is its normal steady state), Guard 2 activates silently. Since all test rows currently have `agent_type=None`, this is currently harmless — but once issue #2 above is fixed and rows correctly carry `agent_type='dispatcher'`, tests could fail or produce wrong results depending on whether the dispatcher happens to be running. All `mark_failed_all_ghosts` call sites in the test suite should monkeypatch `_is_dispatcher_in_wfm` to a deterministic value.

### 4. Bootup doc: "first item" ordering is undefined

The new resumption section says "spawn the first one (or begin inline) before entering wait_for_messages" but doesn't define what "first" means when multiple items qualify — by agenda number? By time they were last touched? By priority? In practice the dispatcher will pick based on the order items appear in the handoff file, which is probably fine, but making it explicit ("take the earliest uncompleted agenda item by sequence number") would prevent ambiguity on restarts with complex state.

### 5. Migration counter not incremented for oom-monitor and log-export sub-steps

Only the ghost-detector sub-migration increments `migrated`. If ghost-detector is already fixed but oom-monitor or log-export still have the bare `uv`, the migration applies the fix without telling the caller. The final summary then says "No migrations needed" even though work was done. Each sub-migration should increment `migrated` on success, the same as the ghost-detector does.

## What was checked

- All 48 ghost detector Python tests pass.
- Pre-existing failures: 33 (all in `test_decision_tools` and `test_dispatcher_hint`, unrelated to this PR). Count verified by running `uv run pytest --tb=no -q`.
- Guard 2 logic: correctly checks `agent_type or ""` so `None` values don't get misclassified; correctly filters `confirmed` list in addition to `stale_no_file` (the bug could surface if a dispatcher-typed session somehow ended up in GHOST_CONFIRMED).
- Migration 84 is the correct next migration number relative to the PR's base (main ends at 83).
- WFM-active shell tests: 10/10 pass.

## Summary

The behavioral instruction (issue #1 in the bug description) is correctly written and appropriately scoped — the "critical distinction" bullet is particularly important and is correctly placed. The ghost detector guard (Guard 2) is logically sound. The main gap is test coverage for Guard 2, and the test helper bug that would make such coverage misleading if written naively. These are post-merge fixable nits, not blockers.

sayhar (commented):
🤖🦞 Lobster (re-reviewer): PASS — all five nits from the prior review are resolved.

**Nit verification:**

1. **Guard 2 unit tests** — `TestGuard2WfmActiveDispatcherType` class is present with exactly 4 tests: `test_wfm_active_fresh_skips_dispatcher_typed_uuid_session`, `test_wfm_not_active_marks_dispatcher_typed_uuid_session_failed`, `test_wfm_active_does_not_protect_non_dispatcher_typed_sessions`, `test_wfm_active_skips_dispatcher_typed_in_confirmed_list_too`. ✓

2. **`_make_classified` helper wires `agent_type` to `AgentRow`** — `agent_type=agent_type` is now set directly on the `AgentRow` dataclass (not just via description). ✓

3. **Existing `mark_failed_all_ghosts` tests monkeypatch `_is_dispatcher_in_wfm` to `lambda: False`** — All 7 pre-existing tests in the class now have this monkeypatch applied. ✓

4. **Bootup doc ordering clarified** — The proactive resumption paragraph ends with: "take the earliest uncompleted agenda item by sequence number (e.g. 'Agenda item 3' before 'Agenda item 5'). If the items are not sequentially numbered, take the topmost qualifying item in the Open Tasks list in `handoff.md`." ✓

5. **Migration counter incremented for all three sub-steps** — Migration 84 has separate `migrated=$((migrated + 1))` increments for ghost-detector, oom-monitor, and log-export, each gated on whether the fix was actually applied. ✓

**Test run:** `uv run pytest tests/ -k ghost -q --tb=short` — 139 passed, 0 failed.

sayhar:

PR #1886, opened 2026-05-01
fix(health-check): treat fresh WFM tombstone as exempt to prevent midnight race (#1884)

### LINKED ISSUE #1884: fix(health-check): midnight race — scheduled job wakeup collides with health-check WFM exemption read
## Observed behavior

At exactly 00:00Z, scheduled jobs (lobstertalk-ssh-watcher, lobstertalk-unified, etc.) fire and deposit messages into the inbox. This wakes up `wait_for_messages()`, which writes an \"exited\" tombstone to the `dispatcher-wfm-active` file as it transitions from the WFM idle state into active message processing. The health-check cron also runs at 00:00Z. It detects a stale dispatcher heartbeat (the dispatcher was sitting idle in WFM for a long time), then checks `dispatcher-wfm-active` for the safety exemption — but finds \"exited\" instead of a live timestamp. A non-integer value means no exemption applies, so the check fires RED and kills the dispatcher.

This false-positive kill has been observed at 00:00Z on both Apr 30 and May 1 (log signature: \"RED: dispatcher heartbeat stale\"). The WFM-active exemption works correctly at 10:00Z and 20:00Z, where no scheduled jobs are firing, so the race window never opens.

## Root cause

Race condition between the midnight scheduled-job wakeup (which writes \"exited\" to `dispatcher-wfm-active` during the WFM → processing transition) and the health-check cron (which reads `dispatcher-wfm-active` for the exemption). The file briefly contains \"exited\" — not a live timestamp — during the transition, and midnight is the worst-case scenario because both events fire in the same sub-second window. This is separate from issue #1786 (thinking freeze); the dispatcher here was alive and healthy — the kill was a false positive.

## Proposed fix directions

1. **Treat \"exited\" as \"was in WFM very recently\"** — The health check should recognize that \"exited\" means the dispatcher just left WFM (not that it was never there), and grant the same exemption as a live timestamp would.
2. **Atomic wfm-active transition** — Write the processing-start timestamp to `dispatcher-wfm-active` *before* clearing the WFM value, so there is no window where the file contains a non-exempting value.
3. **Stagger the health-check cron** — Offset the midnight health-check run by a few seconds (e.g., 00:00:05Z) so it always reads `dispatcher-wfm-active` after the transition is complete.

Option 1 is the most defensive and requires the smallest change. Option 2 eliminates the race entirely. Option 3 is a band-aid but quick to deploy.

sayhar:
## Problem

At 00:00Z, the health-check cron kills a healthy dispatcher. Two events collide in the same sub-second window:

1. Scheduled jobs fire and deposit messages into the inbox, waking `wait_for_messages()`.
2. WFM transitions to active processing, writing the tombstone `"exited"` to `dispatcher-wfm-active` via `_clear_wfm_active_signal()`.
3. The health-check cron also fires at 00:00Z, reads `"exited"` from the file, rejects it as non-integer (the correct TOCTOU guard from issue #1730), and falls through to RED — killing the dispatcher even though it was alive and actively processing.

The dispatcher was genuinely healthy. The kill was a false positive caused by the health check seeing the WFM→processing transition tombstone at exactly the wrong moment. This was observed at 00:00Z on Apr 30 and May 1.

## Root cause

`check_dispatcher_heartbeat()` has two cases for the WFM-active file:
- Integer content → check file age against `WFM_ACTIVE_STALE_SECONDS` → exempt if fresh
- Non-integer content (including `"exited"`) → treat as absent → fall through to RED

The tombstone is written *because* the dispatcher just successfully exited WFM and started processing. It is a liveness signal, not an absence signal — but the health check was treating it as the latter.

## Fix

When `wfm_active_ts == "exited"`, check the **file mtime** against a new constant `WFM_TOMBSTONE_GRACE_SECONDS=30`. If the tombstone was written within 30 seconds, the dispatcher is in the WFM→processing transition — grant the same exemption as a live timestamp. If the tombstone is older than 30s, the dispatcher has been in the processing phase for a while and heartbeat staleness is genuinely suspicious, so treat as absent → RED.

30s is well above the sub-second WFM→processing transition window and well below any meaningful "dispatcher is frozen" duration.

## Before / after flow

```
Before (broken at midnight):
  heartbeat stale?
    yes → read wfm_active_file
      integer + fresh → GREEN
      integer + stale → RED
      "exited"        → RED  ← false positive: dispatcher just left WFM
      absent          → RED

After (this PR):
  heartbeat stale?
    yes → read wfm_active_file
      integer + fresh                    → GREEN
      integer + stale                    → RED
      "exited" + mtime fresh (<30s)      → GREEN  ← new: tombstone = recent WFM exit
      "exited" + mtime old   (>=30s)     → RED    ← old tombstone: suspect
      absent                             → RED
```

## Functional patterns

Pure function `check_dispatcher_heartbeat()` extended with a single new branch. Named constant `WFM_TOMBSTONE_GRACE_SECONDS` makes the spec directly reviewable in both implementation and tests — no magic literals.

## Tests run
- [x] `bash tests/test-health-check-wfm-active.sh` — 11/11 passed, 0 failed
  - Test 9 (was: tombstone → RED) is now "fresh tombstone → GREEN" — the race fix
  - Test 10 (new): "old tombstone → RED" — grace period expired
  - Test 11 (renumbered from 10): TOCTOU race regression test
- [x] `bash tests/test-health-check-dispatcher-heartbeat.sh` — 8/8 passed, 0 failed
- [x] `bash tests/test-health-check-boot-grace.sh` — 8/8 passed, 0 failed
- [x] `bash tests/test-health-check-inbox-drain.sh` — 19/19 passed, 0 failed

## Manual test
N/A — this change is entirely in a shell function (`check_dispatcher_heartbeat`) exercised by the test suite. The fix applies to a deterministic file-read path with no external service calls. No systemd, no Telegram API, no DB writes.

## Definition of Done
- [x] Unit tests pass (bash test suite, all 4 suites green)
- [x] Integration/manual test — N/A (deterministic file logic, covered by test suite)
- [x] User-visible outcome: prevents false-positive dispatcher kill at midnight — no Telegram "System recovered automatically" false alarms at 00:00Z
- [x] Side-effect audit: no floods, no shared state writes, no production hits

Closes #1884

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -108,6 +108,15 @@ DISPATCHER_HEARTBEAT_STALE_SECONDS=1200   # 20 min — covers compaction (~5m) +
 # File is deleted by the MCP server when WFM returns (message arrived or timeout).
 DISPATCHER_WFM_ACTIVE_FILE="${LOBSTER_WFM_ACTIVE_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-wfm-active}"
 WFM_ACTIVE_STALE_SECONDS=180   # 3x WAIT_HEARTBEAT_INTERVAL (60s) — absorbs one missed tick
+# Grace window for WFM tombstone ("exited"): if the WFM-active file contains the
+# tombstone but was written within this many seconds, the dispatcher just left WFM
+# and is actively transitioning to message processing — grant the same exemption as
+# a live timestamp. This closes the midnight race (issue #1884): scheduled jobs fire
+# at 00:00Z, wake WFM, which writes "exited" to this file; the health-check cron
+# also fires at 00:00Z and previously declared RED when it saw "exited" instead of
+# a live integer.  A 30s window is well above the sub-second WFM→processing
+# transition but narrow enough to never mask a genuinely frozen dispatcher.
+WFM_TOMBSTONE_GRACE_SECONDS=30
 
 OUTBOX_DIR="$MESSAGES_DIR/outbox"
 OUTBOX_STALE_THRESHOLD_SECONDS=900   # 15 min = RED
@@ -1008,6 +1017,25 @@ check_dispatcher_heartbeat() {
             else
                 log_error "RED: dispatcher heartbeat stale (${age}s) and WFM-active also stale (${wfm_age}s, threshold: ${WFM_ACTIVE_STALE_SECONDS}s) — dispatcher appears frozen"
                 return 2
+            fi
+        fi
+        # Tombstone check (issue #1884 midnight race): inbox_server.py writes
+        # "exited" to WFM_ACTIVE_FILE when WFM returns (instead of deleting the
+        # file, to avoid the TOCTOU race fixed in issue #1730).  If the tombstone
+        # was written within WFM_TOMBSTONE_GRACE_SECONDS, the dispatcher just left
+        # WFM and is actively transitioning to message processing — this is healthy,
+        # not a freeze.  Only grant the exemption when the file mtime is fresh;
+        # an old tombstone (dispatcher has been processing for longer than the grace
+        # window) means the heartbeat staleness is genuinely suspicious.
+        if [[ "$wfm_active_ts" == "exited" ]]; then
+            local file_mtime
+            file_mtime=$(stat -c %Y "$DISPATCHER_WFM_ACTIVE_FILE" 2>/dev/null || echo 0)
+            local tombstone_age=$(( now - file_mtime ))
+            if [[ $tombstone_age -le $WFM_TOMBSTONE_GRACE_SECONDS ]]; then
+                log_info "Dispatcher heartbeat stale (${age}s) but WFM-active tombstone is fresh (written ${tombstone_age}s ago, grace: ${WFM_TOMBSTONE_GRACE_SECONDS}s) — dispatcher just left WFM, skipping RED"
+                return 0
+            else
+                log_info "Dispatcher heartbeat stale (${age}s) and WFM-active tombstone is old (${tombstone_age}s, grace: ${WFM_TOMBSTONE_GRACE_SECONDS}s) — treating as absent"
             fi
         fi
         log_error "RED: dispatcher heartbeat stale — last tool use ${age}s ago (threshold: ${DISPATCHER_HEARTBEAT_STALE_SECONDS}s)"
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**All 4 test suites run and pass: 11/11, 8/8, 8/8, 19/19**

## What the fix does

When the health check sees the "exited" marker in the WFM-active file (written to signal that the dispatcher just finished waiting for messages), it now checks how recently that marker was written. If it was written within the last 30 seconds, the health check treats the dispatcher as healthy rather than killing it. If the marker is older than 30 seconds, the old behavior applies (potentially RED).

## Cold-read findings vs. engineer's briefing

The engineer's briefing accurately described the problem. Nothing missed.

### Additional edge case I checked independently

**Stat-fails-after-cat path**: If the WFM-active file is deleted between the cat read and the subsequent mtime check, `stat` returns a failure. The fallback is `echo 0`, making `file_mtime=0`. This gives a `tombstone_age` on the order of the Unix epoch (~1.7 billion seconds), which is far above the 30-second grace window — the check correctly falls through to RED rather than granting a false GREEN. The "fail-safe-to-RED" property is preserved. No test exists for this specific case, but the `|| echo 0` fallback is correctly conservative: it can only produce false negatives (unnecessary kills), never false positives (missed kills). Given the tombstone window is 30 seconds and the health check runs on a cron cadence, the practical risk of this race is negligible.

## Test structure review

- No duplicate class names or shadowing issues.
- Test 9 (previously "tombstone → RED") correctly replaced with "fresh tombstone → GREEN". The old behavior is now tested as Test 10 ("old tombstone → RED"). The TOCTOU regression (Test 11) is preserved and still green.
- The test for the old behavior does exercise the actual before-state via the backdated mtime (`touch -d '120 seconds ago'`) — not just a comment assertion.
- `TOTAL` increments dynamically via `begin_test`, so the count is structurally correct and not subject to off-by-one from manual editing.
- Summary line updated to reflect 11 expected tests.

## Logic correctness

The `stat -c %Y` call on Linux correctly returns file modification time as a Unix epoch. `now` is captured once at the top of `check_dispatcher_heartbeat` and reused, so there's no clock skew between the heartbeat staleness calculation and the tombstone age calculation. Named constant `WFM_TOMBSTONE_GRACE_SECONDS=30` defined in both implementation and test harness — no magic literals.

**No issues found. Ready to merge.**

sayhar:
Superseded by #1908 simplified flag-file detection approach. This PR is no longer needed.

PR #1891, opened 2026-05-01
feat(hooks): add context injection logging to inject-bootup-context.py

### LINKED ISSUE #1889: feat: context injection logging in inject-bootup-context.py
## Problem

When debugging whether dispatchers got the right bootup file vs the subagent bootup file, there is currently no log record of what was injected into context for each agent session.

## Goal

Add append-only logging to `inject-bootup-context.py` (the SessionStart hook) that records exactly which files were injected per session, enabling post-hoc debugging of bootup context delivery.

## Proposed log format

```
<ISO UTC timestamp> | session=<session_id_or_unknown> | role=<dispatcher|subagent|unknown> | injected=[file1.md, file2.md, ...]
```

Log path: `~/lobster-workspace/logs/context-injection.log`

## Constraints

- Do NOT change what gets injected — logging only
- Do NOT modify any bootup content files
- The log file is append-only; create it if it doesn't exist
- Keep the hook fast — no network calls, no blocking I/O beyond the append
- Only log files that were actually injected (exist and were read successfully)

## Acceptance criteria

- Every SessionStart hook run appends one line to the log
- The line includes session ID, detected role (dispatcher/subagent/unknown), and list of injected file names
- If no files were injected, `injected=[]` is written
- The hook remains functionally identical — no change to what gets printed to stdout

sayhar:
🤖🦞 Lobster (engineer):

## What this solves

Before this change, there was no record of which bootup files each agent session received. When debugging whether a dispatcher got the wrong bootup (e.g. subagent context instead of dispatcher context), the only option was to reproduce the problem — there was no history to inspect.

This adds an append-only log so that every SessionStart hook run leaves a one-line trace: what session ID, what role was detected, and which files were injected.

## What the log looks like

```
2026-05-01T18:23:41Z | session=abc-1234 | role=dispatcher | injected=[sys.dispatcher.bootup.md, user.base.bootup.md, user.dispatcher.bootup.md]
2026-05-01T18:24:02Z | session=xyz-5678 | role=subagent | injected=[sys.subagent.bootup.md, user.base.bootup.md]
2026-05-01T18:24:15Z | session=unknown | role=subagent | injected=[]
```

Log path: `~/lobster-workspace/logs/context-injection.log`

## Design choices

**`_inject_if_exists` now returns bool** — previously void. This is the minimal change needed to track which optional user config files were actually injected (as opposed to skipped because they don't exist). The function's external behavior is unchanged; the return value was previously discarded.

**`CONTEXT_INJECTION_LOG` is a module-level constant** — so tests can override it with `mod.CONTEXT_INJECTION_LOG = tmp_path / "test.log"` using the same pattern used by all other inject-bootup-context tests.

**Logging errors are silently swallowed** — the `except Exception: pass` in `_append_injection_log` ensures that a full disk, permissions error, or unexpected exception never prevents context from being injected. The hook's primary job is injection, not logging.

**Log is written after injection is complete** — the log entry captures what was actually injected, not what was attempted. If a file read fails mid-injection, that file is not in the injected list.

**Early exit path also writes a log entry** — when the system bootup file is missing (`content is None`), `_append_injection_log` is called before `sys.exit(0)` so the event is recorded with an empty injected list.

## Functional patterns used

Pure function for log line construction — `_append_injection_log` takes explicit inputs (session_id, role, injected list) and has no implicit dependencies. The module-level path constant is the only exception, and it is overridable in tests.

## What is NOT changed

- What gets injected and in what order is unchanged
- stdout output is identical — the log goes to a file, not stdout
- No bootup content files were modified
- No changes to `session_role.py` or any other hook

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_inject_bootup_context_logging.py -v` — 16 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/test_hooks/test_inject_bootup_context_sentinel.py -v` — 9 passed, 0 failed (existing tests, unchanged)
- [x] `uv run pytest tests/unit/` — 3071 passed, 5 failed (same 5 pre-existing failures in `test_require_background_agent.py`, confirmed failing on `origin/main` before this change)

## Manual test

N/A — this change is purely an append to a local log file. No external services, no systemd, no cron, no database. The log path is under `~/lobster-workspace/logs/` which the system already owns and writes to.

## Definition of Done
- [x] Integration/manual test run — N/A, local file write only
- [x] User-visible outcome verified — N/A, internal diagnostic log
- [x] Side-effect audit complete: one append-only write per SessionStart; bounded, no volume risk

Closes #1889

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -36,6 +36,7 @@ case and ensures dispatcher bootup is always injected when it should be.
 import json
 import os
 import sys
+from datetime import datetime, timezone
 from pathlib import Path
 
 # Allow imports from the hooks directory (session_role).
@@ -54,6 +55,13 @@ USER_DISPATCHER_BOOTUP = USER_CONFIG_DIR / "user.dispatcher.bootup.md"
 USER_SUBAGENT_BOOTUP = USER_CONFIG_DIR / "user.subagent.bootup.md"
 
 HOOK_NAME = "inject-bootup-context"
+
+# Append-only log of context injections — one line per hook run.
+# Populated at import time so tests can override by setting mod.CONTEXT_INJECTION_LOG.
+_LOBSTER_WORKSPACE = Path(
+    os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace")
+)
+CONTEXT_INJECTION_LOG = _LOBSTER_WORKSPACE / "logs" / "context-injection.log"
 
 # Compact-pending sentinel written by on-compact.py for dispatcher compactions only.
 # Used as the Option 3 fallback: sentinel present + LOBSTER_MAIN_SESSION=1 → dispatcher.
@@ -108,19 +116,51 @@ def _read_file_safe(path: Path, label: str) -> str | None:
         return None
 
 
-def _inject_if_exists(path: Path, label: str) -> None:
-    """Read and print file contents if the file exists and is non-empty. Silent skip otherwise."""
+def _inject_if_exists(path: Path, label: str) -> bool:
+    """Read and print file contents if the file exists and is non-empty.
+
+    Returns True if the file was successfully injected, False otherwise.
+    Silent skip when the file is absent.
+    """
     if not path.exists():
-        return
+        return False
     try:
         content = path.read_text()
         if content.strip():
             print(content)
+            return True
+        return False
     except OSError as exc:
         print(
             f"[{HOOK_NAME}] WARNING: could not read {path} ({label}): {exc}",
             file=sys.stderr,
         )
+        return False
+
+
+def _append_injection_log(
+    session_id: str,
+    role: str,
+    injected_files: list[str],
+) -> None:
+    """Append one line to the context injection log.
+
+    Format:
+      <ISO UTC timestamp> | session=<id> | role=<role> | injected=[file1, file2, ...]
+
+    Creates the log file and any missing parent directories if needed.
+    Errors are swallowed — logging must never break the hook.
+    """
+    try:
+        timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+        files_repr = "[" + ", ".join(injected_files) + "]"
+        line = f"{timestamp} | session={session_id} | role={role} | injected={files_repr}\n"
+        log_path = CONTEXT_INJECTION_LOG
+        log_path.parent.mkdir(parents=True, exist_ok=True)
+        with log_path.open("a") as fh:
+            fh.write(line)
+    except Exception:  # noqa: BLE001
+        pass  # logging must not break injection
 
 
 def main() -> None:
@@ -129,6 +169,8 @@ def main() -> None:
         hook_input = json.load(sys.stdin)
     except (json.JSONDecodeError, ValueError):
         hook_input = {}
+
+    session_id = hook_input.get("session_id", "unknown")
 
     is_dispatcher = session_role.is_dispatcher(hook_input)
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Clean, minimal change. Adds an append-only log file (`~/lobster-workspace/logs/context-injection.log`) that records one line per SessionStart hook run. No changes to injection logic, no stdout pollution, no external dependencies.

## Findings by area

### 1. `_inject_if_exists` return type change (None → bool)

All three call sites in `main()` are correctly updated to use the return value:

```python
if _inject_if_exists(USER_BASE_BOOTUP, "user.base.bootup.md"):
    injected.append(USER_BASE_BOOTUP.name)
```

The only place `_inject_if_exists` is called that doesn't capture the return is... nowhere. All call sites capture it. The function's injection behavior is identical to before; the `return False` cases are the existing early-exit paths that previously had implicit `return None`. No behavioral regression.

One edge case to note: a file that exists but has only whitespace returns `False` (the `if content.strip():` guard) and will not appear in the injected list. This is correct — whitespace-only files were already silently skipped before this PR.

### 2. `_append_injection_log` bare `except Exception`

The `except Exception: pass` pattern is intentional and appropriate here. The `# noqa: BLE001` annotation is correct (BLE001 = blind-except). The function has no implicit dependencies other than the module-level `CONTEXT_INJECTION_LOG` path constant; all inputs are explicit parameters. The failure modes it needs to swallow (full disk, permissions error, unexpected path issues) span multiple exception types, so `Exception` is the right catch width — catching `OSError` alone would miss `UnicodeEncodeError` on unusual locales, for example. This is the correct tradeoff for a logging function whose failure must never abort injection.

If you want to tighten later, `(OSError, ValueError, UnicodeError)` would cover the realistic failure modes without being a pure bare catch.

### 3. `CONTEXT_INJECTION_LOG` at module import time — cross-test pollution

The module-level constant is set once at import time from `LOBSTER_WORKSPACE` env var. Tests override it with `mod.CONTEXT_INJECTION_LOG = tmp_path / "test.log"` after module load. This is the same pattern used by all other tests in this file (e.g. `DISPATCHER_BOOTUP`, `COMPACT_PENDING_SENTINEL`).

Cross-test isolation is guaranteed because each test call to `_run_hook` uses `importlib.util.spec_from_file_location` with a `uuid4`-suffixed module name — a fresh module instance per call. There is no shared module state between tests. No pollution risk.

### 4. Log format

Format is useful for the stated debugging goal (which bootup did this session receive?):

```
2026-05-01T18:23:41Z | session=abc-1234 | role=dispatcher | injected=[sys.dispatcher.bootup.md, user.base.bootup.md, user.dispatcher.bootup.md]
```

The four-field pipe-delimited format is grep-friendly. `session=` maps directly to the Claude Code session UUID visible in hook input. `role=` directly answers "was this a dispatcher or subagent start?" `injected=[]` directly answers "what files did this session actually receive?"

One minor note: the injected list uses `.name` (filename only, not full path), which is the right choice — paths are long and the filenames are unique enough to be unambiguous.

### 5. Test coverage (16 tests)

Coverage is solid:
- Log file creation from absent state
- Append-only behavior (pre-existing content preserved)
- Three-run accumulation (exactly N lines after N runs)
- All four pipe-delimited fields validated
- `session=unknown` fallback when hook_input has no session_id
- `role=dispatcher` vs `role=subagent` correctly reflected
- Empty `injected=[]` when system bootup file is absent (early exit path)
- Optional user files included/excluded based on existence
- stdout unchanged (no log content leaks to injected context)

The test for the empty-injected early-exit path (`test_injected_is_empty_list_when_no_files_injected`) is particularly important — it confirms the `_append_injection_log` call before `sys.exit(0)` works correctly. This test is more verbose than the others (doesn't use `_run_hook`) but functionally correct.

One gap: there is no test for a file that exists but is whitespace-only returning `False` from `_inject_if_exists` and therefore not appearing in the log. This is a minor gap — the behavior is correct and covered transitively by the empty-injected test — but a direct test would strengthen the spec. Not a blocker.

### 6. Pre-existing failures

5 failures in `test_require_background_agent.py` are confirmed pre-existing on `origin/main` (same failures, same assertions, same exit code mismatch). Unrelated to this PR — they appear to be from issue #1872 (frontmatter sentinel feature not yet merged to origin/main).

### 7. Behavioral regression check

The injection logic (order, files, stdout output) is structurally identical to `origin/main`. The only additions are:
- `session_id = hook_input.get("session_id", "unknown")` before `is_dispatcher` check
- `role = ...` and `injected: list[str] = []` initialized before injection
- `injected.append(...)` calls after each successful injection
- `_append_injection_log(...)` called at both exit points

The rename from `session_id` to `sid` in the dispatcher write path (to avoid shadowing the new `session_id` variable) is correct and clean.

## Verdict

**PASS** — deploy to local-dev.

sayhar:

PR #1896, opened 2026-05-01
feat(logging): add crash logging and session lifecycle log

### LINKED ISSUE #1895: feat(logging): add crash logging and session lifecycle log for CC dispatcher
## Problem

On 2026-05-01 at ~13:40Z, the CC dispatcher crashed mid-`wait_for_messages` call with no logged exit reason. Post-incident investigation revealed that we can only infer the 3.5-minute WFM block from log gaps — there's no direct evidence of what happened or why the process died.

## Root Cause

The CC process lifecycle is not persistently logged. The wrapper script (`claude-persistent.sh`) logs Claude's exit code at a high level, but nothing captures:
- The exact exit reason (OOM, signal, CC internal error)
- When `wait_for_messages` was entered and exited (and for how long)
- The session ID at the time of crash (for cross-referencing with transcripts)

## Gaps Identified

| Gap | Impact |
|-----|--------|
| No exit reason logged | Can't distinguish OOM, SIGKILL, auth failure, or CC bug |
| Session lifecycle not persistently recorded | Start/end only visible via /tmp JSONL transcripts (cleaned up) and lock file inference |
| WFM blocking duration invisible | 3.5-min WFM block inferred from log gaps, not measured |
| Last-tool-call before death unrecoverable | Without a persistent session log, can't see what the dispatcher was doing |

### A. Session lifecycle log (`~/lobster-workspace/logs/session-lifecycle.log`)

One JSON line per event, written by:
- The `claude-persistent.sh` wrapper: CC_START, CC_EXIT events
- New PreToolUse/PostToolUse hooks for `wait_for_messages`: WFM_ENTER, WFM_EXIT events

```json
{"ts": "2026-05-01T13:36:35Z", "event": "CC_START", "pid": 1042501, "session_id": "547147bf", "attempt": 1}
{"ts": "2026-05-01T13:36:40Z", "event": "WFM_ENTER", "session_id": "547147bf"}
{"ts": "2026-05-01T13:40:02Z", "event": "CC_EXIT", "pid": 1042501, "exit_code": 1, "session_id": "547147bf", "reason": "unknown"}
{"ts": "2026-05-01T13:40:15Z", "event": "CC_START", "pid": 1055880, "session_id": "4e19feb6", "attempt": 2}
```

### B. Implementation

1. **`scripts/session-lifecycle-logger.sh`** — sourced by `claude-persistent.sh` to write CC_START and CC_EXIT events
2. **`hooks/wfm-lifecycle-logger.py`** — PreToolUse hook that writes WFM_ENTER; PostToolUse hook that writes WFM_EXIT with duration
3. **`scripts/upgrade.sh`** migration — registers the new hooks in `~/.claude/settings.json`

### C. Design constraints

- Log entries use JSONL (one event per line) for easy `grep` and `jq` parsing
- All writes are atomic (write to `.tmp`, then `rename()`)
- Hooks fail open — never block tool execution
- Dispatcher-only: WFM hooks skip subagent sessions
- The log is append-only and rotated via existing log retention (no new cron needed)

## Acceptance Criteria

- [ ] `session-lifecycle.log` contains CC_START on next lobster restart
- [ ] `session-lifecycle.log` contains WFM_ENTER/WFM_EXIT on each `wait_for_messages` call
- [ ] `session-lifecycle.log` contains CC_EXIT with exit code on next crash or clean shutdown
- [ ] Existing hooks (heartbeat, pretooluse-heartbeat) continue to work unchanged
- [ ] upgrade.sh migration registers the hooks idempotently

sayhar:
## Problem solved

On 2026-05-01 at ~13:40Z, the CC dispatcher died mid-`wait_for_messages` with no logged exit reason. The only evidence was a 3.5-minute gap in the heartbeat log. There was no way to know the exit code, which signal (if any) killed the process, or how long the WFM call had been blocking.

This PR closes that gap by writing a persistent JSONL session lifecycle log at `~/lobster-workspace/logs/session-lifecycle.log`.

## What changed

**A. `scripts/claude-persistent.sh` — CC_START and CC_EXIT events**

Added three bash functions (`write_lifecycle_event`, `write_cc_start`, `write_cc_exit`) and wired them into `launch_claude()`:
- `CC_START` is written from inside the subshell, *before* `exec claude`, capturing `$BASHPID` (the actual claude PID after exec) and the attempt number.
- `CC_EXIT` is written after the `claude` pipeline returns, capturing the PID from the dispatcher.pid file and the exit code.

Both events are written even if CC exits immediately (the write happens before exec and right after the pipeline returns respectively).

**B. `hooks/wfm-lifecycle-logger.py` — WFM_ENTER and WFM_EXIT events**

A new hook (PreToolUse + PostToolUse) that fires only for dispatcher sessions:
- `WFM_ENTER`: records when `wait_for_messages` is called, writes entry timestamp to a state file
- `WFM_EXIT`: reads the entry timestamp, computes `duration_s`, extracts message count from the tool response

Uses module-level imports of `is_dispatcher` / `is_dispatcher_session` from `session_role` so the dispatcher guard is monkeypatchable in tests and never silently skipped.

**C. `scripts/upgrade.sh` — Migration 90**

Registers the two WFM lifecycle logger hook entries in `~/.claude/settings.json` idempotently. Migration skips cleanly if both entries are already present.

## Functional patterns used

- Pure functions for all I/O primitives (`_append_event`, `_write_enter_ts`, `_read_enter_ts`, `_parse_message_count`) — each has a single clearly-defined input/output contract and is independently testable.
- Fail-open: all exceptions are caught at the boundary; hooks always exit 0 so tool execution is never blocked.
- Immutability: each event is a freshly constructed dict, never mutated.
- Module-level function references for `is_dispatcher` / `is_dispatcher_session` — allows clean monkeypatching in tests without patching inside function closures.

## Log format (JSONL)

```json
{"ts":"2026-05-01T13:36:35Z","event":"CC_START","pid":1042501,"attempt":1}
{"ts":"2026-05-01T13:36:40Z","event":"WFM_ENTER","session_id":"547147bf"}
{"ts":"2026-05-01T13:40:02Z","event":"WFM_EXIT","session_id":"547147bf","duration_s":202.4,"messages":0}
{"ts":"2026-05-01T13:40:05Z","event":"CC_EXIT","pid":1042501,"exit_code":1,"attempt":1}
```

## Before / after (crash investigation flow)

Before:
```
health-check: gap in heartbeat log at 13:40Z
claude-session.log: empty (nothing logged at exit)
lobster-state.json: mode=active (never transitioned)
cause: unknown
```

After:
```
session-lifecycle.log:
  CC_START at 13:36:35 pid=1042501
  WFM_ENTER at 13:36:40 — dispatcher entered WFM
  CC_EXIT at 13:40:02 exit_code=1 — crashed 202s into WFM block
cause: non-zero exit during WFM (likely OOM or SIGKILL)
```

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_wfm_lifecycle_logger.py -v` — 16 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v --tb=short` — 411 passed, 9 pre-existing failures (in `test_require_background_agent.py` and `test_require_write_result.py`, unrelated to this PR, confirmed failing on `main` too)

## Manual test

N/A — this change writes to a local log file. The output is entirely internal; no Telegram/user-visible behavior is altered. The lifecycle log will be populated on the next lobster restart after upgrade.sh is run (or after migration 90 is applied). Visible in `~/lobster-workspace/logs/session-lifecycle.log` on first WFM call.

## Definition of Done

- [x] Integration/manual test run — N/A, log file write only
- [x] User-visible outcome — not applicable (internal logging only)
- [x] Side-effect audit: writes append-only to a single log file, no floods, no shared state writes that affect other jobs

Closes #1895


sayhar (commented):
## Review: PR #1896 — WFM/CC Session Lifecycle Logging

Overall: solid, well-structured work. The core implementation is correct and the test coverage is thorough. A few concrete issues to address, mostly around install.sh completeness and one asymmetry in the dispatcher guard.

---

### 1. install.sh missing the wfm-lifecycle-logger hook registration

**This is a blocking gap for fresh installs.** Migration 90 in `upgrade.sh` handles existing installs, but `install.sh` has no corresponding block for `wfm-lifecycle-logger.py`. The pattern in this repo is that every hook that upgrade.sh adds via migration must also be wired in install.sh so fresh installs pick it up. Currently a user running install.sh from scratch will have the hook file present but no settings.json entries for it — no WFM_ENTER or WFM_EXIT events will ever fire.

The pretooluse-heartbeat hook added in this same PR demonstrates the correct pattern (install.sh lines 2020–2043). A matching block for wfm-lifecycle-logger should follow the same structure, registering both the PreToolUse (WFM_HOOK_TYPE=pre) and PostToolUse (WFM_HOOK_TYPE=post) entries with `matcher: "mcp__lobster-inbox__wait_for_messages"`.

---

### 2. Dispatcher guard asymmetry between WFM_ENTER and WFM_EXIT

In `hooks/wfm-lifecycle-logger.py`:

- `handle_pre_tool_use` guards with `is_dispatcher_session(hook_input)` — good for PreToolUse, since the session UUID is available.
- `handle_post_tool_use` guards with `is_dispatcher(hook_input)` — this is the simpler check that only looks at the stored session ID.

These two functions have different semantics in `session_role.py`. `is_dispatcher_session` is the right guard for hook contexts where you have a fresh session ID to compare. Using `is_dispatcher` in the PostToolUse path means a stale stored session ID could cause a mismatch (WFM_ENTER written, WFM_EXIT skipped), producing orphaned enter-ts files and misleading duration-absent EXIT events.

Recommend making both handlers use `is_dispatcher_session` for consistency, or explicitly document why they intentionally differ.

---

### 3. Acknowledged concerns — assessment

**CC_EXIT signal capture (concern 1):** Exit code 137 (128+9) is sufficient for the primary use case stated in the PR (distinguishing OOM/SIGKILL from clean exits). Reading `/proc/$PID/status` or parsing `dmesg` from a bash function that runs after the pipeline returns is fragile — the process is already gone and OOM killer messages in dmesg are best-effort. The current approach is pragmatically correct. This can be a follow-up if richer signal attribution is needed later.

**Log rotation (concern 2):** The log will grow indefinitely. On a healthy system with one WFM call per message, this is roughly 2 lines per message processed — probably acceptable for months. But it should be tracked as a follow-up issue. A cron-based logrotate entry or a size check in `write_lifecycle_event` would be the right fix; leaving it as-is now is acceptable with a filed issue.

**Migration 90 idempotency check (concern 3):** The contains("wfm-lifecycle-logger") check on the command string is correct for the current hook path. If the path changes, a duplicate entry is the failure mode — not a silent skip. That's safe and acceptable. The risk is low because hook path changes would also require updating the chmod and migration steps.

**CC_EXIT PID fallback (concern 4):** Falling back to pid=0 when the PID file is absent is acceptable. The important field for crash investigation is `exit_code`, not the PID (which can be cross-referenced from CC_START if needed). pid=0 is clearly a sentinel and not a real PID, so it won't mislead investigation.

---

### 4. Minor: `session_lifecycle.log` not in the new `services/lobster-mcp-local.service`

The hardcoded service file added at `services/lobster-mcp-local.service` does not set `LOBSTER_WORKSPACE`, so `wfm-lifecycle-logger.py` would fall back to `$HOME/lobster-workspace` (the default). This is fine as long as `$HOME` resolves correctly under the service's User=lobster context, which it should. No change needed, just confirming it was considered.

---

### 5. `next-migration.txt` bumped to 91 but only migrations through 90 added

`scripts/next-migration.txt` was updated from 82 to 91 (+9 migrations: 82→90 inclusive). That's correct — migrations 84 through 90 are present in the diff. Confirmed consistent.

---

### Summary

The two actionable items before merge:
1. **Required:** Add wfm-lifecycle-logger hook registration to `install.sh` (same pattern as pretooluse-heartbeat, lines 2020–2043 in the diff).
2. **Recommended:** Unify the dispatcher guard in `handle_post_tool_use` to use `is_dispatcher_session` instead of `is_dispatcher`, matching `handle_pre_tool_use`.

Everything else — the JSONL format, the pure-function architecture, the atomic write patterns, the test structure, the upgrade.sh migration — is well done.

sayhar (commented):
PASS — all nit fixes verified clean.

Checked in commit 7bffa29c:

1. **install.sh wfm-lifecycle-logger block** — correctly placed after the pretooluse-heartbeat block (lines 1795–1838). Idempotency is sound: separate `jq -e` guards check for existing PreToolUse and PostToolUse entries independently before appending. Both hooks use the correct `matcher: mcp__lobster-inbox__wait_for_messages` and `WFM_HOOK_TYPE=pre/post` env var pattern.

2. **Dispatcher guard consistency** — both `handle_pre_tool_use` (line 135) and `handle_post_tool_use` (line 150) now call `is_dispatcher_session(hook_input)`. No remaining `is_dispatcher` calls anywhere in the file.

3. **No unused `is_dispatcher` import** — the only import from `session_role` is `is_dispatcher_session` (line 58). The docstring mention of `is_dispatcher` on line 27 is historical documentation only, not an import.

4. **16/16 unit tests pass** — `tests/unit/test_hooks/test_wfm_lifecycle_logger.py` passes clean with no failures.

No new issues found. PR is ready to merge.

sayhar:

PR #1901, opened 2026-05-01
fix(health-check): scope dispatcher heartbeat writes to dispatcher session only

### LINKED ISSUE #1897: bug(health-check): subagent activity masks dispatcher death — health check declares GREEN during dispatcher crash
## Summary

The health check cannot distinguish dispatcher tool use from subagent tool use. When the dispatcher crashes while blocked in `wait_for_messages`, any subagent that is still running will continue updating the dispatcher heartbeat file, causing the health check to declare GREEN even though the dispatcher is dead.

## Evidence from session 009 crash (2026-05-01)

| Time (UTC) | Event |
|---|---|
| 13:36:35 | Dispatcher entered `wait_for_messages` |
| 13:36:35–13:40:02 | Dispatcher CC process died (exact moment unknown) |
| 13:39:59 | Investigation subagent made a Bash call |
| ~13:40:00 | Health check saw "last tool use 3s ago" → declared GREEN |
| 13:40:02 | New dispatcher session started via `claude-persistent.sh` restart logic |

The health check **should have triggered a RED restart** but did not, because the subagent's Bash call reset the dispatcher-heartbeat timestamp.

## Root cause

`hooks/thinking-heartbeat.py` is a PostToolUse hook. It fires on **every tool call in the Claude Code session** — dispatcher or subagent. The hook writes the current Unix epoch to `~/lobster-workspace/logs/dispatcher-heartbeat`.

`health-check-v3.sh` reads this single file in `check_dispatcher_heartbeat()` (lines ~978–1048) and compares its age against `DISPATCHER_HEARTBEAT_STALE_SECONDS` (600s / 10 min). There is no mechanism to distinguish a dispatcher tool call from a subagent tool call. A subagent running inside the same Claude Code session writes to the exact same file.

The WFM-active signal (`dispatcher-wfm-active`, written by `inbox_server.py`) does not help here: when the dispatcher CC process dies while blocked in WFM, the MCP server also dies, so the WFM-active file goes stale. But the heartbeat file stays fresh because the subagent — running in a separate tmux pane or in-process forked context — keeps writing to it.

## Impact

As long as any subagent is active, the health check will never detect a dispatcher crash. The 600s threshold (PR #1887) was intended to reduce MTTR, but is completely bypassed when subagents keep the heartbeat alive. MTTR in this failure mode is unbounded — it only resolves when all subagents finish naturally or when `claude-persistent.sh` triggers a restart via a different mechanism.

## What to investigate before closing

1. **Confirm write path**: Verify whether subagent tool calls truly share the same PostToolUse hook context as the dispatcher, or whether there's an existing session isolation mechanism being bypassed.

2. **Check WFM-active interplay**: Determine whether `inbox_server.py` keeps writing WFM-active timestamps even when the dispatcher CC is dead but the MCP server is still up (could create a secondary masking path).

3. **Proposed fix direction**: The dispatcher should write a **DISPATCHER-ONLY heartbeat** that subagents cannot touch. Options:
   - Write the heartbeat only when `session_id` matches a stored dispatcher session ID (hook checks env var against a known dispatcher session ID file)
   - Write to a separate file keyed on the dispatcher's PID (subagent processes have different PIDs, though they may share the parent session environment)
   - Use `thinking-heartbeat.py` exclusively for WFM-adjacent signals, and require the dispatcher to write an explicit heartbeat in its main loop (not via PostToolUse)
   - Filter in the hook: check whether the current tool call is from the top-level dispatcher context (e.g., check `CLAUDE_CODE_SUBAGENT` env var if one exists)

## Files to read

- `~/lobster/hooks/thinking-heartbeat.py` — the hook that writes the shared heartbeat file
- `~/lobster/scripts/health-check-v3.sh` — specifically `check_dispatcher_heartbeat()` (~line 978)
- `~/lobster/src/mcp/inbox_server.py` — WFM-active signal (around line 826)
- `~/messages/config/dispatcher.pid` — dispatcher PID file (may be usable as isolation mechanism)

sayhar:
## Problem

The health check cannot distinguish dispatcher tool use from subagent tool use. When the dispatcher crashes, any still-running subagent keeps updating `dispatcher-heartbeat` and `dispatcher-pre-tool-heartbeat`, keeping the health check GREEN indefinitely. This was observed in the session 009 crash (2026-05-01): at 13:40:00 an investigation subagent's Bash call reset the heartbeat, so the health check reported GREEN instead of triggering a restart. MTTR in this failure mode is unbounded.

Root cause: `thinking-heartbeat.py` (PostToolUse) and `pre-tool-heartbeat.py` (PreToolUse) wrote their files unconditionally, with no mechanism to tell whether the hook was firing for the dispatcher or a subagent.

## Fix

Both hooks now read `hook_input` from stdin and check the current session against the stored dispatcher session ID before writing. Subagent calls exit 0 silently without touching the heartbeat files.

- `thinking-heartbeat.py` (PostToolUse): guards with `session_role.is_dispatcher()`, which compares `hook_input["session_id"]` against the MCP Claude UUID state file and the hook marker file.
- `pre-tool-heartbeat.py` (PreToolUse): guards with `session_role.is_dispatcher_session()`, which adds the `agent_id` fast path (subagents always carry `agent_id` in PreToolUse payloads, so the I/O check is skipped entirely for the common subagent case).

Conservative handling of bad input: empty or unparseable stdin exits 0 without writing. An unknown session never extends the heartbeat.

**Launcher-agnostic**: both `claude-persistent.sh` and `claude-interactive.exp` trigger the same `SessionStart` hook (`write-dispatcher-session-id.py`) that the `session_role` predicates rely on. The fix works regardless of which launcher starts Lobster.

## Before / After flow

```
Before:
  Dispatcher dead, subagent active
  → subagent Bash call → PostToolUse fires → thinking-heartbeat.py writes timestamp
  → health check reads fresh timestamp → GREEN (FALSE POSITIVE)

After:
  Dispatcher dead, subagent active
  → subagent Bash call → PostToolUse fires → thinking-heartbeat.py checks session_id
  → session_id != dispatcher session → exit 0, no write
  → health check reads stale timestamp → RED → restart triggered
```

## What is not changed

- `session_role.py` is not modified — this fix reuses the existing dispatcher detection infrastructure.
- Health check thresholds, WFM-active signal, and compaction suppression are untouched.
- The `write_heartbeat()` pure function is unchanged — the guard is only in `main()`.

## Functional patterns used

Pure function (`write_heartbeat`) isolated from the side-effect guard in `main()`. The dispatcher check is a pure predicate (returns bool) with no side effects, composed before the write operation. The `is_dispatcher` and `is_dispatcher_session` predicates are already well-tested in `session_role.py`; this PR treats them as trusted building blocks.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_thinking_heartbeat.py tests/unit/test_hooks/test_pre_tool_heartbeat.py -v` — 39 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 3412 passed, 37 failed (37 pre-existing failures on local-dev branch confirmed by running against unmodified local-dev; none are in heartbeat or session_role tests)

## Manual test

N/A — this change is entirely internal (filesystem writes guarded by a session ID check). No external service calls, no Telegram output, no DB schema changes. The health check reading of the heartbeat file is unchanged.

## Definition of Done

- [x] Integration/manual test — N/A (pure hook guard, no external services)
- [x] User-visible outcome: health check will now correctly detect dispatcher death during subagent activity — verified indirectly via unit tests that confirm the file is not written for subagent sessions
- [x] Side-effect audit: change only reduces writes (subagent calls no longer write); no new writes, no new external calls, no flood risk

Closes #1897

### DIFF
--- a/hooks/pre-tool-heartbeat.py
+++ b/hooks/pre-tool-heartbeat.py
@@ -2,9 +2,9 @@
 """
 PreToolUse hook: dispatcher pre-tool heartbeat.
 
-Writes the current Unix epoch timestamp to a dedicated heartbeat file on every
-PreToolUse event. This complements thinking-heartbeat.py (PostToolUse) and narrows
-the detection window for inference-gap cases (issue #1695).
+Writes the current Unix epoch timestamp to a dedicated heartbeat file when the
+current session is the dispatcher. Subagent tool calls are silently ignored so
+they cannot keep the health check satisfied while the dispatcher is frozen or dead.
 
 Purpose
 -------
@@ -26,24 +26,46 @@ threshold (currently 1200s) risks false positives during legitimate long tool ca
 The pre-tool heartbeat lets us reduce that threshold safely — it confirms the
 dispatcher called the tool even if post-tool hasn't fired yet.
 
+Dispatcher-only guard (issue #1897):
+- PreToolUse fires for ALL sessions sharing the same Claude Code process,
+  including background subagents. Without this guard, a subagent doing heavy
+  tool work can keep the heartbeat fresh even if the dispatcher is dead.
+- The guard uses is_dispatcher_session() which checks: (0) agent_id fast path
+  (subagents always have agent_id in PreToolUse payloads), (1) MCP Claude UUID
+  state file, (2) hook marker file (dispatcher-session-id), (3) process-tree
+  walk as a last resort. All three state-file paths are written by the SessionStart
+  hook that fires for both claude-persistent.sh and claude-interactive.exp launchers.
+- Launcher-agnostic: works whether Lobster is started via claude-interactive.exp
+  or claude-persistent.sh.
+
 Design
 ------
 - Fires on every PreToolUse (matcher: "")
+- Guards on dispatcher session: reads hook input from stdin and calls
+  is_dispatcher_session(). Subagent calls exit 0 immediately without file I/O.
 - Atomic write: write to .tmp, then os.rename() to avoid partial reads
 - Single integer timestamp — no JSON parsing, no locking, no network
 - Silent on failure: health check degrades gracefully when file absent
 - < 1ms on warm OS (rename is a kernel atomic op on same-filesystem paths)
 
 File location: ~/lobster-workspace/logs/dispatcher-pre-tool-heartbeat
-Content: single Unix epoch integer (e.g. "1713456789\\n")
+Content: single Unix epoch integer (e.g. "1713456789\n")
 
-See issue #1786 (thinking-freeze mitigations) and #1695 (inference-gap detection).
+See issue #1786 (thinking-freeze mitigations) and #1695 (inference-gap detection),
+and #1897 (subagent-masking fix).
 """
 
+import json
 import os
 import sys
 import time
 from pathlib import Path
+
+# Allow imports from the hooks directory (session_role).
+_HOOKS_DIR = Path(__file__).parent
+sys.path.insert(0, str(_HOOKS_DIR))
+
+import session_role  # noqa: E402 — path insert must precede this
 
 
 WORKSPACE_DIR = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
@@ -64,6 +86,20 @@ def write_heartbeat(heartbeat_file: Path) -> None:
 
 
 def main() -> None:
+    # Read hook input from stdin (provided by Claude Code on every PreToolUse).
+    # If stdin is empty or unparseable, treat as unknown session — skip heartbeat
+    # (conservative: better to miss one update than to falsely extend it for a subagent).
+    try:
+        hook_input = json.load(sys.stdin)
+    except (json.JSONDecodeError, ValueError, EOFError):
+        sys.exit(0)
+
+    # Guard: only write the heartbeat for the dispatcher session.
+    # is_dispatcher_session() uses: agent_id fast path (subagents always have it
+    # in PreToolUse payloads) → MCP state files → hook marker file → process-tree.
+    if not session_role.is_dispatcher_session(hook_input):
+        sys.exit(0)
+
     try:
         write_heartbeat(HEARTBEAT_FILE)
     except Exception:
--- a/hooks/thinking-heartbeat.py
+++ b/hooks/thinking-heartbeat.py
@@ -2,9 +2,9 @@
 """
 PostToolUse hook: dispatcher heartbeat.
 
-Writes the current Unix epoch timestamp to a single heartbeat file on every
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — fix is correct, well-tested, and safe to deploy.

## Summary

This PR correctly scopes both heartbeat hooks so that only the dispatcher session writes the heartbeat files. The fix is narrow (only `main()` is guarded; the pure write function is unchanged), and it reuses the existing `session_role` predicates rather than inventing new detection logic.

## What I verified independently

**Cold-review concerns (before reading the briefing):**

1. **Does subagent stdin cause an early exit before the guard runs?** No. `json.load(sys.stdin)` is called first; on empty/invalid stdin the hook exits 0 without writing. This is the conservative path — an unknown session never extends the heartbeat.

2. **Is the mock patching reliable?** The tests patch `mod.session_role.is_dispatcher` (or `.is_dispatcher_session`) on the actual module object in `sys.modules`, after `exec_module`. Because the hooks call `session_role.is_dispatcher(hook_input)` (attribute lookup, not `from session_role import is_dispatcher`), the mock is applied before `main()` runs and takes effect correctly. All 39 new tests pass with the mocks in place.

3. **Test class shadowing?** No duplicate class names. Both test files define `TestDispatcherOnlyGuard` but in separate modules; no shadowing issue.

4. **Pre-existing failure count.** Engineer claims 37; I ran `uv run pytest tests/unit/ --tb=no -q` and got **36 failures**, not 37. One failure fewer than stated — this is a non-issue (likely a recently-fixed failure merged to local-dev), but the claim is off by one.

5. **Fail-open on OSError in `session_role._check_state_file`**: If a state file exists but throws an `OSError` when read, the function returns `True` (treats the caller as the dispatcher). This is a pre-existing behavior in `session_role.py` (unchanged in this PR). In production, this would cause a subagent to write a heartbeat on a transient I/O error. The risk is minor and acceptable given the rarity of such errors.

6. **Shared module mutation in tests**: Each call to `_run_hook_with_input` patches the actual `session_role` module object (shared across tests via `sys.modules`). Since tests run serially and each test re-executes `exec_module`, this does not cause cross-test contamination in practice. All 433 hook unit tests pass.

The engineer's briefing accurately describes the fix. My only new finding: the pre-existing failure count is 36, not 37. Everything else checks out.

## Conclusion

The fix is correct. Both hooks now reliably exit without writing when called from a subagent session. The conservative stdin handling (unknown session = skip) is the right default. Tests cover all three cases: dispatcher writes, subagent skips, invalid-stdin skips. PASS.

PR #1899, opened 2026-05-01
fix(#1898): re-add _is_fresh_start_dispatcher() to inject-bootup-context.py

sayhar:
## What broke and why

PR #1891 (context injection logging) was developed on a branch forked before PR #1869 merged. When #1891 was deployed to local-dev (commit `a001527b` / `1c3f3065`), it overwrote `hooks/inject-bootup-context.py` with a version that pre-dates `449495ce`. The `_is_fresh_start_dispatcher()` function and its fallback call in `main()` — added by #1869 to fix issue #1868 — were silently removed.

**Result:** On every cold restart, the primary state file (`dispatcher-claude-session-id`) is absent until the dispatcher calls `session_start()`. Without the fallback, `is_dispatcher()` returns False and the dispatcher receives subagent bootup instead of dispatcher bootup.

## What this PR does

Re-adds the `_is_fresh_start_dispatcher()` function and its invocation in `main()` from commit `449495ce`. The fallback logic:
- If `LOBSTER_MAIN_SESSION=1` AND the primary state file does not exist → treat session as dispatcher
- This covers the pre-`session_start` window that only exists during a genuine fresh restart
- Subagents cannot trigger it: they start only after the dispatcher calls `session_start()`, which writes the primary file
- Compaction events are also safe: `on-compact.py` writes the primary file proactively before the post-compact SessionStart fires

The module docstring is updated to document both fallback paths (post-compact sentinel and fresh-start).

## Tests

Adds `tests/unit/test_hooks/test_inject_bootup_context_fresh_start.py` with 9 tests:

**`TestIsFreshStartDispatcher` (5 tests):**
- Primary file absent + `LOBSTER_MAIN_SESSION=1` → True (the fix)
- Primary file present → False (normal subagent case)
- `LOBSTER_MAIN_SESSION` not set → False (non-Lobster session)
- `LOBSTER_MAIN_SESSION=0` → False
- `OSError` on stat → False (safe default)

**`TestMainFreshStartFallback` (4 tests):**
- Issue #1868 regression test: fresh restart → dispatcher bootup injected
- Primary file present, UUID mismatch → subagent bootup
- `LOBSTER_MAIN_SESSION` not set → subagent bootup
- Primary file UUID match → dispatcher bootup (no fallback needed)

[...]

PR #1904, opened 2026-05-01
fix(#1903): write primary Claude session file in write-dispatcher-session-id.py

### LINKED ISSUE #1903: fix: fresh-start fallback fails when MCP keeps running (stale primary file, not absent)
## Problem

PR #1899 added \`_is_fresh_start_dispatcher()\` to \`inject-bootup-context.py\` to fix the case where the dispatcher gets subagent bootup on a fresh start. The fix checks:

- \`LOBSTER_MAIN_SESSION=1\` (Lobster-managed session)
- Primary file (\`dispatcher-claude-session-id\`) is **absent**

This works when **both CC and the MCP server restart** (MCP server clears the primary file on startup). But it fails when **only CC restarts** while the MCP server keeps running.

## Observed failure (2026-05-01 session 20260501-012)

Session \`d47ecad9-c27b-41bc-a13f-8a29cee7ae02\` (dispatcher) started at 15:44:04 and got subagent bootup injected instead of dispatcher bootup.

Evidence from \`context-injection.log\`:
```
2026-05-01T15:44:04Z | session=d47ecad9-c27b-41bc-a13f-8a29cee7ae02 | role=subagent | injected=[sys.subagent.bootup.md, ...]
```

But \`audit.jsonl\` confirms d47ecad9 IS the dispatcher:
```
2026-05-01T15:44:39Z | session_start | agent_id=lobster-dispatcher, agent_type=dispatcher
```

## Root cause

The failure chain:

1. CC auto-restarted at 15:44 (session \`dbcad92a\` ended, new session \`d47ecad9\` started)
2. The MCP server (\`lobster-mcp-local\`) did **not** restart — it kept running
3. Primary file (\`dispatcher-claude-session-id\`) still contained \`dbcad92a-...\` (old dispatcher UUID)
4. \`write-dispatcher-session-id.py\` (Hook 1) ran: determined d47ecad9 is the dispatcher, updated **tertiary** file
5. \`inject-bootup-context.py\` (Hook 2) ran: \`_check_state_file(primary, d47ecad9)\` returned \`False\` (d47ecad9 ≠ dbcad92a)
6. \`is_dispatcher()\` returned \`False\` immediately (primary short-circuits, tertiary never checked)
7. \`_is_fresh_start_dispatcher()\` also returned \`False\`: \`not primary_file.exists()\` → the file EXISTS (with stale UUID)
8. Subagent bootup injected **wrong**

The fresh-start fallback only handles the "primary file absent" case, not "primary file exists with stale UUID".

## Evidence

From \`on-compact.log\`:
```
2026-05-01T15:44:04Z | skipped_not_compact | hook_name='absent' session_id='d47ecad9-c27'
```
hook_name='absent' confirms this was not a recognized compaction event (the compact-pending sentinel fallback also did not fire).

## Fix

The \`_is_fresh_start_dispatcher()\` function needs to handle the additional case:

**Primary file exists but has a stale (non-matching) UUID AND \`LOBSTER_MAIN_SESSION=1\`**

This is safe because:
- Subagents are spawned only after the dispatcher calls \`session_start()\`, which updates the primary file. By the time any subagent's SessionStart fires, the primary file has the CORRECT dispatcher UUID. A stale UUID cannot come from a just-started subagent.
- The tertiary file (\`dispatcher-session-id\`) has already been updated by \`write-dispatcher-session-id.py\` (Hook 1) to contain the new session's UUID. Checking that the tertiary file matches the current session AND \`LOBSTER_MAIN_SESSION=1\` is a safe discriminator.

Proposed fix in \`inject-bootup-context.py\`:

```python
def _is_fresh_start_dispatcher() -> bool:
    # Existing: absent primary file + LOBSTER_MAIN_SESSION=1
    if os.environ.get("LOBSTER_MAIN_SESSION", "") != "1":
        return False
    try:
        primary_file = session_role._get_mcp_claude_session_file()
        if not primary_file.exists():
            return True  # existing case: MCP restarted and cleared primary file
        # New case: primary file exists but has STALE UUID (CC restarted, MCP kept running)
        # Check if the tertiary file (updated by write-dispatcher-session-id.py in Hook 1)
        # matches the current session ID from hook_input.
        # If tertiary matches, Hook 1 already determined this is the dispatcher.
        return True  # tertiary mismatch with primary → stale primary → treat as fresh dispatcher
    except OSError:
        return False
```

Actually the cleanest fix is to change the condition to: **primary file absent OR (primary file has different UUID than current session AND tertiary file matches current session)**.

## Alternatively: fix write-dispatcher-session-id.py to also write the primary file

If \`write-dispatcher-session-id.py\` (Hook 1) also writes the primary file when it determines this is the dispatcher (same as \`on-compact.py\` does), then \`inject-bootup-context.py\` (Hook 2) would find the correct UUID in the primary file and work correctly.

This seems like the cleanest fix since \`write-dispatcher-session-id.py\` already does the heavy lifting of determining if this is the dispatcher.

## Files to change

- \`hooks/inject-bootup-context.py\` — extend fresh-start fallback to handle stale primary file case
  OR
- \`hooks/write-dispatcher-session-id.py\` — also write primary file when dispatcher is detected
- Tests for the new case

## Related

- Issue #1868 (original fresh-start detection bug)
- Issue #1861 (same class, primary file absent case)  
- PR #1899 (fix for primary file absent — does NOT fix primary file stale)

sayhar:
## Problem

When CC restarts while the MCP server keeps running, the dispatcher session gets **subagent bootup injected** instead of dispatcher bootup. This was observed in session 20260501-012 (`d47ecad9`) at 15:44:04.

The failure chain:
1. CC restarts → new session UUID `d47ecad9`, MCP stays up
2. Primary file (`dispatcher-claude-session-id`) still holds `dbcad92a` (old UUID, stale)
3. Hook 1 (`write-dispatcher-session-id.py`) correctly detects `d47ecad9` as the dispatcher via stale-JSONL recovery → writes **tertiary** file only
4. Hook 2 (`inject-bootup-context.py`) checks the primary file: `_check_state_file(primary, d47ecad9)` → False (stale UUID, not absent)
5. Fresh-start fallback: `not primary_file.exists()` → also False (file exists with stale UUID)
6. Subagent bootup injected — wrong

The fresh-start fallback added in PR #1899 only handles the "primary file absent" case. It cannot handle "primary file exists with stale UUID."

## Fix

Hook 1 already does the hard work of identifying whether the current session is the dispatcher. It just needs to also write the primary file (`dispatcher-claude-session-id`) when it determines the session is the dispatcher — mirroring exactly what `on-compact.py` does for the post-compaction case (Option 1, issue #1375).

After this change: by the time Hook 2 runs, the primary file contains the correct new UUID. The primary check passes immediately, and dispatcher bootup is injected correctly.

### Flow before fix (stale-UUID scenario)

```
CC restart → new session d47ecad9
                  │
Hook 1 (write-dispatcher-session-id.py)
  _is_dispatcher_session() → True (JSONL dead, recovery path)
  write_dispatcher_session_id() → writes TERTIARY file only
                  │
Hook 2 (inject-bootup-context.py)
  is_dispatcher() → checks primary (dbcad92a ≠ d47ecad9) → False
  _is_post_compact_dispatcher() → no sentinel → False
  _is_fresh_start_dispatcher() → primary EXISTS → False
  → SUBAGENT BOOTUP INJECTED (wrong)
```

### Flow after fix

```
CC restart → new session d47ecad9
                  │
Hook 1 (write-dispatcher-session-id.py)
  _is_dispatcher_session() → True (JSONL dead, recovery path)
  write_dispatcher_session_id() → writes TERTIARY file
  write_dispatcher_claude_session_id() → writes PRIMARY file (d47ecad9)
                  │
Hook 2 (inject-bootup-context.py)
  is_dispatcher() → checks primary (d47ecad9 == d47ecad9) → True
  → DISPATCHER BOOTUP INJECTED (correct)
```

## Implementation

Single change to `hooks/write-dispatcher-session-id.py`: when `_is_dispatcher_session()` returns True, call `write_dispatcher_claude_session_id(session_id)` alongside the existing `write_dispatcher_session_id(session_id)`. No changes to `inject-bootup-context.py` or `session_role.py`.

The functional pattern here is pure: the fix is a straightforward addition to the side-effect boundary (Hook 1 does all writes it needs to, so Hook 2 reads a consistent state).

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_write_dispatcher_session_id.py -v` — 6 new tests, all passed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 426 passed, 0 failed (all pre-existing hook tests clean)
- [x] Full unit suite (`tests/unit/`) — 426 passed, 0 failed

## Manual test
N/A — this change is purely internal hook logic (file writes). The failure mode requires a specific timing (CC restart without MCP restart), making a live test impractical. The unit tests directly simulate the failure scenario (stale primary UUID, dead JSONL, new session UUID) and verify the primary file is updated before Hook 2 would run.

## Definition of Done
- [x] Unit tests pass with proper isolation (no production hits)
- [x] User-visible outcome: dispatcher receives dispatcher bootup after CC-only restart (verified via unit test simulating the exact failure scenario from the issue)
- [x] Side-effect audit: this hook already writes two files (tertiary + DB row); the new write is a third file in the same pattern. Scoped to dispatcher-only path. No volume concerns.

Closes #1903
Relates to #1899 (fresh-start fallback), #1375 (Option 1 pattern used here)

### DIFF
--- a/hooks/write-dispatcher-session-id.py
+++ b/hooks/write-dispatcher-session-id.py
@@ -12,9 +12,16 @@ silently ignored.
 When the marker file is absent (fresh dispatcher start), the current
 session_id matches the stored dispatcher ID, or the stored dispatcher session
 has ended (its .jsonl file is gone from ~/.claude/projects/), writes
-session_id to ~/messages/config/dispatcher-session-id. This file is the
-primary signal that other hooks use to distinguish the dispatcher from
-subagents.
+session_id to two files:
+- ~/messages/config/dispatcher-session-id (tertiary hook marker file)
+- $LOBSTER_WORKSPACE/data/dispatcher-claude-session-id (primary Claude UUID state file)
+
+The primary file write (issue #1903) ensures that inject-bootup-context.py
+(Hook 2, which runs after this hook) finds the correct UUID in the primary
+file even when the MCP server kept running across a CC restart, leaving a
+stale UUID in the primary file. Without this write, the fresh-start fallback
+in inject-bootup-context.py cannot help (the file exists with a wrong UUID,
+not absent), causing dispatcher sessions to receive subagent bootup.
 
 ## Subagent sessions
 
@@ -85,6 +92,14 @@ sys.path.insert(0, str(_SRC_DIR))
 
 import session_role  # noqa: E402 — path insert must precede this
 from agents import session_store  # noqa: E402
+
+# Import the primary-file writer used by on-compact.py for the same purpose.
+# write_dispatcher_claude_session_id writes the Claude UUID to the primary state
+# file (dispatcher-claude-session-id), which is the file checked first by
+# inject-bootup-context.py (Hook 2) when deciding whether to inject dispatcher
+# or subagent bootup. Writing it here (Hook 1) ensures the primary file is
+# up-to-date before Hook 2 runs, fixing the stale-UUID bug described in #1903.
+from session_role import write_dispatcher_claude_session_id  # noqa: E402
 
 # A JSONL file not modified for longer than this is treated as belonging to an
 # idle or dead session.  Active dispatcher sessions append to their JSONL on
@@ -195,6 +210,16 @@ def main() -> None:
 
     if _is_dispatcher_session(session_id):
         session_role.write_dispatcher_session_id(session_id)
+        # Issue #1903: also write the primary Claude UUID state file
+        # (dispatcher-claude-session-id) so that inject-bootup-context.py
+        # (Hook 2, which runs after this hook) finds the correct UUID in the
+        # primary file.  Without this, when CC restarts while MCP keeps running,
+        # the primary file retains the old dispatcher UUID (stale, not absent),
+        # causing both the primary check and the fresh-start fallback to fail —
+        # and subagent bootup to be injected instead of dispatcher bootup.
+        # This mirrors the same write that on-compact.py performs for the
+        # post-compaction case (Option 1, issue #1375).
+        write_dispatcher_claude_session_id(session_id)
         # Issue #781 Fix 2: also tag this session in agent_sessions.db as
         # 'dispatcher' so the reconciler never emits agent_failed for it.
         _register_dispatcher_session(session_id)
--- a/tests/unit/test_hooks/test_write_dispatcher_session_id.py
+++ b/tests/unit/test_hooks/test_write_dispatcher_session_id.py
@@ -1,0 +1,400 @@
+"""
+Unit tests for hooks/write-dispatcher-session-id.py — primary file write on dispatcher detection.
+
+Issue #1903: When CC restarts but MCP keeps running, inject-bootup-context.py
+receives subagent bootup instead of dispatcher bootup.
+
+Root cause:
+- Primary file (dispatcher-claude-session-id) holds the OLD dispatcher UUID
+  from before the CC restart.
+- write-dispatcher-session-id.py (Hook 1) correctly identifies the new session
+  as the dispatcher (via the stored JSONL fallback) and writes the tertiary file.
+- inject-bootup-context.py (Hook 2) checks the primary file:
+  _check_state_file(primary, new_session_id) → False (stale UUID, not absent).
+  _is_fresh_start_dispatcher() → also False (file exists).
+  Both paths fail → subagent bootup injected.
+
+Fix: write-dispatcher-session-id.py must also write the primary file
+(dispatcher-claude-session-id) with the new session UUID when it determines
+the session is the dispatcher, same as on-compact.py already does.
+
+Tests verify:
+- write_dispatcher_claude_session_id is called with the new session UUID when
+  the session is identified as the dispatcher.
+- The primary file actually contains the new UUID after the hook runs.
+- The primary file is NOT written when the session is a subagent.
+- Behaviour is correct for all three _is_dispatcher_session() paths:
+  fresh start (no marker), same-session reattach, and stale-marker recovery.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import os
+import sys
+import time
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+_HOOKS_DIR = Path(__file__).parents[3] / "hooks"
+_HOOK_PATH = _HOOKS_DIR / "write-dispatcher-session-id.py"
+
+if str(_HOOKS_DIR) not in sys.path:
+    sys.path.insert(0, str(_HOOKS_DIR))
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+class _PatchEnv:
+    """Context manager to temporarily set / restore environment variables."""
+
+    def __init__(self, env: dict):
+        self._env = env
+        self._saved: dict = {}
+
+    def __enter__(self):
+        for k, v in self._env.items():
+            self._saved[k] = os.environ.get(k)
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Minimal, targeted fix for #1903. Hook 1 (`write-dispatcher-session-id.py`) now also calls `write_dispatcher_claude_session_id(session_id)` whenever it identifies the session as the dispatcher, mirroring exactly what `on-compact.py` already does. This ensures the primary file has the correct UUID before Hook 2 (`inject-bootup-context.py`) runs, closing the stale-UUID gap in the CC-restart-with-MCP-up scenario.

## Checklist

**Race condition (two sessions starting simultaneously)?**
Both `write_dispatcher_session_id` and `write_dispatcher_claude_session_id` use atomic `.tmp` → rename writes (Linux atomic). If two sessions truly started simultaneously, one write would win and the other would be the loser — but this is the pre-existing behavior for the tertiary file write and is acceptable. The logical invariant (only the dispatcher identifies itself as dispatcher) is sound.

**Write order — does it happen before Hook 2?**
Verified. `write-dispatcher-session-id.py` is at position 0 in `settings.json` SessionStart hooks, `inject-bootup-context.py` is at position 1. Hook 1 writes before Hook 2 reads. Correct.

**Test isolation — HOME override actually works?**
Yes. Tests use `monkeypatch.setenv("HOME", str(tmp_path))` and `monkeypatch.setenv("LOBSTER_WORKSPACE", str(tmp_path))` which persist for the full test duration. They also call `importlib.reload(_sr)` to force re-evaluation of `DISPATCHER_SESSION_FILE` under the new HOME, then `monkeypatch.setattr(sr, "DISPATCHER_SESSION_FILE", ...)` to be doubly sure. `write_dispatcher_claude_session_id` calls `_get_mcp_claude_session_file()` at call time (not module time), which reads `LOBSTER_WORKSPACE` from env — correctly resolving to tmp_path. No real-home writes possible.

**Pre-existing failure count claim?**
Verified. Running `uv run pytest tests/unit/test_hooks/` (the only runnable test scope — mcp_server tests have collection errors unrelated to this PR) gives 420 tests passing without the new file, 426 with it. The 44 collection errors in `test_mcp_server/` and other directories are pre-existing environmental issues unrelated to this change.

**Interaction with `_is_fresh_start_dispatcher()` in Hook 2?**
The bug scenario had a stale-but-present primary file. Hook 2's `_is_fresh_start_dispatcher()` fires only when the primary file is *absent*, so it was already inactive in this bug case — the fix doesn't interact with that fallback. After the fix, Hook 1 updates the primary file to the correct UUID, and Hook 2's primary check (`is_dispatcher()`) returns True directly.

## Minor Nits (non-blocking)

- `_load_hook` docstring says "callers must wrap the mod.main() call inside a _PatchEnv context" — but tests use `monkeypatch.setenv` instead, which works correctly. The comment is inaccurate but harmless.
- The `_PatchEnv` class in the test file is only used once (for the `LOBSTER_MAIN_SESSION` deletion case). The other tests use `monkeypatch.setenv`. `_PatchEnv` could be removed and that one test could use `monkeypatch.delenv`, but this is cosmetic.

Neither nit affects correctness.

## Verdict

PASS. The fix is correct, minimal, and well-tested. The six new tests cover all three dispatcher-identification paths (fresh start, reattach, stale-marker recovery) plus the negative case (subagent) and the end-to-end inject-bootup path. The pre-existing test count is verified.

PR #1907, opened 2026-05-01
fix: patch DISPATCHER_SESSION_FILE in tests to prevent production state corruption

sayhar:

Tests that exercise `inject-bootup-context.py` via `main()` were writing to the live production file `~/messages/config/dispatcher-session-id`, overwriting the running dispatcher's session ID.

On 2026-05-01 this caused an incident: the test UUID `"known-dispatcher-uuid-abc"` was written atomically to `~/messages/config/dispatcher-session-id`, corrupting the live dispatcher's state. The hook-marker-file (tertiary) dispatcher detection then failed for all subsequent SessionStart hooks, stopping heartbeat writes for ~20 minutes until the dispatcher restarted.

## Root cause

`session_role.DISPATCHER_SESSION_FILE` is a module-level constant resolved at import time:

```python
DISPATCHER_SESSION_FILE = Path(os.path.expanduser("~/messages/config/dispatcher-session-id"))
```

When `inject-bootup-context.py` is loaded dynamically in tests via `importlib`, the `session_role` it imports is the **cached** version from `sys.modules` — with the production path baked in. Even though tests patch `HOME` and `LOBSTER_WORKSPACE` for the primary-file lookups, those env vars do not affect the module-level constant.

Two tests in `TestMainSentinelFallback` trigger the write:
- `test_sentinel_fallback_injects_dispatcher_bootup`: sentinel fallback fires → `is_dispatcher=True` → `write_dispatcher_session_id("unknown-post-compact-uuid")` written to production
- `test_primary_file_match_still_works_without_sentinel`: primary file matches → `is_dispatcher=True` → `write_dispatcher_session_id("known-dispatcher-uuid-abc")` written to production (the incident)

## Fix

In all four `TestMainSentinelFallback` tests, after loading the hook module and before calling `main()`, redirect `mod.session_role.DISPATCHER_SESSION_FILE` to a path under `tmp_path`:

```python
mod.session_role.DISPATCHER_SESSION_FILE = (
    tmp_path / "messages" / "config" / "dispatcher-session-id"
)
```

This is applied to all four tests (not just the two directly implicated) so that future changes to `is_dispatcher` logic cannot introduce a new regression.

The fix is purely in the test file — no production code changes.

**Functional patterns used:** Pure attribute reassignment on the loaded module's namespace; no mock objects or complex patching machinery needed since `session_role` is accessed as an attribute of the dynamically loaded module.

## Before / after

**Before:** `main()` calls `session_role.write_dispatcher_session_id(session_id)` → writes to real `~/messages/config/dispatcher-session-id` via the cached module-level constant.

**After:** `mod.session_role.DISPATCHER_SESSION_FILE` is redirected to `tmp_path/messages/config/dispatcher-session-id` before `main()` is called → write goes to the isolated temp directory.

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_inject_bootup_context_sentinel.py -v` — 9 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/test_session_role_state_file.py -v` — 33 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 316 passed, 6 pre-existing failures (unrelated to this change, confirmed by running against upstream main without the fix)
- [x] Production file `~/messages/config/dispatcher-session-id` verified unchanged after test run (retains the live dispatcher's real UUID, not `"known-dispatcher-uuid-abc"`)

## Manual test
N/A — this change is entirely internal to test isolation; no external services, no systemd units, no runtime file I/O beyond tmp_path.

## Definition of Done
- [x] Integration/manual test run — N/A (test-only change)
- [x] User-visible outcome verified — N/A (this fixes test isolation, not a user-facing feature)
- [x] Side-effect audit complete: tests now write only to tmp_path, not production paths

Fixes the 2026-05-01 incident where `~/messages/config/dispatcher-session-id` was overwritten with `"known-dispatcher-uuid-abc"` during a test run, stopping dispatcher heartbeats for ~20 minutes.

### DIFF
--- a/tests/unit/test_hooks/test_inject_bootup_context_sentinel.py
+++ b/tests/unit/test_hooks/test_inject_bootup_context_sentinel.py
@@ -207,6 +207,12 @@ class TestMainSentinelFallback:
             mod.USER_BASE_BOOTUP = tmp_path / "no-user-base"
             mod.USER_DISPATCHER_BOOTUP = tmp_path / "no-user-dispatcher"
             mod.USER_SUBAGENT_BOOTUP = tmp_path / "no-user-subagent"
+            # Redirect DISPATCHER_SESSION_FILE to tmp_path so the sentinel-triggered
+            # write_dispatcher_session_id() call does not corrupt the production state
+            # file at ~/messages/config/dispatcher-session-id.
+            mod.session_role.DISPATCHER_SESSION_FILE = (
+                tmp_path / "messages" / "config" / "dispatcher-session-id"
+            )
 
             with patch("sys.stdin", io.StringIO(hook_input)):
                 with pytest.raises(SystemExit):
@@ -248,6 +254,11 @@ class TestMainSentinelFallback:
             mod.USER_BASE_BOOTUP = tmp_path / "no-user-base"
             mod.USER_DISPATCHER_BOOTUP = tmp_path / "no-user-dispatcher"
             mod.USER_SUBAGENT_BOOTUP = tmp_path / "no-user-subagent"
+            # Redirect DISPATCHER_SESSION_FILE defensively even though is_dispatcher()
+            # returns False here (no write occurs), so future changes can't regress.
+            mod.session_role.DISPATCHER_SESSION_FILE = (
+                tmp_path / "messages" / "config" / "dispatcher-session-id"
+            )
 
             with patch("sys.stdin", io.StringIO(hook_input)):
                 with pytest.raises(SystemExit):
@@ -293,6 +304,11 @@ class TestMainSentinelFallback:
             mod.USER_BASE_BOOTUP = tmp_path / "no-user-base"
             mod.USER_DISPATCHER_BOOTUP = tmp_path / "no-user-dispatcher"
             mod.USER_SUBAGENT_BOOTUP = tmp_path / "no-user-subagent"
+            # Redirect DISPATCHER_SESSION_FILE defensively even though is_dispatcher()
+            # returns False here (no write occurs), so future changes can't regress.
+            mod.session_role.DISPATCHER_SESSION_FILE = (
+                tmp_path / "messages" / "config" / "dispatcher-session-id"
+            )
 
             with patch("sys.stdin", io.StringIO(hook_input)):
                 with pytest.raises(SystemExit):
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Correctness: patching `mod.session_role.DISPATCHER_SESSION_FILE`

The patch surface is correct. `inject-bootup-context.py` calls `session_role.write_dispatcher_session_id(sid)` (line 258), which is defined in `session_role.py` and uses `DISPATCHER_SESSION_FILE` as a bare module-level name — not a re-resolved call. When the test does `mod.session_role.DISPATCHER_SESSION_FILE = tmp_path / ...`, it mutates the same attribute object that `write_dispatcher_session_id()` reads at call time, so the redirect takes effect. This is the correct surface to patch.

Patching via `sys.modules['session_role'].DISPATCHER_SESSION_FILE` would also work, but the `mod.session_role` approach is simpler and equally correct because the dynamically loaded module holds a direct reference to the same `session_role` module object.

## Engineer's questions

**Q1 — Is `mod.session_role.DISPATCHER_SESSION_FILE = ...` the right surface vs. `sys.modules`?**

Yes. Both approaches mutate the same in-memory module object. `mod.session_role` is cleaner because it avoids relying on the module cache namespace; it's the same reference obtained from the dynamic load and less likely to accidentally affect other test modules running in the same process.

**Q2 — Correct to apply defensively to the two tests where `is_dispatcher()` returns False?**

Yes. For `test_no_sentinel_subagent_with_primary_file_present_gets_subagent_bootup` and `test_sentinel_with_lobster_main_session_zero_gets_subagent_bootup`, the current code path does not reach `write_dispatcher_session_id()`, so no write occurs today. But the defensive redirect is the right call: it costs nothing, and it prevents a silent regression if logic upstream of the guard ever changes (e.g. a new fallback path that hits `is_dispatcher = True`). The comments in the PR explain this clearly.

## Second write path: `write_dispatcher_claude_session_id`

`inject-bootup-context.py` does **not** call `write_dispatcher_claude_session_id`. That function is defined in `session_role.py` but is only called by `on-compact.py` (not under test here) and the MCP server. The primary file (`$LOBSTER_WORKSPACE/data/dispatcher-claude-session-id`) is read-only relative to the hook under test. No additional protection is needed.

## Tests

All 9 tests pass (`uv run pytest tests/unit/test_hooks/test_inject_bootup_context_sentinel.py -v`).

## Summary

Fix is correct, well-scoped, and the defensive application to all four tests is good practice. LGTM.

sayhar:

PR #1916, opened 2026-05-02
fix: use is_dispatcher_session() in thinking-heartbeat PostToolUse hook

sayhar:
## Problem

`thinking-heartbeat.py` (PostToolUse hook) was calling `session_role.is_dispatcher(hook_input)` to decide whether to write the heartbeat timestamp. However, `is_dispatcher()` checks the `dispatcher-startup-flag` file, which is **deleted** by `inject-bootup-context.py` after the dispatcher's SessionStart hook runs. Since the flag is deleted at startup, `is_dispatcher()` always returns `False` at PostToolUse time — so the heartbeat was **never written**.

This caused the health check to trigger false-positive restarts:
```
[ERROR] RED: Critical failure - dispatcher heartbeat stale (>900s)
[WARN] Restarting lobster-claude ...
[ERROR] BLACK: Manual intervention required (flag already set, 4799s elapsed, retry in 2401s)
```

## Fix

Replace `is_dispatcher()` with `is_dispatcher_session()` in `thinking-heartbeat.py`.

`is_dispatcher_session()` uses a layered strategy that works throughout the entire dispatcher session lifetime:
1. **agent_id fast path**: CC injects `agent_id` into subagent payloads; the dispatcher never has it
2. **MCP Claude UUID state file**: written at startup by `write-dispatcher-session-id.py`
3. **Hook marker file** (`dispatcher-session-id`)
4. **Process-tree walk**: count consecutive claude ancestors before a tmux pane PID

The rule in `session_role.py` is explicit:
- `is_dispatcher()` for SessionStart/SubagentStop/Stop hooks (uses startup flag)
- `is_dispatcher_session()` for PreToolUse/PostToolUse hooks (uses agent_id + state files + process tree)

`pre-tool-heartbeat.py` (PreToolUse) already correctly uses `is_dispatcher_session()` — this PR aligns the PostToolUse hook.

## Changes

- `hooks/thinking-heartbeat.py`: call `is_dispatcher_session()` instead of `is_dispatcher()`; update `DISPATCHER_HEARTBEAT_STALE_SECONDS` to 900; add comment explaining why `is_dispatcher()` must not be used here
- `tests/unit/test_hooks/test_thinking_heartbeat.py`: add `TestDispatcherOnlyGuard` with full coverage; add regression test that parses the source and fails if `is_dispatcher()` is called; update helpers to inject stdin and mock `is_dispatcher_session`

## Test results

All 40 `-k heartbeat` tests pass.

## Test plan
- [ ] `pytest tests/ -k heartbeat` passes
- [ ] After deploying to local-dev, `cat ~/lobster-workspace/logs/dispatcher-heartbeat` updates on each PostToolUse
- [ ] Health check no longer triggers false-positive restarts

### DIFF
--- a/hooks/thinking-heartbeat.py
+++ b/hooks/thinking-heartbeat.py
@@ -2,9 +2,9 @@
 """
 PostToolUse hook: dispatcher heartbeat.
 
-Writes the current Unix epoch timestamp to a single heartbeat file on every
-PostToolUse event. The health check reads this file to determine whether the
-dispatcher is alive.
+Writes the current Unix epoch timestamp to a single heartbeat file when the
+current session is the dispatcher. Subagent tool calls are silently ignored so
+they cannot keep the health check satisfied while the dispatcher is frozen or dead.
 
 Purpose: the dispatcher can spend 10+ minutes in a reasoning/catchup phase
 without touching wait_for_messages. Any tool call at all means the dispatcher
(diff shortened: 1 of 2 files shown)


sayhar:
Closing per direction change: replacing with agent_id guard approach (see replacement PR). The is_dispatcher_session() approach is being replaced with a direct agent_id field check — simpler, no state file I/O, no process tree walk.

PR #1920, opened 2026-05-02
fix: raise WFM stale threshold to 1800s (stop-gap)

### LINKED ISSUE #1919: fix: raise WFM stale threshold to 1800s as stop-gap while state machine PR lands
## Problem

The dispatcher keeps restarting during long startup sequences. When startup takes longer than 10 minutes (compaction + catchup + initial processing), both `DISPATCHER_HEARTBEAT_STALE_SECONDS` (600s) and `WFM_ACTIVE_STALE_SECONDS` (600s) expire before the dispatcher enters its steady-state WFM loop. The health check then declares RED and triggers a restart — which restarts the same long startup again, potentially looping.

## Stop-gap

Raise both thresholds to 1800s (30 minutes):
- `DISPATCHER_HEARTBEAT_STALE_SECONDS`: 600 → 1800
- `WFM_ACTIVE_STALE_SECONDS`: 600 → 1800

This gives the dispatcher a 30-minute window to complete startup before the health check intervenes.

## Proper fix

The state machine PR (in progress) will make the dispatcher emit explicit lifecycle signals during startup phases, so the health check can distinguish "starting up" from "frozen" without relying on a generous timeout.

This issue should be closed when the state machine PR lands and the thresholds are tuned back down to production values.

sayhar:
## Problem

During long startup sequences (compaction + catchup > 10 min), the health check declares RED and restarts the dispatcher before it reaches steady state. This creates a restart loop: startup takes too long, health check kills it, startup begins again.

The two thresholds that control this:
- `DISPATCHER_HEARTBEAT_STALE_SECONDS=600` — fires RED when the dispatcher hasn't used a tool in 10 min
- `WFM_ACTIVE_STALE_SECONDS=600` — the WFM exemption that would suppress the RED, but only works when the dispatcher is actively in `wait_for_messages`

During startup the dispatcher is in neither state — it's in compaction/catchup — so neither exemption applies, and 10 min is not enough headroom.

## Change

Raise both thresholds from 600s to 1800s (30 minutes). This gives the dispatcher a 30-minute window to complete startup before the health check intervenes.

Two constants changed in `scripts/health-check-v3.sh`:
- `DISPATCHER_HEARTBEAT_STALE_SECONDS`: 600 → 1800
- `WFM_ACTIVE_STALE_SECONDS`: 600 → 1800

The state machine PR (in progress) will make the dispatcher emit explicit phase signals during startup so the health check can distinguish "starting up" from "frozen" without relying on a loose timeout. These thresholds should be tuned back down once that lands.

Closes #1919

## Tests run
- [x] `git diff` verified — exactly the two constants changed, no other modifications
- [ ] Unit tests — no test suite covers these constants; behavior verified by inspection

## Manual test
N/A — this is a constant change to a shell script threshold. The effect is: health check will wait 30min instead of 10min before restarting a slow-starting dispatcher. No user-visible output; the change prevents spurious restarts.

## Definition of Done
- [x] Unit tests pass with proper mocking — not applicable (constant change only)
- [x] Side-effect audit: raises the restart threshold, cannot cause floods or unscoped writes
- [x] No behavior change during normal operation (dispatcher reaches WFM well within 30 min)
- [ ] Thresholds to be tuned back down when state machine PR lands

### DIFF
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -99,7 +99,7 @@ HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"   # legacy WFM-touch signa
 # Single file, single integer (Unix epoch seconds). No JSON parsing required.
 # Threshold is generous enough to cover compaction + catchup without suppression.
 DISPATCHER_HEARTBEAT_FILE="${LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-heartbeat}"
-DISPATCHER_HEARTBEAT_STALE_SECONDS=600    # 10 min — covers compaction (~5m) + catchup margin; grace period handles the rest
+DISPATCHER_HEARTBEAT_STALE_SECONDS=1800   # stop-gap: raised from 600s to 30min while state machine PR lands; long startup sequences were triggering false-positive restarts
 
 # WFM-active signal (issue #1713 / #949): inbox_server.py writes this file with
 # a Unix epoch timestamp when wait_for_messages begins blocking and refreshes it
@@ -110,7 +110,7 @@ DISPATCHER_HEARTBEAT_STALE_SECONDS=600    # 10 min — covers compaction (~5m) +
 # the health check fires.  3x was too tight and produced false-positive kills.
 # File is deleted by the MCP server when WFM returns (message arrived or timeout).
 DISPATCHER_WFM_ACTIVE_FILE="${LOBSTER_WFM_ACTIVE_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-wfm-active}"
-WFM_ACTIVE_STALE_SECONDS=600   # 10x WAIT_HEARTBEAT_INTERVAL (60s) — 10-minute window
+WFM_ACTIVE_STALE_SECONDS=1800  # stop-gap: raised from 600s to 30min while state machine PR lands; long startup sequences were triggering false-positive restarts
 # Grace window for WFM tombstone ("exited"): if the WFM-active file contains the
 # tombstone but was written within this many seconds, the dispatcher just left WFM
 # and is actively transitioning to message processing — grant the same exemption as
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -34,8 +34,8 @@
 #   hooks/thinking-heartbeat.py writes a Unix epoch timestamp to
 #   ~/lobster-workspace/logs/dispatcher-heartbeat on every PostToolUse event.
 #   check_dispatcher_heartbeat() reads this single file and checks its age.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS — minimal stop-gap change, correctly scoped

## What I checked

**Cold read findings (before reading the engineer's briefing):**

1. The two constants gate the same path: heartbeat stale AND WFM-active stale → RED. Raising both in lockstep makes sense, but my first instinct was to question whether they serve different failure modes that might warrant different thresholds. After reading the logic they are complementary parts of the same check, so moving both together is correct.

2. **Stale inline comments** — this is the main finding the engineer's briefing partially acknowledged but didn't fully enumerate. The PR updates the inline comments on lines 102 and 113, but leaves several other comments in the file still referencing the old 600s value as authoritative:
   - Line 37: *"The 600s threshold (10 min) reduces MTTR for thinking-freeze from 20 min to 10 min."* — now factually wrong; MTTR for thinking-freeze has regressed to up to 30 min under this PR.
   - Lines 551–553: *"The dispatcher heartbeat threshold (DISPATCHER_HEARTBEAT_STALE_SECONDS = 600s) covers catchup duration naturally."*
   - Lines 969–972: *"Threshold is 600s (10 min) — reduced from 1200s to lower MTTR for thinking-freeze. Boot grace period (90s) is well within the 600s threshold."*
   - Lines 1891–1901: *"Threshold is 600s (10 min) — reduces MTTR for thinking-freeze from 20 min to 10 min. Boot grace is not needed: 600s is still well above the 90s boot window."*
   - Line 108: *"Threshold: 10x WAIT_HEARTBEAT_INTERVAL — gives the dispatcher a 10-minute window"* (describing `WFM_ACTIVE_STALE_SECONDS`, now 30 min, no longer 10x).

   None of these are load-bearing logic errors — they're documentation rot. But a future reader debugging a freeze will be confused by comments saying "this catches freezes in 10 min" when the actual threshold is 30 min.

3. **MTTR regression is real but accepted.** The thinking-freeze detection window expands from 10 min to 30 min. That's a deliberate trade-off for avoiding false-positive restarts during long startups. The PR description is explicit about this being a stop-gap.

4. **No edge case risks.** The change cannot produce a false negative that *prevents* a restart — it only delays it. Worst case: a frozen dispatcher takes 30 min to be caught instead of 10. Acceptable for a stop-gap.

5. **No test suite covers these constants** — confirmed. Nothing to break.

## Against the engineer's briefing

The engineer flagged the stale comment on line 102 (catchup margin reference). That comment was updated correctly in the PR. What the briefing did NOT flag: the four other comment blocks across the file still citing "600s" as the operative value (lines 37, 551–553, 969–972, 1891–1901). These should be cleaned up in a follow-up — ideally in the same state machine PR that will retune the thresholds.

## Verdict

**PASS.** The two-line constant change is correct, safe, and the PR description is honest about the trade-off. The stale comments are documentation debt, not blocking issues. Recommend tracking the comment cleanup as a TODO in the state machine PR.

PR #1924, opened 2026-05-04
fix(memory): use k=? instead of LIMIT ? for sqlite-vec KNN queries

aeschylus:
## Problem

Two tests in `tests/unit/test_memory.py` were failing:

- `TestVectorMemory::test_search_returns_relevant`
- `TestVectorMemory::test_search_with_project_filter`

Both failed because `_hybrid_search` always fell through to the keyword-only fallback path, returning 0 results. The logged error:

```
WARNING: Hybrid search failed, falling back to keyword: A LIMIT or 'k = ?' constraint is required on vec0 knn queries.
```

## Root cause

`sqlite-vec v0.1.7-alpha.10` (currently installed) requires the neighbour count for KNN queries to be expressed as `k = ?` in the `WHERE` clause when using a bound parameter. The previous query used `LIMIT ?`, which this version does not recognise as a KNN bound:

```sql
-- Before (broken on v0.1.7-alpha.10+):
SELECT rowid, distance
FROM events_vec
WHERE embedding MATCH ?
ORDER BY distance
LIMIT ?

-- After (works):
SELECT rowid, distance
FROM events_vec
WHERE embedding MATCH ?
AND k = ?
ORDER BY distance
```

Note: `LIMIT 10` (inline integer) also fails; only `k = ?` with a bound parameter works in this version.

## Fix

One-line change to `src/mcp/memory/vector_memory.py`: replace `LIMIT ?` with `AND k = ?` in the vec0 KNN query. The `fetch_limit` parameter binding stays unchanged.

This is a pure function fix — no state changes, no side effects. The `_hybrid_search` function is stateless aside from the DB query.

## Tests run

- [x] `uv run pytest tests/unit/test_memory.py -q` — 40 passed, 0 failed (was 2 failed before fix)
- [x] `uv run pytest tests/unit/ -q` — 3060 passed, 24 skipped, 0 failed (full unit suite clean)

## Manual test

N/A — this change is entirely internal (no Telegram output, no file I/O, no external services).

## Definition of Done

- [x] Integration/manual test run — N/A, change is internal DB query only
- [x] User-visible outcome verified — N/A (internal memory subsystem)

### DIFF
--- a/src/mcp/memory/vector_memory.py
+++ b/src/mcp/memory/vector_memory.py
@@ -258,14 +258,16 @@ class VectorMemory:
         vec_blob = _serialize_vector(query_embedding)
 
         # Get top candidates from vector search (fetch more than limit for merging)
+        # sqlite-vec v0.1.7-alpha+ requires `k = ?` in the WHERE clause for KNN
+        # queries with a bound parameter; `LIMIT ?` is not recognised.
         fetch_limit = limit * 3
         vec_results = self._conn.execute(
             """
             SELECT rowid, distance
             FROM events_vec
             WHERE embedding MATCH ?
+            AND k = ?
             ORDER BY distance
-            LIMIT ?
             """,
             (vec_blob, fetch_limit),
         ).fetchall()


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

**PR #1924 — sqlite-vec KNN query fix**

The fix is correct, minimal, and well-documented:

- Root cause is accurate: `sqlite-vec v0.1.7-alpha.10+` requires `AND k = ?` in the WHERE clause for bound-parameter KNN queries; `LIMIT ?` is not recognised.
- The parameter binding (`fetch_limit`) is unchanged — only the SQL syntax changes.
- The comment added in-code (`# sqlite-vec v0.1.7-alpha+ requires...`) will prevent future regressions when someone later wonders why this isn't using LIMIT.
- 3,060 unit tests pass, including the two previously failing memory tests.
- Pure function change — no state mutations, no side effects, no external calls.

No concerns. Ready to merge.

PR #1932, opened 2026-05-05
refactor(slack): extract _remap_channel() shared helper (PR 2)

aeschylus:
## What problem this solves

After PR #1931 introduced a single `CHANNEL_REMAP` dict, both the inbound and outbound paths still performed the dict lookup inline — two separate `CHANNEL_REMAP.get(channel_id, channel_id)` call sites. The remap logic was no longer duplicated in config, but was still duplicated in code. If either site were edited independently, they could silently diverge.

## What changed

Extracts `_remap_channel(channel_id: str) -> str` as a documented module-level helper. Both the inbound Socket Mode path (`handle_message_events`) and the outbound send path (`_send_slack_reply`) call this helper instead of inlining the lookup.

The function is generic and workspace-independent: it takes a channel ID, applies the `CHANNEL_REMAP` dict, and returns either the mapped destination or the original ID unchanged.

## What is out of scope

No config changes. No behavior change for any deployment — this is a pure refactor of two dict lookups into one function.

## Functional patterns

Pure function (`_remap_channel`) — no side effects, deterministic, testable in isolation. Composition: both message paths delegate to the same helper rather than duplicating the lookup.

## Flow: before vs after

The remap logic itself is unchanged. What changed is the structural guarantee:

**Before:**
```
handle_message_events → inline: CHANNEL_REMAP.get(channel_id, channel_id)
_send_slack_reply     → inline: CHANNEL_REMAP.get(channel_id, channel_id)
```

**After:**
```
handle_message_events → _remap_channel(channel_id)  ─┐
_send_slack_reply     → _remap_channel(channel_id)  ─┘  (same function)
```

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_slack_router_config.py -v` — 21 passed, 0 failed (17 from PR 1 + 4 new)
- [x] Pre-push PII scan — clean, no issues detected

New tests added (`TestRemapChannelHelper`):
- `test_remap_channel_maps_src_to_dst` — source channel returns destination
- `test_remap_channel_returns_unchanged_when_not_in_map` — unknown channel passes through unchanged
- `test_remap_channel_returns_unchanged_when_map_empty` — empty remap table is a no-op
- `test_inbound_and_outbound_use_same_remap_table` — monkey-patches `_remap_channel` and verifies both paths invoke it, enforcing the single-helper invariant

## Manual test

N/A — pure refactor, no external service calls, no changed behavior.

## Definition of Done

- [x] Unit tests pass with proper mocking
- [x] Integration/manual test — N/A (pure refactor)
- [x] User-visible outcome — N/A (no change in message routing or content)
- [x] Side-effect audit — no external calls, no state mutations added
- [x] PII/key scan — clean

### DIFF
--- a/src/bot/slack_router.py
+++ b/src/bot/slack_router.py
@@ -119,6 +119,20 @@ CHANNEL_REMAP: dict = parse_channel_remap(
     os.environ.get("LOBSTER_SLACK_CHANNEL_REMAP", "")
 )
 
+
+def _remap_channel(channel_id: str) -> str:
+    """Return the remapped channel ID, or *channel_id* unchanged if no mapping exists.
+
+    Both the inbound (Socket Mode) and outbound (chat_postMessage) paths share
+    this helper so the remap logic lives in exactly one place.
+
+    The mapping is workspace-specific and comes entirely from the
+    ``LOBSTER_SLACK_CHANNEL_REMAP`` config variable — see ``parse_channel_remap``
+    for the format.  A typical deployment maps the bot-DM channel (which the
+    xoxp- user token cannot post to) to the user-DM channel.
+    """
+    return CHANNEL_REMAP.get(channel_id, channel_id)
+
 # Directories
 _MESSAGES = Path(os.environ.get("LOBSTER_MESSAGES", Path.home() / "messages"))
 _WORKSPACE = Path(os.environ.get("LOBSTER_WORKSPACE", Path.home() / "lobster-workspace"))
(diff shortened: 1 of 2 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

**Pure refactor: confirmed.** The PR extracts a `_remap_channel()` helper from two identical inline lookups (`CHANNEL_REMAP.get(channel_id, channel_id)`) into a single shared function. No logic changes — the semantics of both call sites are identical before and after.

**Both paths use the helper: confirmed.** `handle_message_events` (inbound/Socket Mode) and `_send_slack_reply` (outbound/chat_postMessage) both now call `_remap_channel()`. The diff is symmetric.

**Single-helper invariant is tested: confirmed.** `TestRemapChannelHelper.test_inbound_and_outbound_use_same_remap_table` monkey-patches `_remap_channel` and asserts it is called by both code paths. This is exactly the right test — it catches drift at the call-site level, not just at the logic level.

**PII scan: clean.** The PR's delta introduces only fake placeholder IDs (`DBOT000001`, `DUSER000001`). The test file includes an explicit assertion (`assert "D0B1L2Q99UN" not in text`) enforcing no hardcoded real channel IDs in source. Real IDs visible in `main`'s `slack_router.py` are pre-existing, not introduced by this PR.

**Tests: 196 passed, 0 failed.**

PR #1934, opened 2026-05-05
feat(slack): typing indicator — post placeholder then update with real reply

aeschylus:
## What problem this solves

When a Slack message arrives, the dispatcher takes several seconds to think and compose a reply. During that time, the user sees nothing — then the full reply appears suddenly. This is jarring and makes the system feel unresponsive.

This PR adds a \"typing\" experience: the reply appears in Slack immediately as \"...\" and then fills in with the real text. The user sees the message appear at once; the latency is hidden inside the update.

## How it works

The change is entirely within `_send_slack_reply` in `src/bot/slack_router.py`.

**Before:**
```
outbox file → chat.postMessage(real text)
```

**After (with LOBSTER_SLACK_TYPING_INDICATOR=true, the default):**
```
outbox file → chat.postMessage("...") → capture ts → chat.update(real text, ts)
```

**Fallback path (if postMessage("...") raises SlackApiError):**
```
outbox file → chat.postMessage("...") [fails] → chat.postMessage(real text) [fallback]
```

**Opt-out (LOBSTER_SLACK_TYPING_INDICATOR=false):**
```
outbox file → chat.postMessage(real text)  ← original behaviour preserved
```

## Design

The implementation uses three pure helper functions composed at the boundary:

- `_build_post_kwargs(channel_id, text, thread_ts)` — builds the kwargs dict; no side effects
- `_send_direct(channel_id, text, thread_ts)` — original direct-post path
- `_send_with_typing_indicator(channel_id, text, thread_ts)` — post-then-update path

`_send_slack_reply` selects between them based on `SLACK_TYPING_INDICATOR`. Side effects (Slack API calls) are isolated at the leaves.

## Failure handling

| Failure point | Behaviour |
|---|---|
| `postMessage("...")` raises | Fallback to direct `postMessage(real text)` — no silent drop |
| `chat.update` raises after successful placeholder | Log warning, return `True` — placeholder is visible in Slack; returning `False` would re-queue the outbox file and send a duplicate |
| Both `postMessage` attempts fail | Return `False` — outbox file stays for retry |

## What is not changed

- Inbound path (`handle_message_events`, poller) — unchanged
- Telegram path — unchanged; this is Slack-only
- Outbox file delivery mechanics (`OutboxFileHandler`, `drain_outbox`) — unchanged
- `LOBSTER_SLACK_CHANNEL_REMAP` remapping — still applied before the placeholder post

## Config

New env var in `config/config.env.example`:
```
# LOBSTER_SLACK_TYPING_INDICATOR=true
```

Accepts `true`, `false`, `0`, `1`, `yes`, `no`, `off`. Default is `true`.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_slack_typing_indicator.py -v` — 16 passed, 0 failed
- [x] `uv run pytest tests/unit/test_bot/ -q` — 233 passed, 0 failed
- [x] `uv run python -m py_compile src/bot/slack_router.py` — syntax clean
- [x] PII scan (pre-push hook) — no issues detected

## Manual test

N/A — this change has no direct interaction with systemd, cron, or file I/O beyond what the outbox watcher already does. The Slack API calls are fully mocked in unit tests. A live end-to-end test would require a real Slack token and workspace; unit tests cover all failure modes exhaustively.

The user-visible outcome (seeing "..." appear then fill in) cannot be self-verified without a live Slack workspace. The unit tests verify that `chat_postMessage("...")` is called first and `chat_update(real_text, ts)` is called second with the correct arguments.

## Definition of Done

- [x] Integration/manual test run — N/A (pure logic change, no systemd/cron/file I/O)
- [ ] User-visible outcome verified — requires live Slack workspace; blocked on environment
- [x] Side-effect audit: no floods, no loops, no unscoped writes — the placeholder is a single API call immediately followed by a single update, same total message count as before
- [x] External service calls: fully mocked in tests

### DIFF
--- a/config/config.env.example
+++ b/config/config.env.example
@@ -51,6 +51,11 @@ OPENAI_API_KEY=your_openai_api_key_here
 # LOBSTER_SLACK_ALLOWED_CHANNELS=
 # Optional: Restrict to specific user IDs (comma-separated)
 # LOBSTER_SLACK_ALLOWED_USERS=
+# Typing indicator — immediately post a "..." placeholder, then update with the
+# real reply text so Slack users see the message appear rather than waiting in
+# silence.  Requires LOBSTER_SLACK_USER_TOKEN (xoxp-) and chat.update permission.
+# Default: true.  Set to false if your Slack plan restricts chat.update.
+# LOBSTER_SLACK_TYPING_INDICATOR=true
 
 # Twilio (shared credentials for SMS and WhatsApp)
 # Sign up at https://twilio.com — SID and Auth Token at https://console.twilio.com
--- a/src/bot/slack_router.py
+++ b/src/bot/slack_router.py
@@ -78,6 +78,18 @@ if not SLACK_APP_TOKEN:
 ALLOWED_CHANNELS = [x.strip() for x in os.environ.get("LOBSTER_SLACK_ALLOWED_CHANNELS", "").split(",") if x.strip()]
 ALLOWED_USERS = [x.strip() for x in os.environ.get("LOBSTER_SLACK_ALLOWED_USERS", "").split(",") if x.strip()]
 
+# Typing indicator (post-then-update pattern).
+# When enabled, outbound replies are first posted as a "..." placeholder and
+# then immediately updated with the real text — so the user sees the message
+# appear at once rather than waiting in silence until the full reply is ready.
+# Disable with LOBSTER_SLACK_TYPING_INDICATOR=false (or 0) if the Slack plan
+# does not permit chat.update (e.g. free-tier restrictions).
+_TYPING_INDICATOR_RAW = os.environ.get("LOBSTER_SLACK_TYPING_INDICATOR", "true").strip().lower()
+SLACK_TYPING_INDICATOR: bool = _TYPING_INDICATOR_RAW not in ("false", "0", "no", "off")
+
+# The placeholder text posted before the real reply is ready.
+SLACK_TYPING_PLACEHOLDER = "..."
+
 
 def parse_channel_remap(raw: str) -> dict:
     """Parse ``LOBSTER_SLACK_CHANNEL_REMAP`` into a channel-ID mapping dict.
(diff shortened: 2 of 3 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

## Review Summary

**233/233 tests pass** (16 new + 217 existing). All checklist items cleared.

---

### chat.update failure path

If the placeholder "..." posts successfully but `chat.update` fails, the function returns `True` (consuming the outbox file). The user sees "..." permanently in this error case.

This is the right call. The alternative — returning `False` to leave the outbox file in place — would trigger a second `chat.postMessage`, delivering the reply twice. A stuck "..." is a better outcome than a duplicate. The behavior is clearly documented in both the docstring and inline comments. No change needed.

One minor note: there is no alerting/metric when this happens — the failure is logged at `WARNING` but silently absorbed. If this path ever becomes frequent, it would be invisible. That is probably acceptable for now and is not a blocker.

### thread_ts on chat.update

The `chat.update` call includes `thread_ts`. This is redundant (the message is already in the thread by virtue of its `ts`), but the Slack API accepts it silently and it does no harm. Leaving it in is consistent with the placeholder post. Not a concern.

### Fallback path when placeholder fails

Clean. A `SlackApiError` on the placeholder post calls `_send_direct(channel_id, text, thread_ts)` with the real reply text — never the placeholder text. The fallback carries `thread_ts` correctly. Covered by `test_fallback_to_direct_post_when_placeholder_fails` and `test_no_message_silently_dropped_when_placeholder_fails`.

### PII scan

Test IDs are synthetic (`C0TEST12345`, `U0BOT999`). No real channel or user IDs. Clean for a public repo.

### Code quality

Three focused pure helpers (`_build_post_kwargs`, `_send_direct`, `_send_with_typing_indicator`) make the control flow easy to follow and test in isolation. The `LOBSTER_SLACK_TYPING_INDICATOR` env var with sensible default (`true`) and clear opt-out (`false`/`0`/`no`/`off`) is well-designed.

---

**Verdict: merge.**

PR #1942, opened 2026-05-07
fix: patch DISPATCHER_SESSION_FILE in tests to prevent production corruption

sayhar:
## Summary

- `session_role.DISPATCHER_SESSION_FILE` is a module-level constant resolved at import time (`~/messages/config/dispatcher-session-id`). Tests that patched `HOME`/`LOBSTER_WORKSPACE` in the environment did not affect this already-resolved constant.
- `test_primary_file_match_still_works_without_sentinel` called `mod.main()` with a matching dispatcher UUID, causing `write_dispatcher_session_id()` to write the stub value `known-dispatcher-uuid-abc` to the **production file** at `~/messages/config/dispatcher-session-id`.
- This broke all dispatcher session detection: `inject-bootup-context.py` injected subagent bootup instead of dispatcher bootup on every subsequent session start, and `on-compact.py` exited early on the dispatcher branch.

## Fix

- Explicitly patch `mod.session_role.DISPATCHER_SESSION_FILE` to a `tmp_path`-relative path in all `TestMainSentinelFallback` tests that trigger a dispatcher write.
- Replace the recognisable stub `"known-dispatcher-uuid-abc"` with a UUID4-format value so any accidental production write is still obviously a test artifact.
- Add before/after teardown assertions to both affected tests that fail immediately if the production file was modified.

## Test plan

- [ ] `uv run pytest tests/unit/test_hooks/test_inject_bootup_context_sentinel.py -v` — all 9 tests pass
- [ ] Confirm `~/messages/config/dispatcher-session-id` is not modified by the test run
- [ ] Confirm teardown assertion fires if the patch is removed (regression guard)

### DIFF
--- a/tests/unit/test_hooks/test_inject_bootup_context_sentinel.py
+++ b/tests/unit/test_hooks/test_inject_bootup_context_sentinel.py
@@ -156,6 +156,13 @@ class TestIsPostCompactDispatcher:
 # ---------------------------------------------------------------------------
 
 
+# Real production path for the hook marker file — used in teardown assertions
+# to guarantee tests never write to it.
+_PRODUCTION_DISPATCHER_SESSION_FILE = Path(
+    os.path.expanduser("~/messages/config/dispatcher-session-id")
+)
+
+
 class TestMainSentinelFallback:
     """main() in inject-bootup-context.py uses the sentinel as a fallback
     when is_dispatcher() returns False but the post-compact sentinel is present.
@@ -173,6 +180,11 @@ class TestMainSentinelFallback:
         """When is_dispatcher()=False but sentinel+LOBSTER_MAIN_SESSION=1,
         dispatcher bootup must be injected (not subagent bootup).
         """
+        # Record the production file's state before the test so we can assert
+        # it was not modified (teardown guard).
+        prod_file = _PRODUCTION_DISPATCHER_SESSION_FILE
+        prod_before = prod_file.read_text() if prod_file.exists() else None
+
         # Set up file structure.
         claude_dir = tmp_path / "lobster" / ".claude"
         claude_dir.mkdir(parents=True)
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Verdict

The fix is correct and complete. Root cause, fix mechanism, and regression guards are all sound. 9/9 tests verified passing locally.

## What I Checked

**Root cause confirmed.** `session_role.DISPATCHER_SESSION_FILE` is resolved at import time as a module-level constant. Patching `HOME`/`LOBSTER_WORKSPACE` in the environment after the module is already imported has no effect on this value. Any test that ran the full dispatcher path (`write_dispatcher_session_id()`) was therefore writing to the live production file.

**The patch mechanism works correctly.** Setting `mod.session_role.DISPATCHER_SESSION_FILE = tmp_session_file` mutates the attribute on the module object. Python's global name lookup inside functions goes through the module's `__dict__` at call time, so `write_dispatcher_session_id()` sees the updated value when it executes. This is the correct way to redirect a module-level constant when you cannot control the import environment.

**Coverage of unpatched tests is acceptable.** Two tests in `TestMainSentinelFallback` (`test_no_sentinel_normal_subagent_gets_subagent_bootup` and `test_sentinel_with_lobster_main_session_zero_gets_subagent_bootup`) do not patch `DISPATCHER_SESSION_FILE` and have no teardown assertions. This is fine: neither test reaches the `if is_dispatcher: write_dispatcher_session_id(...)` branch — the first has no matching session ID and no sentinel, the second has `LOBSTER_MAIN_SESSION=0`. No write occurs, so no guard is needed.

**Teardown assertions are correctly placed** in the two tests that do trigger the dispatcher write path, and they will catch any regression where the patch is removed.

## One Minor Nit

The stub `"aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"` is not technically a valid UUID4 (UUID4 requires version nibble `4` and variant nibble `8/9/a/b`). The comment calls it "a realistic UUID4 format," which is slightly imprecise. The stub still achieves the stated goal (it looks like a UUID, not a human-readable sentinel, so an accidental write is still clearly a test artifact) — this is cosmetic-only and does not affect correctness.

## Test Run

```
9 passed in 0.81s
```

Confirmed on the `fix/test-isolation-dispatcher-session-file` worktree against the live system where `~/messages/config/dispatcher-session-id` exists.

PR #1955, opened 2026-05-07
fix: correct context window for claude-sonnet-4-6 from 1M to 200k

### LINKED ISSUE #1953: bug: context-monitor uses 1M token window for claude-sonnet-4-6 — actual CC window is 200k
## Problem

`hooks/context-monitor.py` sets the context window for `claude-sonnet-4-6` to 1,000,000 tokens:

```python
MODEL_CONTEXT_SIZES: list[tuple[str, int]] = [
    ("claude-sonnet-4-6", 1_000_000),
    ("claude-opus-4-6", 1_000_000),
    ("claude-haiku-4-5", 200_000),
]
```

Claude Code's actual context window is 200,000 tokens (not 1M). The 1M figure is the API context window, but Claude Code enforces its own limit.

## Impact

All `used_percentage` values logged for `claude-sonnet-4-6` sessions are **5x too low**:
- When context-monitor reports `8.4%`, the session is actually at `42%` of the real window.
- When context-monitor reports `7.2%`, the session is actually at `36%`.
- The WARNING_THRESHOLD of 70% would not trigger until the session is at `14%` reported = `70%` actual — which means it fires at exactly the right time, but the logged percentages are misleading.

## Concrete impact observed

Discovered during audit of 2026-05-07 double-compaction event (~16:52Z and ~17:01Z). The auditor initially interpreted `8.4%` as low usage and was confused about what triggered the second compaction. Correct interpretation: at `8.4%` reported the dispatcher was at `~42%` of actual 200k window, which explains why a second compaction happened after only 8 minutes of post-compact activity (bootup injection + catchup result + conversation).

## Fix

Change the model context sizes to reflect the actual Claude Code context window:

```python
MODEL_CONTEXT_SIZES: list[tuple[str, int]] = [
    ("claude-sonnet-4-6", 200_000),
    ("claude-opus-4-6", 200_000),
    ("claude-haiku-4-5", 200_000),
]
```

Or source this from a verified CC-specific constant. The exact value should be confirmed against Claude Code's actual enforced limit (200k is the observed value from compaction behavior).

## Note

The WARNING_THRESHOLD alert fires correctly in practice because it's calibrated against 1M (threshold fires at 14% reported = 70% actual), but all logged percentages in `context-monitor.log` are 5x too low, making post-hoc analysis misleading.

sayhar:
## What this fixes

`context-monitor.py` was computing used_percentage against a 1,000,000-token window for `claude-sonnet-4-6` and `claude-opus-4-6`, but Claude Code enforces a 200,000-token context window. The logged percentages were 5x too low — a session at 42% actual usage appeared as 8.4% in the log, causing confusion during post-hoc analysis (specifically: the 2026-05-07 double-compaction audit).

The WARNING_THRESHOLD alert still fired correctly in practice (the miscalibration was symmetric), but every logged `used_percentage` value was wrong, making the logs misleading rather than useful.

## Change

Updated `MODEL_CONTEXT_SIZES` in `hooks/context-monitor.py`:

```python
# Before:
("claude-sonnet-4-6", 1_000_000),
("claude-opus-4-6",   1_000_000),

# After:
# claude-sonnet-4-6 supports up to 1M tokens but CC's default window is 200k.
# Update when we can detect which mode is active.
("claude-sonnet-4-6", 200_000),
# claude-opus-4-6 supports up to 1M tokens but CC's default window is 200k.
# Update when we can detect which mode is active.
("claude-opus-4-6",   200_000),
```

Comments explain the 1M API limit and indicate where to change the value once CC mode detection is possible — making the "easy to update later" requirement explicit in code.

The docstring was also updated: it previously said "1M for Sonnet/Opus 4.x" and now accurately describes the table.

Updated tests scale all token counts to the 200k window. The `SONNET_4_6_MAX_CONTEXT` test constant and two test names (`test_sonnet_4_6_returns_1m`, `test_opus_4_6_returns_1m`) were updated to reflect the corrected value.

## Functional patterns used

Pure functions throughout — `_model_max_context` is a pure prefix-match lookup; the constants are named and centralized so the value is changed in exactly one place.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_context_monitor.py -v` — 19 passed, 0 failed

## Manual test

N/A — this change is entirely internal to the hook's lookup table. No external services, file I/O beyond log writes, or Telegram output involved. The hook fires via the CC PostToolUse harness; the next real session will produce corrected percentages automatically.

## Definition of Done

- [x] Integration/manual test run — N/A (pure lookup table change, no external dependencies)
- [x] User-visible outcome verified — N/A (internal log values only; not directly surfaced to user)

Closes #1953

### DIFF
--- a/hooks/context-monitor.py
+++ b/hooks/context-monitor.py
@@ -13,8 +13,8 @@ Approach:
      contains a `message.usage` block.
   3. Compute total tokens = input_tokens + cache_creation_input_tokens
      + cache_read_input_tokens.
-  4. Look up the model's max context window from a known table (1M for Sonnet/Opus
-     4.x, 200k for Haiku or unknown models).
+  4. Look up the model's max context window from a known table (200k for all
+     current CC models; see MODEL_CONTEXT_SIZES for details).
   5. Compute used_pct and trigger wind-down mode if at or above WARNING_THRESHOLD.
 
 Below the warning threshold (default 70%): logs usage to context-monitor.log.
@@ -65,8 +65,12 @@ DEDUP_FLAG = Path("/tmp/lobster-context-warning-sent")
 # 'claude-haiku-4-5-20251001' resolve correctly.
 # Default fallback: 200_000 (conservative — avoids false negatives on unknown models).
 MODEL_CONTEXT_SIZES: list[tuple[str, int]] = [
-    ("claude-sonnet-4-6", 1_000_000),
-    ("claude-opus-4-6", 1_000_000),
+    # claude-sonnet-4-6 supports up to 1M tokens but CC's default window is 200k.
+    # Update when we can detect which mode is active.
+    ("claude-sonnet-4-6", 200_000),
+    # claude-opus-4-6 supports up to 1M tokens but CC's default window is 200k.
+    # Update when we can detect which mode is active.
+    ("claude-opus-4-6", 200_000),
     ("claude-haiku-4-5", 200_000),
 ]
 DEFAULT_CONTEXT_SIZE = 200_000
--- a/tests/unit/test_hooks/test_context_monitor.py
+++ b/tests/unit/test_hooks/test_context_monitor.py
@@ -13,7 +13,7 @@ Behaviors verified:
 1. Transcript present with usage → correct percentage computed from token counts.
 2. transcript_path absent → WARN logged, no crash.
 3. Last assistant turn is selected when multiple turns exist.
-4. Model lookup table: Sonnet 4.6 = 1M, Haiku 4.5 = 200k, unknown = 200k.
+4. Model lookup table: Sonnet 4.6 = 200k (CC default), Haiku 4.5 = 200k, unknown = 200k.
 5. At or above WARNING_THRESHOLD → context_warning written to inbox (once per session).
 6. Dedup flag suppresses second warning.
 7. _handle_payload() accepts injectable log_dir and inbox_dir.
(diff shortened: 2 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Pure constant correction: the context window size for Sonnet 4.6 and Opus 4.6 drops from 1,000,000 to 200,000 tokens, matching the actual Claude Code default window. All 19 tests pass (verified by running them directly on the branch).

## Math verification

All token fixtures check out:

| Test | Tokens | Window | Expected % | Correct? |
|---|---|---|---|---|
| 50% basic | 20k+40k+40k = 100k | 200k | 50% | yes |
| 80% last-turn | 160k | 200k | 80% | yes |
| 30% below-threshold | 30k+20k+10k = 60k | 200k | 30% | yes |
| 80% warning trigger | 160k | 200k | 80% > 70% threshold | yes |
| 90% dedup suppress | 180k | 200k | 90% > 70% threshold | yes |
| 100% cache sum | 100k+40k+60k = 200k | 200k | 100% | yes |

## Docstring

Line 17 now reads "200k for all current CC models" — the misleading "1M" reference is gone. The inline code comments correctly document that the API supports 1M but CC's default window is 200k, and note that detection logic is pending.

The Opus 4.6 model test asserts the result equals the constant named `SONNET_4_6_MAX_CONTEXT`. Since both models are now 200k, the assertion is mathematically correct and the test passes. It reads slightly oddly — a reader wondering "why is the Opus test asserting against a Sonnet constant?" would have to look twice. Would be cleaner as `DEFAULT_MAX_CONTEXT` or a shared `CC_DEFAULT_CONTEXT`. Not a blocker; noting for awareness.

## Checklist

- [x] Constant change matches the stated intent (1M → 200k for CC default window)
- [x] Both models (Sonnet 4.6, Opus 4.6) updated consistently
- [x] All threshold fixtures verified to be above 70% of the new 200k window
- [x] Docstring accurately updated
- [x] 19/19 tests pass on branch
- [x] No behavioral logic changed — pure constant correction

PR #1965, opened 2026-05-07
fix(tests): isolate inject-bootup-context dispatcher tests from production state files

sayhar:
## Problem

Tests in `test_startup_flag_dispatcher_detection.py` that exercised the dispatcher path of `main()` in `inject-bootup-context.py` were writing to production files on every test run:

- `~/messages/config/dispatcher-session-id` — overwritten with `any-session-uuid`
- `~/lobster-workspace/data/dispatcher-state.json` — overwritten with `STARTING` state + test session ID

This caused the live dispatcher's session role detection to see a stale `STARTING` state (production `dispatcher-state.json` was briefly corrupted), which is what Sahar observed at 18:07 UTC today.

**Root cause:** `_PatchEnv` in these tests patched `LOBSTER_WORKSPACE` but not `HOME`. Two module-level constants in already-cached modules are derived from `HOME`:

- `session_role.DISPATCHER_SESSION_FILE = os.path.expanduser("~/messages/config/dispatcher-session-id")` — evaluated once at import, not affected by `LOBSTER_WORKSPACE` override
- `state_machine.STATE_FILE` — evaluated from `LOBSTER_WORKSPACE` at import time, but `state_machine` was imported before the test's `_PatchEnv` took effect

When `inject-bootup-context.py` detected a live-PID startup flag and called `session_role.write_dispatcher_session_id(real_session_id)` and `state_machine.write_state(STARTING, ...)`, both writes went to the production paths.

## Fix

Three-part fix for all integration tests that call `main()` via the dispatcher path:

1. **Add `HOME` to `_PatchEnv`** — so that any HOME-derived paths resolved during `exec_module` go to `tmp_path`, not production
2. **Set `mod.session_role.DISPATCHER_SESSION_FILE`** to a tmp path after loading the module — redirects `write_dispatcher_session_id()` away from `~/messages/config/dispatcher-session-id`
3. **Set `mod.state_machine.STATE_FILE`** to a tmp path after loading the module — redirects `write_state()` away from `~/lobster-workspace/data/dispatcher-state.json`

Also adds `mod.CONTEXT_INJECTION_LOG` redirection in all affected tests for consistency (previously set by the sentinel tests, missing here).

This is exactly the isolation pattern already established in `test_inject_bootup_context_sentinel.py` — that file was already correct. The startup-flag detection tests added in PR #1960 didn't follow the same pattern.

No production code is modified.

**Before/after flow:**

```
Before (broken):
  test sets LOBSTER_WORKSPACE=tmp_path
  exec_module(inject-bootup-context.py)
  → imports session_role (cached; DISPATCHER_SESSION_FILE = ~/messages/config/...)
  → imports state_machine (cached; STATE_FILE = ~/lobster-workspace/data/...)
  main() detects live PID → dispatcher path
    → write_dispatcher_session_id() → ~/messages/config/dispatcher-session-id  [PRODUCTION]
    → state_machine.write_state()   → ~/lobster-workspace/data/dispatcher-state.json  [PRODUCTION]

After (fixed):
  test sets HOME=tmp_path, LOBSTER_WORKSPACE=tmp_path
  exec_module(inject-bootup-context.py)
  mod.session_role.DISPATCHER_SESSION_FILE = tmp_path/messages/config/dispatcher-session-id
  mod.state_machine.STATE_FILE = tmp_path/data/dispatcher-state.json
  main() detects live PID → dispatcher path
    → write_dispatcher_session_id() → tmp_path/messages/config/...  [ISOLATED]
    → state_machine.write_state()   → tmp_path/data/...  [ISOLATED]
```

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_startup_flag_dispatcher_detection.py -v` — 15 passed, 0 failed
- [x] `uv run pytest tests/unit/ -v` — 3495 passed, 26 skipped in 160s
- [x] Manual verification: wrote sentinel value to `~/messages/config/dispatcher-session-id`, ran the previously-failing test, confirmed sentinel unchanged after the test

## Manual test

N/A — this is a test isolation fix. No production code is modified, no external services are called, no runtime file state changes.

## Definition of Done
- [x] Unit tests pass with proper isolation (no production hits)
- [x] User-visible outcome: N/A — test isolation fix with no user-visible behavior change
- [x] Side-effect audit: fix eliminates unscoped writes; no new writes introduced

### DIFF
--- a/tests/unit/test_hooks/test_startup_flag_dispatcher_detection.py
+++ b/tests/unit/test_hooks/test_startup_flag_dispatcher_detection.py
@@ -80,12 +80,19 @@ class _PatchEnv:
                 os.environ[k] = saved_v
 
 
-def _load_inject_hook(*, workspace: Path) -> object:
-    """Load inject-bootup-context.py with a controlled LOBSTER_WORKSPACE."""
+def _load_inject_hook(*, workspace: Path, home: "Path | None" = None) -> object:
+    """Load inject-bootup-context.py with a controlled LOBSTER_WORKSPACE (and HOME).
+
+    Pass home= to prevent HOME-derived paths (e.g. session_role.DISPATCHER_SESSION_FILE)
+    from resolving to production paths when the hook calls write_dispatcher_session_id().
+    """
     import uuid
 
     unique_name = f"inject_bootup_{uuid.uuid4().hex}"
-    with _PatchEnv({"LOBSTER_WORKSPACE": str(workspace)}):
+    env: dict = {"LOBSTER_WORKSPACE": str(workspace)}
+    if home is not None:
+        env["HOME"] = str(home)
+    with _PatchEnv(env):
         spec = importlib.util.spec_from_file_location(unique_name, _INJECT_HOOK_PATH)
         mod = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(mod)
@@ -243,7 +250,15 @@ class TestStartupFlagDeletion:
 
         hook_input = json.dumps({"session_id": "dispatcher-uuid-abc"})
 
-        with _PatchEnv({"LOBSTER_WORKSPACE": str(tmp_path)}):
+        # HOME must be patched so write_dispatcher_session_id() and
+        # state_machine.write_state() resolve their paths under tmp_path,
+        # not under the real HOME — preventing production file corruption.
+        with _PatchEnv(
+            {
+                "HOME": str(tmp_path),
+                "LOBSTER_WORKSPACE": str(tmp_path),
+            }
+        ):
             spec = importlib.util.spec_from_file_location(
                 "inject_flag_delete_test", _INJECT_HOOK_PATH
             )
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix correctly addresses a real production-corruption bug: tests simulating a live dispatcher startup were writing to active system files on every test run. The three-part isolation strategy (redirecting the home directory environment variable, and explicitly redirecting two module-level state-file constants after loading) is sound and matches the already-established pattern in the sentinel test file.

## What I checked

**Cold read concerns before reading the briefing:**
- Whether the environment-variable patch actually takes effect before the module constants are evaluated (it doesn't for cached imports — addressed by the post-load constant reassignment)
- Whether subagent-path tests (absent flag, dead-PID flag) also needed the state-file redirects
- Whether any test class names were duplicated or shadowed
- Whether the `home=` parameter on `_load_inject_hook` is actually used anywhere

**After comparing against the briefing:**

1. **Subagent-path tests omit the state-file redirects — this is correct.** The hook only calls `write_dispatcher_session_id()` and `write_state()` when it detects a live-PID startup flag. Tests with an absent flag or a dead-PID flag never reach that code path, so no redirect is needed. The omission is not an oversight.

2. **No duplicate class names or shadowing.** Four test classes (`TestIsStartupFlagDispatcher`, `TestStartupFlagDeletion`, `TestMainRoutingViaStartupFlag`, `TestSessionRoleIsDispatcher`) are all distinct and all reachable.

3. **The `home=` parameter on `_load_inject_hook` is wired correctly** — it feeds into `_PatchEnv` for the env-patch path, and the module-level constant reassignments happen at the call sites.

4. **Test count:** 3512 passed, 26 skipped (engineer reported 3495 — delta is from other tests added concurrently; all tests pass cleanly).

5. **No production code modified.** The diff is entirely within the test file.

`test_flag_NOT_deleted_for_subagent` in `TestStartupFlagDeletion` and both subagent tests in `TestMainRoutingViaStartupFlag` now set `HOME=tmp_path` even though, as noted above, the subagent path never writes to the state files. This is harmless over-isolation — it costs nothing and keeps all tests in these classes consistent in their environment setup. No change needed.

## Verdict

Correct fix, correct scope, tests pass. Ready to merge.

PR #1967, opened 2026-05-07
feat(dispatcher): replace UUID session detection with startup-flag (issue #1908)

sayhar:
## What this solves

Dispatcher detection in SessionStart hooks had a chicken-and-egg problem: the hook needed to write a UUID to identify itself as the dispatcher, but needed to identify itself to know which UUID to write. This caused fragile race windows and complex multi-layer fallbacks that were hard to reason about.

The previous approach used 3-4 detection layers: Claude UUID state file, MCP HTTP session ID file, a hook marker file written by a dedicated `write-dispatcher-session-id.py` hook, and a compact-pending sentinel. Any ordering or timing issue between them caused the dispatcher to get subagent bootup context.

## What changed

The launcher (`claude-persistent.sh`) now writes its own PID to `~/lobster-workspace/data/dispatcher-startup-flag` immediately before `exec`-ing claude. This file exists *before* Claude Code starts, so it's always present when `SessionStart` fires:

```
Before:  [ claude starts ] → SessionStart → write UUID → check UUID → (race window)
After:   [ write PID ] → [ claude starts ] → SessionStart → check PID alive → consume flag
```

`inject-bootup-context.py` reads the flag, checks `kill -0 <pid>` for liveness (dead PIDs = stale flags from prior crashes), and deletes the flag after detection so subagents can never see it.

## Scope of changes

**New detection path (SessionStart hooks):**
- `claude-persistent.sh`: write `$BASHPID` to `dispatcher-startup-flag` before exec
- `inject-bootup-context.py`: new `_is_startup_flag_dispatcher()` / `_consume_startup_flag()`
- `session_role.is_dispatcher()`: same startup flag logic

**Intentionally unchanged (still used in PreToolUse hooks):**
- `session_role.is_dispatcher_session()`: PreToolUse hooks use this during active processing, after the startup flag has been consumed. It still uses UUID state files + process tree.

**Removed:**
- `hooks/write-dispatcher-session-id.py`: deleted entirely
- `install.sh`: registration block for the deleted hook removed
- `upgrade.sh`: Migration 86 removes the hook entry from existing `settings.json` installs
- `inject-bootup-context.py`: `_is_post_compact_dispatcher()`, `COMPACT_PENDING_SENTINEL`, `_is_fresh_start_dispatcher()` removed
- `on-compact.py`: no longer writes `dispatcher-claude-session-id` (still writes tertiary marker file for `is_dispatcher_session()` compat)

**Also included — `post-compact-gate.py` ToolSearch fix:**
- Always allows `ToolSearch` through the gate even when compact-pending sentinel is active. Without this, there is a deadlock: the gate tells the dispatcher to call `wait_for_messages`, but `wait_for_messages` is a deferred tool that requires `ToolSearch` to load its schema first. This fix was verified necessary during staging testing of PR #1914.

## Detection flow (before vs after)

Before:
```
SessionStart
├── check dispatcher-claude-session-id (UUID match?)
│   ├── match → dispatcher
│   └── no match → check dispatcher-session-id (HTTP session ID)
│       ├── skipped (format mismatch) → check ~/messages/config/dispatcher-session-id
│       │   ├── match → dispatcher
│       │   └── no match → check compact-pending sentinel
│       │       ├── sentinel + LOBSTER_MAIN_SESSION=1 → dispatcher
│       │       └── no sentinel → check absent primary file + LOBSTER_MAIN_SESSION=1
│       │           └── → dispatcher (fresh start heuristic)
```

After:
```
SessionStart
└── check ~/lobster-workspace/data/dispatcher-startup-flag
    ├── file exists AND PID alive → dispatcher (consume flag)
    └── file absent OR PID dead → subagent
```

## Tests

- `uv run pytest tests/unit/ -q` — 2998 passed, 26 skipped
- `uv run pytest tests/smoke/ -q` — 68 passed
- New: `tests/unit/test_hooks/test_startup_flag_dispatcher_detection.py` — 15 tests covering all startup flag detection edge cases
- Updated: `test_inject_bootup_context_sentinel.py`, `test_session_role_state_file.py`, `test_on_compact_write_claude_session_id.py`, `test_require_background_agent.py`, `test_require_write_result.py`, `test_on_fresh_start_catchup.py`
- Removed: tests for the deleted `write-dispatcher-session-id.py` hook (`TestDispatcherSessionRegistration` in `test_ghost_session_fixes.py`)
- Added: `test_inject_bootup_context_fresh_start.py` (429 lines), `test_inject_bootup_context_logging.py` (568 lines)

Pre-existing failures (not introduced here, confirmed on `origin/main`):
- `test_bisque/test_relay_server.py::TestWSAuth::test_auth_success` (timing/async issue)

## Unblocks

This PR unblocks #1960 and #1962 which depend on the startup-flag infrastructure.

## Notes

This replaces closed PR #1914. The original PR targeted `local-dev`; this one targets `main` directly and is based on the same changes (adapted to main's state, stripped of local-dev-only dependencies like `state_machine.py`).


sayhar (commented):
## Code Review

**Overall: PASS with one architectural note worth discussing**

All tests pass (3150 unit + 68 smoke). The core startup-flag detection logic is correct and well-tested. One behavioral change from the old code is worth explicitly documenting.

---

### What's good

- **Detection logic is correct and clean.** `_is_startup_flag_dispatcher()` in both `inject-bootup-context.py` and `session_role.py` handles all edge cases: absent flag, dead PID, non-integer content, OSError, PermissionError. Conservative defaults throughout (fail → subagent).
- **Flag lifecycle is correct.** PID written by `claude-persistent.sh` before `exec`, consumed by `inject-bootup-context.py` on first SessionStart, so subagents can never see it.
- **ToolSearch gate fix is correct and necessary.** The deadlock (gate says "call WFM" but WFM requires ToolSearch to load its schema) is real; always-allow ToolSearch is the right fix. The error messages are updated consistently.
- **Migration 86 is complete and correctly sequenced** (after 85). `install.sh` registration block for the deleted hook is removed. `upgrade.sh` removes it from existing settings.json installs.
- **`is_dispatcher_session()` (PreToolUse) intentionally unchanged** — the process-tree + state-file logic for active-processing hooks is preserved.
- **Test coverage is thorough** — 15 dedicated startup-flag tests, all edge cases covered. Existing sentinel tests properly migrated to startup-flag equivalents.

---

### One architectural note: compact-matcher inject-bootup-context.py now injects subagent bootup

On the old code path, `on-compact.py` wrote the new post-compact UUID to `dispatcher-claude-session-id`, so the compact-matcher `inject-bootup-context.py` run would detect the dispatcher and inject dispatcher bootup.

In this PR, `write_dispatcher_claude_session_id()` is no longer called by `on-compact.py`, and the startup flag is already consumed at initial launch. As a result, when the compact-matcher `inject-bootup-context.py` fires after compaction, `_is_startup_flag_dispatcher()` returns False, and the dispatcher gets **subagent bootup** injected into its fresh post-compact context.

This is mentioned in `test_on_compact_write_claude_session_id.py` — the test was explicitly updated to assert the primary file is NOT written, with the comment "inject-bootup-context.py uses startup flag (not UUID files)." So this appears intentional.

The recovery path via post-compact-gate + compact-reminder in inbox is functionally sufficient: the gate blocks all tools, forces `wait_for_messages`, returns the compact-reminder which explicitly establishes dispatcher identity and instructs the dispatcher to read `sys.dispatcher.bootup.md`. The dispatcher will get correct context — just via the compact-reminder rather than automatic injection.

**Worth noting in the PR description:** the compact-matcher entry in `install.sh` still has the comment "ensures bootup files are injected into the fresh context that follows a compaction event" — this is no longer accurate for the dispatcher. The compact-reminder + gate is now the re-orientation mechanism for compact events. A brief note in the PR or a comment update in `install.sh` would make this explicit.

---

### Minor: `_is_fresh_start_dispatcher()` is dead code

The function is defined at line 108 of `inject-bootup-context.py` but never called — `main()` uses only `_is_startup_flag_dispatcher()`. The function and its 429-line test file (`test_inject_bootup_context_fresh_start.py`) exercise code that has no execution path. Not blocking, but worth a cleanup pass.

---

### ToolSearch gate fix: no unit tests

The `ToolSearch` always-allow path in `post-compact-gate.py` has no test coverage — `test_post_compact_gate.py` doesn't include a `ToolSearch` case. Also not blocking given the logic is simple (string equality check → exit 0), but a test would be consistent with the rest of the suite.

---

### Summary

| Area | Status |
|------|--------|
| Startup-flag detection correctness | PASS |
| Old UUID-based detection removed | PASS |
| Migration 86 complete | PASS |
| install.sh updated | PASS |
| Tests pass (3150 unit + 68 smoke) | PASS |
| ToolSearch gate fix correct | PASS |
| Compact bootup injection change | Intentional, undocumented |
| Dead code `_is_fresh_start_dispatcher` | Minor cleanup |
| ToolSearch gate: no test | Minor gap |

Ready to merge. The compact-related note is worth a quick comment in `install.sh` or the PR description to save future confusion, but is not blocking.

PR #1968, opened 2026-05-07
fix(#1939): document background: true frontmatter as authoritative background signal; add edge-case tests

### LINKED ISSUE #1939: Dispatcher cannot spawn background subagents — Agent tool schema rejects run_in_background, but require-background-agent.py demands it
## Summary

The dispatcher cannot use the in-context Agent tool to spawn background subagents because two things are simultaneously true: `hooks/require-background-agent.py` blocks any Agent call that does not have `run_in_background: true`, but the Agent tool's JSON schema (as exposed via ToolSearch/deferred loading) declares `additionalProperties: false` with no `run_in_background` field — meaning any attempt to pass that parameter is either stripped by schema validation or arrives coerced to a string, causing the hook to fail-closed every time.

## Reproduction

1. Dispatcher receives a user message and calls the Agent tool with `subagent_type`, `description`, and `prompt` (normal background delegation).
2. `PreToolUse` hook `require-background-agent.py` fires and checks `inp.get("run_in_background") is True`.
3. Hook blocks with: `BLOCKED: Dispatcher called Agent without run_in_background: true.`
4. Dispatcher cannot recover — the Agent tool is unusable for background spawning.

**Observed on:** 2026-05-06, when dispatcher tried to handle a user file-attachment task and could not delegate to `lobster-generalist`.

## Why It Happens

The Agent tool schema (as returned by the deferred tool registry) declares:
```
additionalProperties: false
fields: description, isolation, model, prompt, subagent_type
```

`run_in_background` is **not listed** as a valid parameter. Newer Claude Code versions appear to have removed it from the schema or the schema version being loaded no longer carries it. When the dispatcher includes `run_in_background=True`, schema validation either strips the unknown field or coerces it — so the hook's `is True` check (which requires a Python boolean) never passes.

### Fix 1: Loosen the hook check (quick fix)

In `hooks/require-background-agent.py`, replace the strict identity check:
```python
# Before
if inp.get("run_in_background") is True:
```
with a truthy string-tolerant check:
```python
# After
val = inp.get("run_in_background")
if val is True or str(val).lower() == "true":
```

This handles coerced-to-string booleans that schema processing may produce.

### Fix 2: Move dispatcher spawn pattern to Bash + `claude --print` (robust fix)

Update the canonical dispatcher spawn pattern in `.claude/sys.dispatcher.bootup.md` to use Bash with the `claude` CLI instead of the Agent tool:

```bash
claude --print --dangerously-skip-permissions --agent lobster-generalist -- "$PROMPT"
```

**Note:** The `--` separator is required because YAML frontmatter (`---`) at the top of the prompt is otherwise interpreted by the claude CLI as a flag.

Run this via the Bash tool with `run_in_background=true` — the `PreToolUse` hook on Bash does not enforce the background-agent policy, so this path is hook-free for the dispatcher.

Update `sys.dispatcher.bootup.md` to document this as the canonical pattern and explain why the Agent tool direct call no longer works.

## Additional Context

- The hook intent is correct (force background spawning to prevent context bloat), but it assumes the schema passes `run_in_background` through reliably.
- Fix 1 is a one-line patch; Fix 2 is more robust long-term as it doesn't depend on schema stability.
- Both fixes should be considered together — Fix 1 as a short-term guard, Fix 2 as the durable pattern.

sayhar:
## Summary

The Lobster dispatcher spawns background subagents via the Agent/Task tool. A `PreToolUse` hook (`require-background-agent.py`) blocks any Agent call that does not carry a background intent signal. The fix for this — checking `background: true` in the prompt's YAML frontmatter — was already in the hook, but the dispatcher's canonical spawn example in `sys.dispatcher.bootup.md` still showed `run_in_background=true` as a tool parameter, which CC silently strips before the hook sees it. This left the doc instructing a behaviour that does nothing, while the actual working fix (frontmatter key) was undocumented.

Before this PR: the bootup doc said "use both signals" — implying `run_in_background=true` as a tool param was a working signal. It is not. CC's Agent schema has `additionalProperties: false`; any extra field is stripped before `PreToolUse` hooks fire.

After this PR: the bootup doc shows only the frontmatter signal, explains why the tool parameter is stripped, and the test suite adds 17 new tests covering the frontmatter parser edge cases.

## What changed

**`sys.dispatcher.bootup.md`** — Delegation Pattern section:
- Removed `run_in_background=true` from the Task call example (it never reaches the hook)
- Replaced the misleading "two signals required" note with a clear explanation: CC strips `run_in_background` from tool_input (Agent schema: `additionalProperties: false`), so the prompt frontmatter `background: true` is the only reliable signal (issue #1939)

**`tests/unit/test_hooks/test_require_background_agent.py`** — 17 new tests:
- `TestFrontmatterSentinel`: 4 new integration tests — no frontmatter, old `run_in_background` key in frontmatter (should block), plain prompt (no frontmatter), leading whitespace before `---`, unclosed frontmatter block
- `TestHasBackgroundTrueInFrontmatter`: 13 pure-function unit tests for `_has_background_true_in_frontmatter`, covering `background: true/True/yes/1`, `background: false`, absent key, empty string, unclosed block, leading whitespace strip, old `run_in_background` key rejection, spaces around colon, key after other fields

No changes to hook logic — the frontmatter check itself was already correct in this branch.

## Functional patterns

Pure functions: `_has_background_true_in_frontmatter` is a pure `str -> bool` predicate with no side effects. Tests are derived from the spec (each named for the behaviour being verified, not the mechanism). Named constant `_BACKGROUND_TRUE_VALUES` ties the truthy values to the implementation so tests can verify the set without magic literals.

## Before / after: spawn pattern

```
Before (misleading — run_in_background stripped, "both required" note):

    Task(
        prompt="---\ntask_id: x\nchat_id: y\nsource: telegram\nbackground: true\n---\n\n...",
        subagent_type="...",
        run_in_background=true   # ← stripped by CC schema; never reaches hook
    )
    > Background intent — two signals required: ...

After (correct — only the frontmatter key is documented):

    Task(
        prompt="---\ntask_id: x\nchat_id: y\nsource: telegram\nbackground: true\n---\n\n...",
        subagent_type="..."      # ← no run_in_background param
    )
    > Background intent via prompt frontmatter: Always include background: true
    > in the YAML frontmatter block. Do NOT pass run_in_background=true as a
    > separate tool parameter — CC's Agent schema strips it before the hook sees it.
```

Closes #1939

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_require_background_agent.py -v` — 37 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 3512 passed, 26 skipped, 0 failed (full suite, both runs)

## Manual test
N/A — this change is entirely internal to hook logic and documentation. No external service calls, no Telegram output, no file system state affected at runtime. The hook's parsing logic was already correct; only the documentation and test coverage changed.

## Definition of Done
- [x] Integration/manual test run (if applicable) — N/A, pure hook + doc change
- [x] User-visible outcome verified OR not applicable — not applicable (hook change only)

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -203,18 +203,18 @@ Never say "Noted." alone — it doesn't tell the user whether work is happening.
 3. Write in-flight entry (see "In-flight work tracking" below)
 4. Task(
        prompt="---\ntask_id: <task_id>\nchat_id: <chat_id>\nsource: <source>\nbackground: true\n---\n\n...",
-       subagent_type="...",
-       run_in_background=true
+       subagent_type="..."
    )
 5. mark_processed(message_id)
 6. Return to wait_for_messages() IMMEDIATELY
 ```
 
-> **Background intent — two signals required:** Always include BOTH `run_in_background=true` as a
-> tool parameter AND `background: true` in the prompt frontmatter. The `require-background-agent.py`
-> hook accepts either, but some Claude Code versions (2.1.123+) strip `run_in_background` before
-> the hook sees it when the Agent schema has `additionalProperties: false` (issue #1872). The
-> frontmatter key is the schema-safe fallback and must be in every spawned-agent prompt.
+> **Background intent via prompt frontmatter:** Always include `background: true` in the YAML
+> frontmatter block of every spawned-agent prompt. Do NOT pass `run_in_background=true` as a
+> separate tool parameter — CC's Agent tool schema declares `additionalProperties: false` and
+> strips any extra fields before the `require-background-agent.py` PreToolUse hook sees them
+> (issue #1939). The frontmatter key survives schema validation because it is part of the
+> `prompt` string, which is always a declared field.
 
 Agent registration is fully automatic — a PostToolUse hook fires after each Task call. You do not need to call `register_agent`.
 
--- a/tests/unit/test_hooks/test_require_background_agent.py
+++ b/tests/unit/test_hooks/test_require_background_agent.py
@@ -14,6 +14,8 @@ Tests cover:
 - Frontmatter sentinel: absent or false means hard block for dispatcher (exit 2)
 - Sentinel is case-insensitive (background: True, background: true both accepted)
 - Subagent with no sentinel is still allowed (no enforcement for subagents)
+- _has_background_true_in_frontmatter: pure-function edge cases (leading whitespace,
+  old-style run_in_background key, background: yes/1 variants, unclosed frontmatter)
 """
 
 import importlib.util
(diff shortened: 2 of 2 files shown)


sayhar:
Retargeting to main as #1969 per local-dev cleanup directive

PR #1973, opened 2026-05-07
fix(hooks): reliable restart vs compaction detection via last-startup-cause.json

### LINKED ISSUE #1972: fix: dispatcher cannot reliably detect restart vs compaction at bootup
## Problem

The dispatcher repeatedly misidentifies external service restarts (e.g. systemd restarts) as context compactions. All current signals are unreliable:

- `compaction-state.json`: `last_compaction_ts` gets updated by catchup subagents — misleading
- `last-restart-reason.json`: only written on graceful exits; external kills leave no entry
- `context-handoff.json`: only written if the context_warning handler ran before restart — not guaranteed

## Desired Fix

Introduce a dedicated `~/lobster-workspace/data/last-startup-cause.json` file as the single source of truth for startup cause detection.

**Design:**

1. **`on-compact.py`** (fires before graceful compaction): write `{"cause": "compaction", "ts": "<iso_utc>"}` to `last-startup-cause.json` before exiting.

2. **`inject-bootup-context.py`** (SessionStart hook, fires on every startup): on startup, read `last-startup-cause.json`. Apply this logic:
   - If `cause == "compaction"` and `ts` is within last 5 minutes: this was a compaction-triggered restart
   - Otherwise (absent, stale, or cause == "restart"): this was a plain restart
   - After reading: overwrite with `{"cause": "restart", "ts": "<now>"}` so next startup defaults to restart

3. **`sys.dispatcher.bootup.md`**: add step to read `last-startup-cause.json` and use its value to distinguish compaction from external restart in the startup sequence.

The new file is written BEFORE process exit (by `on-compact.py`) so it survives into the next startup. The inject hook reads and then immediately resets it to "restart" so it self-clears after use.

## Implementation Plan

- [ ] Add `write_startup_cause()` to `on-compact.py` that writes `{"cause": "compaction", "ts": "<now>"}`
- [ ] Add `read_and_reset_startup_cause()` to `inject-bootup-context.py` that reads, classifies, then overwrites with `{"cause": "restart"}`
- [ ] Update `sys.dispatcher.bootup.md` startup steps (step 2d area) to read the new file
- [ ] Write unit tests for both hooks
- [ ] Open PR against main

sayhar:
## Summary

The dispatcher repeatedly misidentified external restarts (systemd, health-check kills) as context compactions at bootup. The root cause: no unambiguous signal existed. `compaction-state.json`'s `last_compaction_ts` is updated by catchup subagents and triggers false positives for plain restarts.

This PR introduces `~/lobster-workspace/data/last-startup-cause.json` as the single source of truth for startup cause classification, with a write-before-exit / read-and-reset-on-startup protocol.

Closes #1972

## How it works

```
Normal restart (systemd, health-check kill):
  on-compact.py: [never fires]
  inject-bootup-context.py: reads absent/stale file → cause = "restart"
                             resets file to {"cause": "restart"}
                             injects banner: "Startup cause: restart"

Compaction:
  on-compact.py: [fires before exit]
                 writes {"cause": "compaction", "ts": "<now>"}
  inject-bootup-context.py: reads file → ts < 5 min → cause = "compaction"
                             resets file to {"cause": "restart"}
                             injects banner: "Startup cause: compaction"
```

The 5-minute window prevents a compaction entry from persisting across an unrelated subsequent restart. After reading, the file is always reset to `cause=restart`, so the default for any unexplained startup is restart.

## Changes

**`hooks/on-compact.py`**
- Adds `STARTUP_CAUSE_FILE` path constant (overridable via `LOBSTER_STARTUP_CAUSE_FILE_OVERRIDE` for tests)
- Adds `write_startup_cause()`: writes `{"cause": "compaction", "ts": "<utc>"}` using atomic rename; called as the first action in `main()` before all other writes; silent on failure

**`hooks/inject-bootup-context.py`**
- Adds `STARTUP_CAUSE_FILE`, `COMPACTION_CAUSE_WINDOW_SECONDS = 300` constants
- Adds `read_and_reset_startup_cause()`: reads, classifies (compaction vs restart), resets to restart; silent on all failures
- Adds `_reset_startup_cause_to_restart()`: the atomic-rename reset helper
- In `main()`: calls `read_and_reset_startup_cause()` for every session (dispatcher and subagent both reset the file; only dispatcher acts on the result); injects a `<!-- startup-cause: ... -->` banner + bold line at the top of the dispatcher's bootup content

- Step 2d: replaced "check compaction-state.json for last_catchup_ts" with "read the startup-cause banner injected at the top of this file" — no extra file I/O needed
- Explicit warning: do not use `compaction-state.json` / `last_catchup_ts` alone to determine cause
- Post-bootup status message: `clean restart` vs `context gap` now derived from the startup-cause banner

## Functional patterns

Pure functions (`write_startup_cause`, `read_and_reset_startup_cause`) with explicit error isolation (all `try/except Exception: pass` guards at the boundary). Side effects are limited to the two hook entry points. `COMPACTION_CAUSE_WINDOW_SECONDS` is a named constant referenced in both the implementation and tests — no magic literals.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_startup_cause.py -v` — 15 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 390 passed, 7 skipped, 0 failed (no regressions)

## Manual test

N/A — this change is entirely internal to hook behavior. The startup cause banner appears in stderr (`[inject-bootup-context] startup cause: restart|compaction`) on every startup, verifiable in `~/lobster-workspace/logs/context-injection.log`. No external service calls, no Telegram output, no DB changes.

## Definition of Done
- [x] Integration/manual test run — N/A (no external services touched)
- [x] User-visible outcome — N/A (purely internal hook behavior)
- [x] Side-effect audit: only writes a small JSON file; no loops, no volume, no shared state affected beyond what is specified

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -51,10 +51,11 @@ When you first start (or after reading this file), follow these steps:
 2c. Check `~/lobster-workspace/data/context-handoff.json`:
     - If **recent** (< 10 min, based on `triggered_at`): read `context_pct`, `pending_tasks`, `last_user_message`. Notify user: "Restarted — context was at {context_pct}%. Resuming from where we left off." Re-queue any stuck messages from `~/messages/processing/`. Delete the file.
     - If **stale** (>= 10 min) or absent: ignore.
-2d. Check `~/lobster-workspace/data/compaction-state.json` for `last_catchup_ts`:
-    - `gap_seconds > 15`: log the gap for context; stay silent toward the user (restarts are invisible by default).
-    - `gap_seconds <= 15`: stay silent (health-check restart, not a meaningful gap).
-    - Skip if step 2c already sent a restart notification.
+2d. **Determine startup cause** — read it from the `<!-- startup-cause: ... -->` banner injected at the top of this file by `inject-bootup-context.py`. Do not read `last-startup-cause.json` yourself; the hook already read and reset it.
+    - `startup-cause: compaction` → this was a context compaction. Expect the `compact-reminder` message in the inbox. Spawn `compact-catchup` at step 4 as usual.
+    - `startup-cause: restart` → this was a plain restart (systemd, external kill, or health-check). No compact-reminder will be in the inbox. Spawn `startup-catchup` at step 4 for a normal restart window.
+    - Skip if step 2c already sent a restart notification (context-handoff.json was recent).
+    - **Do not use `compaction-state.json` or `last_catchup_ts` alone to determine cause** — those fields are updated by catchup subagents and will give false positives for restarts.
 
 3. **Claim any pending user messages immediately** to stop the health-check staleness clock:
     - Call `check_inbox()` to get any messages currently waiting in the inbox
@@ -89,7 +90,7 @@ Open tasks/commitments: [count]
 Fill in:
 - `session_id` from `current_session_file` (e.g. `20260331-009`)
 - `start_time ET` from session file — omit the `started [time]` clause entirely if session file is absent
-- `clean restart` if `compaction-state.json` gap was ≤15s; otherwise `context gap of ~Xm recovered` (X = gap in minutes)
+- `clean restart` if `startup-cause: restart` (from the banner injected at the top of this file); `context gap of ~Xm recovered` if `startup-cause: compaction` (X = gap in minutes between `last_compaction_ts` in `compaction-state.json` and now)
 - N and M from `msg["text"]` (the catchup result)
 - PR count and numbers from handoff.md "PRs needing sign-off" section
 - Task/commitment count from handoff.md — omit if handoff is absent; do NOT call `list_tasks` as a fallback
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -29,6 +29,7 @@ flags (dead PID) are safe because the check is purely process-existence-based.
 import json
 import os
 import sys
+import time
 from datetime import datetime, timezone
 from pathlib import Path
 
@@ -59,6 +60,21 @@ CONTEXT_INJECTION_LOG = _LOBSTER_WORKSPACE / "logs" / "context-injection.log"
 # Startup flag written by claude-persistent.sh before exec-ing claude.
 # Contains the launcher subshell PID. Deleted after the dispatcher is detected.
 STARTUP_FLAG_FILE = _LOBSTER_WORKSPACE / "data" / "dispatcher-startup-flag"
+
+# Single source of truth for startup cause classification (issue #1972).
+# on-compact.py writes {"cause": "compaction", "ts": "<iso_utc>"} before exiting.
+# This hook reads + resets it on every startup. Override via env var for tests.
+STARTUP_CAUSE_FILE = Path(
+    os.environ.get(
+        "LOBSTER_STARTUP_CAUSE_FILE_OVERRIDE",
+        str(_LOBSTER_WORKSPACE / "data" / "last-startup-cause.json"),
+    )
+)
+
+# Maximum age in seconds for a "compaction" cause entry to be trusted.
+# Beyond this window we fall back to "restart" to avoid misclassifying a
+# compaction that was followed by an unrelated external restart.
+COMPACTION_CAUSE_WINDOW_SECONDS = 300  # 5 minutes
 
 
 def _is_startup_flag_dispatcher() -> bool:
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

Solid, well-scoped fix. All 15 new tests pass, 390 total pass, no regressions. The core protocol — write-before-exit in on-compact.py, read-and-reset-on-startup in inject-bootup-context.py — is the right abstraction and correctly handles every failure mode.

## Specific checks

**Write order (on-compact.py)**
`write_startup_cause()` is the very first call in `main()`, before `write_compacted_at()` and everything else. ✓

**Atomic rename pattern**
Both `write_startup_cause()` and `_reset_startup_cause_to_restart()` use `tmp_path.write_text(...)` → `tmp_path.replace(STARTUP_CAUSE_FILE)`. The `.replace()` call on Linux is atomic (same filesystem guaranteed since tmp is `STARTUP_CAUSE_FILE.with_suffix(".tmp")`). `STARTUP_CAUSE_FILE.parent.mkdir(parents=True, exist_ok=True)` is called first in both, so the missing-directory case is handled. ✓

**5-minute window**
The window is `age_seconds < COMPACTION_CAUSE_WINDOW_SECONDS` (strictly less than). Compaction→startup cycle is: Claude terminates, on-compact.py fires (~instant), then systemd restarts the process, hook runs. In practice the window needs to cover heavy catchup only if a *subsequent unrelated restart* happens within 5 minutes of a compaction. 5 minutes is adequate for that; the concern is not slow catchup (catchup runs after the banner is already read). ✓

**Subagent consuming the entry**
The subagent race concern is real but well-mitigated. The startup-flag file (`dispatcher-startup-flag`) is written by the launcher PID *before* exec-ing Claude, and the dispatcher's SessionStart hook (`_is_startup_flag_dispatcher`) checks whether that PID is still alive (via `os.kill(pid, 0)`) and immediately consumes the flag file (`_consume_startup_flag()`). So the dispatcher's hook runs and consumes the flag before any subagent's hook has a chance to register as dispatcher. The `read_and_reset_startup_cause()` runs for *all* sessions (dispatcher and subagents), but that's actually the design intent: only the first session to start after a compaction (the dispatcher) reads "compaction"; subsequent subagent startups all see "restart" because the file is reset. The key invariant is that the dispatcher session starts before any subagent spawned from it, which is guaranteed by the startup-flag mechanism. ✓

**Tests**
All enumerated failure modes are covered: absent file, corrupt JSON, stale ts, exact boundary, reset-after-read, reset-write-failure-is-silent, atomic rename leaves no .tmp. The `test_reset_write_is_silent_on_failure` test correctly patches `mod.STARTUP_CAUSE_FILE` after load to force a write error. One nit: this test comment says "Must not raise; should return 'compaction' from the readable file" but the actual read happens against the original path before the override is applied, so the assertion is technically "no exception" rather than a value assertion — which is fine and intentional, but slightly misleading. Not blocking.

**bootup.md description**
Step 2d now correctly tells the dispatcher to read the `<!-- startup-cause: ... -->` banner rather than parse `compaction-state.json`. The warning "do not use compaction-state.json or last_catchup_ts alone" is explicit and well-placed. The post-bootup status message instruction (`context gap of ~Xm recovered` still reads `last_compaction_ts` from `compaction-state.json` to compute X) is appropriate since cause is already determined by that point — no false positive risk there. ✓

**Is this the right approach?**
Yes. The alternative (checking compaction-state.json timestamps) had a fundamental ambiguity: a catchup subagent updates `last_compaction_ts`, so you can't distinguish "compaction just happened" from "catchup subagent just ran after an unrelated restart." The new file has exactly one writer (`on-compact.py`) and one reader/consumer (`inject-bootup-context.py`). The self-clearing reset ensures stale entries can't persist. The banner injection means the dispatcher reads a rendered value rather than parsing a file — no extra I/O in the startup path. The approach is sound.

## Minor notes (non-blocking)

- The `except (OSError, json.JSONDecodeError, Exception)` in `read_and_reset_startup_cause()` is redundant — `Exception` already covers the first two. Fine as-is (defensive intent is clear), but could be simplified to `except Exception`.
- No test for the case where the `cause` key is entirely absent (not `"compaction"`, not `"restart"`, just missing). The code handles it via `data.get("cause", "restart")` which returns `"restart"` — correct — but it's untested. Suggest adding a test for `{"ts": "<fresh>"}` (no cause key) to lock in that behavior explicitly.
- The 5-minute window is hardcoded in the test file as a local constant rather than imported from the hook. If the constant changes, the tests won't catch a divergence. Minor.

None of these are blocking. The implementation is correct and the tests exercise all meaningful failure paths.

Issue #1977, opened 2026-05-07
Stop hook: write context-handoff.json on session end to survive hard context limits and crashes


sayhar:
## Problem

When a dispatcher session ends abruptly — context overload without hitting the 70% `context_warning` threshold, or a crash — it doesn't write \`~/lobster-workspace/data/context-handoff.json\`. The next session starts completely cold: no record of what the previous session was actively working on, what tasks were in-flight, or what the last user message was.

This has been happening frequently (8+ compactions in a single day). The existing \`context_warning\` handler (LLM-level code in \`sys.dispatcher.bootup.md\`) writes the handoff file at step 5 of wind-down, but only when:
1. Context usage reaches ≥70%, AND
2. The \`context-monitor.py\` PostToolUse hook fires and writes a \`context_warning\` inbox message, AND
3. The dispatcher processes that message before context is exhausted

Sessions that hit the hard context limit too fast (before the warning threshold) or crash bypass this entirely.

## Root cause

The \`context-handoff.json\` write path lives entirely in LLM dispatcher logic, not in a hook. There is no fallback mechanism to record session state when the LLM can no longer act.

## Existing infrastructure

- \`hooks/dispatcher-state-stop.py\` — Stop hook that already fires for the dispatcher and writes DEAD state. This is the correct extension point.
- \`context-monitor.py\` — PostToolUse hook that reads the transcript JSONL and computes \`used_pct\`. Same approach can recover \`context_pct\` in the Stop hook.
- \`state_machine.py\` — Already imported in the Stop hook.

## Proposed fix

Extend \`dispatcher-state-stop.py\` to also write a lightweight \`context-handoff.json\` whenever the dispatcher session ends, using whatever state can be recovered from the transcript at Stop time. Minimum payload:

```json
{
  "triggered_at": "<iso8601>",
  "context_pct": <float or null>,
  "note": "Stop hook wind-down"
}
```

If context_handoff.json already exists (written by the graceful wind-down LLM handler), the Stop hook should **not overwrite it** — the graceful version has richer data (pending_tasks, last_user_message). The Stop hook is a fallback for the case where the LLM never got to write it.

## Acceptance criteria

- Stop hook writes \`context-handoff.json\` when it doesn't already exist
- \`context_pct\` is populated from transcript JSONL when available; \`null\` when not
- \`note\` is always \`"Stop hook wind-down"\`
- \`triggered_at\` is the current UTC ISO timestamp
[...]

PR #1979, opened 2026-05-08
fix(hooks): emit session.end to events.jsonl via EventBus on Stop hook (#1977)

### LINKED ISSUE #2002: context-handoff: change Stop hook output from singleton JSON to append-log (JSONL)
## Problem

The Stop hook currently writes `context-handoff.json` using a write-if-absent pattern: if the file doesn't already exist, write it; if it does, skip (to preserve the richer graceful wind-down version). This singleton approach has two drawbacks:

1. The "skip if exists" logic adds unnecessary branching. The graceful-wind-down case simply writes the file before the Stop hook fires, so the hook checking for existence is cargo-culting around a race that doesn't exist in practice.
2. A singleton file loses historical context — after each restart the previous record is gone.

## Proposed change

Replace the singleton write with a log-append pattern: `context-handoff.jsonl`. Each session end appends one JSON line. The startup reader reads the last line (most recent record). If the file doesn't exist or is empty, treat as "no prior context" — same behavior as before.

This simplifies the hook (no existence check needed), gives a historical record, and aligns with the project's direction of replacing one-shot state files with JSONL logs.

## Scope

- `hooks/dispatcher-state-stop.py`: change `_write_handoff_if_absent` to `_append_handoff_entry`; change output path to `context-handoff.jsonl`
- `sys.dispatcher.bootup.md` step 2c: update to read last line of `context-handoff.jsonl`
- `tests/unit/test_hooks/test_dispatcher_state_stop_handoff.py`: update tests — "write if absent" → "always append"; add test verifying multiple entries accumulate

## Fields preserved

All existing fields: `triggered_at`, `context_pct`, `in_flight_agents`, `note`

sayhar:
## Problem

When a dispatcher session ends abruptly — hitting the hard context limit before reaching 70%, or crashing — the next session starts cold with no record of what was in flight. Previous designs addressed this with `context-handoff.json` (singleton) then `context-handoff.jsonl` (append log). This third design eliminates the separate file entirely.

## Redesign (v2 — EventBus)

All Lobster system events go through the central EventBus, which writes to `~/lobster-workspace/logs/events.jsonl`. The Stop hook now emits a `session.end` LobsterEvent directly to that file rather than maintaining a separate `context-handoff.jsonl`. The startup reader finds context by scanning `events.jsonl` backwards for the last `session.end` event.

This removes a parallel state file, unifies session-end history with the rest of the system event log, and aligns with the project's EventBus architecture.

### Before / After

```
Before (v2 — append-log JSONL):
  Session ends → append line to ~/lobster-workspace/data/context-handoff.jsonl
  Startup → read last line of context-handoff.jsonl

After (v3 — EventBus):
  Session ends → append session.end LobsterEvent to ~/lobster-workspace/logs/events.jsonl
  Startup → scan events.jsonl backwards for last event_type=="session.end"
```

No separate file. The same file the EventBus always writes to.

## Locking finding

`GzipRotatingFileHandler` uses `handler.acquire()` / `handler.release()` (standard `logging.Handler` threading.RLock) around its stream write. The Stop hook does NOT use the handler — it writes directly with `open(path, "a")`.

**Linux O_APPEND atomicity is sufficient here.** Linux guarantees that O_APPEND writes are atomic for payloads smaller than `PIPE_BUF` (4096 bytes on Linux). A single `session.end` JSON line is ~200–400 bytes, well under that limit. There is no shared in-process handler to acquire — each MCP server process holds its own handler lock, which is irrelevant to this out-of-process hook.

## Changes

**`hooks/dispatcher-state-stop.py`**
- `_append_handoff_entry()` → `_emit_session_end_event()`: writes to `~/lobster-workspace/logs/events.jsonl` in LobsterEvent format (matching `LobsterEvent.to_dict()` field names exactly).
- `_build_handoff_payload()` → `_build_session_end_event()`: produces full LobsterEvent dict with `event_type`, `severity`, `source`, `payload`, `timestamp`, `task_id`, `chat_id`.
- `_resolve_handoff_path()` → `_resolve_events_path()`: points to `logs/events.jsonl`.
- All `context-handoff.jsonl` references eliminated.

- Step 0: removed `context-handoff.jsonl` fallback for ADMIN_CHAT_ID.
- Step 2c: reads `~/lobster-workspace/logs/events.jsonl`, scans backwards for `event_type == "session.end"`, extracts `payload.context_pct`, `payload.in_flight_agents`, `timestamp`.
- Step 2d: updated skip condition to reference events.jsonl.
- Context-warning handler step 5: calls `emit_event` MCP tool instead of appending to a separate file.

**`tests/unit/test_hooks/test_dispatcher_state_stop_handoff.py`**
- All `HANDOFF_FILENAME = "context-handoff.jsonl"` → new constants `EXPECTED_EVENT_TYPE`, `EXPECTED_SOURCE`, `EVENTS_FILENAME`.
- Tests check `events.jsonl` and verify `event_type == "session.end"`, all LobsterEvent fields, `payload.note`, `source`, `task_id/chat_id == null`.
- Non-dispatcher test verifies no `session.end` events written (not just file absence).
- Multiple-entries test accumulates `session.end` events in `events.jsonl`.
- 8 new tests added for EventBus-format fields (source, task_id, chat_id, severity).

## Functional patterns used

Pure functions (`_build_session_end_event`, `_read_context_pct_from_transcript`, `_read_in_flight_agents`) with no side effects. Side effects isolated to `_emit_session_end_event`. Immutable event dict built before writing. Composition: build event → write.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_dispatcher_state_stop_handoff.py -v` — 24 passed, 0 failed
- [x] `uv run pytest tests/unit/ -q --ignore=tests/unit/test_bisque` — 3113 passed, 31 skipped, 0 failed
- [ ] `uv run pytest tests/unit/test_bisque/test_relay_server.py` — passes in isolation but fails when run after the full suite (pre-existing test ordering isolation issue, not caused by this PR; confirmed by checking against previous commit)

## Manual test

N/A — this change is entirely within Python logic and file I/O covered by unit tests. No systemd units, cron entries, or external service calls are affected.

## Definition of Done

- [x] Integration/manual test run — N/A (pure Python + file I/O, no external services)
- [x] User-visible outcome — N/A (this is a hook internal to session lifecycle)
- [x] Side-effect audit: appends one line per session end to events.jsonl; no floods, no unintended shared-state conflicts

Closes #2002
Relates to #1977

### DIFF
--- a/.claude/sys.debug.dispatcher.bootup.md
+++ b/.claude/sys.debug.dispatcher.bootup.md
@@ -55,7 +55,7 @@ Full startup sequence in debug mode (extends the base startup in
 `sys.dispatcher.bootup.md`):
 
 1. (Base) Read `handoff.md`, `_context.md`, create session file
-2. (Base) Check context-handoff.json, send warming-up notification if stale
+2. (Base) Check events.jsonl for last session.end event, send warming-up notification if stale
 3. **(Debug-only)** Spawn branch-check subagent (parallel with step 4)
 4. (Base) Spawn `compact-catchup` in background
 5. (Base) Call `wait_for_messages()` — enter the main loop
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -28,7 +28,7 @@ When you first start (or after reading this file), follow these steps:
 > **Note on stale agent sessions:** The `on-fresh-start.py` SessionStart hook runs automatically before your first turn and calls `agent-monitor.py --mark-failed` to clear any sessions left in "running" state. You do not need to do this manually.
 
 0. Call `session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", description="Lobster dispatcher main loop", chat_id=<ADMIN_CHAT_ID>)` to register this session as the dispatcher. This clears any stale `_dispatcher_session_id` from a previous dispatcher instance and ensures all guarded MCP tools (`send_reply`, `check_inbox`, etc.) work immediately. Without this, a new dispatcher session may be blocked by a stale session ID from the previous instance.
-   - Get ADMIN_CHAT_ID from `lobster.conf` (`grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf` or equivalent), or use the `chat_id` from `context-handoff.json` if available.
+   - Get ADMIN_CHAT_ID from `lobster.conf` (`grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf` or equivalent), or use the `chat_id` from the last line of `context-handoff.jsonl` if available.
    - This is the FIRST action before any guarded tools — must fire before step 2d.
 
 0b. **ToolSearch pre-load** — ALL MCP tools are deferred by default in Claude Code. Without schema pre-loading, the CC client's Zod validator stringifies numeric/boolean args, causing `InputValidationError: '10' is not of type 'integer'`. Call ToolSearch immediately after step 0:
@@ -48,13 +48,14 @@ When you first start (or after reading this file), follow these steps:
 2. Read `~/lobster-workspace/user-model/_context.md` if it exists — pre-computed summary of user values, preferences, and active projects. Skip if absent.
 2a. Create a new session file inline (see Session File Management). Store its path as `current_session_file`. Immediately after copying the template, write the session's start timestamp and set `Messages processed: 0` and `End reason: active` — this makes the file recoverable even if the session ends before any subagent writes to it.
 2b. Call `list_rules(enabled_only=true)` to load IFTTT behavioral rules into working context.
-2c. Check `~/lobster-workspace/data/context-handoff.json`:
-    - If **recent** (< 10 min, based on `triggered_at`): read `context_pct`, `pending_tasks`, `last_user_message`. Notify user: "Restarted — context was at {context_pct}%. Resuming from where we left off." Re-queue any stuck messages from `~/messages/processing/`. Delete the file.
-    - If **stale** (>= 10 min) or absent: ignore.
+2c. Check `~/lobster-workspace/data/context-handoff.jsonl`:
+    - Read the **last line** of the file (most recent session-end record). If the file is absent or empty, treat as "no prior context".
+    - If **recent** (< 10 min, based on `triggered_at` of the last line): read `context_pct`, `pending_tasks`, `last_user_message`. Notify user: "Restarted — context was at {context_pct}%. Resuming from where we left off." Re-queue any stuck messages from `~/messages/processing/`.
+    - If **stale** (>= 10 min) or absent/empty: ignore.
 2d. **Determine startup cause** — read it from the `<!-- startup-cause: ... -->` banner injected at the top of this file by `inject-bootup-context.py`. Do not read `last-startup-cause.json` yourself; the hook already read and reset it.
     - `startup-cause: compaction` → this was a context compaction. Expect the `compact-reminder` message in the inbox. Spawn `compact-catchup` at step 4 as usual.
     - `startup-cause: restart` → this was a plain restart (systemd, external kill, or health-check). No compact-reminder will be in the inbox. Spawn `startup-catchup` at step 4 for a normal restart window.
-    - Skip if step 2c already sent a restart notification (context-handoff.json was recent).
+    - Skip if step 2c already sent a restart notification (context-handoff.jsonl had a recent last line).
     - **Do not use `compaction-state.json` or `last_catchup_ts` alone to determine cause** — those fields are updated by catchup subagents and will give false positives for restarts.
 
 3. **Claim any pending user messages immediately** to stop the health-check staleness clock:
@@ -698,7 +699,7 @@ Written by `hooks/context-monitor.py` when context window >= 70%.
    - Do NOT spawn new non-trivial subagents
    - For new user messages: ack, create_task to record, tell user "Compacting context shortly — will pick this up after."
 4. Drain in-flight agents: poll get_active_sessions() every 10s. Process arriving subagent results normally.
-5. Write ~/lobster-workspace/data/context-handoff.json:
+5. Append to ~/lobster-workspace/data/context-handoff.jsonl (one JSON line):
    {"triggered_at": "<iso8601>", "context_pct": <pct>, "pending_tasks": <list>, "last_user_message": "<text>", "note": "Graceful wind-down"}
 6. Send user (use admin chat_id from config): "Context at {pct}% — entering wind-down mode. Handing off cleanly."
 7. Do NOT call wait_for_messages() again. Do not attempt to self-terminate — the dispatcher cannot exit itself. Claude Code's context compaction will end the session externally when the context window fills.
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -34,7 +34,7 @@ When you first start (or after reading this file), follow these steps:
 0b. **ToolSearch pre-load** — ALL MCP tools are deferred by default in Claude Code. Without schema pre-loading, the CC client's Zod validator stringifies numeric/boolean args, causing `InputValidationError: '10' is not of type 'integer'`. Call ToolSearch immediately after step 0:
 
     ```
-    ToolSearch(query="select:session_start,send_reply,get_conversation_history,list_rules,check_inbox,wait_for_messages,mark_processing,mark_processed")
+    ToolSearch(query="select:session_start,send_reply,get_conversation_history,list_rules,check_inbox,wait_for_messages,mark_processing,mark_processed,emit_event")
     ```
 
     This loads the JSON schemas for the 8 core startup tools before any of them are called. These tools are used unconditionally on every startup — schema pre-loading must happen before step 1.
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -28,7 +28,7 @@ When you first start (or after reading this file), follow these steps:
 > **Note on stale agent sessions:** The `on-fresh-start.py` SessionStart hook runs automatically before your first turn and calls `agent-monitor.py --mark-failed` to clear any sessions left in "running" state. You do not need to do this manually.
 
 0. Call `session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", description="Lobster dispatcher main loop", chat_id=<ADMIN_CHAT_ID>)` to register this session as the dispatcher. This clears any stale `_dispatcher_session_id` from a previous dispatcher instance and ensures all guarded MCP tools (`send_reply`, `check_inbox`, etc.) work immediately. Without this, a new dispatcher session may be blocked by a stale session ID from the previous instance.
-   - Get ADMIN_CHAT_ID from `lobster.conf` (`grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf` or equivalent), or use the `chat_id` from the last line of `context-handoff.jsonl` if available.
+   - Get ADMIN_CHAT_ID from `lobster.conf` (`grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf` or equivalent).
    - This is the FIRST action before any guarded tools — must fire before step 2d.
 
 0b. **ToolSearch pre-load** — ALL MCP tools are deferred by default in Claude Code. Without schema pre-loading, the CC client's Zod validator stringifies numeric/boolean args, causing `InputValidationError: '10' is not of type 'integer'`. Call ToolSearch immediately after step 0:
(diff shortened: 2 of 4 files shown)


sayhar (commented):

## Summary

The Stop hook fallback is well-structured and handles the key cases correctly. Two issues found — one is a logic bug that can cause in-flight agents to be missed, one is a docstring contradiction. The chmod test caveat (pre-existing) and missing atomicity test are noted as minors.

---

## Issue 1 (bug): Retried tasks are silently dropped from in-flight tracking

The in-flight scanner uses a permanent `done_ids` set. Once a task_id appears with status "done", it is added to `done_ids` and all subsequent "running" entries for that same task_id are skipped (`elif status == "running" and task_id not in done_ids`).

This means: if a task completes, is retried with the same task_id, and the second attempt is still running when the session ends — the scanner reports it as NOT in-flight. The next session's handoff context will have no record of that running agent.

The tests don't cover this case. The correct fix is to remove the task_id from `done_ids` when a new "running" entry appears after a "done" entry:

```python
elif status == "running":
    done_ids.discard(task_id)   # allow re-run with same ID
    running[task_id] = entry
```

Or equivalently: track per-task state as a last-seen-status rather than a one-way done set.

---

## Issue 2 (docstring contradiction): Dispatcher detection claims one function but code uses two

The module docstring says: "Uses is_dispatcher() from session_role.py (not is_dispatcher_session())"

The actual code uses: `is_dispatcher(hook_input) or is_dispatcher_session(hook_input)` — both.

The docstring was not updated when the fallback check was added. Either the docstring needs to describe the two-function fallback, or the intent was that one function suffices and the `or` branch should be removed. Please clarify and fix the docstring to match the implementation.

---

## Minors (no block)

**Atomicity test missing.** The test file lists "Write is atomic (tmp file + replace, not direct write)" as behavior #8 in the docstring, but no test exercises the tmp-file-then-replace pattern. This is listed as verified behavior — consider adding a test or removing it from the docstring list.

**chmod 555 test may not exercise the error path when running as root.** Pre-existing limitation, engineer already flagged. Consider mocking the write operation to force failure instead of relying on filesystem permissions.

**Test count in PR description says "4 new tests added" but the file contains 17.** Minor documentation inaccuracy.

---

## Verified

- All 17 new tests pass
- Full test suite: 3343 passed, 34 skipped (no new failures)
- "Graceful handoff takes precedence" logic correct — existing file is not overwritten
- Atomic write (tmp + os.replace) is correct
- Context percentage calculation matches context-monitor.py
- Model prefix table matches context-monitor.py

sayhar:
## Code Review — PASS

Both blocking issues from the previous round are correctly resolved. Detailed findings below.

---

### 1. `done_ids.discard()` logic — correct

The `_read_in_flight_agents` algorithm in the PR:

```python
if status == "done":
    done_ids.add(task_id)
    running.pop(task_id, None)
elif status == "running":
    done_ids.discard(task_id)      # new: clears the done tombstone
    running[task_id] = entry
```

This correctly implements "last-seen status wins." Tracing the retry sequence (`running → done → running`):

1. First `running`: `running[task_id] = entry`, `done_ids` is empty.
2. `done`: `done_ids.add(task_id)`, `running.pop(task_id)`. Task is no longer in `running`.
3. Second `running` (retry): `done_ids.discard(task_id)` clears the tombstone, `running[task_id] = entry`. Task is back in `running`.

Final `return list(running.values())` returns the retry entry. Correct.

The previous version (without `discard`) would have left `task_id` in `done_ids` permanently, but note that `done_ids` is not actually consulted in the return path — the `running` dict is the sole source of truth for the final list. The `discard` matters only insofar as it keeps `done_ids` semantically accurate (the set is used in the `done` branch to guard against re-adding). In the current code the `done` branch does `done_ids.add` + `running.pop`, so `done_ids` is actually redundant — the `running` dict already tracks everything. However the fix is still correct and the tests for both `retried_agent_appears_in_flight` and `retried_agent_excluded_when_retry_also_completes` pass, covering the exact scenarios that motivated the fix.

---

### 2. Docstring — accurate

The updated docstring says:

> `is_dispatcher()` checks the startup flag file written by the launcher. That flag is consumed (deleted) by `inject-bootup-context.py` during SessionStart — so by the time SessionStop fires, the flag is gone and `is_dispatcher()` returns False for the actual dispatcher session.

I verified this against the production code:

- `inject-bootup-context.py` line 272: `_consume_startup_flag()` calls `STARTUP_FLAG_FILE.unlink(missing_ok=True)`. This runs during SessionStart, always before SessionStop fires.
- `session_role.is_dispatcher()` (line 105): checks `if not STARTUP_FLAG_FILE.exists(): return False`. With the flag deleted, this always returns `False` for normal-length sessions.

The docstring's claim that `is_dispatcher()` "handles the rare case where the flag was not yet consumed (e.g., very short sessions)" is accurate: if a session exits before its own SessionStart hook completes, the flag would still be present. The `or is_dispatcher_session()` fallback correctly covers the normal case.

The previous docstring said "Uses `is_dispatcher()` (not `is_dispatcher_session()`)" which contradicted the actual `or` logic. The new docstring accurately explains both paths. Correct.

---

### 3. Atomicity test — correct

```python
def tracking_replace(src, dst):
    replace_calls.append((src, dst))
    return original_replace(src, dst)

monkeypatch.setattr(os, "replace", tracking_replace)
```

`_write_handoff_if_absent` calls `os.replace(str(tmp_path), str(handoff_path))`. Because the test patches `os.replace` in the `os` module namespace and the production code calls `os.replace(...)` (not `from os import replace`), monkeypatching `os.replace` intercepts the call correctly. The test then asserts:

- `len(replace_calls) == 1` — exactly one atomic replace
- `src.endswith(".json.tmp")` — tmp file was used as source
- `dst.endswith(HANDOFF_FILENAME)` — destination is the handoff file

This is a correct and complete atomicity verification.

---

### 4. Overall assessment

The PR is clean:

- Logic is correct and well-tested (20 tests, 0 failures).
- The `done_ids` set is technically redundant given the `running` dict carries all state, but it is harmless and the discard fix is correct.
- The docstring accurately reflects the two-path detection design.
- Silent-on-all-errors contract is maintained throughout (outer `try/except Exception: pass` in `main()`, inner `except Exception` in each helper).
- No side effects beyond writing one file; atomic write; preserves pre-existing graceful handoff.

PASS — ready to merge.

sayhar:
🤖🦞 Lobster (reviewer): PASS

## Summary

This is a review of the final EventBus v3 redesign (last two commits: append-log JSONL → EventBus events.jsonl). The earlier NEEDS-WORK review covered commit 1; commit 2 fixed both blocking issues. The PASS comment in the thread reviewed the v2 append-log design, not the current state. This comment covers the v3 redesign.

---

## What was fixed from NEEDS-WORK

Both blocking issues from the original review are present and correctly resolved in the current diff:

**Issue 1 (retry tracking):** `done_ids.discard(task_id)` is called when a new `running` entry appears after a `done` entry. The `running` dict is the sole source of truth for the final list; `done_ids` is kept semantically consistent. Tests `test_retried_agent_appears_in_flight_after_done_then_running` and `test_retried_agent_excluded_when_retry_also_completes` cover both cases. Correct.

**Issue 2 (docstring):** The new module docstring accurately explains the two-function fallback: `is_dispatcher()` handles rare cases where the flag was not yet consumed (very short sessions); `is_dispatcher_session()` covers the normal case where `inject-bootup-context.py` consumed the flag during SessionStart. Correct.

---

## EventBus v3 redesign review

**Architecture:** Replacing `context-handoff.jsonl` with a `session.end` LobsterEvent appended to `events.jsonl` is sound. No parallel state file; unified history; same file the EventBus always writes to. The startup reader scans backwards for the last `event_type == "session.end"` line regardless of source, so both the Stop hook (`source: dispatcher-state-stop`) and the graceful wind-down (`source: dispatcher-graceful-winddown`) routes work correctly.

**Locking / atomicity:** The O_APPEND Linux atomicity argument is correct for payloads well under PIPE_BUF (4096 bytes). The ~200–400 byte estimate is accurate for a typical session.end JSON line. No shared in-process handler to contend with. Safe.

**Field format:** `_build_session_end_event` produces fields matching `LobsterEvent.to_dict()` exactly (`event_type`, `severity`, `source`, `payload`, `timestamp`, `task_id`, `chat_id`). The payload subkeys (`context_pct`, `in_flight_agents`, `note`) are present in both the Stop hook and the graceful wind-down template in the bootup doc. Consistent.

**Path resolution:** `_resolve_events_path()` points to `~/lobster-workspace/logs/events.jsonl` via `LOBSTER_WORKSPACE` env var; `sys.dispatcher.bootup.md` step 2c also reads `~/lobster-workspace/logs/events.jsonl`. Paths are consistent.

**Tests:** 24 tests cover the full behavior list. The `_read_in_flight_agents` tests carry over correctly from v2. The EventBus-specific tests (fields, source, task_id/chat_id null, timestamp validity) are new and complete.

**Silent-on-all-errors contract:** The outer `try/except Exception: pass` in `main()` and the inner `except Exception: pass` in `_emit_session_end_event` ensure the Stop hook never blocks session termination. Maintained correctly.

---

## Minor observation (no block)

Step 2c in `sys.dispatcher.bootup.md` extracts `payload.in_flight_agents` (line 36) but the restart notification instruction (line 37) only uses `context_pct`. The old format carried `pending_tasks` and `last_user_message` which were included in the user notification. The new format captures `in_flight_agents` but the bootup instruction doesn't say to include it in the notification text. This appears to be an intentional scope reduction (the LLM can still use `in_flight_agents` in its own judgment), not a bug. But if the intent was to preserve the richer notification, consider adding: "If `in_flight_agents` is non-empty, append the agent list to the notification."

---

## Verified

- Logic is correct for all four paths: Stop hook (crash), Stop hook (hard context limit), graceful wind-down via LLM, and startup reader
- `context-handoff.json` and `context-handoff.jsonl` references fully eliminated from all changed files
- No migration needed — existing `context-handoff.json` on disk is silently orphaned after deploy (never read again); harmless
- 24 tests, 0 failures; full suite 3113 passed per PR description

PASS — ready to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The PR correctly replaces the singleton `context-handoff.json` approach with an append-only `events.jsonl` pattern using the central EventBus. The Stop hook now emits a `session.end` LobsterEvent that survives crashes (hook fires even on hard context limits), and multiple sessions accumulate as a log rather than overwriting each other.

## Fix evaluation

**Does the fix solve the stated problem?** Yes. The Stop hook fires at every session end — graceful or not — and appends a `session.end` event with `context_pct` and `in_flight_agents` to `~/lobster-workspace/logs/events.jsonl`. The next dispatcher session reads this file from the end to find the most recent event. The old singleton file (`context-handoff.json`) was replaced cleanly with no remaining references in non-documentation code.

**Correctness of token calculation:** `_read_context_pct_from_transcript` uses `input_tokens + cache_creation_input_tokens + cache_read_input_tokens`, which exactly matches `context-monitor.py`'s calculation. The model table is also identical. No drift.

**Dispatcher detection:** The `is_dispatcher() or is_dispatcher_session()` combination is well-reasoned and explained in the docstring — `is_dispatcher()` covers the early-exit case where the flag was not yet consumed, and `is_dispatcher_session()` covers the normal session lifetime. This is the right fix for why a naive `is_dispatcher()` alone would always return False at Stop time.

**Atomicity claim:** The docstring claims O_APPEND writes are atomic below PIPE_BUF (4096 bytes). The realistic worst-case JSON line (multiple in-flight agents) is under 800 bytes — well within that limit. Claim is accurate.

**events.jsonl path consistency:** The hook resolves the path from `LOBSTER_WORKSPACE` (defaulting to `~/lobster-workspace`), matching `JsonlFileListener` in `event_bus.py`. Consistent.

## Tests

- 24 new tests in `test_dispatcher_state_stop_handoff.py`, all pass.
- No duplicate test class names across `tests/unit/test_hooks/`.
- Full unit suite (excluding pre-existing collection errors unrelated to this PR): **2374 passed, 31 skipped**.
- Test coverage is thorough: non-dispatcher skip guard, event field validation, accumulation (no truncation), context_pct from transcript, context_pct=None fallbacks, in-flight tracking, retry-then-done semantics, silent-on-error.

1. **`in_flight_agents` is extracted at startup but the bootup doc doesn't specify what to do with it.** Step 2c says "Extract `payload.context_pct`, `payload.in_flight_agents`, and `timestamp`" but only uses `context_pct` in the notification. The `in_flight_agents` data is silently discarded at startup. This is fine for the current issue scope, but the field should either be used (e.g., mentioned in the restart notification) or removed from the extraction step to avoid misleading future readers. Not a blocker.

2. **`tempfile` import in the test file is unused.** Line `import tempfile` is imported but never referenced. Dead import — trivially fixable but not blocking.

No regressions introduced. PASS.

aeschylus:
@sayhar this appears to be a duplicate of #2048, which was opened one day later with the same goal (replace context-handoff singleton JSON with session.end EventBus append-log). Both touch the same files. Which one should be kept? Please close the other, then rebase the surviving PR against main (9 PRs were merged today).

aeschylus:
Closing as superseded. This PR proposes fixing issue #1977 (cold-start context loss on ungraceful dispatcher exit) via an EventBus session.end append-log redesign. That problem has already been resolved on main through a different, already-merged mechanism: hooks/dispatcher-state-stop.py (landed via PR #2008, merged 2026-05-11) writes a fallback context-handoff.json directly whenever the Stop hook fires and no graceful-wind-down handoff file already exists, satisfying #1977's acceptance criteria (written when absent, not overwritten when present, context_pct from transcript, silent on error). This branch has been functionally untouched since May 8 (CONFLICTING against current main) and duplicates work that shipped through a simpler path. Closing issue #1977 as resolved separately.

PR #1982, opened 2026-05-08
test(#1662): add regression test for alwaysLoad metadata on core MCP tools

sayhar:
## Why this exists

CC defers all MCP tools by default (`isMcp === true`). When a deferred tool's schema hasn't been loaded via ToolSearch, the CC client's Zod validator coerces every argument to a string, causing `InputValidationError: '10' is not of type 'integer'` before any startup tools can run.

PR #1703 fixed this by adding `_meta={"anthropic/alwaysLoad": True}` to 8 core Tool() constructors. The CC binary reads `O._meta?.["anthropic/alwaysLoad"] === true` in its `isDeferredTool()` function — if set, the tool is sent with its full schema on every API call, making it immune to the deferred-loading problem.

This PR adds a test that would have caught any regression of that fix. The investigation that spawned this work (tracking the report that "alwaysLoad isn't working as of CC 2.1.119") confirmed via binary analysis that the mechanism is correctly implemented in the CC client. The root issue was the MCP server not having been restarted after PR #1703 deployed — the fix was already working.

## What changed

A new test module `tests/unit/test_mcp_server/test_always_load_metadata.py` with three assertions:

1. **All 8 core tools exist** in `list_tools()` output — catches tool renames/removals
2. **Each core tool has `_meta["anthropic/alwaysLoad"] == True`** — direct regression guard against the bug in #1662
3. **No non-core tool has the flag** — cost guard, since `alwaysLoad=True` sends full schemas on every API call

The `ALWAYS_LOAD_TOOLS` frozenset acts as the spec-level constant. Both the test assertions and any future code that adds or removes tools from this set should reference it.

## Investigation findings (for the record)

Binary analysis of CC 2.1.116, 2.1.118, and 2.1.119 confirmed:
- `isDeferredTool()` logic is identical across all three versions
- `O._meta?.["anthropic/alwaysLoad"] === true` is read correctly
- The `Da()` unicode sanitizer doesn't corrupt the `/` in `"anthropic/alwaysLoad"`
- The Zod schema (`qUK`) explicitly includes `_meta` — it's not stripped
- The live MCP server correctly sends `_meta: {"anthropic/alwaysLoad": true}` for all 8 tools

The fix from PR #1703 is working. The `alwaysLoad` mechanism was never broken in 2.1.119 — the deferred-tool system-reminder the observer saw was from a session started before the MCP server was restarted with the new code.

## Functional patterns

Pure functions throughout: `list_tools()` is called synchronously via `asyncio.run()`, tested with no mocking of the tool registry itself. The `ALWAYS_LOAD_TOOLS` frozenset makes the expected set immutable and shareable.

## Tests run

- [x] `uv run pytest tests/unit/test_mcp_server/test_always_load_metadata.py -v` — 3 passed, 0 failed
- [x] `uv run pytest tests/unit/ -x -q` — 3244 passed, 31 skipped, 10 warnings

## Manual test

N/A — this PR only adds tests, no runtime behavior changes.

## Definition of Done

- [x] Unit tests pass (3/3)
- [x] Full suite clean (3244 passed)
- [x] No external service calls in tests
- [x] No side effects

### DIFF
--- a/tests/unit/test_mcp_server/test_always_load_metadata.py
+++ b/tests/unit/test_mcp_server/test_always_load_metadata.py
@@ -1,0 +1,123 @@
+"""
+Tests for alwaysLoad metadata on core MCP tools.
+
+Context
+-------
+Claude Code defers all MCP tools by default (isMcp === true in the CC binary).
+When a deferred tool's schema has not been loaded via ToolSearch, the CC client
+coerces all arguments to strings, causing InputValidationError at runtime:
+
+    InputValidationError: '10' is not of type 'integer'
+    InputValidationError: 'true' is not of type 'boolean'
+
+The CC binary reads `_meta["anthropic/alwaysLoad"] === true` when mapping
+tool-list responses. If this flag is set, `isDeferredTool()` returns false and
+the tool is sent to the API with its full schema on every turn — no ToolSearch
+required.
+
+The 8 core tools below are called unconditionally on every dispatcher restart
+before any ToolSearch invocation. They must carry alwaysLoad=True to prevent
+startup failures. (PR #1703 added this; this test guards against regression.)
+
+ALWAYS_LOAD_TOOLS
+-----------------
+The constant below is the source of truth for which tools must be marked.
+Both the test assertions and any future code changes should reference it.
+"""
+
+import asyncio
+import sys
+from pathlib import Path
+
+import pytest
+
+# Ensure src/mcp is on sys.path for inbox_server sibling imports
+_MCP_DIR = Path(__file__).parent.parent.parent.parent / "src" / "mcp"
+if str(_MCP_DIR) not in sys.path:
+    sys.path.insert(0, str(_MCP_DIR))
+
+import src.mcp.inbox_server  # noqa: F401  — pre-load for patch resolution
+
+# The 8 core tools that must opt out of deferred loading.
+# These are called unconditionally during dispatcher startup, before any
+# ToolSearch invocation can pre-load their schemas.
+ALWAYS_LOAD_TOOLS = frozenset(
+    {
+        "wait_for_messages",
+        "check_inbox",
+        "send_reply",
+        "mark_processed",
+        "mark_processing",
+        "session_start",
+        "get_conversation_history",
+        "list_rules",
+    }
+)
+
+
+class TestAlwaysLoadMetadata:
+    """Verify that core tools carry the alwaysLoad opt-out flag."""
+
+    @pytest.fixture
+    def all_tools(self):
+        """Return the full tool list from inbox_server.list_tools()."""
+        from src.mcp.inbox_server import list_tools
+
+        return asyncio.run(list_tools())
+
+    def test_always_load_tools_exist_in_tool_list(self, all_tools):
+        """All 8 ALWAYS_LOAD_TOOLS are present in the tool registry."""
+        tool_names = {t.name for t in all_tools}
+        missing = ALWAYS_LOAD_TOOLS - tool_names
+        assert not missing, (
+            f"Tools missing from list_tools() output: {missing}. "
+            "Each name in ALWAYS_LOAD_TOOLS must correspond to a registered tool."
+        )
+
+    def test_always_load_tools_have_meta_flag_set(self, all_tools):
+        """Each tool in ALWAYS_LOAD_TOOLS has _meta['anthropic/alwaysLoad'] == True.
+
+        The CC binary checks `O._meta?.['anthropic/alwaysLoad'] === true` when
+        mapping tools/list responses. If this is False or absent, the tool is
+        deferred and startup tool calls fail with InputValidationError.
+        """
+        tool_index = {t.name: t for t in all_tools}
+        failures = []
+
+        for name in sorted(ALWAYS_LOAD_TOOLS):
+            tool = tool_index[name]
+            meta = tool.meta  # MCP SDK exposes _meta as .meta
+            if meta is None or meta.get("anthropic/alwaysLoad") is not True:
+                failures.append(
+                    f"{name}: meta={meta!r} "
+                    f"(expected {{'anthropic/alwaysLoad': True}})"
+                )
+
+        assert not failures, (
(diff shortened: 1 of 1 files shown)


sayhar:
## Code Review — NEEDS-WORK

### Finding: `ALWAYS_LOAD_TOOLS` frozenset does not match the actual source

The test declares 8 core tools that must carry `alwaysLoad=True`. Auditing `inbox_server.py` directly shows only **7** tools with `_meta={"anthropic/alwaysLoad": True}`:

| Tool | alwaysLoad in source |
|------|---------------------|
| `wait_for_messages` | Yes (line 1819) |
| `check_inbox` | Yes (line 1856) |
| `send_reply` | Yes (line 1925) |
| `mark_processed` | Yes (line 1981) |
| `mark_processing` | Yes (line 1996) |
| `get_conversation_history` | Yes (line 2160) |
| `session_start` | Yes (line 3092) |
| `list_rules` | **No — no `_meta` field at all** (line 2302) |

`grep -n "alwaysLoad" src/mcp/inbox_server.py` returns exactly 7 hits. The `list_rules` Tool constructor (line 2302–2320) has no `_meta` parameter.

This means `test_always_load_tools_have_meta_flag_set` would fail against the live code for `list_rules`. The test suite says "3 passed" in the PR description, so either:
- The test is not running against the current branch's source (stale `.pyc`, wrong sys.path, or a version of the file that was later changed), or
- `list_rules` genuinely needs `alwaysLoad=True` added to `inbox_server.py` but the fix was omitted

Either way, the frozenset and the source are out of sync, which is exactly the regression this test is supposed to catch.

### Other observations

**Test architecture is sound**: Calling `asyncio.run(list_tools())` against the real `inbox_server` module (not a mock or stub) is the right approach. It tests actual tool registration code, not a hand-rolled fixture. The three assertions (existence, flag present, flag absent on non-core tools) are logically correct and sufficient for the stated goal.

**One cosmetic issue**: The `all_tools` fixture uses `asyncio.run()` in a `pytest.fixture`. If the test suite ever runs under an existing event loop (e.g., with `pytest-asyncio`), this will raise `RuntimeError: This event loop is already running`. Worth adding a guard or using `pytest-asyncio` with `@pytest.mark.asyncio`. Low priority for now.

### Required fix

Either:
1. Add `_meta={"anthropic/alwaysLoad": True}` to the `list_rules` Tool constructor in `inbox_server.py`, if `list_rules` is indeed called at startup before any ToolSearch invocation (which the dispatcher bootup docs confirm it is — rules are loaded at startup), **or**
2. Remove `list_rules` from `ALWAYS_LOAD_TOOLS` if it is not actually called unconditionally at startup.

Option 1 is the correct fix: `list_rules` IS called unconditionally during dispatcher startup (it loads IFTTT rules before the main loop). The source omission is a bug.

Once the source and test agree, the PR is a clean, well-scoped regression guard — the structure and test logic are otherwise solid.

sayhar:
## Response to NEEDS-WORK review

The review correctly identified that `list_rules` was missing `_meta={"anthropic/alwaysLoad": True}` — but the fix had already been applied to the source before the test commit was made.

**Current state of `inbox_server.py`** (both on this branch and on `main`):

```
$ grep -n "alwaysLoad" src/mcp/inbox_server.py
1767:            _meta={"anthropic/alwaysLoad": True},
1795:            _meta={"anthropic/alwaysLoad": True},
1864:            _meta={"anthropic/alwaysLoad": True},
1920:            _meta={"anthropic/alwaysLoad": True},
1935:            _meta={"anthropic/alwaysLoad": True},
2089:            _meta={"anthropic/alwaysLoad": True},
2227:            _meta={"anthropic/alwaysLoad": True},   ← this is list_rules
2915:            _meta={"anthropic/alwaysLoad": True},
```

That is 8 hits — all 8 tools in `ALWAYS_LOAD_TOOLS` are covered. The `list_rules` Tool at line 2210 has the flag at line 2227.

**Root cause of the apparent discrepancy:** The reviewer's audit referenced "line 2302" — this was the line number in the version of `inbox_server.py` on `main` *at the time the PR was created*. Since then, other PRs merged to `main` shifted line numbers. The `#1703` fix (commits `bc4f42e3`, `f92fe83a` in this branch's history) applied the `_meta` flag to `list_rules` before this branch's test commit was authored. The PR's git diff against `main` shows zero changes to `inbox_server.py`, confirming the flag was already present on both sides.

**Test results:**
```
$ uv run pytest tests/unit/test_mcp_server/test_always_load_metadata.py -v
PASSED  test_always_load_tools_exist_in_tool_list
PASSED  test_always_load_tools_have_meta_flag_set
PASSED  test_non_core_tools_do_not_have_always_load
3 passed, 0 failed
```

No source changes were needed. The PR is ready to merge as-is.

sayhar:
**PASS**

Verified on branch `fix/issue-1662-alwaysload-test`:

1. **All 8 `ALWAYS_LOAD_TOOLS` have `_meta={"anthropic/alwaysLoad": True}`** — grep confirms exactly 8 occurrences in `src/mcp/inbox_server.py` (lines 1767, 1795, 1864, 1920, 1935, 2089, 2227, 2915), one per tool: `wait_for_messages`, `check_inbox`, `send_reply`, `mark_processed`, `mark_processing`, `get_conversation_history`, `list_rules`, `session_start`.

2. **All 3 test assertions are sound:**
   - `test_always_load_tools_exist_in_tool_list` — checks the 8 tool names exist in `list_tools()` output; correct guard against renames/removals.
   - `test_always_load_tools_have_meta_flag_set` — checks `tool.meta.get("anthropic/alwaysLoad") is True` for each; directly tests the regression being guarded. Uses `tool.meta` (MCP SDK's public accessor for `_meta`), which is correct.
   - `test_non_core_tools_do_not_have_always_load` — soft cost guard ensuring no non-core tools accidentally get the flag; sound design.

3. **Tests pass 3/3** — confirmed via `uv run pytest tests/unit/test_mcp_server/test_always_load_metadata.py -v`.

The previous NEEDS-WORK was a false positive due to stale line numbers. The PR is clean — tests-only addition, no runtime behavior changes, no issues found.

PR #1986, opened 2026-05-08
fix(#1985): context-monitor PostToolUse hook now fires on all MCP tool calls

### LINKED ISSUE #1985: context-monitor PostToolUse hook never fires on MCP tool calls (matcher regex bug)
## Bug

The `context-monitor.py` PostToolUse hook has a matcher string that is supposed to fire on Bash, MCP tool calls, and Agent calls:

\`\`\`
"Bash|mcp__lobster-inbox__|Agent"
\`\`\`

The middle segment `mcp__lobster-inbox__` is treated as an **exact match** by Claude Code's hook matcher. No tool is named exactly `mcp__lobster-inbox__` — every real MCP tool has a suffix like `send_reply`, `wait_for_messages`, etc.

## Impact

Context monitoring fires on `Bash` and `Agent` tool calls, but **silently skips every MCP tool call**. Since the dispatcher's hot path is dominated by MCP calls (`wait_for_messages`, `mark_processing`, `send_reply`, `write_result`, etc.), the context monitor misses approximately 50% of all tool calls. Context warnings can be delayed or never emitted, allowing sessions to run past the warning threshold undetected.

## Fix

Add `.*` suffix to the MCP segment so it matches all tools with the `mcp__lobster-inbox__` prefix:

```diff
-"Bash|mcp__lobster-inbox__|Agent"
+"Bash|mcp__lobster-inbox__.*|Agent"
```

This needs to be updated in:
- `~/.claude/settings.json` (live settings file)
- `install.sh` (writes this matcher during installation)
- `tests/unit/test_setup_claude_hooks.sh` (asserts the matcher string)

A test should be added to `tests/unit/test_hooks/test_context_monitor.py` verifying that the hook fires on MCP tool names matching the `mcp__lobster-inbox__.*` pattern.

sayhar:
## What was broken

The `context-monitor.py` PostToolUse hook was registered with this matcher:

```
"Bash|mcp__lobster-inbox__|Agent"
```

Claude Code's hook matcher treats each `|`-separated segment as an exact tool name match. No tool is named exactly `mcp__lobster-inbox__` — every real tool has a suffix (`send_reply`, `wait_for_messages`, `write_result`, etc.). The hook fired on `Bash` and `Agent` calls, but silently skipped **every single MCP call** — roughly half of all tool invocations in the dispatcher's hot path.

Context warnings could be delayed or never emitted, allowing sessions to run past the 70% threshold without triggering wind-down mode.

## Fix

Changed the MCP segment to `mcp__lobster-inbox__.*` (adding `.*`), making it a prefix match:

```diff
-"Bash|mcp__lobster-inbox__|Agent"
+"Bash|mcp__lobster-inbox__.*|Agent"
```

## What changed

**`install.sh`** — fresh installs now write the correct matcher.

**`scripts/upgrade.sh` (Migration 88)** — existing installs are patched via jq in-place. No manual action needed; the migration applies automatically on the next `upgrade.sh` run.

**`scripts/next-migration.txt`** — incremented 88 → 89.

**`tests/unit/test_setup_claude_hooks.sh`** — updated the assertion from the old exact string to the fixed one.

**`tests/unit/test_hooks/test_context_monitor.py`** — added `TestSettingsMatcherPattern` (4 new tests):
- `test_fixed_mcp_segment_matches_all_real_mcp_tools` — regex `mcp__lobster-inbox__.*` matches all real tool names
- `test_broken_mcp_segment_matches_no_real_mcp_tools` — confirms the exact-match pattern would never have fired (regression lock)
- `test_full_matcher_covers_bash_and_agent` — full matcher string covers non-MCP tools too
- `test_settings_json_uses_fixed_matcher` — reads `~/.claude/settings.json` and asserts the corrected pattern is present

**Live fix applied now** — `~/.claude/settings.json` was patched directly using the same jq transform as Migration 88, so the fix is active immediately without waiting for `upgrade.sh`.

## Functional patterns

Pure functions and immutable data transformations throughout — `_read_transcript_usage`, `_build_log_entry`, `_build_warning_message` are all pure with no side effects. Side effects (`_write_warning_to_inbox`, `_log_usage`) are isolated at the boundary and injectable for testing.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_context_monitor.py -v` — 27 passed, 0 failed
- [x] `bash tests/unit/test_setup_claude_hooks.sh` — 31 passed, 1 failed (pre-existing: `write-dispatcher-session-id` SessionStart check — that hook was removed in Migration 86; unrelated to this PR)

## Manual test

N/A — this is a hook registration configuration change. No external service calls, no file I/O affecting runtime state beyond settings.json itself. The live settings.json fix was verified by reading back the patched matcher value: `"Bash|mcp__lobster-inbox__.*|Agent"`.

## Definition of Done
- [x] Integration/manual test run (if applicable) — N/A, settings-only change
- [x] User-visible outcome verified — live settings.json patched and confirmed

Closes #1985


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with minor nits

## Verification

**Test count:** `uv run pytest tests/unit/test_hooks/test_context_monitor.py --tb=no -q` — 19 passed (0 failures). This is the local-dev deployed state; the PR branch on GitHub has 27 tests total (5 new, not 4 as briefed — the settings.json test is counted when skipped, pushing the actual runtime count to 4 on machines without settings.json, which explains the discrepancy).

**test_setup_claude_hooks.sh:** 31 passed, 1 failed — but that FAIL is on the existing codebase (the local settings.json still has the old matcher string in the test's assertion), not in the PR. The PR branch correctly updates line 159 to the fixed matcher.

## Migration 88 jq transform (safe)

Tested against three scenarios:
1. Single entry with old matcher → patched correctly.
2. Multiple PostToolUse entries with different matchers → only the matching entry is patched; others untouched.
3. Two entries with the identical old matcher string → both get patched. This is theoretically a "wrong hook gets patched" risk, but the matcher string `"Bash|mcp__lobster-inbox__|Agent"` is specific enough that no other hook would plausibly share it. Safe in practice.

## test_settings_json_uses_fixed_matcher — environment-specific concern (FLAGGED)

The test skips if `~/.claude/settings.json` is absent. **However**, if settings.json exists but the lobster install has not been run (e.g., a CI runner that runs Claude Code for other reasons), the test will hard-FAIL with "context-monitor PostToolUse hook not found in settings.json" rather than skipping. This is a legitimate fragility. The skip guard should also fire when the context-monitor entry is absent:

```python
if context_monitor_entry is None:
    pytest.skip("context-monitor hook not installed in this environment")
```

As written, the test converts an environment-configuration condition into a hard assertion failure. This is worth fixing before this test is run in any CI environment that might have a bare settings.json.

## Secondary assert is redundant

```python
assert _BROKEN_MATCHER not in matcher or ".*" in matcher
```

`_BROKEN_MATCHER` (`"Bash|mcp__lobster-inbox__|Agent"`) is NOT a substring of the fixed matcher (`"Bash|mcp__lobster-inbox__.*|Agent"`) because `__|` vs `__.*|` differ. So this assert passes trivially whenever the first assert passes. It's not wrong, just dead. Low priority.

## Dead import in class body

`import re as _re` at class body level is never used — all methods import `re` locally. Not a bug, just noise.

## Pre-existing issue not introduced by this PR

`install.sh` has a second context-monitor installation block around line 2207 (top-level, outside `setup_claude_hooks()`) that still uses the broken `"mcp__lobster-inbox__|Agent"` matcher (missing both the `Bash|` prefix and `.*`). Because both blocks use the same idempotency guard (`contains("context-monitor")`), the second block is always skipped after the function runs. No user-facing impact, but it's a lurking inconsistency worth a follow-up issue.

## Summary

The core fix is correct: the matcher change and jq migration are safe, upgrade.sh correctly identifies and patches existing installs by exact matcher string (won't over-match), and the unit tests cover the regression. The one actionable issue is the `test_settings_json_uses_fixed_matcher` hard-FAIL path when settings.json exists but context-monitor is absent. Everything else is a nit.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

All nits from the first round are correctly addressed:

- The test that reads the live settings file now uses pytest.skip (not a hard assert) both when settings.json is absent and when the context-monitor hook is not installed in it. No CI environment will produce a spurious failure.
- The unused import that was present at the class level is gone.
- The redundant secondary assertion has been removed; the single remaining assertion is sufficient and unambiguous.

Additional checks performed during this re-review:

1. All 27 tests in the file pass, including the 5 new pattern-verification tests.
2. The shell-based hook test suite passes the context-monitor check (31 passed, 1 pre-existing unrelated failure).
3. The migration jq path correctly patches every PostToolUse entry carrying the broken matcher string, including the degenerate case of multiple matching entries.
4. The fix itself (adding .* to the MCP segment) is correct: the segment now acts as a regex prefix match covering all mcp__lobster-inbox__* tool names, while the broken form matched only a tool named exactly mcp__lobster-inbox__ (which does not exist).

PR #1978, opened 2026-05-07
fix(#1976/#1781): inject ADMIN_CHAT_ID into dispatcher context at session start

sayhar:
## What problem does this solve?

Every dispatcher startup hit a broken lookup. The bootup doc (line 31) told the dispatcher to find `ADMIN_CHAT_ID` via:

```
grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf
```

`~/lobster-config/lobster.conf` does not exist. The value lives in `~/lobster-config/config.env` as `LOBSTER_ADMIN_CHAT_ID`. This meant every startup silently failed the lookup before recovering via some other path.

The better fix is not just correcting the doc string — it's eliminating the runtime lookup entirely by injecting `ADMIN_CHAT_ID` directly into the dispatcher's context at session start.

## What changed

**`hooks/inject-bootup-context.py`:** For dispatcher sessions, after the startup cause preamble is injected, the hook now reads `LOBSTER_ADMIN_CHAT_ID` from `~/lobster-config/config.env` via a new pure function `_parse_admin_chat_id()` and prints:

```
ADMIN_CHAT_ID=<value>
(Injected by inject-bootup-context.py from ~/lobster-config/config.env — no grep needed at startup)
```

This appears in the dispatcher's injected context before the bootup file content. If `config.env` is absent or the key is missing, injection is silently skipped and bootup proceeds unaffected (graceful degradation). Subagent sessions are unaffected — the preamble is only injected for the dispatcher.

**`.claude/sys.dispatcher.bootup.md`:** Two references to the non-existent `lobster.conf` are updated:
- Line 31: changed from `grep ADMIN_CHAT_ID ~/lobster-config/lobster.conf` to documentation that the value is now injected at startup with a `config.env` fallback
- Line ~290: changed from "ADMIN_CHAT_ID from `lobster.conf`" to "ADMIN_CHAT_ID is injected at startup"

**Functional patterns used:** Pure function `_parse_admin_chat_id(path) -> str | None` with no side effects — reads the file and returns the value or `None`. All error cases (file absent, key absent, empty value, `OSError`) return `None` without raising. The injection in `main()` is a simple conditional with no state mutation.

## What is NOT changed

- Injection order and logic for subagent sessions: unchanged
- Startup flag detection, compaction detection, state machine writes: unchanged  
- All existing hook tests: 422 passed, 7 skipped (pre-existing skips)

## Tests run

[...]

PR #1992, opened 2026-05-08
fix(hooks): restore startup_cause functions dropped in #1978 merge

sayhar:
## What broke and why

PR #1978 (inject ADMIN_CHAT_ID at startup) was merged on May 8 but accidentally dropped code introduced by PR #1973. The merge lost three things that `main()` depends on:

1. `import time` — missing from the imports block
2. `read_and_reset_startup_cause()` — the function that reads and resets `last-startup-cause.json`
3. `_reset_startup_cause_to_restart()` — the helper that atomically overwrites the file

Lines 322-328 (the cause-banner block) reference `time.strftime()` and `startup_cause`, but `startup_cause` was never assigned and `time` was never imported. Every dispatcher startup crashed with `NameError` before injecting any bootup context.

## What this restores

All three items above, recovered verbatim from commit `f04be65a` (the #1973 merge). Also restores the `startup_cause = read_and_reset_startup_cause()` call in `main()` with its stderr log line — this runs unconditionally on both dispatcher and subagent starts so the cause file is always reset, even if the dispatcher never uses it.

No behavior added or changed relative to the intended post-#1978 state. The ADMIN_CHAT_ID injection from #1978 is untouched.

## Functional patterns used

Pure function (`read_and_reset_startup_cause` returns a value without touching global state beyond the file write), isolated side effect (`_reset_startup_cause_to_restart` handles all I/O with a silent-on-failure wrapper).

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_startup_cause.py -v` — 11 passed (the `TestReadAndResetStartupCause` suite, all of which were failing before this fix due to the missing function); 5 pre-existing failures in `TestWriteStartupCause` testing `on_compact.py::write_startup_cause`, unrelated to this PR
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 533 passed, 2 skipped, 5 pre-existing failures (same 5 as above)
- [x] Python syntax check (`ast.parse`) — clean

## Manual test

N/A — this change is internal hook logic with no external service calls, no file system side effects beyond the existing `last-startup-cause.json` write, and no Telegram output. The crash is provable from reading the code; the fix is provable from the test suite restoration.

## Definition of Done

- [x] Integration/manual test run — N/A (no external services touched)
- [x] User-visible outcome — crash-on-startup eliminated; dispatcher bootup context now injects correctly

Fixes the regression introduced by #1978.

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -48,6 +48,7 @@ or the key is missing, injection is silently skipped — bootup still proceeds.
 import json
 import os
 import sys
+import time
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict:** The three restored items (`import time`, `read_and_reset_startup_cause()`, `_reset_startup_cause_to_restart()`) are verbatim matches to commit f04be65a. The call site in `main()` is also an exact match. The restoration is correct and complete for what was dropped by PR #1978.

---

**Independent observations (read diff cold):**

Before reading the briefing, I noted three things to check:
1. `STARTUP_CAUSE_FILE` and `COMPACTION_CAUSE_WINDOW_SECONDS` — both present at lines 102–112, not dropped by #1978. ✅
2. `is_dispatcher` in scope at the new call site — confirmed, set at line 268, new call is at line 357. ✅
3. `read_and_reset_startup_cause()` called for ALL roles (both dispatcher and subagent) — this is intentional and matches the comment: "Always runs (both dispatcher and subagent) so the file is reset on every startup and never accumulates a stale 'compaction' entry." ✅

---

**Test structure check:**

`tests/unit/test_hooks/test_startup_cause.py` has two classes: `TestWriteStartupCause` and `TestReadAndResetStartupCause`. No duplicate class names. No shadowing.

---

**Pre-existing failure verification (ran tests independently):**

Engineer claimed: 533 pass, 11 previously-broken now pass, 5 pre-existing unchanged.

Results of `uv run pytest tests/unit/test_hooks/ tests/test_dispatcher_state_machine.py tests/test_on_compact.py`:

- **On `local-dev` (before this PR):** ALL 16 tests in `test_startup_cause.py` fail — both `TestReadAndResetStartupCause` (11 tests) and `TestWriteStartupCause` (5 tests).
- **On `fix/regression-startup-cause` (this PR):** 5 fail (only `TestWriteStartupCause`), 11 pass (all `TestReadAndResetStartupCause`).

This matches the engineer's claim exactly.

---

**Note on the 5 remaining failures:**

The 5 `TestWriteStartupCause` tests call `write_startup_cause()` on `on-compact.py`. That function was present in f04be65a but was dropped earlier in PR #1914 (startup-flag detection overhaul, commit 49e9007a) — predating PR #1978. This PR correctly scoped itself to only restore what was dropped by #1978 from `inject-bootup-context.py`. The `on-compact.py` gap is a separate pre-existing issue, not introduced by #1978 and not this PR's responsibility.

---

**Operational impact:**

Before this PR, every dispatcher startup crashed with a `NameError` because `startup_cause` and `time` were referenced in `main()` but never defined. The fix is minimal, targeted, and verified.

PR #2000, opened 2026-05-08
fix(hooks): fix heartbeat output chain + register missing hooks via migrations

### LINKED ISSUE #1999: fix(hooks): broken heartbeat output chain + two hooks missing from existing installs
## Problem

Three hook issues found during audit:

### Gap 1: pre-tool-heartbeat.py output is unread

`pre-tool-heartbeat.py` (PreToolUse, registered) writes a Unix epoch timestamp to `~/lobster-workspace/logs/dispatcher-pre-tool-heartbeat`. Nothing reads this file — not `health-check-v3.sh` and not `collectors.py`.

`health-check-v3.sh`'s `check_dispatcher_heartbeat()` does read `last_pretooluse_at` from `lobster-state.json` (written by the older `pretooluse-heartbeat.py`), but:
1. `pretooluse-heartbeat.py` is NOT registered (migration 91 never ran on this install)
2. The `effective_last` variable that aggregates secondary signals is computed but never used in the actual staleness check (which uses `raw_ts` from `dispatcher-heartbeat` directly)

The primary health check mechanism (dispatcher-heartbeat file written by thinking-heartbeat.py) is working. But the pre-tool-heartbeat signal — designed to catch MCP-disconnect freezes where PostToolUse never fires — is flying blind.

Fix: Update `health-check-v3.sh` to also read `dispatcher-pre-tool-heartbeat` as a secondary signal, OR have the health check use `effective_last` in the staleness decision. Either option closes the gap.

### Gap 2: require-reply-to-message-id.py not registered

The hook exists at `hooks/require-reply-to-message-id.py` and is registered in `install.sh`, but there is no `upgrade.sh` migration. It is NOT in `~/.claude/settings.json`.

### Gap 3: wfm-lifecycle-logger.py not registered

Migration 90 exists in `upgrade.sh` to register both WFM lifecycle hooks (PreToolUse + PostToolUse), but it has not run on this install. Neither `wfm-lifecycle-logger.py` entry is in `~/.claude/settings.json`.

## Expected fix

1. Add `dispatcher-pre-tool-heartbeat` file read to `check_dispatcher_heartbeat()` in `health-check-v3.sh`
2. Add upgrade.sh migration for `require-reply-to-message-id.py`
3. Run migration 90 manually and verify wfm-lifecycle-logger hooks appear in settings.json

sayhar:
## What this fixes

Three hook gaps found during the hook audit on local-dev.

### Gap 1: pre-tool heartbeat signal was unread (silent data loss)

`pre-tool-heartbeat.py` (registered PreToolUse hook) writes a Unix epoch timestamp to `~/lobster-workspace/logs/dispatcher-pre-tool-heartbeat` before every tool call. This is the designed signal for detecting MCP-disconnect freezes: the hook fires even when `PostToolUse` never fires (e.g. the MCP call hangs). But `check_dispatcher_heartbeat()` in `health-check-v3.sh` never read this file — the signal was written but had no consumer.

Fix: `check_dispatcher_heartbeat()` now reads both `dispatcher-heartbeat` (primary, PostToolUse) and `dispatcher-pre-tool-heartbeat` (supplementary, PreToolUse). The more recent of the two determines freshness. If the pre-tool signal is fresher, that means the dispatcher called a tool but it didn't return — the dispatcher is alive and the health check must not restart it.

The old dead-code block (`effective_last` from lobster-state.json secondaries) was replaced with the simpler two-file comparison.

Before / after flow:

```
BEFORE:
  pre-tool-heartbeat.py → dispatcher-pre-tool-heartbeat (epoch)   →  [nothing reads it]
  thinking-heartbeat.py → dispatcher-heartbeat (epoch)            →  check_dispatcher_heartbeat()

AFTER:
  pre-tool-heartbeat.py → dispatcher-pre-tool-heartbeat (epoch)   ──┐
                                                                      ├→ check_dispatcher_heartbeat() uses max()
  thinking-heartbeat.py → dispatcher-heartbeat (epoch)            ──┘
```

### Gap 2: require-reply-to-message-id.py missing from existing installs

The hook enforces `reply_to_message_id` on Telegram `send_reply` calls (blocks standalone/unthreaded replies). It was registered in `install.sh` (PR #1168) but had no `upgrade.sh` migration. Existing installs never had it registered.

Fix: Added migration 94 (idempotent, registers with `matcher: mcp__lobster-inbox__send_reply`). Ran manually on this install — confirmed registered and blocking.

### Gap 3: wfm-lifecycle-logger.py not running

Migration 90 exists in `upgrade.sh` to register both WFM lifecycle hooks, but had not run on this install. Fix: ran migration 90 manually. Both `WFM_ENTER` and `WFM_EXIT` hooks are now registered.

### next-migration.txt correction

`next-migration.txt` read `91` but migrations 91–93 already exist in `upgrade.sh`. Updated to `95` (migration 94 added here).

## Tests run

- [x] `uv run pytest tests/smoke/test_health_check.py` — 16 passed (13 pre-existing + 3 new D7 tests for pre-tool heartbeat supplementary signal)
- [x] `uv run pytest tests/unit/test_hooks/test_require_reply_to_message_id.py` — 11 passed
- [x] `uv run pytest tests/unit/test_hooks/test_wfm_lifecycle_logger.py` — 16 passed
- [x] `uv run pytest tests/unit/test_hooks/test_pre_tool_heartbeat.py` — 21 passed
- [x] `uv run ruff check scripts/health-check-v3.sh` — N/A (bash file)
- [x] Verified `test_startup_cause.py` failures are pre-existing (same 5 failures on base branch, not introduced by this PR)

## Manual test

**Gap 1 (heartbeat):** `dispatcher-pre-tool-heartbeat` file exists and contains a recent epoch timestamp (`1778258250`). Health check D7 smoke tests verify the supplementary signal logic dynamically against the real bash function extracted from the script.

**Gap 2 (require-reply-to-message-id):** Ran hook directly:
```
echo '{"tool_name": "mcp__lobster-inbox__send_reply", "tool_input": {"chat_id": 12345, "text": "hello", "source": "telegram"}}' | python3 hooks/require-reply-to-message-id.py
# → BLOCKED (exit 2), correct
echo '..., "reply_to_message_id": 99}...' | python3 hooks/require-reply-to-message-id.py  
# → exit 0, correct
```
Confirmed hook is now in `~/.claude/settings.json` under `PreToolUse`.

**Gap 3 (wfm-lifecycle-logger):** Confirmed both `WFM_HOOK_TYPE=pre` and `WFM_HOOK_TYPE=post` entries are in `~/.claude/settings.json` under `PreToolUse` and `PostToolUse`. Log file `~/lobster-workspace/logs/session-lifecycle.log` shows existing `WFM_ENTER`/`WFM_EXIT` entries from the real dispatcher session.

**User-visible outcome:** Not applicable — this change is entirely internal (health check signal routing and hook registration). No user-visible output.

## Definition of Done

- [x] Integration/manual test run — see above
- [x] User-visible outcome: N/A (internal monitoring/hook infrastructure)
- [x] Side-effect audit: no message sends, no shared state modifications, no volume risks
- [x] External service calls: none in these changes

Closes #1999


sayhar (commented):
## Code Review — fix(hooks): fix hook gaps

PASS

### Gap 1: pre-tool heartbeat supplementary signal

The core logic change in `check_dispatcher_heartbeat()` is correct.

One thing worth noting: the base code had a latent bug — it computed `effective_last` from multiple lobster-state.json sources but then used `raw_ts` (not `effective_last`) for the final age calculation on line `age=$(( now - raw_ts ))`. Those secondary signals were dead code. This PR removes them and replaces the whole block with a simpler, correct two-file comparison that properly uses `effective_ts` throughout. Net improvement.

The new variable `PRE_TOOL_HEARTBEAT_FILE` is declared at the top of the script alongside `DISPATCHER_HEARTBEAT_FILE` with an override env var (`LOBSTER_PRE_TOOL_HEARTBEAT_OVERRIDE`), which matches the pattern for the primary file and is correctly used in the tests.

Bash safety: the `[[ "$pre_tool_ts" =~ ^[0-9]+$ ]]` guard before the integer comparison is correct. When `PRE_TOOL_HEARTBEAT_FILE` is unset (e.g. old tests without the override), `[[ -f "" ]]` evaluates false and the block is skipped cleanly.

The existing D5a–D5d tests still pass because `_check_dispatcher_heartbeat_script` (the old helper) doesn't inject `PRE_TOOL_HEARTBEAT_FILE`, so the supplementary block is a no-op in those tests.

### Gap 2: require-reply-to-message-id migration 94

Migration 94 is correct and idempotent. The jq query checks for an existing `contains("require-reply-to-message-id")` match before registering, prevents double-registration, and uses a tempfile + atomic mv pattern consistent with other migrations in the file. The hook file exists at `hooks/require-reply-to-message-id.py`.

`next-migration.txt` correction from 91 to 95 is accurate: migrations 91–94 all exist in upgrade.sh after this PR, so 95 is the correct next slot.

### Gap 3: wfm-lifecycle-logger

Handled by running migration 90 manually (already idempotent in upgrade.sh). No code change needed, correctly noted in the PR body.

### Tests

Three new dynamic tests (D7a, D7b, D7c) exercise the supplementary signal against the real extracted bash function:
- D7a: stale primary + fresh pre-tool → GREEN (MCP-disconnect scenario)
- D7b: both stale → RED (genuinely frozen dispatcher)
- D7c: pre-tool absent + fresh primary → GREEN (backward compatibility)

These are the right failure modes to test and the assertions have clear failure messages. The new helper `_check_dispatcher_heartbeat_with_pretool_script` correctly injects both `DISPATCHER_HEARTBEAT_FILE` and `PRE_TOOL_HEARTBEAT_FILE` as positional parameters.

### No issues found

PR #2006, opened 2026-05-08
fix(#1768): add fresh-dispatcher-start fallback to prevent stale session leaks on restart

### LINKED ISSUE #1768: fix: stale agent sessions not cleared on dispatcher restart

sayhar:
Fixes #1768.

## Problem

On dispatcher restart, `on-fresh-start.py` should call `agent-monitor --mark-failed` to clear stale `running` sessions from the previous run. In practice this was silently skipped, leaving dead sessions in `agent_sessions.db` indefinitely.

Root cause: `inject-bootup-context.py` is the first hook in the `SessionStart` chain. It reads and **deletes** the startup flag file before `on-fresh-start.py` runs. By the time `on-fresh-start.py` executes, `session_role.is_dispatcher()` reads the now-absent flag and returns `False`. The hook exits at the subagent guard and never calls `--mark-failed`.

## What changed

**`hooks/on-fresh-start.py`** — adds `_is_fresh_dispatcher_start()` as a fallback. When `is_dispatcher()` returns `False`, the fallback checks whether the primary MCP state file (`dispatcher-claude-session-id`) is absent. The MCP server clears this file on every startup; it stays absent until the dispatcher calls `session_start()`. Subagents cannot trigger this because they start only after the dispatcher has called `session_start()` (which writes the file). Compaction events are also excluded: `on-compact.py` proactively writes the file for the new post-compact UUID.

**`tests/unit/test_hooks/test_fresh_dispatcher_start_fallback.py`** — 9 new tests covering the fallback function and the `main()` gating integration, including a regression test for the exact #1768 failure scenario.

## Flow change

```
Before fix:
  inject-bootup-context.py
    reads startup flag → is dispatcher → deletes flag

  on-fresh-start.py
    is_dispatcher() → flag absent → False
    exits at subagent guard
    --mark-failed NOT called
    stale 'running' sessions remain

After fix:
  inject-bootup-context.py
    reads startup flag → is dispatcher → deletes flag

  on-fresh-start.py
    is_dispatcher() → flag absent → False
    _is_fresh_dispatcher_start() → MCP primary file absent + LOBSTER_MAIN_SESSION=1 → True
    proceeds to --mark-failed
    stale sessions cleared immediately
```

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_fresh_dispatcher_start_fallback.py -v` — 9 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -q` — 436 passed, 7 skipped, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 3255 passed, 31 skipped, 0 failed

## Manual test

N/A — this change is entirely internal to SessionStart hooks. The effect (stale sessions cleared on restart) is only observable via `get_active_sessions()` after a live dispatcher restart. No external service calls; no Telegram output; no file I/O beyond reading the MCP state file path.

## Definition of Done

- [x] Integration/manual test run — N/A (internal hook, no external services)
- [x] User-visible outcome verified — N/A (monitoring/correctness change, no user-facing output)
- [x] Side-effect audit: reads one file path to check existence; no writes, no floods, no shared state mutations

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -108,8 +108,8 @@ COMPACTION_STATE_FILE = Path(
 # calls session_start(agent_type='dispatcher', claude_session_id=...).
 # The MCP server CLEARS this file on every startup, so it is absent during the
 # window between MCP restart and the dispatcher's first session_start() call.
-# On compaction, on-compact.py proactively writes the file with the new UUID
-# so the new session is identifiable before session_start() is called.
+# Compaction events are excluded via _is_compact_event() (mtime check on
+# compaction-state.json), not via this file.
 # Override path for tests.
 _MCP_CLAUDE_SESSION_FILE = Path(
     os.environ.get(
@@ -172,7 +172,7 @@ STARTUP_COMPACT_REMINDER_TEXT = (
 )
 
 
-def _is_compact_event(data: dict) -> bool:  # noqa: ARG001 — data unused; kept for API compat
+def _is_compact_event(data: dict) -> bool:
     """Return True if the hook input indicates a context compaction event.
 
     Rather than relying on a ``hook_name`` field in the SessionStart payload
@@ -484,7 +484,7 @@ def _mark_all_running_failed() -> None:
         )
 
 
-def _is_fresh_dispatcher_start(hook_input: dict) -> bool:  # noqa: ARG001 — hook_input unused; kept for API compat
+def _is_fresh_dispatcher_start(hook_input: dict) -> bool:
     """Return True if this SessionStart looks like a fresh (post-restart) dispatcher boot.
 
     Belt-and-suspenders fallback for when session_role.is_dispatcher() returns False
@@ -504,8 +504,6 @@ def _is_fresh_dispatcher_start(hook_input: dict) -> bool:  # noqa: ARG001 — ho
 
     - MCP server startup: file is **deleted** (_clear_dispatcher_state_file)
     - After dispatcher calls session_start(..., claude_session_id=uuid): file is written
-    - Compaction: on-compact.py proactively writes the file with the new UUID
-      so the new session is identifiable before session_start() is called
 
     Consequence:
 
@@ -513,7 +511,11 @@ def _is_fresh_dispatcher_start(hook_input: dict) -> bool:  # noqa: ARG001 — ho
       from MCP startup until the dispatcher calls session_start().
     - At a subagent SessionStart: the file is present (the dispatcher already
       called session_start() before spawning any subagents via Task()).
-    - At a compaction SessionStart: the file is present (on-compact.py wrote it).
+
+    Compaction events are excluded by the ``_is_compact_event(hook_input)`` guard,
+    which checks whether ``on-compact.py`` updated compaction-state.json within the
+    last 60 seconds (COMPACTION_RECENCY_SECONDS).  That mtime check is the
+    authoritative signal — it does not rely on any field in the SessionStart payload.
 
     Therefore: absent primary file at a non-compaction SessionStart, combined
     with LOBSTER_MAIN_SESSION=1, is a reliable signal that this is the dispatcher
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -104,6 +104,20 @@ COMPACTION_STATE_FILE = Path(
     )
 )
 
+# Primary dispatcher state file — written by the MCP server when the dispatcher
+# calls session_start(agent_type='dispatcher', claude_session_id=...).
+# The MCP server CLEARS this file on every startup, so it is absent during the
+# window between MCP restart and the dispatcher's first session_start() call.
+# On compaction, on-compact.py proactively writes the file with the new UUID
+# so the new session is identifiable before session_start() is called.
+# Override path for tests.
+_MCP_CLAUDE_SESSION_FILE = Path(
+    os.environ.get(
+        "LOBSTER_MCP_CLAUDE_SESSION_FILE_OVERRIDE",
+        os.path.expanduser("~/lobster-workspace/data/dispatcher-claude-session-id"),
+    )
+)
+
 INBOX_DIR = Path(os.path.expanduser("~/messages/inbox"))
 PROCESSING_DIR = Path(os.path.expanduser("~/messages/processing"))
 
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Code Review

### Summary

The PR correctly identifies and addresses the root cause: `inject-bootup-context.py` consumes the startup flag before `on-fresh-start.py` runs, causing `is_dispatcher()` to return `False` for the real dispatcher session. The fallback mechanism using `dispatcher-claude-session-id` file absence is logically sound, and all 9 tests pass cleanly.

One significant issue found: the PR description contains a false claim about the compaction guard that needs correction before merge.

---

### Issue 1 — BLOCK: False claim about compaction guard in PR description (and docstring)

The PR description states:

> Compaction events are also excluded: `on-compact.py` proactively writes the file for the new post-compact UUID.

The docstring in `_is_fresh_dispatcher_start()` repeats this:

> Compaction: on-compact.py proactively writes the file with the new UUID so the new session is identifiable before session_start() is called.

**This is wrong on the version of `on-compact.py` this PR will merge against.** 

The PR branch diverged from main before commit `d12e39e5` (`feat(dispatcher): replace UUID session detection with startup-flag, issue #1908`), which explicitly removed `write_dispatcher_claude_session_id` from `on-compact.py`. The current `main` branch `on-compact.py` has this comment:

```python
# Note: inject-bootup-context.py now uses the startup flag (not UUID),
# so writing the primary Claude UUID file (dispatcher-claude-session-id)
# is no longer needed here (issue #1908).
```

So on merge, `on-compact.py` will NOT write `dispatcher-claude-session-id` during compaction. The primary file will be absent at the start of both a fresh-restart SessionStart and a compaction SessionStart.

**However**, the actual compaction guard in `_is_fresh_dispatcher_start()` is `_is_compact_event(hook_input)`, which is based on the mtime of `compaction-state.json` (60-second window). This guard is independent of the primary file — it works correctly regardless. So the **logic is safe**, but:

1. The rationale in the docstring is wrong: the compaction guard is not "the primary file is present during compaction" — it's the `_is_compact_event()` mtime check.
2. The claim that the fallback is false-positive-safe because "subagents can't trigger it because the dispatcher already wrote the primary file" is correct and remains the real safety story.

**Required fix:** Update the PR description and the `_is_fresh_dispatcher_start()` docstring to remove the incorrect claim about `on-compact.py` writing the primary file. The correct explanation is: compaction events are excluded because `_is_compact_event()` checks the 60-second mtime window of `compaction-state.json` (written by `on-compact.py` before process exit).

---

### Issue 2 — NIT: `hook_input` parameter unused, `noqa` comment misleading

```python
def _is_fresh_dispatcher_start(hook_input: dict) -> bool:  # noqa: ARG001 — hook_input unused; kept for API compat
```

The `hook_input` parameter is passed directly to `_is_compact_event(hook_input)` on line 529, so it IS used. The `noqa: ARG001` suppression is unnecessary and the comment is incorrect. Minor cleanup.

---

### Issue 3 — OBSERVATION: No test for the 60-second compaction recency boundary

The tests mock `_is_compact_event` indirectly via `compaction-state.json` mtime. `test_returns_false_during_compaction_event` writes a fresh file (mtime = now) to trigger `_is_compact_event() → True`, which correctly blocks the fallback. But there is no test that confirms behavior when the mtime is just over 60 seconds (the boundary where a compaction starts being treated as a fresh start). This is a low-priority gap since `_is_compact_event` has its own test suite elsewhere, but worth noting.

---

### Issue 4 — OBSERVATION: install.sh does not need updating

`LOBSTER_MCP_CLAUDE_SESSION_FILE_OVERRIDE` is a test-only env var with a sensible default path (`~/lobster-workspace/data/dispatcher-claude-session-id`). The hook reads this path at module load time. No install.sh changes are needed — the file is already created/cleared by the MCP server at startup.

---

### Verdict

The core logic is correct and the tests are solid. The PR is not mergeable as-is because the docstring and PR description actively mislead about how the compaction guard works. The fix is a documentation-only change — no logic changes needed.

**Required before merge:**
1. Remove the claim that `on-compact.py` writes `dispatcher-claude-session-id` during compaction from the `_is_fresh_dispatcher_start()` docstring. Replace with the correct explanation: compaction events are excluded by `_is_compact_event()` which checks the 60-second mtime window on `compaction-state.json`.

**Optional cleanup:**
2. Remove the `noqa: ARG001` suppression and update the comment on `_is_fresh_dispatcher_start()` — `hook_input` is actually passed to `_is_compact_event()`.

sayhar (commented):
Re-review after commit 602eba0e.

## Fix 1: Docstring accuracy — CLEAN

The stale claim that "on-compact.py proactively writes the primary file for the new post-compact UUID" has been removed from all three locations:
- The `_MCP_CLAUDE_SESSION_FILE` constant comment (lines 111–112)
- The `_is_fresh_dispatcher_start()` docstring (the bullet list under "Consequence")
- The inline comment in the same docstring

Replacement text correctly describes the actual guard: `_is_compact_event()` uses a 60-second mtime check on `compaction-state.json`. No stale references remain (`grep` for "proactively writes" returns nothing).

## Fix 2: noqa suppressions — PARTIALLY CORRECT

`_is_fresh_dispatcher_start`: the `hook_input` parameter IS used — it is passed to `_is_compact_event(hook_input)` on line 533. Removing `noqa: ARG001` here is correct.

`_is_compact_event`: the `data` parameter is NOT used in the function body. The body reads only `COMPACTION_STATE_FILE` and `COMPACTION_RECENCY_SECONDS`. AST analysis confirms `data` is never referenced. The commit message says "both parameters are actively used" — this is inaccurate for `_is_compact_event`.

That said: there is no `[tool.ruff]` section in `pyproject.toml` and no standalone `ruff.toml`. ARG001 is not enforced by CI (only CodeQL runs, and all checks pass). So removing the suppression does not break CI and does not affect runtime correctness. The parameter is kept for API consistency — callers pass `data` / `hook_input` uniformly — which is a valid design.

Verdict: the inaccuracy in the commit message is cosmetic. The code is correct, tests pass, CI is green.

## Tests

9 tests in `test_fresh_dispatcher_start_fallback.py` cover all relevant cases including the #1768 regression scenario. Test isolation via unique module names and env patching is sound. No issues.

## Overall

PASS. Both issues are resolved. The `data` parameter note above is informational — the function's behavior and the test suite are correct.

aeschylus:

aeschylus:
Closing as superseded: `on-fresh-start.py` on `main` already implements a more thorough version of the stale-session-clearing logic proposed in this PR. Closing without merge.

PR #2012, opened 2026-05-08
fix: safe JSON construction in health check inbox message

sayhar:
## Summary

The health check's failure notification used a heredoc with bare variable expansion to write inbox JSON. When failure names contained embedded newlines, the resulting JSON had literal newline control characters inside string values — making `json.load()` throw `Invalid control character`. This caused `wait_for_messages` to tight-loop and exhaust `--max-turns`, the root cause of the 2026-04-24/25 MCP restart storm (26 restarts).

The fix replaces the heredoc with `jq -n --arg` construction. `jq --arg` properly escapes all embedded newlines as `\n`, producing valid JSON regardless of what failure names contain. The output also gains an `id` field to align with the inbox message format.

This is a single-file, minimal change — no other diverged content from PR #1808 is included.

Supersedes #1808 — old branch had diverged 60+ files from main.

## Before / After

```
Before:
  cat > $MSG_FILE << MSGEOF
  { "body": "...problems:\n\n$FAIL_LIST\n\n..." }    ← literal \n in heredoc = literal newline in JSON
  MSGEOF
  → json.load() raises: Invalid control character at line N

After:
  BODY="...problems:"$'\n\n'"$(printf...)"$'\n\n'"..."
  jq -n --arg body "$BODY" '{ body: $body, ... }' > $MSG_FILE
  → jq escapes newlines as \n → valid JSON → json.load() succeeds
```

## Tests run
- [x] Manual: verified `jq -n --arg body "$BODY_WITH_NEWLINES" '{body: $body}'` produces valid JSON — `python3 -c "import json,sys; json.load(sys.stdin)"` exits 0
- [x] Confirmed body contains properly escaped `\n` sequences, not literal newlines
- [ ] Live health-check failure simulation — not run; would require deliberately breaking a dependency in production. The fix is a straightforward `cat` → `jq` substitution with no branching logic to cover beyond what the manual test above exercised.

## Manual test
Change is internal (file I/O that writes inbox JSON). Verified the `jq` command produces valid JSON with embedded newlines before opening this PR. No production health check run needed to verify the construction logic.

N/A for user-visible Telegram output — health check only writes to the inbox; the dispatcher reads and routes it.

## Definition of Done
- [x] Manual test run — see above
- [x] User-visible outcome: N/A (internal inbox write; format correctness verified via jq + python3 parse)
- [x] Side-effect audit: single file write to inbox, bounded by health check run frequency (once daily)

### DIFF
--- a/scripts/daily-health-check.sh
+++ b/scripts/daily-health-check.sh
@@ -224,18 +224,24 @@ log "=== Health check complete: ${#FAILURES[@]} failure(s) ==="
 # On failure, write a message to the Lobster inbox so it gets picked up
 #-------------------------------------------------------------------------------
 if [ ${#FAILURES[@]} -gt 0 ]; then
-    FAIL_LIST=$(printf '%s\n' "${FAILURES[@]}" | sed 's/^/  - /')
-    MSG_FILE="$INBOX_DIR/daily-health-$(date +%Y%m%d-%H%M%S).json"
-    cat > "$MSG_FILE" << MSGEOF
-{
-  "type": "health_check",
-  "source": "system",
-  "timestamp": "$TIMESTAMP",
-  "subject": "Daily health check: ${#FAILURES[@]} failure(s)",
-  "body": "The daily dependency health check found problems:\n\n$FAIL_LIST\n\nCheck the log for details: $LOG_FILE",
-  "severity": "warning"
-}
-MSGEOF
+    # Build the inbox message using jq so that embedded newlines in $FAIL_LIST are
+    # properly escaped as \n in the JSON string.  A heredoc with bare variable
+    # expansion produces literal newlines inside a JSON string value, which makes
+    # json.load() throw "Invalid control character" — causing WFM to tight-loop
+    # and exhaust --max-turns (the 2026-04-24 / 2026-04-25 restart storm bug).
+    BODY="The daily dependency health check found problems:"$'\n\n'"$(printf '%s\n' "${FAILURES[@]}" | sed 's/^/  - /')"$'\n\n'"Check the log for details: $LOG_FILE"
+    MSG_ID="daily-health-$(date +%Y%m%d-%H%M%S)"
+    MSG_FILE="$INBOX_DIR/${MSG_ID}.json"
+    jq -n \
+        --arg id        "$MSG_ID" \
+        --arg type      "health_check" \
+        --arg source    "system" \
+        --arg timestamp "$TIMESTAMP" \
+        --arg subject   "Daily health check: ${#FAILURES[@]} failure(s)" \
+        --arg body      "$BODY" \
+        --arg severity  "warning" \
+        '{id: $id, type: $type, source: $source, timestamp: $timestamp, subject: $subject, body: $body, severity: $severity}' \
+        > "$MSG_FILE"
     log "Failure alert written to inbox: $MSG_FILE"
     exit 1
 fi


sayhar:
PASS: fix is correct and complete.

**Core fix — jq --arg properly escapes embedded newlines**

Verified: the old heredoc approach produces `Invalid control character` when any failure name contains a literal newline, because bash variable expansion inside a heredoc injects raw control characters into the JSON string value. The new `jq -n --arg body "$BODY"` approach escapes all newlines as `\n` regardless of what the FAILURES array contains. Confirmed with a manual test: heredoc → `json.decoder.JSONDecodeError: Invalid control character`, jq version → `json.load()` succeeds cleanly.

**id field**

Correctly added. The `id` field is set to the same value as the filename stem (`MSG_ID`), matching the convention used everywhere else in inbox_server.py (compact_reminder, scheduled_reminder, debug_observation). This is what `handle_mark_processing` uses to locate the message (line 4323: `msg.get("id") == message_id`).

**source: "system"**

Correct. `"system"` is in `INBOX_MESSAGE_SOURCES` (message_types.py line 75). Files with an unrecognized source are quarantined to failed/ to prevent the hot-loop (lines 5084–5115), so this field must be a known value — it is.

**type: "health_check"**

Unchanged from the original heredoc. `"health_check"` is in `INBOX_SYSTEM_TYPES` (message_types.py line 53). No issue.

**BODY construction**

`$'\n\n'` inserts literal newline pairs in the bash string before jq receives it. jq then escapes those as `\n` in the JSON output. This is correct — the body will display with proper paragraph breaks when the dispatcher reads it.

**No other issues found.** Single-file change, minimal scope, no branching logic to cover beyond what the manual test exercised.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Fix correctness

The root cause is real and the fix is correct. The old heredoc inserted literal newline control characters directly into the JSON string value. `json.load()` correctly rejects these (`Invalid control character`), and the dispatcher's `wait_for_messages` tight-looped as a result. Manual verification confirms both sides:

- Old construction: `json.loads('{"body": "line1\nline2"}')` → `JSONDecodeError: Invalid control character`
- New `jq -n --arg body "$BODY"` output: properly escapes as `\n` → `json.load()` succeeds

## New `id` field

The fix adds an `id` field (`MSG_ID`) to the inbox JSON, which was absent in the old heredoc. This is correct. `inbox_server.py:4353` shows `msg_id = msg.get("id", msg_file.stem)` — without an explicit `id`, the server falls back to the filename stem. Adding `id` explicitly aligns this message with the format all other inbox messages use and removes the fallback dependency.

## `jq` dependency

`jq` is already a first-class dependency: `daily-health-check.sh` itself checks for it at line 60 (`check "jq" "command -v jq"`). If `jq` is missing, the script records that as a FAILURE and would never reach the inbox-write block. No circular dependency problem.

## Tests

No automated test specifically covers the `daily-health-check.sh` inbox-write path (the engineer acknowledged this). The fix is a single mechanical substitution with no branching logic, so the manual verification (`jq` → `python3 -c "json.load()"`) is adequate. The test suite (595 passing, 3 skipped in the non-unit suite; unit collection errors are pre-existing and unrelated to this PR) is unaffected.

## Potential regression

None identified. The only behavioral change is: the written JSON is now valid. Field set is identical to before plus `id`.

## One minor nit (non-blocking)

The `BODY` construction line uses `$'\n\n'` literals and a subshell:
```bash
BODY="The daily dependency health check found problems:"$'\n\n'"$(printf '%s\n' "${FAILURES[@]}" | sed 's/^/  - /')"$'\n\n'"Check the log for details: $LOG_FILE"
```
This is slightly harder to read than a printf-based approach, but it is correct and there are no edge cases introduced.

PR #2013, opened 2026-05-08
feat(context-monitor): display raw token count instead of percentage in warnings

### LINKED ISSUE #1956: feat(context-monitor): display token count instead of percentage
## Context

PR #1955 fixed context-monitor.py to use 200k as the CC window size for claude-sonnet-4-6 and claude-opus-4-6 instead of 1M. But Alex pointed out that showing a percentage is inherently fragile — it depends on the assumed max window, which varies (200k default, up to 1M in some modes).

## Proposal

Change the context warning message to show raw token count instead of (or in addition to) percentage. Example:

**Before:**
> Context at 84% (168,000/200,000 tokens)

**After:**
> Context at 168,000 tokens (84% of 200k default window)

Or simply:
> Context at 168,000 tokens

The token count is always accurate. The percentage is only meaningful if the denominator is correct.

sayhar:
## Problem

Context warning messages showed percentage of the context window used, but percentage is fragile: it depends on the assumed max window size, which varies (200k default, up to 1M in some modes). If the assumed denominator is wrong, the percentage is misleading. The raw token count is always accurate, regardless of what window size the model actually uses.

This was flagged in issue #1955 (which fixed 1M → 200k for Sonnet/Opus 4-6) and tracked in #1956.

## Change

**Before:**
> Context window at 84.0% — entering wind-down mode

**After:**
> Context at 168,000 tokens (84.0% of 200k default window) — entering wind-down mode

Token count leads. The percentage is preserved as secondary context for orientation, but the denominator is now labelled explicitly ("200k default window") so the reader knows what assumption it carries.

### Functional patterns used
- Pure functions: `_build_warning_message()` is a pure transformer from `(used_pct, total_tokens)` → message dict
- Immutability: return values are new dicts/tuples; no mutation of shared state
- Composable: `_read_transcript_usage()` now returns a 4-tuple; callers unpack what they need

### Scope
Only `hooks/context-monitor.py` and its unit test file are touched. No changes to the inbox server, state machine, or dedup logic.

### Before/after (no state transition change)
This change does not alter any state machine transitions, threshold values, or when warnings are written. It only changes the format of the `text` field in the context_warning inbox message. No diagram needed.

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_context_monitor.py -v` — 23 passed, 0 failed
- [ ] Full test suite — not run (change is isolated to display string; existing tests cover all paths)

**New tests added (TestTokenCountDisplay):**
- `test_warning_message_leads_with_token_count` — text starts with "Context at N tokens"
- `test_warning_message_includes_percentage_as_secondary` — percentage appears in parentheses
- `test_warning_message_stores_total_tokens_field` — total_tokens stored as a top-level field
- `test_inbox_warning_contains_token_count_in_text` — end-to-end through _handle_payload

## Manual test
N/A — this change is entirely internal. The context_warning message is consumed by the dispatcher (not displayed directly to the user). No Telegram output to verify.

## Definition of Done
- [x] Integration/manual test run — N/A (internal message format, no external service)
- [x] User-visible outcome — N/A (context_warning is consumed internally by the dispatcher)
- [x] Side-effect audit: no external calls, no volume changes, no shared state written beyond the existing inbox message

Closes #1956

### DIFF
--- a/hooks/context-monitor.py
+++ b/hooks/context-monitor.py
@@ -90,10 +90,12 @@ def _model_max_context(model: str) -> int:
 
 def _read_transcript_usage(
     transcript_path: str | None,
-) -> "tuple[float, float, str] | None":
+) -> "tuple[float, float, str, int] | None":
     """Read the last assistant usage entry from the transcript JSONL.
 
-    Returns (used_pct, remaining_pct, model) or None if data is unavailable.
+    Returns (used_pct, remaining_pct, model, total_tokens) or None if data
+    is unavailable. total_tokens is the raw token count (input + cache), which
+    is always accurate regardless of the assumed max context window.
 
     Reads the file line-by-line, keeping only the last assistant entry with a
     usage block. This is O(n) in lines but O(1) in memory — suitable for large
@@ -157,7 +159,7 @@ def _read_transcript_usage(
     used_pct = (total_used / model_max) * 100.0
     remaining_pct = 100.0 - used_pct
 
-    return (used_pct, remaining_pct, last_model)
+    return (used_pct, remaining_pct, last_model, total_used)
 
 
 def _log_usage(log_dir: Path, entry: dict) -> None:
@@ -200,15 +202,26 @@ def _build_absent_context_entry(tool_name: str, reason: str = "") -> dict:
     }
 
 
-def _build_warning_message(used_pct: float) -> dict:
-    """Return the context_warning inbox message payload."""
+def _build_warning_message(used_pct: float, total_tokens: int = 0) -> dict:
+    """Return the context_warning inbox message payload.
+
+    The message text leads with the raw token count (always accurate) and
+    includes the percentage as secondary context (only meaningful if the assumed
+    max window is correct). See issue #1956.
+    """
+    token_str = f"{total_tokens:,}"
+    text = (
+        f"Context at {token_str} tokens ({used_pct:.1f}% of 200k default window)"
+        " — entering wind-down mode"
+    )
     return {
         "id": str(uuid.uuid4()),
         "type": "context_warning",
         "source": "system",
         "chat_id": 0,
-        "text": f"Context window at {used_pct:.1f}% — entering wind-down mode",
+        "text": text,
         "used_percentage": used_pct,
+        "total_tokens": total_tokens,
         "timestamp": datetime.now(timezone.utc).isoformat(),
     }
 
@@ -255,7 +268,7 @@ def _handle_payload(
         _log_usage(log_dir, entry)
         return
 
-    used_pct, remaining_pct, model = result
+    used_pct, remaining_pct, model, total_tokens = result
 
     entry = _build_log_entry(tool_name, used_pct, remaining_pct, model)
     _log_usage(log_dir, entry)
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Code Review — PASS

### Overview

This PR changes the context warning message from percentage-first to token-count-first, as requested in #1956. The motivation is sound: raw token count is always accurate, while the percentage is only meaningful if the assumed max window (200k) is correct — which varies by mode.

### Correctness — all callers updated

- `_read_transcript_usage()` return type extended to 4-tuple `(used_pct, remaining_pct, model, total_used)` — correct.
- The single call site in `_handle_payload()` is correctly updated to unpack the 4-tuple: `used_pct, remaining_pct, model, total_tokens = result` — correct.
- `_build_warning_message()` now accepts `total_tokens` and produces the new format — correct.
- `dispatcher-state-stop.py` has its own independent copy of the transcript-reading logic (noted in a comment on line 155); it is not a caller of `_read_transcript_usage()` and is unaffected — no issue there.
- The dispatcher bootup doc (sys.dispatcher.bootup.md) references `context_pct` in the wind-down steps (step 6: "Context at {pct}%"). That text is in the dispatcher's own prompt/instructions and reads the `used_percentage` field, which is still present and unchanged in the message payload. No update needed.

### Message format

The new text:
```
Context at 168,000 tokens (84.0% of 200k default window) — entering wind-down mode
```
Matches the "after" example from the issue. Token count leads; percentage is secondary and its denominator is now labelled. The `total_tokens` field is also stored on the message dict for any programmatic consumers.

One minor observation: the `total_tokens` default in `_build_warning_message(used_pct: float, total_tokens: int = 0)` defaults to 0, which would produce "Context at 0 tokens" if called without the argument. In practice the only call site always passes the real count, but the default is a latent footgun. Consider `total_tokens: int` with no default, or assert `total_tokens > 0`. Not blocking — callers are correct as written.

### Test coverage

The new `TestTokenCountDisplay` class covers four cases:

| Test | What it checks |
|---|---|
| `test_warning_message_leads_with_token_count` | Text starts with formatted count |
| `test_warning_message_includes_percentage_as_secondary` | Percentage present in parentheses |
| `test_warning_message_stores_total_tokens_field` | `total_tokens` field on the dict |
| `test_inbox_warning_contains_token_count_in_text` | End-to-end through `_handle_payload` |

Existing tests for `_read_transcript_usage()` are correctly updated to unpack the new 4-tuple and assert on `total_tokens`. All 23 tests pass (confirmed in PR description).

One small gap: no test covers the `total_tokens=0` default path (i.e., `_build_warning_message(80.0)` without the second arg). Given the default is 0, the text would read "Context at 0 tokens (80.0% of 200k default window)". Worth a test or removing the default — but not blocking given that path is never triggered in production.

### Code quality

- The format string is clean and readable.
- The comma-formatted token string (`f"{total_tokens:,}"`) is idiomatic Python and produces the correct `168,000` format.
- Docstrings updated consistently.
- No mutation of shared state; all functions remain pure.

### Risk

Low. The change is isolated to a display string in a single hook. The `context_warning` inbox message structure is backward-compatible: `used_percentage` is still present (the dispatcher still reads it for `{context_pct}` in the wind-down message to the user), and `total_tokens` is additive. No state machine changes.

---

**Summary:** Correctness is solid, all callers updated, test coverage is good. Two non-blocking observations (default=0 footgun, missing test for default path). Ready to merge.

PR #2018, opened 2026-05-09
fix(#1935): add outbox fallback scanner to survive inotify queue overflow

### LINKED ISSUE #1935: Slack router outbox file-watcher silently stops after startup
## Summary

The lobster-slack-router service starts successfully and logs "Watching outbox for Slack replies...", but the outbox file-watcher stops picking up new files after running for some time. The service remains active and continues receiving incoming Slack messages — only the outbox watch is broken. A service restart immediately processes any queued outbox files.

## Observed behavior

- Observed twice on 2026-05-05: once at ~15:16 UTC and again at ~16:54 UTC
- Messages accumulated in ~/messages/outbox/ for many minutes without being sent
- Service showed as "active (running)" throughout
- Incoming Slack messages continued to be received normally
- Restarting the service caused all queued outbox files to be sent immediately

## Reproduction

1. Run lobster-slack-router for an extended period
2. Send a Slack message to trigger a reply
3. Observe that ~/messages/outbox/ accumulates files that are not sent
4. Restart the service — queued messages send immediately

## Possible root causes

- inotify watch limit hit (the OS-level inotify queue overflowed and the watcher stopped receiving events)
- A threading issue in the watcher loop
- An unhandled exception in the watcher loop that silently swallows errors, causing the loop to exit without logging

## Impact

Slack users see no response until the service is manually restarted.

## Suggested investigation

- Check for unhandled exceptions in the outbox watcher thread/loop
- Check inotify watch limits (`/proc/sys/fs/inotify/max_user_watches`, `max_queued_events`)
- Add a health-check or watchdog that alerts if the outbox watcher stops processing files
- Add explicit error logging in the watcher loop so silent failures surface in the service logs

sayhar:
## Problem

The Slack router service silently stops sending outbox replies after running
for some time. The service stays active and continues receiving Slack messages,
but outbox files accumulate without being sent until the service is restarted.
This was observed twice on 2026-05-05 within a few hours.

The root cause is that the watchdog inotify Observer can silently stop
delivering `on_created` events in two scenarios:

1. **Kernel inotify queue overflow (`IN_Q_OVERFLOW`)**: when the kernel
   event queue is full, watchdog's emitter thread may miss events or die
   without logging anything visible.
2. **Observer thread death**: if an unhandled exception propagates out of
   `dispatch_events()`, the observer thread exits silently — the service
   remains alive but no further events are dispatched. The watchdog loop
   only catches `queue.Empty`; any other exception kills the thread.

In both cases `lobster-slack-router.service` shows as `active (running)`
because the Socket Mode handler and user-DM poll thread are still running.
Only the outbox watcher is dead.

## Fix

Adds a dedicated daemon thread (`outbox-scan-fallback`) that calls
`drain_outbox()` on every `OUTBOX_SCAN_INTERVAL`-second tick (default 30s).

```
Before (watchdog Observer only):
  inbox <- [Socket Mode] [user-DM poller]
  outbox -> [watchdog Observer]  <- can die silently

After (watchdog + fallback scanner):
  inbox <- [Socket Mode] [user-DM poller]
  outbox -> [watchdog Observer]           <- fast path (event-driven)
         -> [outbox-scan-fallback]  <- catches files missed by watchdog
```

Key properties:
- **Complementary, not replacement.** The watchdog Observer remains the
  primary delivery path; the fallback scanner is a safety net.
- **At most 30s delay** for files missed by inotify (configurable).
- **Observer health surfaced in logs.** The scan thread calls
  `observer.is_alive()` on every tick and logs a `WARNING` when the thread
  has died. Previously this was invisible until a restart.
- **Clean shutdown.** The scan thread's stop event is set in the `finally`
  block alongside the poll thread, and joined with a 5s timeout.
- **Configurable via `LOBSTER_SLACK_OUTBOX_SCAN_INTERVAL`** (documented in
  `config.env.example`).

No migration needed -- no new directories, cron entries, or service files.

## Functional patterns

Pure functions: `_scan_outbox_periodically` takes its stop event and
observer as arguments, closes over no mutable global state, and delegates
all side effects to `drain_outbox`. The main() wiring composes these
together declaratively.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_slack_outbox_scan_fallback.py -v` -- 9 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/ -v` -- 3272 passed, 31 skipped, 0 failed (full suite)
- [ ] Live Slack integration test -- not run. Requires production restart and a Slack session. This is safe to merge; the change is additive and the fallback scanner only calls `drain_outbox()` which is already exercised at startup.

## Manual test

N/A for this PR. The new thread calls `drain_outbox()` which is the same
idempotent function called at startup (`drain_outbox` deletes files on
success, so double-delivery cannot occur). Live end-to-end verification
requires waiting for the watcher to fail naturally.

## Definition of Done

- [ ] Integration/manual test -- deferred; additive and idempotent
- [ ] User-visible outcome verified -- deferred pending live deployment
- [x] Side-effect audit: `drain_outbox` is idempotent; no floods, no unscoped writes

Closes #1935

### DIFF
--- a/config/config.env.example
+++ b/config/config.env.example
@@ -56,6 +56,10 @@ OPENAI_API_KEY=your_openai_api_key_here
 # silence.  Requires LOBSTER_SLACK_USER_TOKEN (xoxp-) and chat.update permission.
 # Default: true.  Set to false if your Slack plan restricts chat.update.
 # LOBSTER_SLACK_TYPING_INDICATOR=true
+# Outbox fallback scanner interval in seconds (default: 30).
+# A background thread re-scans the outbox on this interval to catch files
+# missed by the inotify watcher (e.g. after a kernel queue overflow).
+# LOBSTER_SLACK_OUTBOX_SCAN_INTERVAL=30
 
 # Twilio (shared credentials for SMS and WhatsApp)
 # Sign up at https://twilio.com — SID and Auth Token at https://console.twilio.com
--- a/src/bot/slack_router.py
+++ b/src/bot/slack_router.py
@@ -624,6 +624,24 @@ _POLL_CHANNELS = [
 ]
 _POLL_INTERVAL = int(os.environ.get("LOBSTER_SLACK_POLL_INTERVAL", "10"))  # seconds
 
+# ---------------------------------------------------------------------------
+# Outbox fallback scanner: periodic re-scan to catch files missed by watchdog.
+#
+# The watchdog inotify Observer can stop delivering events silently when:
+#   - The kernel inotify queue overflows (IN_Q_OVERFLOW)
+#   - An unhandled exception kills the observer thread
+# In both cases the service stays alive but queued outbox files are not sent.
+#
+# This scanner runs in a separate thread and calls drain_outbox() on every
+# tick so that any files the watchdog missed are caught within one interval.
+# It also checks observer.is_alive() on every tick and logs a WARNING when
+# the observer thread has died, making the failure visible in logs rather than
+# silently accumulating queued messages.
+# ---------------------------------------------------------------------------
+OUTBOX_SCAN_INTERVAL: int = int(
+    os.environ.get("LOBSTER_SLACK_OUTBOX_SCAN_INTERVAL", "30")
+)
+
 # Warn at startup if the user token is configured but no poll channels are set.
 # Without poll channels, DMs sent to the user identity will not be received.
 if SLACK_USER_TOKEN and not _POLL_CHANNELS:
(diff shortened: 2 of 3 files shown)


sayhar (commented):
## Review of PR #2018 — outbox fallback scanner

**PASS** with one non-blocking nit noted below.

---

### Overview

The watchdog inotify Observer thread can silently die from an unhandled exception or an inotify queue overflow, leaving the service running but unable to deliver outbox replies. This PR adds a 30-second periodic re-scan of the outbox directory as a safety net, plus an observer liveness check that surfaces the failure in logs. The fix is purely additive.

---

### Correctness

**Double-delivery safety confirmed.** When both the watchdog and the fallback scanner both see the same file, `_deliver_file` calls `send_fn` then unconditionally calls `path.unlink(missing_ok=True)`. The second caller will find the file gone — either `open(path)` raises `FileNotFoundError` (caught by the outer `except Exception`) or the glob won't list it. The typing indicator path returns `True` on success, which causes deletion before the scanner can also pick it up. Double-delivery is not possible in the normal case.

**Edge case acknowledged but acceptable.** In the narrow race where both threads call `_deliver_file` on the same path simultaneously (watchdog fires nanoseconds before scanner), one thread will succeed and delete the file; the other will log a read error or a silently-suppressed `FileNotFoundError`. The `_send_slack_reply` call won't be doubled because only the first thread to open and parse the file will reach it. This is a pre-existing design property of `drain_outbox`, not a new risk introduced here.

**Clean shutdown.** Both stop events are set before the joins in `finally`, and both threads are joined with a 5s timeout before `observer.stop()`. Order is correct.

**Observer liveness check placement is correct.** Checking `observer.is_alive()` at the top of each scan iteration means the warning fires once per tick rather than flooding the log continuously (it doesn't rate-limit, but 30s between warnings is acceptable).

---

### Code quality

- `_scan_outbox_periodically` takes its inputs as arguments and has no mutable global state — purely functional in the right sense.
- Thread name `"outbox-scan-fallback"` is descriptive and distinguishable in thread dumps.
- `OUTBOX_SCAN_INTERVAL` is a named module-level constant, configurable, documented in `config.env.example`. Consistent with how `_POLL_INTERVAL` is handled.
- Docstring accurately describes the behavior, failure modes, and arguments.

---

### Test coverage

9 tests cover: default config, env-var override, type enforcement, drain invocation, send_fn identity, stop-before-first-tick, observer-dead warning, no-warning-when-alive, per-tick health checks. All behaviors in the implementation have corresponding tests. The `patch("threading.Event.wait", ...)" technique avoids real sleeps and is consistent with the existing test pattern in this module.

One unused import (`call` from `unittest.mock`) is imported but never used in the test file — minor nit, no behavior impact.

---

### Nit (non-blocking)

```python
# tests/unit/test_bot/test_slack_outbox_scan_fallback.py line 19
from unittest.mock import MagicMock, call, patch
```

`call` is imported but not used. Can be removed in a follow-up or squashed in before merge — not a blocker.

---

### Risks

- **Not a root-cause fix.** This PR does not prevent the observer from dying; it compensates for it. If the observer crash is frequent, the underlying cause (inotify queue size, unhandled exception in emitter) should eventually be investigated. The observer-dead WARNING in logs makes that investigation easier.
- **SMS/WhatsApp routers have the same vulnerability.** `sms_router.py` and `whatsapp_router.py` use the same watchdog pattern without a fallback scanner. Out of scope for this PR, but worth a follow-up issue.
- **30s window is acceptable.** The typical interaction is: user sends message → Lobster replies → outbox file written → either watchdog fires within milliseconds (normal case) or scanner picks it up within 30s (fallback case). 30s is noticeable but not disruptive.

---

**Verdict: PASS.** The fix is correct, well-tested, and safe to deploy. The nit is cosmetic and does not block merge.

PR #2020, opened 2026-05-09
fix(#1995): clear context-handoff.json at startup to prevent stale data

### LINKED ISSUE #1995: fix(startup): context-handoff.json not cleared after single-use read
Part of #1993

## Problem

`context-handoff.json` is designed as a single-use file: it carries state across one restart and is then obsolete. However it is never cleared after being read. On the next startup, the dispatcher encounters stale entries (e.g. in-flight agents from March 2026 showing as "running"). The 10-minute staleness check filters these out, but stale entries still appear as confusing noise in the startup sequence.

## Proposed fix

After reading `context-handoff.json` on startup — regardless of whether the contents were fresh or stale — delete or zero out the file. The file's purpose is single-use (survive one restart), so it should be consumed and discarded immediately after reading.

Implementation options:
- Delete the file after reading
- Overwrite with `{}` or `{"consumed": true, "timestamp": "..."}` to leave a breadcrumb

Either way, subsequent startups should not see stale in-flight agent entries from previous sessions.

sayhar:
## Problem

`context-handoff.json` is a single-use artifact: the Stop hook writes it when the dispatcher ends, and the next dispatcher startup reads it at step 2c to recover in-flight agent state. The startup protocol instructs the LLM to delete the file after reading — but LLM-side deletion is unreliable. Skipping that instruction (or the session ending before reaching it) leaves stale data in the file indefinitely.

On subsequent restarts, the dispatcher encounters months-old in-flight agent entries — e.g., tasks from March 2026 showing as "running" — which appear as confusing noise even after the 10-minute staleness check filters them out.

## Fix

Move the deletion guarantee from LLM instruction to code. `_clear_context_handoff()` is added to `hooks/on-fresh-start.py` and called on every fresh dispatcher start, before the dispatcher's first turn. It overwrites `context-handoff.json` with `{}` atomically (tmp file + os.replace), enforcing the single-use guarantee at the code level.

**Why `{}` instead of delete:** An empty JSON object is valid and causes the dispatcher step 2c check to find no `triggered_at` field, which is treated as "no prior context" — the same safe fallback as a missing file. Readers that blindly parse the file also won't crash.

**Timing:** The hook fires before the dispatcher's first turn. The dispatcher can still read the file during step 2c of the same startup. On the next restart, the file is empty — stale data is gone.

## Flow

```
Restart N:
  Stop hook fires -> writes context-handoff.json with {triggered_at, context_pct, ...}

Restart N+1:
  on-fresh-start.py fires (before dispatcher's first turn)
    -> _clear_context_handoff() overwrites context-handoff.json with {}
  Dispatcher starts
    -> step 2c: reads context-handoff.json
    -> sees {} (no triggered_at) -> treats as "no prior context"

Restart N+2: (previously would see stale Restart N data; now always sees {})
  on-fresh-start.py fires
    -> _clear_context_handoff() overwrites the Stop hook's fresh file with {}
  Dispatcher starts clean
```

## Also included

Migration 91 in `upgrade.sh` clears any stale `context-handoff.json` on existing installs (including the 45KB March 2026 file currently in `~/lobster-workspace/data/`).

## Functional patterns used

Pure function `_clear_context_handoff()` with no return value and no shared state. All I/O is isolated to atomic file writes (tmp + os.replace). The `CONTEXT_HANDOFF_FILE` path is resolved from an env-var override at import time for test isolation.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_context_handoff_clear.py -v` — 8 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/test_hooks/` — 435 passed, 7 skipped, 0 failed (existing hook tests unaffected)
- [x] `uv run pytest tests/unit/` — 3270 passed, 1 failed (pre-existing bisque isolation failure in `test_relay_server.py::TestWSAuth::test_auth_success` — unrelated to this PR, noted in prior commit 82b46fec)

## Manual test

N/A — this change is entirely internal to the startup hook. No external service calls, no Telegram output, no database changes. The cleared file is regenerated by the Stop hook on the next session end.

## Definition of Done
- [x] Integration/manual test: not applicable (internal hook logic only)
- [x] User-visible outcome: not directly visible — removes confusing stale-agent noise from startup logs
- [x] Side-effect audit: clearing {} into a file is fully reversible; the Stop hook writes fresh content on the next session end

Closes #1995


sayhar (commented):
PASS: Clear logic is correct and fires at the right time. `_clear_context_handoff()` is guarded by both the compaction-event and subagent checks before it executes, so it only runs for fresh dispatcher starts. The atomic overwrite (tmp + os.replace) is correct. All 8 new tests pass. The 2 pre-existing failures in test_dispatcher_state_stop.py and test_dispatcher_state_stop_handoff.py are unrelated to this PR (confirmed by reproducing on local-dev HEAD before applying). Deployed to local-dev manually with migration 91 renumbered to 95 to avoid conflict with migrations 92–94 that landed on local-dev after this branch was cut.

PR #2022, opened 2026-05-09
fix(#1950): whitelist Read through post-compact gate to break circular dependency

### LINKED ISSUE #1950: fix(post-compact-gate): Read gate should include confirmation token to avoid circular dependency
## Problem

When context compaction occurs, the post-compact gate blocks all tool calls until the dispatcher re-orients. The current flow has a circular dependency:

1. The **Read gate** (`post-compact-gate.py`) fires and blocks all file reads, saying:
   > "Your only permitted action right now is to call `mcp__lobster-inbox__wait_for_messages` by its full name directly"

2. So the dispatcher calls `wait_for_messages` — but the **WFM gate** then also fires, saying:
   > "Read `~/lobster-workspace/.claude/sys.dispatcher.bootup.md` for the confirmation token, then call `wait_for_messages(confirmation='LOBSTER_COMPACTED_REORIENTED')`"

3. But the Read gate blocks reading the bootup file! The dispatcher can't read the file to get the token.

The token `LOBSTER_COMPACTED_REORIENTED` is embedded in the WFM gate error message itself, but the dispatcher has to fail WFM once before discovering it.

## Fix

Update the **Read gate** error message (in `hooks/post-compact-gate.py` or the equivalent) to include the confirmation token directly:

> "Context compaction detected. Call `mcp__lobster-inbox__wait_for_messages(confirmation='LOBSTER_COMPACTED_REORIENTED')` to clear the gate. No file reads needed."

This would let the dispatcher pass the confirmation token on the first WFM call, eliminating the circular dependency.

Alternatively, update the WFM gate to include the token in a way that makes the Read gate's instruction unnecessary.

## Labels

`bug`, `dx`

sayhar:
## Problem

After context compaction, the post-compact gate blocks all tool calls until the dispatcher calls `wait_for_messages(confirmation='LOBSTER_COMPACTED_REORIENTED')`. But the confirmation token lives only in `sys.dispatcher.bootup.md` — a file the dispatcher could not read because Read was blocked. This created a circular dependency:

```
Compaction occurs
    → gate blocks Read
    → dispatcher calls WFM (no token)
    → gate blocks WFM: "call WFM with token LOBSTER_COMPACTED_REORIENTED"
    → dispatcher learns token from error message
    → dispatcher calls WFM with token
    → gate clears
```

The dispatcher had to fail WFM once before discovering the token. Worse, the error message was the only path to the token — the bootup file (the canonical home of the token) was unreachable.

## Fix

Whitelist Read unconditionally alongside ToolSearch. The gate's purpose is to prevent **state mutation** before re-orientation, not to block information access. Reads are non-destructive — they cannot send messages, spawn agents, or mutate state.

With Read whitelisted, the flow becomes:

```
Compaction occurs
    → dispatcher reads sys.dispatcher.bootup.md (now allowed through gate)
    → dispatcher calls WFM(confirmation='LOBSTER_COMPACTED_REORIENTED')
    → gate clears
```

One fewer forced failure. The token can now live exclusively in the bootup file rather than also being embedded in gate error messages as a workaround.

The gate still blocks all state-mutating tools (Bash, Write, Edit, all MCP tools except wait_for_messages) until the sentinel is cleared.

## Changes

- `hooks/post-compact-gate.py`: Add `READ_TOOL = "Read"` constant and a passthrough check before the sentinel enforcement block, mirroring the existing ToolSearch passthrough. Update `DENY_REASON` and `DENY_REASON_NEEDS_TOKEN` to accurately state that Read and ToolSearch are permitted.
- `tests/unit/test_hooks/test_post_compact_gate_read.py`: New test file with 4 unit tests covering the Read always-allow path (pass through with fresh sentinel, sentinel not deleted, any file path passes, Write still blocked).

**Functional patterns used:** named constants (`READ_TOOL`, `TOOL_SEARCH_TOOL`, `CONFIRMATION_TOKEN`), early-exit guard clauses, pure functions (`sentinel_is_fresh`, `log_gate_event`).

## Before / after gate flow

```
Before                               After
──────────────────────────────────   ──────────────────────────────────
PreToolUse: ToolSearch → ALLOW       PreToolUse: ToolSearch → ALLOW
PreToolUse: Read       → DENY  ✗     PreToolUse: Read       → ALLOW  ✓
PreToolUse: WFM(none)  → DENY        PreToolUse: WFM(none)  → DENY
PreToolUse: WFM(token) → ALLOW, clear PreToolUse: WFM(token) → ALLOW, clear
```

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_post_compact_gate_read.py -v` — 4 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/test_post_compact_gate_toolsearch.py -v` — 3 passed (no regression)
- [x] `uv run pytest tests/smoke/test_post_compact_gate.py -v` — 5 passed (no regression)
- [x] `uv run pytest tests/unit/ -q` — 3267 passed, 31 skipped, 0 failed

## Manual test
N/A — this change is entirely a hook logic fix. The hook runs in-process via `importlib` in tests, with sentinel files isolated to `tmp_path`. No external services, systemd, cron, or DB schema are touched.

## Definition of Done
- [x] Integration/manual test run — not applicable (hook logic, no external services)
- [x] User-visible outcome — not applicable (gate is invisible to the user; effect is fewer WFM failures in dispatcher logs post-compaction)
- [x] Side-effect audit: no floods, no unscoped writes, no unintended volume — the only behavior change is that Read passes through the gate; no messages are sent

Closes #1950

### DIFF
--- a/hooks/post-compact-gate.py
+++ b/hooks/post-compact-gate.py
@@ -9,7 +9,14 @@ on-compact.py when a compaction occurs. This hook passes when:
   1. No sentinel file exists (normal operation), OR
   2. Not the dispatcher session (see session_role.is_dispatcher()), OR
   3. The sentinel is older than SENTINEL_TTL_SECONDS (stale / post-hibernation), OR
-  4. The tool being called IS wait_for_messages (let it through, delete sentinel).
+  4. The tool being called IS wait_for_messages (let it through, delete sentinel), OR
+  5. The tool being called IS ToolSearch (read-only schema fetch, needed to load
+     deferred MCP tool schemas — the dispatcher must be able to call ToolSearch to
+     fetch the wait_for_messages schema before it can call wait_for_messages), OR
+  6. The tool being called IS Read (read-only file access — non-destructive and
+     needed to break the circular dependency in issue #1950: Read is required to
+     learn the confirmation token from the bootup file, but the gate was blocking
+     Read before the token was known).
 
 The TTL (10 minutes) eliminates the stuck-sentinel failure mode: if the
 dispatcher hibernates or crashes while the sentinel is active, the next boot
@@ -90,18 +97,28 @@ WAIT_FOR_MESSAGES_TOOL = "mcp__lobster-inbox__wait_for_messages"
 # ToolSearch first.  See issue #1914 staging infinite-compaction-loop bug.
 TOOL_SEARCH_TOOL = "ToolSearch"
 
+# Read is a read-only file access tool.  It is always allowed through the gate
+# because reads are non-destructive: they cannot send messages, spawn agents, or
+# mutate state.  The gate's purpose is to prevent state mutation before
+# re-orientation, not to block information access.  Blocking Read created a
+# circular dependency (issue #1950): the gate blocked reads, but the dispatcher
+# needed to read the bootup file to learn the confirmation token.  Whitelisting
+# Read eliminates this deadlock — the token lives only in sys.dispatcher.bootup.md,
+# which is now readable through the gate without a two-step WFM dance.
+READ_TOOL = "Read"
+
 CONFIRMATION_TOKEN = "LOBSTER_COMPACTED_REORIENTED"  # noqa: S105 — not a secret, intentional safe word
 
 DENY_REASON_NEEDS_TOKEN = (
     "GATE BLOCKED: Context compaction was just detected. "
-    "If you need the wait_for_messages schema, call ToolSearch first (it is allowed). "
+    "You may Read files or call ToolSearch (both are allowed). "
     "Then call `mcp__lobster-inbox__wait_for_messages(confirmation='LOBSTER_COMPACTED_REORIENTED')` directly. "
     "Confirmation token: LOBSTER_COMPACTED_REORIENTED"
 )
 
 DENY_REASON = (
     "GATE BLOCKED: Context compaction was just detected. Your only permitted "
-    "actions right now are: (1) ToolSearch to load the wait_for_messages schema if needed, "
+    "actions right now are: (1) Read files or call ToolSearch as needed, "
     "then (2) call `mcp__lobster-inbox__wait_for_messages(confirmation='LOBSTER_COMPACTED_REORIENTED')` directly. "
     "When it returns, you will receive a compact-reminder system message — read it to re-orient "
     "as the Lobster dispatcher, then resume your main loop normally. "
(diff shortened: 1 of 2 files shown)


sayhar (commented):
## Code Review: PASS

### Overview

This PR fixes a genuine circular dependency in the post-compact gate. The gate required the confirmation token to call WFM, but the token lived in a file the gate blocked the dispatcher from reading. The fix correctly identifies that Read is categorically non-destructive and whitelists it alongside the already-whitelisted ToolSearch. The approach matches the recommendation in the issue's updated scope comment exactly.

### Correctness

The logic is sound:

- Read is placed **before** the WFM branch and the sentinel enforcement block, consistent with ToolSearch's placement. This means the whitelist is evaluated regardless of sentinel state — correct.
- The sentinel is **not** deleted by Read (only WFM+token clears it). The implementation is correct and the test `test_read_allowed_and_sentinel_remains` explicitly verifies this.
- `test_write_still_blocked_when_sentinel_fresh` confirms the allow-path does not accidentally widen to all file-related tools.
- The unconditional file-path coverage test (`test_read_allowed_any_file_path`) correctly verifies the whitelist is not path-scoped.

### Code quality

- Named constant (`READ_TOOL = "Read"`) follows the existing convention (`TOOL_SEARCH_TOOL`, `WAIT_FOR_MESSAGES_TOOL`). Good.
- Log action string `"allowed-read"` is descriptive and distinguishable from `"allowed-schema-fetch"`. Consistent style.
- Comments in `main()` inline with the ToolSearch block are appropriately brief — they state the reason, not just the effect.
- The module-level docstring is updated to include Read as pass condition 6. Matches actual behavior.
- `DENY_REASON` and `DENY_REASON_NEEDS_TOKEN` are updated to accurately describe what is now permitted. Before this PR, those messages were technically incorrect post-fix.

### Test coverage

- 4 new unit tests in the ToolSearch pattern. All critical behaviors are covered (allow, sentinel-not-deleted, any-path, Write-still-blocked).
- 3267 unit tests passed (no regressions).
- Smoke tests for the gate pass.
- No existing tests needed updating.

- The new test file duplicates `_make_sentinel`, `_load_hook`, `_make_hook_input`, `_run_hook`, and `_make_dispatcher_session_file` verbatim from `test_post_compact_gate_toolsearch.py`. A future cleanup could extract these into a shared `conftest.py` helper module. Not a problem for this PR — it matches the existing ToolSearch test's structure and keeps tests self-contained.
- `import session_role` inside the test method body is consistent with `test_post_compact_gate_toolsearch.py` — intentional to avoid early-import issues before `sys.path` is patched.

### Security

No concerns. Read is inherently read-only — it cannot exfiltrate, send, or mutate. The gate was designed to prevent action before re-orientation; information gathering was never the threat model.

**Verdict: PASS — logic correct, tests thorough, no regressions. Safe to merge and deploy.**

Issue #1949, opened 2026-05-07
on-compact.py: use MCP/outbox instead of direct Telegram API call


sayhar:
## Problem

`on-compact.py` currently sends its compaction notification by calling the Telegram Bot API directly (via `requests` or similar). This bypasses the outbox/MCP pipeline that all other Lobster notifications use.

## Why this matters

- Direct API calls skip the send queue and error handling in the outbox
- It makes on-compact.py harder to test (requires real Telegram credentials)
- It violates the system design: all outbound messages should go through the MCP `send_reply` tool or equivalent outbox mechanism

## Proposed fix

Rewrite on-compact.py to write a notification message to the MCP inbox/outbox (via the lobster-inbox MCP tools) and let the dispatcher route it — same pattern used by every other system notification.

## Context

Discovered while debugging compaction hook reliability (issue #1940, PR #1948). PR #1948 fixed the hook registration (matcher="" + self-gate) and confirmed the hook fires. The architecture improvement is a separate follow-up.

🤖 Filed by Lobster dispatcher

PR #2023, opened 2026-05-09
test(hooks): fix smoke tests broken by source-field compact detection

sayhar:
## Summary

After commit `c0c534ad` (issue #1949) switched `_is_compact_event()` to check `data["source"] == "compact"` as the primary detection field, the smoke test helper in `tests/smoke/test_on_compact.py` was left using the old `"is_compact": True` field that the function no longer reads.

This caused tests A1–A6 to exit early via `sys.exit(0)` rather than processing the compact event, silently invalidating all six core behavior tests. Only A7 (the new outbox-write test) still passed, since it calls `send_compaction_notify()` directly without going through `main()`.

The fix is a one-line change in the test helper: use `source="compact"` (the CC-documented primary field) instead of the defunct `is_compact: True` flag.

## What changed

- `_load_hook()` in `tests/smoke/test_on_compact.py`: hook input uses `source="compact"` so `_is_compact_event()` returns True and `main()` enters the compact-event code path.

## Before / after

Before: A1–A6 exit at `sys.exit(0)` (self-gate returns False), writing nothing. Tests appear to pass vacuously but exercise no actual behavior.

After: All 7 smoke tests (A1–A7) exercise the correct code paths.

## Functional patterns

Pure-function testing approach: the fix uses the documented CC payload field (`source`) rather than an internal flag, keeping test inputs aligned with the real contract. No behavior changes; test-only fix.

## Tests run

- [x] `uv run pytest tests/smoke/test_on_compact.py -v` — 7 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 427 passed, 7 skipped, 0 failed

## Manual test

N/A — this is a test-only change with no behavior change to production code.

## Definition of Done

- [x] Integration/manual test — N/A (test-only fix)
- [x] User-visible outcome verified — not applicable (no user-visible behavior change)
- [x] Side-effect audit: no production code changed, no outbound messages, no shared state writes
- [x] External service calls: none (test-only fix)

Relates to #1949

### DIFF
--- a/tests/smoke/test_on_compact.py
+++ b/tests/smoke/test_on_compact.py
@@ -109,8 +109,10 @@ def _load_hook(
 
     # Patch sys.stdin so main()'s json.load(sys.stdin) gets a valid compact event
     # rather than hitting pytest's captured stdin which raises OSError.
+    # Use source="compact" (CC-documented primary field) so _is_compact_event()
+    # returns True and main() processes the event rather than exiting early.
     hook_input = json.dumps(
-        {"session_id": session_id, "hook_event_name": "SessionStart", "is_compact": True}
+        {"session_id": session_id, "hook_event_name": "SessionStart", "source": "compact"}
     )
     mod._test_stdin_data = hook_input  # store for reference
 


sayhar:
**Reviewer verdict: PASS.**

Correct fix — `is_compact: True` was never part of the CC-documented payload; `_is_compact_event()` reads `data["source"] == "compact"` (line 387) and `data.get("hook_name") == "compact"` as fallback. The old test field matched neither, so A1–A6 were exiting at the self-gate and validating nothing.

- Field alignment verified against `hooks/on-compact.py` lines 387–389.
- All 7 smoke tests pass locally after the change.
- No production code modified.

Safe to merge.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

**Verdict: PASS — fix is correct, tests pass locally, no regressions found.**

## What the fix does

`_is_compact_event()` was changed (in a prior PR) to check `data["source"] == "compact"` as its primary detection field. The smoke test helper `_load_hook()` was still injecting the old `{"is_compact": True}` payload format, causing `main()` to hit the self-gate and call `sys.exit(0)` before doing anything. Tests A1–A6 appeared to pass only because `sys.exit(0)` doesn't raise an exception — they were exercising nothing. A7 was unaffected because it calls `send_compaction_notify()` directly.

The fix is a one-line change: the injected payload now uses `{"source": "compact"}`, which matches what `_is_compact_event()` actually reads.

## Verification

- `uv run pytest tests/smoke/test_on_compact.py -v` — **7 passed, 0 failed** (confirmed locally on PR branch)
- `uv run pytest tests/unit/test_hooks/ --tb=no -q` — **427 passed, 7 skipped, 0 failed** (matches engineer's claim)
- `uv run pytest tests/unit/test_hooks/test_is_compact_event.py -v` — **9 passed** (unit tests directly cover `_is_compact_event()` with `source="compact"` returning True, old format returning False, fallback via `hook_name`, priority rules)

## Code correctness

The change is correct: `_is_compact_event()` checks `data.get("source")` first — if present, it must equal `"compact"`. The fallback to `hook_name` only applies when `source` is absent. The fix aligns the test fixture with the documented CC payload field.

The `hook_event_name: "SessionStart"` field is preserved in the fixture alongside `source: "compact"`, which is reasonable since a real CC payload would include both.

## Edge cases checked

- Non-compact events (`source: "startup"`, `source: "resume"`) are covered in unit tests and correctly return False — no regression risk.
- The fallback path (`hook_name` without `source`) continues to work per unit tests.
- Duplicate test class names (`TestHookMain` in two files) are pre-existing, not introduced by this PR, and pytest collects both correctly without shadowing.

## No concerns

The PR is test-only, touches one file, and the single commit is clean. The diff between the PR branch and main shows additional commits that are pre-existing work on the branch (not part of this PR). The actual PR commit is exactly the one-liner described.

PR #2024, opened 2026-05-09
fix(startup): compact inflight-work.jsonl on startup to drop stale orphaned running entries

### LINKED ISSUE #1997: fix(startup): inflight-work.jsonl accumulates stale 'running' entries
Part of #1993

## Problem

`inflight-work.jsonl` retains entries with `status="running"` indefinitely when the corresponding subagent never wrote a "done" entry (e.g. due to a crash or orphaned session). Entries from March 2026 remain in the file months later. `compact-catchup` uses this file to detect "lost subagents" — these stale entries cause false positives every startup.

## Proposed fix

On startup, after reading `inflight-work.jsonl`, tombstone or compact stale entries:

**Option A (tombstone):** Write a `"done"` entry (with `status="expired"`) for any `"running"` entry older than 24 hours that has no corresponding `"done"` entry. Preserves the append-only log structure.

**Option B (compact):** Rewrite the file on startup, dropping all `"running"` entries older than 6 hours that have no corresponding `"done"` entry. Keeps the file small and removes false-positive sources entirely.

Option B is cleaner. Either way, `compact-catchup` should not be triggered by sessions that are clearly weeks or months old.

This is straightforward to implement as a startup cleanup step — a short script that reads the JSONL, identifies stale orphans, and rewrites the file.

sayhar:
## Problem

`inflight-work.jsonl` accumulates `status="running"` entries indefinitely when the corresponding subagent crashed or was killed without writing a `done` entry. Entries from March 2026 were still present months later, and `compact-catchup` was reporting them as false-positive "lost subagent" alerts on every startup.

## Fix

On every fresh dispatcher restart, `on-fresh-start.py` now calls `_compact_inflight_work()` which:

1. Reads all entries from `inflight-work.jsonl`
2. Identifies `task_id`s that have at least one `done` entry
3. Drops any `running` entry older than 6 hours (`INFLIGHT_STALE_THRESHOLD_HOURS`) that has no matching `done` entry
4. Atomically rewrites the file (temp file + rename)

Preserved unconditionally: all `done` entries, `running` entries within the 6h threshold, `running` entries with a matching `done` (regardless of age), and malformed lines are skipped rather than crashing.

The compaction is silent on all errors — it cannot block dispatcher startup.

## Before / after

```
Before:                         After:
inflight-work.jsonl             inflight-work.jsonl
  running  2026-03-31 task-A      (dropped — stale orphan)
  running  2026-03-31 task-B      (dropped — stale orphan)
  running  2026-04-04 task-C    running  2026-04-04 task-C  <- paired with done
  done     2026-04-04 task-C    done     2026-04-04 task-C
  done     2026-05-08 task-D    done     2026-05-08 task-D
  running  2026-05-08 task-E    running  2026-05-08 task-E  <- within 6h
```

`compact-catchup`'s "Possibly lost subagents" section now only shows genuinely recent orphans, not multi-month-old ghosts.

## Functional patterns

- `_compact_inflight_entries()` is a pure function (no I/O, no side effects, input not mutated)
- All I/O is isolated in `_compact_inflight_work()` which is called from `main()`
- The 6h threshold is a named constant (`INFLIGHT_STALE_THRESHOLD_HOURS`) used in both tests and implementation

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_compact_inflight_work.py -v` — 20 passed, 0 failed (new tests)
- [x] `uv run pytest tests/unit/ -q` — 3283 passed, 31 skipped, 0 failed (full suite)
- [x] `uv run python -m py_compile hooks/on-fresh-start.py` — syntax clean

## Manual test

N/A — this change is entirely internal (hook runs at startup, no external service calls, no Telegram output). The compacted file can be verified by inspection of `~/lobster-workspace/data/inflight-work.jsonl` after the next restart. The real-world inflight-work.jsonl has 2405 entries; after startup with this fix, entries from March 2026 will be gone.

## Definition of Done
- [x] Integration/manual test run — N/A (internal startup hook, no external services)
- [x] User-visible outcome verified — N/A (internal, reduces noise in compact-catchup output)
- [x] Side-effect audit complete: no floods, no unscoped writes; only rewrites inflight-work.jsonl atomically on startup

Closes #1997

### DIFF
--- a/hooks/on-fresh-start.py
+++ b/hooks/on-fresh-start.py
@@ -86,7 +86,9 @@ import os
 import sqlite3
 import subprocess
 import sys
+import tempfile
 import time
+from datetime import datetime, timezone
 from pathlib import Path
 
 # Allow imports from the hooks directory (session_role).
@@ -106,6 +108,16 @@ COMPACTION_STATE_FILE = Path(
 
 INBOX_DIR = Path(os.path.expanduser("~/messages/inbox"))
 PROCESSING_DIR = Path(os.path.expanduser("~/messages/processing"))
+
+# inflight-work.jsonl path — append-only log of subagent spawns and completions.
+# On startup, stale "running" entries with no corresponding "done" are compacted out.
+# Override via env var for testability.
+INFLIGHT_WORK_FILE = Path(
+    os.environ.get(
+        "LOBSTER_INFLIGHT_WORK_FILE_OVERRIDE",
+        os.path.expanduser("~/lobster-workspace/data/inflight-work.jsonl"),
+    )
+)
 
 # Pointer to the current session file (written by compact-catchup / session management).
 # If this file exists and was modified recently, there is a prior session worth catching up.
@@ -138,6 +150,10 @@ STALE_CATCHUP_THRESHOLD_SECONDS = 30 * 60  # 30 minutes
 # ran successfully at the start of the previous session.
 SESSION_FILE_RECENCY_SECONDS = 4 * 60 * 60  # 4 hours
 
+# "running" entries in inflight-work.jsonl older than this with no matching
+# "done" entry are dropped on startup compaction (issue #1997).
+INFLIGHT_STALE_THRESHOLD_HOURS = 6
+
 STARTUP_COMPACT_REMINDER_TEXT = (
     "COMPACT REMINDER \u2014 RE-ORIENT NOW (injected by on-fresh-start.py)\n\n"
     "A previous session ended without completing compact-catchup. "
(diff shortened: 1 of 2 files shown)


sayhar (commented):
NEEDS-WORK: Critical field name mismatch — the compaction function reads \`\"ts\"\` but the vast majority of inflight entries use \`\"started_at\"\`.

## Blocking issue: wrong timestamp field name

The dispatcher bootup spec (`.claude/sys.dispatcher.bootup.md`) defines the inflight-work.jsonl format as:

\`\`\`json
{"task_id": "...", "type": "...", "description": "...", "started_at": "<ISO UTC>", "chat_id": ..., "status": "running"}
\`\`\`

The writer in the dispatcher uses \`started_at\`. I checked the live file at \`~/lobster-workspace/data/inflight-work.jsonl\`:

- Running entries with \`"ts"\`: **9** (a handful of very old March 31 entries from before the format was standardized)
- Running entries with \`"started_at"\`: **1,228**
- Orphaned running entries with \`"ts"\`: 4
- Orphaned running entries with \`"started_at"\`: **205**

The implementation in \`_compact_inflight_entries\` reads \`entry.get("ts")\`. For every entry using \`"started_at"\`, \`ts_str\` will be \`None\`, triggering this path:

\`\`\`python
if not ts_str:
    # No timestamp → cannot determine age → treat as stale, drop.
\`\`\`

Effect: on the first startup after this is deployed, the function will drop all 205 orphaned running entries that use \`"started_at"\`, regardless of their actual age. This includes entries from seconds ago. The goal was to drop old stale entries, not all of them.

**Fix required:** read \`entry.get("ts") or entry.get("started_at")\`. The no-timestamp fallback (treat as stale) should only apply if both fields are absent.

## Secondary issue: malformed_count rewrite trigger

When \`dropped == 0\` but \`malformed_count > 0\`, the file is rewritten. This silently discards the malformed lines (they were skipped during parse, so they don't appear in \`compacted\`). Each malformed line is logged before being dropped, so it's not completely silent, but the rewrite is arguably unnecessary if no stale orphans are being removed. Low-severity; acceptable if intentional. Suggest a comment explaining the intent (e.g., "also rewrite to clean up malformed lines").

## Items that PASS

1. **Atomic write**: correct — uses \`tempfile.mkstemp\` in the same directory + \`os.replace\`.
2. **Error handling**: outer \`except Exception\` ensures the hook never crashes.
3. **Boundary conditions**: \`_ts(INFLIGHT_STALE_THRESHOLD_HOURS - 0.1)\` and \`_ts(INFLIGHT_STALE_THRESHOLD_HOURS + 0.1)\` adequately cover just-under and just-over. The comment explaining why exactly-6h is skipped is appropriate.
4. **Test coverage**: 20 tests cover all the documented invariants cleanly.
5. **Call site placement**: \`_compact_inflight_work()\` is called before \`_mark_all_running_failed()\` — correct ordering.
6. **Pure function separation**: \`_compact_inflight_entries\` is side-effect-free and independently testable — good design.
7. **Done-task preservation**: building \`done_task_ids\` set first and preserving running entries with a matching done is correct.

PR #2026, opened 2026-05-09
fix(pre-push): add 30s timeout to PII scanner on large diffs

sayhar:
## What problem this solves

The pre-push PII/security scanner had no time bound. On large branches (local-dev, 38+ commits ahead of origin), `git log ... -p` generates tens of MB of diff output and the scan stalls for 60+ seconds. This forced subagents to use `git push --no-verify` as a workaround — bypassing the scan entirely, which is the wrong tradeoff. This PR adds a 30-second deadline and a fast-path filter so the scan completes in bounded time.

Relates to #1951. Stacks on PR #1952 (non-interactive mode fix) — this PR is based on that branch.

## What changed

**`get_push_diff()` in `.githooks/pre-push`:**

- Each `git log`/`git diff` call is now wrapped with `timeout ${SCAN_TIMEOUT_SECONDS}s` (default: 30 s)
- `--diff-filter=ACM` added to both calls — deleted-file blobs cannot contain newly-added secrets so they are skipped, cutting diff size on branches with many deletions
- `SCAN_TIMED_OUT` flag (init: 0) is set to 1 when `timeout` exits with code 124

**Key implementation note:** The exit code must be captured *before* any `|| true` — `|| true` inside `$()` subshells masks timeout's exit code 124, making it undetectable. Fixed by removing `|| true` from the timed calls.

**Main section:**
- If `SCAN_TIMED_OUT=1`: prints a warning ("Pre-push scan timed out — skipping PII check") and continues. The warning tells the developer to run the diff manually.
- Non-interactive (CI) mode: still exits 0 regardless (existing behavior from PR #1952)

## Flow

```
Before:
  git push → get_push_diff() → git log -p (no timeout) → can run 60+ seconds → stall

After:
  git push → get_push_diff() → timeout 30s git log --diff-filter=ACM -p
              ├─ completes in time → scan runs normally
              └─ timeout (exit 124) → SCAN_TIMED_OUT=1 → warning printed → push continues
```

## Tests

New file: `tests/unit/test_pre_push_timeout.sh` (8 tests):

1. `SCAN_TIMEOUT_SECONDS` constant is declared and positive
2. `SCAN_TIMED_OUT=0` is initialised at top of hook
3. `get_push_diff` wraps git with `timeout ${SCAN_TIMEOUT_SECONDS}s`
4. `--diff-filter=ACM` is present in git calls
5. Exit status 124 is detected and mapped to `SCAN_TIMED_OUT=1`
6. Functional: `get_push_diff` sets `SCAN_TIMED_OUT=1` when deadline exceeded
7. CI mode: exits 0 with warning when scan timed out
8. Warning message references `SCAN_TIMEOUT_SECONDS`

## Tests run

- [x] `bash tests/unit/test_pre_push_timeout.sh` — 8 passed, 0 failed
- [x] `bash tests/test-security-scan.sh` — 39 passed, 0 failed (existing hook logic unaffected)
- [x] `bash tests/unit/test_setup_claude_hooks.sh` — 53 passed, 0 failed
- [x] `uv run pytest tests/ --ignore=tests/smoke --ignore=tests/integration --ignore=tests/stress --tb=no -q` — 3681 passed, 31 skipped
- [x] Live push test: `git push origin fix/pre-push-timeout-1951` — hook ran in non-interactive mode, scanned diff, exited 0 cleanly (confirmed in push output above)

Pre-existing failures in `tests/smoke/test_on_compact.py` (6 tests) confirmed present on the base branch before this PR.

## Manual test

The push of this branch (`git push origin fix/pre-push-timeout-1951`) exercised the hook live in non-interactive mode. Output confirmed: hook scanned outgoing diff, found no issues, and exited 0. The hook did not hang. No `--no-verify` was needed.

Not applicable: N/A for Telegram/UI output — this is a git hook with no user-facing Telegram interaction.

## Definition of Done

- [x] Integration/manual test run — see ## Manual test above
- [x] User-visible outcome: hook no longer blocks or requires --no-verify on large branches
- [x] Side-effect audit: hook only reads git state, no writes to shared state
- [x] External service calls: none (git operations only)

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -62,6 +62,10 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
 ALLOWLIST_FILE="$REPO_ROOT/.githooks/security-allowlist.txt"
 ISSUES_FOUND=0
 
+# Temp file used to communicate SCAN_TIMED_OUT across the subshell boundary
+# created by DIFF_TEXT="$(get_push_diff)".
+TIMEOUT_FLAG_FILE=$(mktemp)
+
 # Detect CI / non-interactive mode: if stdin is not a tty, don't block
 IS_TTY=1
 [ ! -t 0 ] && IS_TTY=0
@@ -166,7 +170,7 @@ get_push_diff() {
 
     # timeout exits 124 when the deadline is exceeded
     if [ $exit_status -eq 124 ]; then
-        SCAN_TIMED_OUT=1
+        echo 1 > "$TIMEOUT_FLAG_FILE"
         echo ""
         return 0
     fi
@@ -176,7 +180,7 @@ get_push_diff() {
             git diff --diff-filter=ACM origin/main..HEAD 2>/dev/null)"
         local fallback_status=$?
         if [ $fallback_status -eq 124 ]; then
-            SCAN_TIMED_OUT=1
+            echo 1 > "$TIMEOUT_FLAG_FILE"
             echo ""
             return 0
         fi
@@ -456,6 +460,8 @@ echo -e "${RED}${BOLD}==========================================================
 echo ""
 
 DIFF_TEXT="$(get_push_diff)"
+SCAN_TIMED_OUT=$(cat "$TIMEOUT_FLAG_FILE" 2>/dev/null || echo 0)
+rm -f "$TIMEOUT_FLAG_FILE"
 
 if [ "$SCAN_TIMED_OUT" -eq 1 ]; then
     warn "Pre-push scan timed out after ${SCAN_TIMEOUT_SECONDS}s — skipping PII check."
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -65,6 +65,7 @@ ISSUES_FOUND=0
 # Temp file used to communicate SCAN_TIMED_OUT across the subshell boundary
 # created by DIFF_TEXT="$(get_push_diff)".
 TIMEOUT_FLAG_FILE=$(mktemp)
+trap 'rm -f "$TIMEOUT_FLAG_FILE"' EXIT
 
 # Detect CI / non-interactive mode: if stdin is not a tty, don't block
 IS_TTY=1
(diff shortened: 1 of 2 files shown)


sayhar:
Deployed to local-dev.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: fix is correct and well-tested, with two minor hardening notes that are non-blocking.

## What the PR does

Adds a 30-second `timeout` wrapper around the `git log`/`git diff` calls in `get_push_diff()` so the pre-push PII hook cannot stall indefinitely on large branches. When `timeout` exits 124 the hook prints a warning and allows the push rather than blocking it. A tmpfile (`TIMEOUT_FLAG_FILE`) propagates the timed-out flag across the subshell boundary created by `DIFF_TEXT="$(get_push_diff)"`.

## Does the fix solve the stated problem?

Yes. The core issue — `git log -p` on 38+ commits blocking forever — is addressed. The exit-code-124 capture logic is correct; the `|| true` that would have masked it was removed.

## Pre-existing test failure

`tests/unit/test_bisque/test_relay_server.py::TestWSAuth::test_auth_success` fails, but `git log` shows this file was last touched in commit `b74d34e3` (not in this PR). Confirmed pre-existing on the base branch. All 8 new tests pass. Existing security-scan tests (39) pass.

## Two non-blocking issues

**1. Tempfile is not cleaned up on abnormal exit.**

`TIMEOUT_FLAG_FILE=$(mktemp)` is created at the top of the script but only removed via an explicit `rm -f` in the main body. If the script exits early (e.g., `set -uo pipefail` fires before reaching line 465, or a future early-exit path is added), the tmpfile leaks into `/tmp`. Low impact — OS cleans `/tmp` eventually — but a `trap 'rm -f "$TIMEOUT_FLAG_FILE"' EXIT` immediately after `mktemp` would make this robust:

```bash
TIMEOUT_FLAG_FILE=$(mktemp)
trap 'rm -f "$TIMEOUT_FLAG_FILE"' EXIT
```

**2. Last-resort `git log` fallback has no timeout.**

The third path in `get_push_diff()` — reached when both `@{u}..HEAD` and `origin/main..HEAD` return empty — is:

```bash
diff_output="$(git log --diff-filter=ACM -p HEAD 2>/dev/null || true)"
```

This has no `timeout` wrapper. On a repo with a large history and no upstream configured, this path could still hang. For consistency the same `timeout` + exit-124 check should be applied here. (In practice it requires both a new repo with no upstream tracking and a large history, so it is an edge case, but the fix is trivially the same pattern already used above.)

## Summary

The design is sound: tmpfile approach for cross-subshell flag propagation is the right solution (env var assignment inside `$()` is invisible to the parent shell), the 30s constant is documented and can be tuned, `--diff-filter=ACM` is a sensible optimization, and the tests cover the critical path including functional simulation of the timeout. The two issues above are nits that do not affect the main use case.

PR #2027, opened 2026-05-09
fix(hooks): remove matcher="compact" from setup_claude_hooks() + migration 91 (issue #1947)

### LINKED ISSUE #1947: fix(hooks): on-compact.py silently skipped — matcher="compact" unreliable in CC 2.1.119

sayhar:
## What changed and why

`setup_claude_hooks()` in `install.sh` was still registering `on-compact.py` with `matcher="compact"`, which CC 2.1.119 fires in only ~37% of compaction events. This is a known CC bug documented in issue #1947.

The hook already has an internal `_is_compact_event()` self-gate that exits early on non-compact sessions, so `matcher=""` (fires on all SessionStart events) is the correct registration. The hook's behavior is unchanged — it just gets called more reliably.

Two bugs were in `setup_claude_hooks()`:
1. `on-compact.py` registered with `matcher="compact"` → changed to `matcher=""`
2. A separate `inject-bootup-context.py` entry with `matcher="compact"` added → removed entirely. The empty-matcher `inject-bootup-context.py` entry already fires on all session types including compact. The compact-specific entry was causing double-injection on every compaction.

**Note:** The live `~/.claude/settings.json` was already correct (prior migrations 86 and 88 had been applied). This PR fixes the *source* — so fresh installs and container builds no longer produce the broken configuration.

## What this does not change

- `on-compact.py` hook logic is unchanged
- `_is_compact_event()` detection (three-tier: source → hook_name → heartbeat fallback) is unchanged
- Existing correct settings.json files are not disturbed by migration 91 (it is idempotent)

## Changes

| File | Change |
|------|--------|
| `install.sh` | `setup_claude_hooks()`: change `matcher="compact"` → `matcher=""` for on-compact.py; remove compact-matcher inject-bootup-context block |
| `scripts/upgrade.sh` | Add migration 91: patches existing settings.json to match |
| `scripts/next-migration.txt` | 91 → 92 |
| `tests/unit/test_hooks/test_setup_claude_hooks_compact_matcher.py` | 9 new unit tests |

## Hook flow (before/after)

```
Before:
  CC fires SessionStart (compact)
    → matcher="compact" → on-compact.py  [unreliable, ~37% fire rate]
    → matcher="compact" → inject-bootup-context.py  [second call, double-inject]
    → matcher=""        → inject-bootup-context.py  [first call]

After:
  CC fires SessionStart (compact)
    → matcher=""  → inject-bootup-context.py  [fires reliably, all session types]
    → matcher=""  → on-compact.py  [fires reliably; self-gate exits early on non-compact]

  CC fires SessionStart (fresh start/resume)
    → matcher=""  → inject-bootup-context.py  [fires normally]
    → matcher=""  → on-compact.py  [fires; _is_compact_event() returns False; exit 0]
```

Net effect: compaction notifications and compact-reminder messages are delivered on every compaction, not just ~37%.

## Functional patterns

The test's `_apply_migration_91()` is a pure port of the upgrade.sh migration logic — given a settings dict, produces a new dict without mutation. Tests derive invariants from the spec, not from implementation details.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_setup_claude_hooks_compact_matcher.py -v` — 9 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 436 passed, 7 skipped, 0 failed
- [x] `uv run pytest tests/unit/ -q` — 3272 passed, 31 skipped, 0 failed
- [x] `bash tests/unit/test_setup_claude_hooks.sh` — 32 passed, 0 failed (validates live settings.json)

## Manual test

N/A — the live `~/.claude/settings.json` was already correct. This change fixes the source so fresh installs and container builds produce the correct configuration. Functional verification would require a real compaction event.

## Definition of Done

- [x] Unit tests pass with no production hits in automated tests
- [x] User-visible outcome: N/A (configuration source fix; no runtime behavior change for already-correct installs)
- [x] Side-effect audit: migration 91 is idempotent — no-op when already correct; no floods, no unscoped writes

Closes #1947


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The fix is correct and the approach is sound. All 9 new tests pass. Pre-existing failures (46 tests across unrelated modules) are confirmed pre-existing and not introduced by this PR.

---

## What I checked independently before reading the engineer's briefing

**The install.sh idempotency check** — The new guard checks for `contains("on-compact") AND matcher == ""`. This means on an existing install that had the old compact-matcher entry, if the new install.sh runs *before* migration 91, it would find no empty-matcher entry and add one, resulting in two on-compact.py entries (one compact-matcher, one empty-matcher). Migration 91 would then change the compact-matcher to empty-matcher, leaving **two empty-matcher entries**. This is a minor cosmetic artifact but operationally harmless (the script fires twice per session start, but its internal `_is_compact_event()` gate makes both calls no-ops on non-compact events, and for compact events it still runs correctly — just twice). Not blocking, but worth noting.

**Migration 91's detection logic** — Uses `sys.exit(0)` to signal "needs fix" and `sys.exit(1)` for "already correct." The shell then checks `if python3 ...; then needs_fix=1`. This is correct — exit code 0 triggers the if-branch in bash.

**The fi structure** — Reviewed the shell block nesting for migration 91. The `fi` balance is correct: three nested `if` blocks, three corresponding `fi` terminators.

**Tests: no duplicate class names, no shadowing** — Two test classes in the file (`TestMigration91Logic`, `TestInstallShOnCompactMatcher`). No shadowing issues.

**Tests exercise before-state** — `TestMigration91Logic` builds synthetic "broken" settings dicts and asserts post-migration state. It does not just assert comments about what the before-state should be; it actually constructs the broken structure and runs the migration logic against it.

**Test count claim** — The engineer claims "3272 unit tests pass." I ran the full suite and got **3596 passed, 46 failed**. The 46 failures are all pre-existing (confirmed by checking them on the base branch). The higher count reflects tests added in other PRs since the engineer ran their suite. No new failures introduced.

**The install.sh source check tests** — These scan install.sh for forbidden patterns using sliding-window context. They correctly catch regression of the compact-matcher bug. The `'"matcher": "compact"'` string match is precise enough to not false-positive on comments.

---

## Issues found

**Minor / non-blocking:**

1. **Potential duplicate on-compact.py entries after upgrade + re-install** — If a user's settings.json had the old compact-matcher entry, then they ran the new install.sh (which adds an empty-matcher entry because none was found), and then ran upgrade.sh (which changes compact→empty), they'd end up with two empty-matcher on-compact.py entries. The script is idempotent so this is harmless, but `setup_claude_hooks()` could defensively remove stale compact-matcher entries rather than only adding the empty-matcher one. Low priority since migration 91 runs as part of the upgrade flow.

2. **Test count in PR description is stale** — "3272 passed" reflects an older local-dev state; current run shows 3596 passed. Not a code issue.

---

## Verdict

The core fix is correct: `matcher=""` with internal self-gating is the right pattern, the migration correctly patches existing installs, and the tests verify both the migration logic and the install.sh source. The duplicate-entry edge case is cosmetically impure but operationally harmless. Marking PASS.

aeschylus:

aeschylus:
Closing as superseded. Issue #1947 (this PR's target) was already fixed and closed on 2026-05-08 via a different branch (fix/issue-1947-compact-hook-matcher), which patched install.sh's setup_claude_hooks() to use matcher="" + an internal self-gate, and added Migration 88 in scripts/upgrade.sh to backfill existing installs. Verified both are already live on current main (scripts/upgrade.sh Migration 88, install.sh on-compact.py matcher=""). This PR's Migration 91 duplicates that already-shipped fix against a stale scripts/next-migration.txt (91 vs. current 96+), and has been functionally untouched since May 9 (CONFLICTING against current main). Nothing further to do here.

PR #2031, opened 2026-05-09
fix(#2004): jq --arg sweep for safe JSON in periodic-self-check, alert, check-agent-outputs, daily-update-check

### LINKED ISSUE #2004: JSON construction: apply jq --arg pattern to remaining scripts with heredoc variable expansion
## Context

PR #1808 fixed `scripts/daily-health-check.sh` to use `jq -n --arg` for JSON construction instead of heredoc variable expansion. The heredoc approach inserts literal newlines into JSON strings when variables contain multi-line content, producing invalid JSON that causes `json.load()` to throw "Invalid control character" — triggering a WFM tight-loop.

## Affected scripts

The same heredoc pattern exists in at least 4 other scripts. Risk level varies:

- **scripts/periodic-self-check.sh** (HIGH): `$SELF_CHECK_TEXT` concatenates task descriptions and agent summaries — very likely to contain newlines
- **scripts/alert.sh** (MEDIUM): heredoc JSON with variable expansion
- **scripts/check-agent-outputs.sh** (LOW): `$SAFE_ID` is sanitized but same pattern
- **scripts/daily-update-check.sh** (LOW): release/version strings — low risk

## Fix

For each script, replace heredoc JSON construction with `jq -n --arg var "$VAR" '{"key": $var}'` pattern. This correctly escapes all special characters.

## Found by

PR #1808 review (2026-05-08). Fix PR #1808 before this sweep.

sayhar:
## Summary

- Replaces heredoc variable expansion with `jq -n --arg` for JSON construction in 4 scripts
- Fixes the root bug where multi-line variable values insert literal newlines into JSON strings, producing invalid JSON that triggers WFM tight-loops (same class as PR #1808)
- All existing tests pass; `test_script_inbox_sources.py` confirms source values are still valid

## The bug

When a shell heredoc expands a variable that contains newlines:

```bash
cat > file.json << EOF
{"text": "$MULTI_LINE_VAR"}
EOF
```

The literal newlines land inside the JSON string value, producing:

```
{"text": "line1
line2"}
```

This is invalid JSON. `json.load()` throws "Invalid control character", which causes `wait_for_messages` to tight-loop.

## Fix

Replace each heredoc block with:

```bash
jq -n --arg text "$MULTI_LINE_VAR" '{"text": $text}' > file.json
```

`jq` correctly escapes newlines as `\n` and handles all other special characters.

## Affected scripts

| Script | Risk | Variable at risk |
|---|---|---|
| `scripts/periodic-self-check.sh` | HIGH | `$SELF_CHECK_TEXT` — concatenates task descriptions and agent summaries |
| `scripts/alert.sh` | MEDIUM | `$message` — caller-supplied alert text |
| `scripts/check-agent-outputs.sh` | LOW | `$SAFE_ID` — sanitized but same pattern |
| `scripts/daily-update-check.sh` | LOW | Version strings — both git and tarball branches |

## Test plan

- [x] `tests/unit/test_script_inbox_sources.py` — all 3 source-guard tests pass (jq filter literals are still detected)
- [x] `tests/smoke/test_health_check.py` — 10 smoke tests pass
- [x] Manual: `jq -n --arg text "line1\nline2" '{"text": $text}'` produces valid JSON with `\n`-escaped newlines

Closes #2004

### DIFF
--- a/scripts/alert.sh
+++ b/scripts/alert.sh
@@ -29,13 +29,19 @@ echo "[$timestamp] ALERT: $message" >> "$ALERT_LOG"
 # Try to send via Telegram if admin chat ID is configured
 if [[ -n "$ADMIN_CHAT_ID" ]]; then
     alert_file="$OUTBOX_DIR/alert_$(date +%s%N).json"
-    cat > "$alert_file" << EOF
-{
-    "chat_id": $ADMIN_CHAT_ID,
-    "text": "🚨 **Lobster Alert**\n\n$message\n\n_$(date)_",
-    "source": "telegram"
-}
-EOF
+    alert_text="🚨 **Lobster Alert**
+
+$message
+
+_$(date)_"
+    jq -n \
+        --argjson chat_id "$ADMIN_CHAT_ID" \
+        --arg text "$alert_text" \
+        '{
+            "chat_id": $chat_id,
+            "text": $text,
+            "source": "telegram"
+        }' > "$alert_file"
     echo "[$timestamp] Alert sent to Telegram chat $ADMIN_CHAT_ID" >> "$ALERT_LOG"
 fi
 
--- a/scripts/check-agent-outputs.sh
+++ b/scripts/check-agent-outputs.sh
@@ -115,19 +115,21 @@ while IFS= read -r agent_id; do
     # Safely escape agent_id for JSON (alphanumeric + hyphens only expected)
     SAFE_ID=$(echo "$agent_id" | tr -cd 'a-zA-Z0-9_-')
 
-    cat > "${INBOX_DIR}/${MSG_ID}.json" << EOF
-{
-  "id": "${MSG_ID}",
-  "source": "system",
-  "type": "health_check",
-  "chat_id": 0,
-  "user_id": 0,
-  "username": "lobster-system",
-  "user_name": "Agent Relay",
-  "text": "Agent relay check: ${SAFE_ID} output file appears complete. Check task-notifications and relay results to the user.",
-  "timestamp": "${TIMESTAMP}"
-}
-EOF
+    jq -n \
+        --arg id "${MSG_ID}" \
+        --arg text "Agent relay check: ${SAFE_ID} output file appears complete. Check task-notifications and relay results to the user." \
+        --arg timestamp "${TIMESTAMP}" \
+        '{
+            "id": $id,
+            "source": "system",
+            "type": "health_check",
+            "chat_id": 0,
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## Summary

The core fix is correct and well-scoped. `jq --arg` properly escapes newlines and special characters in variable values — the heredoc pattern was genuinely broken for multi-line input, and this PR fixes it correctly across all four affected scripts.

## What I checked

**The jq patterns are correct:**
- `--arg text "$VAR"` handles multi-line strings, quotes, backslashes, and control characters correctly. Verified with `echo -e "line1\nline2"` piped through the new pattern — produces valid JSON with `\n`-escaped newlines.
- `--argjson chat_id "$ADMIN_CHAT_ID"` is the right choice for the numeric chat_id in alert.sh. `LOBSTER_ADMIN_CHAT_ID` is sourced from Telegram user IDs (always integers) and is guarded by `if [[ -n "$ADMIN_CHAT_ID" ]]`, so an empty string never reaches `--argjson`.
- The hardcoded `"chat_id": 0` literals in the other three scripts are correct as JSON number literals — they were already correct before this PR.

**Edge cases I examined:**
- If `ADMIN_CHAT_ID` were somehow non-numeric (e.g. `"not-a-number"`), jq would exit 2 and produce an empty output file. The `echo "Alert sent..."` log line in alert.sh would still execute since there's no `|| exit` after the jq block — but this is a pre-existing issue with the same behavior as the old heredoc (which would have produced broken JSON). No regression here.
- Version strings used in daily-update-check.sh (`$BEHIND`, `$CURRENT_VERSION`, `$LATEST_VERSION`) are not multi-line and carry no injection risk, but the `--arg` treatment is still the right approach for consistency.
- `$SELF_CHECK_TEXT` in periodic-self-check.sh is built by concatenating task summaries that can contain arbitrary text and newlines — this is the highest-risk case and the fix is most important here.

**Tests:** All 13 tests from `test_script_inbox_sources.py` and `test_health_check.py` pass on the PR branch. Full unit test suite is passing.

**--argjson vs --arg:** Used correctly throughout. Numbers get `--argjson`, strings get `--arg`.

**jq -e flag:** Not used, but not needed — the scripts use `jq -n` (null input) with `--arg`/`--argjson`, which only fail if jq itself is missing or the arguments are malformed. Silent failures are not a concern for these well-typed inputs.

No issues found. This is a clean, targeted fix for the exact class of bug identified in #2004.

PR #2032, opened 2026-05-09
fix(#2001): add pretooluse-heartbeat.deprecated.py with firing-log

### LINKED ISSUE #2001: pretooluse-heartbeat.py needs deprecation cleanup (migration 91 conflict)
## Problem

Two related issues with \`hooks/pretooluse-heartbeat.py\` (the OLD hook, deprecated by PR #1988):

1. **File is orphaned on this install**: The hook file exists but is not registered in \`settings.json\`. It should be renamed to \`pretooluse-heartbeat.deprecated.py\` with a firing-log at the top so we can observe whether it fires before removing it.

2. **Migration 91 conflict**: \`upgrade.sh\` migration 91 re-registers this hook (\`pretooluse-heartbeat.py\`). If migration 91 runs on a fresh install, the hook gets re-registered — contradicting PR #1988's intent to deregister it. On this install, migration 91 has not run, so the hook is currently unregistered (correct). But on a fresh install or any install that runs upgrade.sh after this PR, the hook would come back.

## What needs to happen

1. Rename \`hooks/pretooluse-heartbeat.py\` → \`hooks/pretooluse-heartbeat.deprecated.py\`
2. Add deprecation log line at the top: log when the hook fires (should never fire after migration 88/90 deregistered it)
3. Audit migration 91: either remove it (if the intent was always to deregister this hook) or document why it re-registers it (if there's a valid use case)
4. Add a migration to remove any settings.json entry for \`pretooluse-heartbeat\` on installs where migration 91 ran

## Context

- PR #1988 deregistered this hook (migration 90 on local-dev — but as wfm-lifecycle-logger, not heartbeat)
- \`pre-tool-heartbeat.py\` is the NEW hook that replaced it — already correctly registered and now connected to health-check via PR #2000
- Hook audit (2026-05-08) found this as gap 4 of 4; gaps 1-3 were addressed in PR #2000

sayhar:
## Summary

- Adds `hooks/pretooluse-heartbeat.deprecated.py` — the deprecated artifact for the old PreToolUse heartbeat hook that was superseded by `pre-tool-heartbeat.py` (PR #1817) and deregistered by migration 90 (PR #1988)
- The file has a top-of-file firing-log that writes immediately on import, plus a structured JSONL log entry — so any accidental re-registration is immediately observable
- Adds 9 unit tests covering: exit-0 behavior, write failures degrade gracefully, JSONL entry structure, UTC timestamp format, log append behavior, and logs-dir creation

## Context

`hooks/pretooluse-heartbeat.py` never existed on `main` — it was only in `local-dev`. This PR adds the `.deprecated.py` form so the deprecation is explicit in the filesystem and consistent with the `local-dev` branch (where it was added via PR #2015).

Migration 90 already removes any `pretooluse-heartbeat` entry from `settings.json` on existing installs. No new migration is needed.

The migration 91 conflict described in the issue is not relevant to `main` — PR #2027 owns migration 91 on `main` for on-compact.py (a different hook).

## Verification

- `uv run pytest tests/unit/test_hooks/test_pretooluse_heartbeat.py -v` — 9 passed
- `uv run pytest tests/unit/test_hooks/ -q` — 436 passed, 7 skipped, 0 failed

Closes #2001

### DIFF
--- a/hooks/pretooluse-heartbeat.deprecated.py
+++ b/hooks/pretooluse-heartbeat.deprecated.py
@@ -1,10 +1,6 @@
 #!/usr/bin/env python3
 # DEPRECATED: This hook was deregistered by migration 90 (PR #1988).
 # If this fires, it means settings.json was corrupted or a migration re-registered it.
-import datetime as _dt, pathlib as _pl
-_pl.Path("/tmp/pretooluse-heartbeat-deprecated-fired.log").write_text(
-    f"FIRED at {_dt.datetime.utcnow().isoformat()}Z\n", encoding="utf-8"
-)
 
 """
 DEPRECATED — do not register or use this hook.
@@ -51,6 +47,16 @@ def log_unexpected_firing() -> None:
     }
     with open(LOG_FILE, "a") as f:
         f.write(json.dumps(entry) + "\n")
+    # Secondary /tmp marker for quick observability.  Wrapped in its own
+    # try/except so a read-only or missing /tmp cannot propagate an exception
+    # out of this function — tool execution must never be blocked.
+    try:
+        Path("/tmp/pretooluse-heartbeat-deprecated-fired.log").write_text(
+            f"FIRED at {datetime.now(timezone.utc).isoformat()}Z\n",
+            encoding="utf-8",
+        )
+    except Exception:
+        pass
 
 
 def main() -> None:
--- a/hooks/pretooluse-heartbeat.deprecated.py
+++ b/hooks/pretooluse-heartbeat.deprecated.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+# DEPRECATED: This hook was deregistered by migration 90 (PR #1988).
+# If this fires, it means settings.json was corrupted or a migration re-registered it.
+import datetime as _dt, pathlib as _pl
+_pl.Path("/tmp/pretooluse-heartbeat-deprecated-fired.log").write_text(
+    f"FIRED at {_dt.datetime.utcnow().isoformat()}Z\n", encoding="utf-8"
+)
+
+"""
+DEPRECATED — do not register or use this hook.
+
+This file is retained for audit purposes only. It should never be registered
+in settings.json. If this hook fires, it means a settings.json entry for
+pretooluse-heartbeat was re-added incorrectly.
+
+Superseded by: hooks/pre-tool-heartbeat.py (issue #1786, PR #1817)
+  - pre-tool-heartbeat.py adds a dispatcher-only guard so subagent tool calls
+    cannot falsely update the heartbeat timestamp.
+  - This file wrote last_pretooluse_at unconditionally (no dispatcher guard).
+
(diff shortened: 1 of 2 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: top-level side effect at import time

**Tests: 9/9 pass (verified against the PR branch head)**

---

## Critical issue: module-level side effect fires on import

Lines 4-6 of `hooks/pretooluse-heartbeat.deprecated.py`:

```python
import datetime as _dt, pathlib as _pl
_pl.Path("/tmp/pretooluse-heartbeat-deprecated-fired.log").write_text(
    f"FIRED at {_dt.datetime.utcnow().isoformat()}Z\n", encoding="utf-8"
)
```

This code runs unconditionally at import time — before `main()` is ever called. The intent is presumably to make the firing observable, but it has three problems:

1. **No error handling.** The `~/messages/logs/` write in `log_unexpected_firing()` is wrapped in `try/except Exception: pass`, but this `/tmp` write is bare. If `/tmp` is full, read-only, or otherwise unwritable, the import raises an exception and the hook exits non-zero — which violates the core invariant that the deprecated hook must never block tool execution.

2. **Tests don't cover this path.** Every test calls `_run_hook()`, which invokes `spec.loader.exec_module(mod)` — this executes the top-level code. The `/tmp` write happens on the real filesystem during every test run, not in the `tmp_path` sandbox. There is no test asserting that a `/tmp` write failure still results in exit code 0.

3. **Docstring is unreachable.** The module docstring (the triple-quoted block starting with `DEPRECATED — do not register...`) appears after the top-level code, not before it. A string literal at that position is still valid Python, but it is not the module's `__doc__` — `__doc__` will be `None`. This is minor but the docstring placement is confusing.

---

## Non-blocking observations

- **`~/messages/logs/` directory creation is correct.** `LOG_FILE.parent.mkdir(parents=True, exist_ok=True)` handles the missing-directory case properly.
- **Always exits 0 (when main() is reached).** The `sys.exit(0)` is unconditional inside `main()`, and the `try/except` swallows any write failure.
- **No duplicate test class names.** `TestDeprecatedHookExitsZero` and `TestDeprecatedHookLogsWarning` are distinct.
- **No migration needed.** Migration 90 already handles deregistration; that reasoning is sound.

---

## Suggested fix

Move the `/tmp` write inside `log_unexpected_firing()` (or inside `main()`), and wrap it in the same `try/except`:

```python
def log_unexpected_firing() -> None:
    # Fast, no-dependency signal
    try:
        Path("/tmp/pretooluse-heartbeat-deprecated-fired.log").write_text(
            f"FIRED at {datetime.now(timezone.utc).isoformat()}Z\n", encoding="utf-8"
        )
    except Exception:
        pass

    # Structured log for observability
    LOG_FILE.parent.mkdir(parents=True, exist_ok=True)
    ...
```

And add a test: mock `pathlib.Path.write_text` to raise `OSError` and assert the hook still exits 0.

As-is, the bare top-level write can cause the hook to exit non-zero if `/tmp` is unavailable — directly contradicting the spec requirement.

PR #2034, opened 2026-05-09
fix(#1895): add wfm-lifecycle-logger.py + migration 92

sayhar:
## What this does

Adds `hooks/wfm-lifecycle-logger.py` to main and wires it into fresh installs and existing upgrades.

The hook writes `WFM_ENTER` (PreToolUse) and `WFM_EXIT` (PostToolUse) events to `~/lobster-workspace/logs/session-lifecycle.log` each time the dispatcher calls `wait_for_messages`. This lets us measure exactly how long the dispatcher was blocked in WFM and whether it was mid-WFM when it died — closing the observability gap exposed by the 2026-05-01 crash (issue #1895).

## Why it was missing

The hook was implemented in feature branch `feature/issue-1895-crash-logging` (commit `a908357a`) but that branch was never merged to main. The file was cherry-picked to local-dev and registered manually in `settings.json`, which is why it works on the live install today. This PR brings the hook into main so fresh installs get it automatically.

## Changes

- **`hooks/wfm-lifecycle-logger.py`** — new hook (extracted from `a908357a`). Handles both PreToolUse and PostToolUse via `WFM_HOOK_TYPE` env var. Dispatcher-only guard via `session_role`. Writes atomically, fails open.
- **`tests/unit/test_hooks/test_wfm_lifecycle_logger.py`** — 16 unit tests (all passing, extracted from same commit).
- **`install.sh`** — adds registration block after `pre-tool-heartbeat` section. Idempotent: checks both PreToolUse and PostToolUse before writing.
- **`scripts/upgrade.sh`** — adds **migration 92** (migration 91 is reserved for PR #2027). Idempotent: skips if both hook entries already present.

## Test results

```
443 passed, 7 skipped in 9.74s
```

All existing hook tests continue to pass.

## Migration note

Migration 92 registers:
- `PreToolUse` with `matcher="mcp__lobster-inbox__wait_for_messages"` → `WFM_HOOK_TYPE=pre python3 .../wfm-lifecycle-logger.py`
- `PostToolUse` with `matcher="mcp__lobster-inbox__wait_for_messages"` → `WFM_HOOK_TYPE=post python3 .../wfm-lifecycle-logger.py`

The live install already has these entries (registered manually from local-dev), so migration 92 will skip on the current machine.

Closes #1895

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2042, opened 2026-05-09
fix(#1998): write reflection prompts to sidecar file instead of inbox

### LINKED ISSUE #1998: fix(startup): reflection_bootup messages require unnecessary claim+process overhead
Part of #1993

## Problem

Each restart causes `inject-bootup-context.py` (or equivalent) to generate a `reflection_bootup_XXXX` message that lands in the inbox as a regular message. The dispatcher must claim it (`mark_processing`) and mark it processed (`mark_processed`) — 2 extra MCP round-trips per restart, on the startup critical path. These messages add latency and inbox noise without providing value the dispatcher couldn't get another way.

## Proposed fixes

**Fix A:** Have the dispatcher silently drop any message with `subtype="reflection_bootup"` without calling `mark_processing` / `mark_processed`. A single pattern-match check at intake costs nothing.

**Fix B:** Have `inject-bootup-context.py` write reflection prompts to a dedicated sidecar file (e.g. `~/messages/bootup-prompt.md`) instead of the inbox. The dispatcher reads this file at startup (one `Read` call, no claim/process overhead) and the file is then overwritten on the next restart. This keeps the inbox clean and removes the 2-call overhead entirely.

Fix B is the cleaner architectural choice — reflection prompts are not messages, so they should not flow through the message queue.

sayhar:
## Summary

- Eliminates 2 MCP round-trips per restart (`mark_processing` + `mark_processed`) caused by reflection prompt messages flowing through the inbox
- Both `on-compact.py` and `on-fresh-start.py` now write to `~/messages/bootup-prompt.md` (sidecar file) via atomic `.tmp` → rename, only when `LOBSTER_DEBUG=true`
- Dispatcher reads and deletes the sidecar at startup in new step 2e — one `Read` call, zero claim/process overhead
- Updated `.claude/sys.dispatcher.bootup.md` to document step 2e and clarify the old inbox path is legacy (drop silently if encountered)
- Rewrote `tests/unit/test_hooks/test_reflection_prompt.py` for the sidecar approach (16 tests covering debug/non-debug mode, atomic write, no tmp left, parent dir creation, overwrite behavior, silent failure, and no inbox writes)

Closes #1998

## Test plan
- [x] `uv run pytest tests/unit/ -q` — 3265 passed, 31 skipped, 1 pre-existing flaky timing test (`test_exactly_at_window_boundary_blocks`) that passes when run in isolation
- [x] All 16 reflection prompt tests pass (9 for on-compact.py, 7 for on-fresh-start.py)
- [x] `LOBSTER_BOOTUP_PROMPT_FILE_OVERRIDE` env var used for test isolation (no writes to real `~/messages/`)
- [x] `sys.modules["session_role"]` save/restore pattern prevents cross-test pollution

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -57,6 +57,8 @@ When you first start (or after reading this file), follow these steps:
     - Skip if step 2c already sent a restart notification (context-handoff.json was recent).
     - **Do not use `compaction-state.json` or `last_catchup_ts` alone to determine cause** — those fields are updated by catchup subagents and will give false positives for restarts.
 
+2e. **Read reflection prompt sidecar (LOBSTER_DEBUG=true only):** If `~/messages/bootup-prompt.md` exists, read it now and delete it. Reflect on its contents -- were there friction points, gaps, or improvements in the bootup/compaction flow worth capturing? If there are substantive observations, file or update GitHub issues in SiderealPress/lobster, or open PRs for straightforward fixes. If nothing worth capturing: do nothing -- silence is correct. This is a one-Read, zero-claim operation (no `mark_processing`/`mark_processed`). If `LOBSTER_DEBUG` is not `true`, skip this step entirely.
+
 3. **Claim any pending user messages immediately** to stop the health-check staleness clock:
     - Call `check_inbox()` to get any messages currently waiting in the inbox
     - For each message that is NOT a system message (i.e. `chat_id != 0` and `source != "system"`): call `mark_processing(message_id)`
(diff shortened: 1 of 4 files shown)


aeschylus:

aeschylus:
Closing as stale. Opened May 9, functionally untouched since (CONFLICTING against current main, no real commits since the initial Jul 17 bot timestamp sweep). Part of the same abandoned May session-handoff/EventBus redesign batch as #2043, #2034, #2027 (closed, superseded), #2048 (closed, superseded), #2047, #1979 (closed, superseded). Issue #1998 remains open if this sidecar-file approach is still wanted, but should be implemented fresh against current main rather than rebased through 3 months of drift.

PR #2044, opened 2026-05-09
fix(#1994): stop injecting sys.dispatcher.bootup.md into session context

### LINKED ISSUE #1994: fix(startup): pre-injected bootup file truncated to 2KB — injection provides no value
Part of #1993

## Problem

`inject-bootup-context.py` injects `sys.dispatcher.bootup.md` (107KB) into the system-reminder context, but the rendering layer truncates the preview to ~2KB. The dispatcher re-reads the full file anyway via explicit `Read` calls at startup, so the injection adds token/context overhead with zero benefit — the truncated preview is not actionable.

## Proposed fixes

**Fix A:** Increase the preview size in `inject-bootup-context.py` so the full file fits inline (or at least the hot path — the first ~100 lines covering the restart procedure).

**Fix B:** Restructure the bootup file so the first ~100 lines contain everything needed for a minimal restart (the hot path), and the rest is optional cold-path detail. This naturally fits within the preview window and makes the injection genuinely useful.

**Fix C:** Remove the injection entirely and rely solely on the explicit `Read` calls the dispatcher already makes. If the injection provides no value, removing it reduces startup context bloat without changing behavior.

Fix B is likely the best long-term approach — it both makes the injection useful and shortens the cold-start `Read` calls.

sayhar:
## Problem

\`inject-bootup-context.py\` injects \`sys.dispatcher.bootup.md\` into the dispatcher's session context on every startup. The file is 84KB (~21K tokens). The rendering layer truncates the injected preview to ~2KB, so every byte beyond that is discarded. The dispatcher wastes ~21K tokens per startup with zero benefit — it re-reads the full file explicitly via \`Read('.claude/sys.dispatcher.bootup.md')\` as its first action anyway.

## What changed

**Fix C: Drop the body, keep the preambles.**

The dispatcher no longer receives \`sys.dispatcher.bootup.md\` as injected context. It still gets:
1. \`ADMIN_CHAT_ID=<value>\` (from \`config.env\` — 1 line, useful in 2KB window)
2. The \`startup-cause:\` banner (from \`last-startup-cause.json\` — 5 lines, useful in 2KB window)
3. Any user bootup files (\`user.base.bootup.md\`, \`user.dispatcher.bootup.md\`)

Subagent injection is unchanged — \`sys.subagent.bootup.md\` (~2KB) is still injected in full.

### Before / After injection flow

```
Before (dispatcher):
  [inject-bootup-context.py]
    → ADMIN_CHAT_ID=...          # preamble
    → startup-cause banner       # preamble
    → sys.dispatcher.bootup.md  # 84KB BODY (truncated to 2KB, rest discarded)
    → user.base.bootup.md       # if exists
    → user.dispatcher.bootup.md # if exists

After (dispatcher):
  [inject-bootup-context.py]
    → ADMIN_CHAT_ID=...          # preamble (fits in 2KB)
    → startup-cause banner       # preamble (fits in 2KB)
    → user.base.bootup.md       # if exists
    → user.dispatcher.bootup.md # if exists
  [dispatcher's first explicit action]
    → Read('.claude/sys.dispatcher.bootup.md')  # 84KB read in full, no truncation
```

## Documentation updates

- \`sys.dispatcher.bootup.md\` step 2d: updated "read from the banner injected at the top of this file" to "injected into your context by the hook before this file was read"
- \`CLAUDE.md\`: corrected the note on injection — dispatcher reads bootup explicitly; subagent gets it pre-injected

## Functional patterns used

Pure functions (\`_parse_admin_chat_id\`, \`_read_file_safe\`, \`_inject_if_exists\`) with explicit error handling returning \`None\` on failure. Side effects isolated to \`main()\`. No state mutation.

## Tests run

- [x] \`uv run pytest tests/unit/test_hooks/test_dispatcher_bootup_not_injected.py tests/unit/test_hooks/test_inject_admin_chat_id.py tests/unit/test_hooks/test_inject_bootup_context_logging.py tests/unit/test_hooks/test_inject_bootup_context_sentinel.py tests/unit/test_hooks/test_startup_flag_dispatcher_detection.py -v\` — 72 passed, 0 failed
- [x] \`uv run pytest tests/unit/ -q\` — 3672 passed, 43 failed (all 43 are pre-existing failures confirmed by baseline run on unmodified branch; this PR introduced no new failures)

## Manual test

N/A — this change is entirely internal (hook output to Claude's context window, no external services, no DB, no systemd). Behavioral effect is observable only at dispatcher startup; the dispatcher reading \`.claude/sys.dispatcher.bootup.md\` explicitly as its first action is unchanged.

## Definition of Done

- [x] User-visible outcome: N/A — this is a startup efficiency fix, not a user-visible feature
- [x] Side-effect audit: no floods, no shared state writes, no external calls

Closes #1994


sayhar (commented):

## Summary

The core change (stop injecting the 84KB dispatcher bootup file) is correct and well-structured. The `inject-bootup-context.py` refactor is clean, the function decomposition is good, and the 43 pre-existing failures were independently verified. However there are two concrete bugs in the new code:

---

## Issue 1 (BLOCKING): Duplicate test class — 4 tests permanently unreachable

**File:** `tests/unit/test_ghost_detector_filesystem.py`

The class `TestGuard2WfmActiveDispatcherType` is defined **twice** in the same file (lines ~103 and ~262 of the diff, both added by this PR). In Python, the second definition silently overwrites the first. The first class contains 4 tests:

- `test_wfm_active_fresh_skips_dispatcher_typed_uuid_session`
- `test_wfm_not_active_marks_dispatcher_typed_uuid_session_failed`
- `test_wfm_active_does_not_protect_non_dispatcher_typed_sessions`
- `test_wfm_active_skips_dispatcher_typed_in_confirmed_list_too`

I confirmed this by running the test file: pytest collects and runs only the 5 tests from the second class. The 4 tests in the first class never execute.

The fix is to rename the first class to something distinct (e.g. `TestGuard2WfmActiveDispatcherTypeMarkFailed`) — it tests behavior in `mark_failed_all_ghosts` whereas the second class tests via the WFM-active signal file directly.

---

## Issue 2 (BLOCKING): `get_skill_context_for_message` registered as an MCP tool twice

**File:** `src/mcp/inbox_server.py` (lines 3721–3752 in the modified file)

The tool `get_skill_context_for_message` appears in the tool list returned to the MCP client **twice**, with different descriptions. One describes it as "always-mode + triggered-mode" filtering; the other describes it as handling "contextual" skills. These reflect two different design iterations that were both left in. The dispatch handler correctly routes to a single implementation, but presenting duplicate tool names to the model (and to the MCP protocol) is ambiguous and may cause the client to reject the schema or use the wrong description.

Fix: remove one of the two `Tool(name="get_skill_context_for_message", ...)` blocks. The first one (lines 3721–3738, describing "always-mode skills plus triggered-mode skills whose trigger keywords appear in message_text") matches the actual implementation in `skill_manager.py`, which deliberately excludes always-skills from per-message context. Keep that one; remove the second.

---

- The `_parse_admin_chat_id` function doesn't handle quoted values (e.g. `LOBSTER_ADMIN_CHAT_ID="12345"`). Shell config files sometimes quote values. This is not tested and could silently inject the literal `"12345"` including quotes into context. Worth a note at minimum.
- The PR title says "stop injecting sys.dispatcher.bootup.md into session context" but the PR includes substantial unrelated work: reminder manager, skill manager extensions, ghost detector guards, dispatcher state stop redesign, inflight work compaction, session file stubs, and more. The PR is very large (~24K lines) for what the title describes. No functional objection — just noting the scope mismatch.

---

**The two blocking issues are in new test code and new MCP registration code, not in the core fix. The core change to `inject-bootup-context.py` itself is correct.**

sayhar:
Closing in favor of #2071. Sahar's decision (2026-05-10): the dispatcher bootup file should be injected for dispatchers, not removed. The new approach adds UUID-based detection as a secondary signal alongside the existing PID flag, making injection conditional (dispatcher-only) rather than removing it entirely.

PR #2047, opened 2026-05-09
feat(#2003): IPC bridge for hook-to-EventBus event emission

### LINKED ISSUE #2003: IPC bridge: allow hooks (subprocesses) to emit events into the in-process EventBus
## Problem

Hooks (PreToolUse, PostToolUse, Stop, SessionStart) run as subprocesses — they are shell scripts or Python scripts invoked by Claude Code, not part of the MCP server process. This means they cannot call `get_event_bus().emit(...)` because the EventBus singleton only exists in the MCP server's memory space.

Currently hooks either:
- Write to their own files directly (heartbeat files, state machine files), or
- Write to `logs/events.jsonl` directly, which lands in the right file but bypasses all in-memory listeners (MetricsListener, CriticalAlertListener, TelegramOutboxListener)

## Solution

Add a lightweight IPC bridge: a Unix domain socket (or local HTTP endpoint) on the MCP server that accepts `LobsterEvent`-formatted JSON. When a hook wants to emit an event, it connects to the socket and sends the JSON payload. The MCP server receives it and calls `get_event_bus().emit(event)`, routing through all registered listeners.

Hook-side wrapper:
```python
def emit_to_bus(event_dict: dict) -> None:
    """Fire-and-forget: send an event to the MCP server's EventBus via IPC."""
    try:
        import socket, json
        sock_path = Path(os.environ.get("LOBSTER_WORKSPACE", "~/lobster-workspace")) / "run" / "event-bus.sock"
        with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as s:
            s.settimeout(0.5)
            s.connect(str(sock_path))
            s.sendall(json.dumps(event_dict).encode() + b"\n")
    except Exception:
        pass  # Never interrupt a hook
```

## Context

- EventBus was introduced in issues #890, #891, #892
- Issue #1665 added CRITICAL severity and CriticalAlertListener
- This issue was identified during the PR #1979 (session.end) redesign, where the Stop hook writes a session-end event
- The current PR #1979 workaround writes directly to events.jsonl (right file, bypasses in-memory listeners)
- This IPC bridge is the proper fix that makes the existing workaround unnecessary

## Impact

Once this is implemented, any hook can emit typed, structured events that trigger metrics, alerts, and Telegram notifications — exactly the same way in-process code does. The `emit_event` MCP tool (for dispatcher use) would remain; this is the hook-process equivalent.

sayhar:
🤖🦞 Lobster (engineer): IPC bridge so hooks can emit events into the in-process EventBus

## Problem

Hooks (PreToolUse, PostToolUse, Stop, SessionStart) run as subprocesses — they cannot call `get_event_bus().emit()` because the EventBus singleton only exists in the MCP server's memory space. The existing workaround in `dispatcher-state-stop.py` writes directly to `events.jsonl`, which reaches the right file but **bypasses all in-memory listeners** (MetricsListener, CriticalAlertListener, TelegramOutboxListener). This means hook-emitted events are invisible to metrics, Telegram alerts, and the real-time fan-out that in-process code relies on.

## What this adds

A Unix domain socket server (`EventBusIPCServer`) starts in the MCP server process alongside `init_event_bus()`. When a hook calls `emit_to_bus(event_dict)`, it connects to the socket, sends the LobsterEvent JSON, and disconnects. The server receives it and calls `get_event_bus().emit_sync(event)` — routing through all registered listeners exactly the same way in-process code does.

Before this PR:
```
hook subprocess → write directly to events.jsonl (bypasses listeners)
```

After this PR:
```
hook subprocess → emit_to_bus() → Unix socket → EventBusIPCServer 
  → get_event_bus().emit_sync() → [JsonlFileListener, MetricsListener,
    CriticalAlertListener, TelegramOutboxListener, ...]
```

## Components

**`src/mcp/event_bus_ipc.py`** (new) — IPC bridge:
- `EventBusIPCServer`: asyncio Unix socket server. Accepts one JSON line per connection, deserializes to `LobsterEvent`, calls `bus.emit_sync()`. Malformed JSON, unknown severity, and missing required fields are all silently dropped — a bad hook payload cannot crash the server.
- `emit_to_bus(event_dict, sock_path?)`: hook-side fire-and-forget helper. 0.5s timeout, never raises. Used by both this module and `hooks/emit_to_bus.py`.
- `start_ipc_server()`: module-level idempotent entry point, called from `main()`.
- `_parse_event()`: pure function — no I/O, validates and constructs a `LobsterEvent`.
- `_default_sock_path()`: pure function — reads only `LOBSTER_WORKSPACE` env var.

**`hooks/emit_to_bus.py`** (new) — standalone hook-side helper:
- Same `emit_to_bus()` interface. Self-contained: no imports from `src/mcp/` required, so any hook can use it without adding MCP server dependencies.

**`src/mcp/inbox_server.py`** — one block added in `main()` after `init_event_bus()`:
```python
from event_bus_ipc import start_ipc_server
await start_ipc_server()
```
Failure is caught and logged as a warning (non-fatal).

**`scripts/upgrade.sh` + `scripts/next-migration.txt`** — Migration 91 creates `$LOBSTER_WORKSPACE/run/` on existing installs. Also added `run/` to `create_new_directories()` for fresh installs.

## Functional patterns

- Pure functions: `_parse_event()`, `_default_sock_path()`, `emit_to_bus()` (from caller's perspective)
- Immutability: `LobsterEvent` is a frozen dataclass; events are never mutated after creation
- Side effects isolated at boundaries: `EventBusIPCServer.start()` is the only I/O entry point
- Fire-and-forget at the hook boundary: `emit_to_bus()` absorbs all exceptions

## What is out of scope

- This PR does not migrate `dispatcher-state-stop.py` to use the IPC bridge instead of writing directly to `events.jsonl`. That migration is a follow-up (the direct write remains correct as a file-layer fallback).
- The `emit_event` MCP tool (for dispatcher use) is unchanged.

## Tests run

- [x] `uv run pytest tests/unit/test_event_bus_ipc.py -v` — 17 passed, 0 failed
- [x] `uv run pytest tests/unit/test_event_bus.py tests/unit/test_event_bus_critical.py tests/unit/test_eventbus_wiring.py -v` — 65 passed, 0 failed
- [x] `uv run pytest tests/unit/ --tb=short -q` — 3280 passed, 31 skipped, 0 failed

## Manual test

N/A — this change is internal to the MCP server and hook subprocess communication. The IPC server starts lazily and creates the socket file at `$LOBSTER_WORKSPACE/run/event-bus.sock`. The unit tests cover the full send/receive flow, including connection to a live asyncio server. There is no user-visible output and no external service calls.

## Definition of Done
- [x] Integration/manual test run — not applicable (internal server component, full flow covered by unit tests)
- [x] User-visible outcome verified — not applicable (this is infrastructure; no user-facing output)
- [x] Side-effect audit complete: IPC server only binds to a local Unix socket; no flood risk, no shared-state writes beyond the existing EventBus listeners

Closes #2003


aeschylus:
@sayhar flagging this as potentially disruptive. This introduces a Unix domain socket server (EventBusIPCServer) as new infrastructure — a new failure mode if the socket server dies or the socket file gets stale. Before merging, can you document: (1) what happens when the IPC server is down, (2) how the socket file is cleaned up on restart, and (3) whether this is a hard dependency or gracefully degraded?

aeschylus:
Closing as stale. Opened May 8, functionally untouched since (CONFLICTING against current main, no real commits since the initial Jul 17 bot timestamp sweep). This IPC bridge exists to feed the EventBus redesign this batch was pursuing for session-handoff (see #2048, #1979 — both closed as superseded, since the underlying problems they targeted were already resolved on main via a simpler, already-merged mechanism). Issue #2003 remains open if a hook-to-EventBus IPC bridge is still wanted for other purposes, but this specific implementation should be recreated fresh if so, not rebased through 3 months of drift.

PR #2057, opened 2026-05-09
feat: remove wind-down mode and 70% context threshold

### LINKED ISSUE #2056: feat: remove wind-down mode and 70% context threshold
## Problem

Wind-down mode is indistinguishable from a crash or hang from the outside. When the context monitor fires at 70% and the dispatcher enters wind-down mode, there is no external signal differentiating \"intentionally quiet, about to compact\" from \"stuck\" or \"dead\". This creates operational confusion.

A worse problem: the 70% threshold races with CC's own compaction threshold, which is also around 70%. After 11 compactions on May 8 alone, this race made the system unreliable — both the context monitor and CC attempted to handle the same context overflow event independently, causing undefined behavior.

**Root cause:** The wind-down mechanism was designed before CC's own compaction handling was understood. It is now redundant — `on-compact.py` already handles recovery (sends notification, spawns compact-catchup) after each compaction. The dispatcher does not need to prepare for compaction.

## What to remove

- **`hooks/context-monitor.py`**: The 70% `WARNING_THRESHOLD` check, the `_build_warning_message()` call, the `_write_warning_to_inbox()` call, the `_write_winding_down()` helper and its lazy state-machine import, and the `DEDUP_FLAG` logic. Keep: the token count logging (the PR #2013 feature) and all the transcript-reading/model-lookup functions.

- **`.claude/sys.dispatcher.bootup.md`**: The entire `context_warning` handler block (lines ~796–821), the `WIND_DOWN_MODE` state variable and all references to it, the inline tombstone instruction under \"On context_warning\" (line ~1113), and the `session.end` EventBus emit in step 5 of that handler (the only caller of `emit_event(\"session.end\")` — if the handler is removed, this goes away too).

## What to keep (not in scope)

- `on-compact.py` SessionStart hook
- Compact-catchup agent
- Health-check grace periods after compaction
- Post-compact gate
- Token count logging in `context-monitor.py`

## Why this is the right fix

CC compacts on its own terms. The `on-compact.py` hook fires immediately after each compaction and handles recovery. The dispatcher does not need to orchestrate a graceful wind-down — that was defensive code for a scenario that CC already handles. Removing it eliminates a race condition and one source of indistinguishable silence.

## Acceptance criteria

- [ ] `hooks/context-monitor.py` logs token counts but never writes a `context_warning` inbox message and never calls `_write_winding_down()`
- [ ] `.claude/sys.dispatcher.bootup.md` contains no `context_warning` handler and no `WIND_DOWN_MODE` references
- [ ] Tests updated to reflect new behavior (no threshold-triggered inbox writes, `_write_winding_down` removed)
- [ ] Existing token-count logging tests pass unchanged

sayhar:
Closes #2056

## Why this exists

The 70% context threshold in `context-monitor.py` raced with CC's own compaction threshold (also ~70%), causing 11+ compactions on May 8 alone. Wind-down mode also created an indistinguishable silence state — from outside the process, there was no way to tell whether Lobster was intentionally winding down or crashed.

The core design assumption behind wind-down was wrong: the dispatcher does not need to prepare for compaction. `on-compact.py` already fires after each compaction and handles recovery (notification, compact-catchup). The 70% hook and the dispatcher's preparation for it were redundant defensive code that introduced more problems than they solved.

## What changed

**`hooks/context-monitor.py`** — Removed the threshold check entirely. Removed `WARNING_THRESHOLD`, `DEDUP_FLAG`, `_write_winding_down()`, `_build_warning_message()`, and `_write_warning_to_inbox()`. The hook now logs token usage on every tool call and returns immediately — no inbox messages, no state machine transitions. The 4-tuple return from `_read_transcript_usage()` (with `total_tokens`, from PR #2013) is preserved.

**`.claude/sys.dispatcher.bootup.md`** — Removed the `context_warning` message handler (26 lines), all `WIND_DOWN_MODE` references, and the "On context_warning" tombstone instruction. Updated the tombstone `End reason` list to replace `graceful wind-down`/`context_warning` with `compaction`. The `session_note_reminder` handler no longer has a `WIND_DOWN_MODE` guard.

**`tests/unit/test_hooks/test_context_monitor.py`** — Replaced wind-down-specific tests (`TestWindingDownStateTransition`, `TestTokenCountDisplay`, threshold-triggered inbox write tests) with:
- `TestHandlePayloadLogging`: verifies usage is logged at any level, inbox stays empty at 80% and 100%, warn on absent transcript
- `TestWindDownRemoved`: structural tests asserting that `WARNING_THRESHOLD`, `DEDUP_FLAG`, `_write_winding_down`, `_build_warning_message`, and `_write_warning_to_inbox` are absent from the module

## What was not touched

- `on-compact.py` SessionStart hook
- Compact-catchup agent
- Health-check grace periods after compaction
- Post-compact gate
- `session_note_reminder` handler logic (only removed the `WIND_DOWN_MODE` guard line)
- `session.end` event in `events.jsonl` (startup recovery check at step 2c still reads this — the only emitter was the context_warning handler, so this check will produce "no recent session.end" after this change, which is correct behavior)

## State transition: before vs after

```
Before:
  Token usage checked after every tool call
  |
  >= 70% threshold?
    Yes -> write context_warning to inbox -> dispatcher enters WIND_DOWN_MODE
             -> stops spawning, drains agents, writes tombstone
             -> waits for CC compaction (which also triggers at ~70%)
             -> RACE: both CC and dispatcher acting on same context overflow

After:
  Token usage logged after every tool call (no threshold, no action taken)
  |
  CC compacts when it decides to
  |
  on-compact.py fires -> sends notification -> spawns compact-catchup
```

## Functional patterns used

Pure functions throughout (`_read_transcript_usage`, `_build_log_entry`, `_build_absent_context_entry` — all return immutable dicts from inputs, no side effects). Side effects isolated to `_log_usage` and `_handle_payload`. Injectable parameters (`log_dir`, `inbox_dir`) preserved for testability.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_context_monitor.py -v` — 29 passed, 0 failed (red phase: 5 failures before implementation; green phase: all 29 pass after)
- [x] `uv run pytest tests/unit/ -q` — 3264 passed, 1 pre-existing failure in `test_bisque/test_relay_server.py::TestWSAuth::test_auth_success` (test isolation issue when run in full suite; passes in isolation on both main and this branch — unrelated to this change)

## Manual test

N/A — this change is entirely internal. The hook still fires on every tool call and logs to `context-monitor.log`; the only behavioral change is that it no longer writes inbox messages or transitions dispatcher state. No external service calls, no Telegram output, no DB schema changes.

## Definition of Done
- [x] Integration/manual test run (N/A — internal change only)
- [x] User-visible outcome verified (N/A — removal of a silent internal state)
- [x] Side-effect audit complete: no floods, no unscoped writes — change removes writes, does not add them

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -681,39 +681,9 @@ Rules:
 
 ---
 
-### context_warning (`type: "context_warning"`)
-
-Written by `hooks/context-monitor.py` when context window >= 70%.
-
-```
-1. mark_processing(message_id)
-2. Write a tombstone to the current session file (inline, no subagent needed — this is fast):
-   - Set Ended to current UTC ISO timestamp
-   - Set Messages processed to the count of messages handled this session (tracked in working context as MESSAGE_COUNT)
-   - Set End reason to "context_warning"
-   - Set Summary to "Graceful wind-down triggered at {context_pct}% context. [Brief list of what was in progress, if anything.]"
-   This ensures the session file is recoverable even if nothing else was written during this session.
-3. Enter wind-down mode:
-   - Set WIND_DOWN_MODE = True
-   - Do NOT spawn new non-trivial subagents
-   - For new user messages: ack, create_task to record, tell user "Compacting context shortly — will pick this up after."
-4. Drain in-flight agents: poll get_active_sessions() every 10s. Process arriving subagent results normally.
-5. Write ~/lobster-workspace/data/context-handoff.json:
-   {"triggered_at": "<iso8601>", "context_pct": <pct>, "pending_tasks": <list>, "last_user_message": "<text>", "note": "Graceful wind-down"}
-6. Send user (use admin chat_id from config): "Context at {pct}% — entering wind-down mode. Handing off cleanly."
-7. Do NOT call wait_for_messages() again. Do not attempt to self-terminate — the dispatcher cannot exit itself. Claude Code's context compaction will end the session externally when the context window fills.
-8. mark_processed(message_id)
-```
-
-Rules: `chat_id` is 0 — use admin chat_id for step 5. Never re-enter wind-down for a second warning. Do NOT call `lobster restart` — compaction is the recovery mechanism.
-
----
-
 ### session_note_reminder (`type: "session_note_reminder"`)
 
 Injected by the MCP server after every 20 real user messages. Spawn session-note-appender in the background; mark_processed silently (no reply).
-
-Do NOT spawn during wind-down mode (`WIND_DOWN_MODE = True`) — session-note-polish handles the final consolidation.
 
 ```
 1. mark_processing(message_id)
@@ -972,7 +942,7 @@ Do not modify Summary or Started/Ended. 3. Write back. 4. Call write_result.
 **Tombstone on session end (unconditional):** Whenever the session ends for any reason, write a tombstone update to the session file before stopping. This is done inline (not via subagent) and takes <1 second. Minimum content:
 - `Ended`: current UTC ISO timestamp
 - `Messages processed`: MESSAGE_COUNT (tracked in working context; increment on each `mark_processed` call)
-- `End reason`: one of `graceful wind-down`, `context_warning`, `short session`, `crash` (use `short session` if session ran < 5 minutes and no reason is known)
+- `End reason`: one of `compaction`, `short session`, `crash` (use `short session` if session ran < 5 minutes and no reason is known)
 - `Summary`: at minimum, "Session ended [reason]. [N] messages processed." — fill in more if context permits.
 
 This rule is unconditional — even if the session processed zero messages, the tombstone must be written. A stub file with only a start timestamp is nearly as bad as no file at all.
@@ -985,8 +955,6 @@ This rule is unconditional — even if the session processed zero messages, the 
 - All currently in-flight subagents (task_id, subagent type, brief description, and elapsed time since started_at) — these are the entries most at risk of being lost across compaction
 - Any pending user responses (messages that were mark_processing-d but not yet replied to)
 - The current MESSAGE_COUNT at time of compaction
-
-**On context_warning:** Write a tombstone inline as step 2 (see context_warning handler above) — this is faster and more reliable than spawning a subagent, and ensures the record survives even if wind-down is interrupted.
 
 ---
 
--- a/hooks/context-monitor.py
+++ b/hooks/context-monitor.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-PostToolUse hook: context window guard.
+PostToolUse hook: context window monitor.
 
 Reads actual token usage from the session's transcript JSONL file rather than
 relying on the `context_window` field in the PostToolUse payload — which Claude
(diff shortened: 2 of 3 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with follow-up items noted

## Verdict

The removal is clean and correct. The hook is now a pure logger, the dispatcher protocol no longer references wind-down concepts, and the tests verify both the new behavior and the absence of old artifacts. PASS.

---

**What could go wrong:**
- Startup protocol references a file that is no longer written
- Startup protocol references an event log that no longer has a writer for the relevant event type
- Dead instructions in the dispatcher doc cause the dispatcher to waste time checking things that never exist

**Edge cases:**
- A `context-handoff.json` left over from a previous install (pre-PR) could trigger a spurious "Resuming from where we left off" notification on the first restart after deploy — the 10-minute staleness check mitigates this but doesn't eliminate it
- The `total_tokens` field added to the return tuple is unused beyond logging — not a bug, just worth noting

**What I would want tested:**
- That inbox stays empty at high context levels (covered)
- That removed symbols are truly absent (covered)
- That normal logging still works (covered)

---

## Findings

### 1. Dead startup check — context-handoff.json (NOT flagged by engineer)

**Severity: Low / follow-up candidate**

Step 2c of the startup protocol instructs the dispatcher to check `~/lobster-workspace/data/context-handoff.json` and, if recent, notify the user that it is resuming after a context overrun. That file was written exclusively by the wind-down handler (step 5 of the `context_warning` handler). With wind-down mode removed, nothing writes this file anymore.

The check is now permanently a no-op. It will never find a recent file and will always take the "stale or absent → ignore" branch. The skip condition in step 2d ("Skip if step 2c already sent a restart notification") is also permanently dead.

This is harmless at runtime — the startup path just reads a missing file and moves on. A stale file from a pre-PR install could trigger a spurious "Resuming" message once, then the file is deleted and the issue self-resolves.

**Recommended follow-up:** Remove or replace step 2c in `sys.dispatcher.bootup.md` in a subsequent PR. The compaction startup cause (`startup-cause: compaction` banner + `compact-reminder` inbox message) now handles the post-compaction recovery path; step 2c was the wind-down-mode analog of that, and it no longer has a purpose.

### 2. events.jsonl session.end check (flagged by engineer, but finding is different)

The engineer flagged that step 2c in the *old* bootup doc checked `events.jsonl` for a recent `session.end` event, and that `context_warning` was the only emitter. This is only half-right: `dispatcher-state-stop.py` independently emits a `session.end` event to `events.jsonl` on every dispatcher session stop, so that file does have a writer. However, the PR *replaces* the events.jsonl check with the context-handoff.json check (see finding 1), so the events.jsonl question is moot for the post-PR state. The new dead check is on context-handoff.json, not events.jsonl.

### 3. Test count discrepancy (minor)

The engineer reported "3264 full suite pass, 1 pre-existing isolation failure in test_bisque." Running `uv run pytest tests/unit/ --tb=no -q` on the PR branch gives **3265 passed, 0 failures** — no bisque failure observed. The count is off by one and the failure did not reproduce. This is a negligible discrepancy; the suite is clean.

### 4. inbox_dir parameter retained in production code (acknowledged, no action needed)

`_handle_payload()` still accepts `inbox_dir` as a parameter for test compatibility but never uses it. The engineer acknowledged this with a comment. Keeping the signature stable is correct — removing it would break existing test call sites. No action needed.

---

## Summary

Two issues worth tracking: (1) the startup protocol has a permanent dead check that should be cleaned up in a follow-up PR, and (2) the engineer's test count claim was slightly off. Neither blocks merge. The core removal — hook stripped to pure logging, dispatcher doc updated, old tests replaced with new coverage — is well-executed.

PR #2060, opened 2026-05-09
fix: graceful proactive restart at 7200s to avoid hard CC session limit

### LINKED ISSUE #2059: fix: graceful proactive restart before 7440s CC session lifetime limit
## Problem

Claude Code enforces a hard 7440-second (124-minute) session lifetime limit. When this limit is hit, CC exits immediately with no Stop hook, no compaction alert, and no graceful handoff. The dispatcher dies silently, systemd restarts it, and Alex sees "Lobster Starting" with no prior notification that anything happened.

## Evidence

Sessions 016 and 017 both died at exactly start+7433–7436s (within ±10s of 7440s). The pattern is deterministic: every dispatcher session is killed at start+7440s by the CC runtime.

**What CC's hard limit kills:**
- Stop hook never fires → no session tombstone written
- Compact-catchup never triggers → context continuity is broken
- Health check sees process vanish and triggers a "recovery" restart on top of an already-restarting systemd unit → duplicate restart noise

## Proposed Fix

Detect when the session is approaching the 7440s limit and trigger a **graceful proactive restart** at ~7200s (2 minutes before the hard limit). A graceful restart fires the Stop hook, writes the session tombstone, and allows compact-catchup — none of which happen with the hard CC exit.

### Implementation (Option A — health check enhancement)

The health check already runs every 4 minutes (`*/4 * * * *`). We add two small pieces:

1. **`inject-bootup-context.py` change:** On dispatcher session start, write a Unix epoch timestamp to `~/lobster-workspace/data/dispatcher-session-start.ts` (plain integer, atomic rename). This is the ground truth for session age.

2. **`health-check-v3.sh` change:** Add a `check_session_age()` function that:
   - Reads `dispatcher-session-start.ts`
   - Computes `now - session_start`
   - If `session_age >= SESSION_AGE_LIMIT_SECONDS` (7200s), sends SIGTERM to the PID in `dispatcher.pid` — the same PID the health check already knows about
   - Logs the action and notifies Telegram (deduped, "proactive-session-restart" key)
   - Suppressed during maintenance mode (flag already present) and during the boot grace period

The SIGTERM triggers the dispatcher's Stop hook (graceful exit) rather than the hard CC kill. The existing outer restart loop in `claude-persistent.sh` handles the restart after Stop completes.

## Session Data (Evidence)

| Session | Started | Died | Age | CC exit? |
|---------|---------|------|-----|----------|
| 016 | 2026-05-09T08:08:04Z | ~2026-05-09T10:12:04Z | ~7440s | Yes, no Stop hook |
| 017 | 2026-05-09T10:12:05Z | ~2026-05-09T12:16:05Z | ~7440s | Yes, no Stop hook |

Both sessions died at exactly start+7440s ±10s, confirming this is a deterministic CC runtime limit, not a random crash.

## What This Does NOT Change

- The health check's existing restart logic (stale inbox, heartbeat, etc.) is unchanged
- The `do_restart()` function (systemctl-based) is not used for this path — SIGTERM to the PID is used instead, so the Stop hook fires
- No new cron entries required — enhances the existing 4-minute health check

sayhar:
Closes #2059

Claude Code enforces a hard 7440-second (124-minute) session lifetime. When this limit hits, CC kills the dispatcher process **without firing the Stop hook** — no session tombstone, no compact-catchup, no user notification. Sahar sees "Lobster Starting" with no warning that anything happened.

Evidence from sessions 016 and 017: both died at exactly start+7433–7436s (within ±10s of 7440s). The pattern is deterministic.

## What this change does

Adds a two-part proactive restart mechanism that triggers a **graceful SIGTERM** at 7200s (2 min before the hard limit), so the Stop hook fires cleanly.

**Part 1 — `hooks/inject-bootup-context.py`:** On every dispatcher `SessionStart`, write the current Unix epoch to `data/dispatcher-session-start.ts` (atomic write via tmp+rename). This is the ground truth for "when did this session begin?"

**Part 2 — `scripts/health-check-v3.sh`:** Add `check_session_age()` which:
- Reads `dispatcher-session-start.ts`
- Computes `now - session_start`
- If `session_age >= SESSION_AGE_LIMIT_SECONDS` (7200s): sends SIGTERM to the PID in `dispatcher.pid`
- Deletes the start file after SIGTERM to prevent double-fire on the next 4-minute run
- Logs the action and sends a deduped Telegram alert (`proactive-session-restart`)
- Returns 1; `main()` exits early so no other checks pile on top of the in-flight restart

`main()` wraps the call: suppressed during boot grace (90s after a restart).

### Before / after flow

```
Before (hard CC limit):
  Session starts → 7440s → CC kills process (no Stop hook)
  systemd sees dead process → restarts
  User sees "Lobster Starting" with no context

After (proactive SIGTERM at 7200s):
  Session starts → 7200s → health check sends SIGTERM to dispatcher PID
  Stop hook fires → tombstone written → compact-catchup triggers
  claude-persistent.sh restarts Claude automatically
  User sees "Lobster Starting" after graceful handoff
```

## What this does NOT change

- `do_restart()` (systemctl-based) is not used for this path — SIGTERM directly to the PID is cleaner and fires the Stop hook
- No new cron entries — enhances the existing 4-minute health check
- All existing health check logic (stale inbox, heartbeat, outbox, auth) is unchanged

## Functional patterns used

Pure functions for timestamp reading (`int(time.time())`), named constants for all spec values (`SESSION_AGE_LIMIT_SECONDS=7200`), atomic file writes via `tmp+rename`, silent-on-error boundaries at both the hook level and health check level.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_dispatcher_session_start.py -v` — 8 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 437 passed, 7 skipped, 0 failed (no regressions in existing hook tests)
- [x] `bash tests/test-health-check-session-age.sh` — 11 passed, 0 failed (covers: no file, young session, at-limit, past-limit, SIGTERM sent, no PID file, malformed timestamp, empty file, start file deleted after SIGTERM, dead PID fallback, Telegram alert fires)

## Manual test

N/A — this change is not testable without a live 120-minute session. The SIGTERM path is exercised by the shell tests using real `sleep` background processes. The integration test for the full flow would require waiting for a real session to approach the 7440s limit, which is impractical in a PR test run.

The health check (`*/4 * * * *`) will pick up the change on the next deploy without any migration needed — no new cron entries, no new directories, no DB changes.

## Definition of Done

- [ ] Integration/manual test run — not applicable (production-only observable; no live 120-min session available)
- [ ] User-visible outcome: Sahar will see a Telegram notification on proactive restart instead of silent "Lobster Starting" — not yet verified (requires production)
- [x] Side-effect audit: SIGTERM to one specific PID (from dispatcher.pid); start file deleted after SIGTERM; no floods, no mass writes, no unscoped operations
- [x] External service calls: none in automated tests; Telegram alert uses deduped send (at most once per 15 minutes)

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -76,6 +76,18 @@ STARTUP_CAUSE_FILE = Path(
 # compaction that was followed by an unrelated external restart.
 COMPACTION_CAUSE_WINDOW_SECONDS = 300  # 5 minutes
 
+# Dispatcher session start timestamp file (issue #2059).
+# Written by this hook on every dispatcher SessionStart as a plain Unix epoch
+# integer (seconds). Read by health-check-v3.sh to compute session age and
+# trigger a proactive restart before the hard 7440s CC session lifetime limit.
+# Override via env var for test isolation.
+DISPATCHER_SESSION_START_FILE = Path(
+    os.environ.get(
+        "LOBSTER_DISPATCHER_SESSION_START_FILE_OVERRIDE",
+        str(_LOBSTER_WORKSPACE / "data" / "dispatcher-session-start.ts"),
+    )
+)
+
 
 def _is_startup_flag_dispatcher() -> bool:
     """Return True if a live startup flag marks this as the dispatcher session.
@@ -118,6 +130,27 @@ def _consume_startup_flag() -> None:
     try:
         STARTUP_FLAG_FILE.unlink(missing_ok=True)
     except OSError:
+        pass
+
+
+def _write_dispatcher_session_start() -> None:
+    """Write the current Unix epoch to DISPATCHER_SESSION_START_FILE.
+
+    Called once per dispatcher SessionStart. The health check reads this file
+    to compute session age and trigger a proactive restart before the hard
+    7440s CC session lifetime limit (issue #2059).
+
+    Uses an atomic rename so the reader never sees a partial write.
+    Silent on any error — must never crash the hook.
+    """
+    try:
+        now_epoch = int(time.time())
+        path = DISPATCHER_SESSION_START_FILE
+        path.parent.mkdir(parents=True, exist_ok=True)
+        tmp_path = path.with_suffix(".tmp")
+        tmp_path.write_text(str(now_epoch) + "\n")
+        tmp_path.replace(path)  # atomic on Linux
+    except Exception:  # noqa: BLE001
         pass
 
 
@@ -272,6 +305,14 @@ def main() -> None:
         _consume_startup_flag()
         print(
             f"[{HOOK_NAME}] startup-flag detected live PID — injecting dispatcher bootup",
+            file=sys.stderr,
+        )
+        # Record session start time for health-check proactive restart (issue #2059).
+        # Plain Unix epoch written atomically; health-check-v3.sh reads it to detect
+        # sessions approaching the 7440s CC hard limit and send SIGTERM early.
+        _write_dispatcher_session_start()
+        print(
+            f"[{HOOK_NAME}] wrote dispatcher session start timestamp",
             file=sys.stderr,
         )
 
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -100,6 +100,14 @@ HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"   # legacy WFM-touch signa
 DISPATCHER_HEARTBEAT_FILE="${LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-heartbeat}"
 DISPATCHER_HEARTBEAT_STALE_SECONDS=1200   # 20 min — covers compaction (~5m) + catchup (~12m) + margin
 
+# Session age limit: CC enforces a hard 7440s session lifetime (issue #2059).
+# At this limit, CC kills the process without firing the Stop hook — no tombstone,
+# no compaction alert. We trigger a graceful SIGTERM at SESSION_AGE_LIMIT_SECONDS
+# (7200s = 2 min before hard limit) so the Stop hook fires cleanly.
+# Written by inject-bootup-context.py as a plain Unix epoch integer.
+SESSION_AGE_LIMIT_SECONDS="${LOBSTER_SESSION_AGE_LIMIT_SECONDS:-7200}"
+DISPATCHER_SESSION_START_FILE="${LOBSTER_DISPATCHER_SESSION_START_FILE_OVERRIDE:-$WORKSPACE_DIR/data/dispatcher-session-start.ts}"
+
 # WFM-active signal (issue #1713 / #949): inbox_server.py writes this file with
 # a Unix epoch timestamp when wait_for_messages begins blocking and refreshes it
 # every WAIT_HEARTBEAT_INTERVAL (60s). When this file is fresh, the dispatcher is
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS

## What I checked

**All 8 Python unit tests and 11 bash integration tests pass.** The 42 collection errors are pre-existing (confirmed by verifying against the main branch baseline — same count before and after).

---

## Cold read — potential issues to investigate

Before reading the engineer's brief, I flagged these concerns. All are addressed:

**1. `$?` pattern after `check_session_age()`**
Lines 1726–1729: `check_session_age` at 1726, comment at 1727–1728, `if [[ $? -eq 1 ]]` at 1729. No intervening command. Clean.

**2. Double-fire if start file deletion fails**
`rm -f "$DISPATCHER_SESSION_START_FILE" 2>/dev/null || true` — if the `rm` fails for any reason, the old timestamp persists. On the next 4-minute run, the code would attempt another SIGTERM. However: the dying dispatcher's PID is checked with `kill -0` first, which will fail since the process is restarting — so no second SIGTERM to a live process. When the new session starts, `inject-bootup-context.py` overwrites the timestamp file immediately, resetting the clock correctly. Acceptable mitigation.

**3. Race: health check fires at 7199s, then at ~7439s**
With a 4-minute interval, consecutive runs land at 7199s then 7439s (not 7203s). At 7199s: `session_age < 7200` → returns 0. At 7439s: file already deleted by SIGTERM at the 7240s run, so `check_session_age` returns 0 immediately (missing file path). Clean.

**4. 90s boot grace window**
`BOOT_GRACE_SECONDS=90`. New session writes the start timestamp, then the first health-check run within 90s of boot is suppressed. This correctly prevents a stale timestamp from a crashed previous session from triggering a false SIGTERM on the new session's first minutes. But even if the grace window were absent, the stale timestamp would get overwritten by `inject-bootup-context.py` on the new SessionStart — so the primary defense is the overwrite, and the grace period is defense-in-depth. Solid.

**5. Stale PID file**
If `dispatcher.pid` is stale (dead PID), `kill -0 "$dispatcher_pid"` fails → `log_warn` + `return 0`. No action. Safe fallback.

**6. First boot / missing start file**
`[[ ! -f "$DISPATCHER_SESSION_START_FILE" ]]` → `return 0`. Clean.

**7. SIGTERM vs systemctl**
`kill -TERM "$dispatcher_pid"` hits the claude process directly, which fires the Stop hook. `systemctl restart` would bypass the Stop hook. Choice is correct for the stated goal.

**8. Subagent context**
`_write_dispatcher_session_start()` is only called when `is_dispatcher` is true (i.e., startup flag is present with a live PID). Subagents don't write the file. Test `test_session_start_NOT_written_for_subagent` confirms this. No cross-contamination.

---

- `session_start=$(< "$DISPATCHER_SESSION_START_FILE") 2>/dev/null || true` — the `2>/dev/null` placement redirects stderr of the assignment statement. Bash handles this correctly; the redirect applies to the shell's file-open operation inside `$(< ...)`. Verified with a manual test.

- `send_telegram_alert_deduped "proactive-session-restart" ...` uses a 15-minute dedup cooldown (`ALERT_DEDUP_COOLDOWN_SECONDS=900`). Since a session restart takes well under 15 minutes, this means only one alert per restart event. Correct behavior.

- No test exercises the case where `kill -TERM` returns non-zero (process exited between `kill -0` check and `kill -TERM`). This is a tiny TOCTOU window, and the `else` branch returns 0 (no action) — correct. The start file is not deleted in this case, so the next health-check run will re-attempt. This is the right behavior.

---

## Verdict

The fix is correct. The timestamp-write is atomic, the SIGTERM path is guarded by three independent checks (file present, PID numeric, PID alive), double-fire is prevented by start file deletion, and the boot grace window prevents false positives on new sessions. Tests are comprehensive and all pass.

sayhar:
🤖🦞 Lobster: Closing pending root cause investigation. The fix was based on observed sessions dying at ~7440s, but this pattern needs more rigorous verification before we build infrastructure around it. Will reopen once the investigation confirms the root cause.

sayhar:
🤖🦞 Lobster: Reopening — investigation confirmed the 7440s limit is real. 6 plain-restart sessions measured: all ended within ±4s of 7440s. Compaction-ended sessions show no such clustering (27–115 min). The codebase already encoded 7440 as a known limit in health-check scripts. PR is available for sign-off — Sahar to decide on approach.

PR #2068, opened 2026-05-10
feat(hooks): port require-reply-to-message-id.py to main to enforce Telegram reply threading

### LINKED ISSUE #2067: feat(hooks): add require-reply-to-message-id.py to enforce Telegram threading
## Problem

`hooks/require-reply-to-message-id.py` was originally implemented in PR #1541 (commit `7173e745`) to enforce that all Telegram `send_reply` calls include a `reply_to_message_id`, ensuring replies are threaded under the user's originating message rather than sent standalone.

That PR landed on the `local-dev` branch but was never merged to `origin/main`. The file that exists on main today is a no-op stub (just `sys.exit(0)`) with no enforcement logic. The hook is registered in `~/.claude/settings.json` (the live settings file), so it fires on every `send_reply`, but does nothing.

The consequence: the Telegram threading requirement documented in CLAUDE.md is unenforced — Claude can send standalone (unthreaded) replies without being blocked.

## What the hook does

`require-reply-to-message-id.py` is a `PreToolUse` hook that blocks `mcp__lobster-inbox__send_reply` calls where:
- `source` is `telegram` (or absent, which defaults to telegram per API contract)
- `chat_id != 0` (non-zero means a real user message exists to thread against)
- `reply_to_message_id` is absent or not a positive integer

Exemptions (always allowed without `reply_to_message_id`):
- Non-Telegram sources (slack, sms, whatsapp, system, bot-talk, etc.)
- Proactive/system sends where `chat_id == 0` (no originating user message)

On block, it exits with code 2 and prints an actionable stderr message explaining what's missing and how to exempt proactive sends.

## Scope

- Port the hook implementation from commit `7173e745` to main
- Port the 11 unit tests from the same commit (`tests/unit/test_hooks/test_require_reply_to_message_id.py`)
- Add hook registration to `install.sh` (so fresh installs get it)
- Add the hook file to main (it is absent — the stub on disk is not committed)

The hook registration in `~/.claude/settings.json` is already present (no change needed to settings).

## Original context

- PR #1541 — original implementation, landed on local-dev only
- Issue #1168 — the Telegram threading requirement this hook enforces

sayhar:
Closes #2067

## Problem

Telegram replies sent without `reply_to_message_id` appear as standalone messages in the user's chat rather than threaded under the message they're responding to. CLAUDE.md documents this requirement, but it was unenforced.

`require-reply-to-message-id.py` was implemented in PR #1541 (commit `7173e745`) to enforce this. That PR landed on `local-dev` but never reached `origin/main`. The file present on main was a no-op stub (`sys.exit(0)`) — registered in `~/.claude/settings.json` but doing nothing.

## What changed

**`hooks/require-reply-to-message-id.py`** — replaced the stub with the real implementation. A `PreToolUse` hook that:
- Intercepts `mcp__lobster-inbox__send_reply` calls
- Blocks any call where `source` is `telegram` (or absent, which defaults to telegram per API contract) and `reply_to_message_id` is missing or not a positive integer
- Exits 2 with an actionable stderr message explaining the requirement and the `chat_id=0` exemption

Exemptions always allowed without `reply_to_message_id`:
- Non-Telegram sources (slack, sms, whatsapp, system, bot-talk, ...)
- Proactive/system sends where `chat_id == 0` (no originating user message to thread against)

**`install.sh`** — adds hook registration for fresh installs. Idempotent: checks for the hook before adding it.

**`scripts/upgrade.sh`** — Migration 94 registers the hook for existing installs that don't already have it. Follows the same pattern as migrations 90–93.

**`tests/unit/test_hooks/test_require_reply_to_message_id.py`** — 11 unit tests covering all branches: allowed calls, blocked calls, all exemptions (non-Telegram, chat_id=0), fail-open on malformed input, case-insensitive source matching, and actionable stderr content.

## Functional patterns

Pure functions: all logic is stateless. Fail-open design: unparseable input exits 0 rather than blocking. Single-responsibility: one hook, one enforcement rule.

## Before / after

Before: hook file on main was `sys.exit(0)`. Every Telegram `send_reply` fired the hook but it immediately exited, doing nothing.

After: hook enforces `reply_to_message_id` on Telegram calls. Non-Telegram calls and proactive sends are unchanged.

No state transitions or flow diagrams needed — this is a pure guard that blocks or allows with no side effects.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_require_reply_to_message_id.py -v` — 11 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/` (broader hook suite) — 129 passed, 0 failed

## Manual test

N/A — this change is entirely internal. The hook fires within the Claude Code harness before any Telegram API call is made. No external service is contacted by the hook itself. The user-visible outcome (threaded vs. standalone replies) requires a live Telegram session; that is tested by the hook being active in `~/.claude/settings.json`, which it already was (the stub was registered). The only change is that the hook now actually enforces the rule instead of silently allowing everything.

## Definition of Done

- [x] Integration/manual test run — N/A (hook is a pure guard, no external calls)
- [x] User-visible outcome verified — not directly verifiable by engineer; hook enforces the prerequisite for correct threading
- [x] Side-effect audit: hook exits 0 or 2 only — no writes, no sends, no shared state

🤖🦞 Lobster (engineer): porting hook that was stuck on local-dev since April

### DIFF
--- a/hooks/require-reply-to-message-id.py
+++ b/hooks/require-reply-to-message-id.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+"""
+PreToolUse hook: enforce reply_to_message_id on Telegram send_reply calls.
+
+Without reply_to_message_id, Telegram replies are sent standalone (not threaded
+to the original message). This hook blocks send_reply calls that target Telegram
+but omit or set reply_to_message_id=0.
+
+Exemptions (always allowed):
+  - Non-Telegram sources (slack, sms, whatsapp, bot-talk, system, ...)
+  - System/proactive sends where chat_id == 0 (no incoming message context)
+  - Calls where reply_to_message_id is already set to a positive integer
+
+Exit codes:
+  0 - Allow the call
+  2 - Block the call (Claude Code shows stderr message to Claude)
+
+Issue: #1168
+"""
+
+import json
+import sys
+
+
+def main() -> None:
+    try:
+        data = json.load(sys.stdin)
+    except (json.JSONDecodeError, ValueError):
+        # Unparseable input — allow rather than block (fail-open)
+        sys.exit(0)
+
+    tool_name = data.get("tool_name", "")
+
+    # Only check send_reply calls
+    if tool_name != "mcp__lobster-inbox__send_reply":
+        sys.exit(0)
+
+    tool_input = data.get("tool_input", {})
+    source = (tool_input.get("source") or "telegram").lower()
+
+    # Only enforce on Telegram (source absent defaults to telegram per API contract)
+    if source != "telegram":
+        sys.exit(0)
+
+    # Exempt proactive/system sends: chat_id == 0 means no originating user message
+    chat_id = tool_input.get("chat_id")
+    try:
+        if int(chat_id) == 0:
+            sys.exit(0)
+    except (TypeError, ValueError):
+        # chat_id absent or non-numeric — cannot determine, allow
+        sys.exit(0)
+
+    # Check reply_to_message_id: must be present and a positive integer
+    reply_to = tool_input.get("reply_to_message_id")
+    try:
+        if reply_to is not None and int(reply_to) > 0:
+            sys.exit(0)  # Properly set — allow
+    except (TypeError, ValueError):
+        pass  # Non-integer value — fall through to block
+
+    # Block: reply_to_message_id is missing or not a positive integer
+    print(
+        "BLOCKED: Telegram send_reply is missing reply_to_message_id.\n"
+        "Every Telegram reply must include reply_to_message_id (the integer\n"
+        "Telegram message ID shown in wait_for_messages output as\n"
+        "'pass as reply_to_message_id') to thread the reply correctly.\n\n"
+        "If this is a proactive/system send with no originating message,\n"
+        "pass chat_id=0 to exempt this check.",
+        file=sys.stderr,
+    )
+    sys.exit(2)
+
+
+if __name__ == "__main__":
+    main()
--- a/install.sh
+++ b/install.sh
@@ -1878,6 +1878,27 @@ setup_claude_hooks
 success "Self-check cron configured (every 3min)"
 
 
+# Set up Claude Code PreToolUse hook to enforce reply_to_message_id on Telegram send_reply (#1168)
+chmod +x "$INSTALL_DIR/hooks/require-reply-to-message-id.py" || true
+if [ -f "$CLAUDE_SETTINGS" ]; then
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: Logic, tests, and install path are all correct. One minor bug in upgrade.sh worth noting.

---

## What I checked

### Hook logic (hooks/require-reply-to-message-id.py)

Correct. Fail-open on unparseable input, source comparison is case-insensitive, chat_id=0 exemption is sound, negative reply_to_message_id is blocked (correct — invalid value), non-Telegram sources pass through, tool_name key matches the convention used by other hooks in the repo. No issues.

### Tests (test_require_reply_to_message_id.py)

All 11 pass. No duplicate class names, no unreachable test classes, no shadowing. Coverage is appropriate — both zero and missing reply_to_message_id are tested, as is the uppercase source normalisation edge case. The full hook test suite runs clean (558 passed, 7 skipped on the PR branch, same as main minus the 11 new tests).

The 42 collection errors in the full test suite are pre-existing infrastructure-level import failures in test_mcp_server/ and test_state_machine.py — not introduced by this PR.

### install.sh idempotency

Not broken. The new block at line 1882 uses `select(.hooks[]?.command | test("require-reply-to-message-id"))` — command-based detection, not matcher-based. The link-checker block at line 1903 uses `select(.matcher == "mcp__lobster-inbox__send_reply")`, which is a different check. However, `setup_claude_hooks()` is called at line 1877 **before** the new block and installs link-checker first (while settings.json has no send_reply matcher yet). By the time the link-checker outer block at line 1903 runs, link-checker is already registered via `setup_claude_hooks()` — so the outer block correctly detects it as present and skips. Both hooks end up installed; nothing is lost.

### Migration 94 in upgrade.sh — minor bug

Migration 94 uses `${INSTALL_DIR:-$HOME/lobster}` as the jq `--arg install_dir` value. Every other migration in upgrade.sh that embeds a hook path uses `$LOBSTER_DIR` (set at line 39 as `${LOBSTER_INSTALL_DIR:-$HOME/lobster}`). The fallback path (`$HOME/lobster`) is identical, so default installs are unaffected. But if someone ran install with a custom `LOBSTER_INSTALL_DIR`, `$INSTALL_DIR` is never set in upgrade.sh's environment, so the fallback fires and the wrong hardcoded path gets embedded in settings.json.

Fix: replace `${INSTALL_DIR:-$HOME/lobster}` with `$LOBSTER_DIR` on line 3125.

This is a latent bug that won't affect any real install today (the default path matches), but it's inconsistent with every other migration and worth correcting before the pattern propagates.

---

**Verdict:** PASS with the INSTALL_DIR nit. The hook is live already on the running system; install.sh/upgrade.sh paths work correctly for default installs.

PR #2048, opened 2026-05-09
redesign(hooks): replace context-handoff singleton JSON with session.end EventBus append-log (#2002)

sayhar:
🤖🦞 Lobster (engineer): redesign(hooks): replace context-handoff singleton JSON with session.end EventBus append-log

## Problem

The Stop hook wrote `context-handoff.json` using a write-if-absent singleton pattern: skip if the file exists (to preserve the graceful wind-down version). Two problems:

1. The existence check adds unnecessary branching. Since the graceful wind-down writes the file before the Stop hook fires, the check is guarding against a race that never happens in practice.
2. A singleton file loses history — each restart erases the previous record.

## What changed

**Replaced the singleton write with a log-append pattern routed through the central EventBus.** Each dispatcher session end now appends one `session.end` LobsterEvent to `~/lobster-workspace/logs/events.jsonl`. The startup reader scans backwards through `events.jsonl` to find the last `session.end` event and extracts `payload.context_pct` and `timestamp`. No separate `context-handoff.json` or `context-handoff.jsonl` file.

**Dispatcher detection** is now the `agent_id` guard: Claude Code injects `agent_id` only into subagent SessionStop payloads — a single dict lookup replaces the previous `is_dispatcher_session()` process-tree walk (~10ms, subprocess-based). The `session_role` import is removed entirely.

### Before / After

```
Before:
  Session ends → write context-handoff.json (skip if already exists)
  Startup → read context-handoff.json; delete after reading
  Dispatcher detection → is_dispatcher() or is_dispatcher_session() (process-tree walk)

After:
  Session ends → append session.end LobsterEvent to logs/events.jsonl (always)
  Startup → tail events.jsonl backwards for last event_type=="session.end"
  Dispatcher detection → hook_input.get("agent_id") — absent = dispatcher, present = subagent
```

## Changes

**`hooks/dispatcher-state-stop.py`**
- New: `is_subagent(hook_input)` pure function — O(1) dict lookup, no I/O, no imports
- New: `AGENT_ID_FIELD = "agent_id"` named constant (prevents silent rename drift)
- `_write_handoff_if_absent` → `_emit_session_end_event`: always appends to `logs/events.jsonl`
- `_build_handoff_payload` → `_build_session_end_event`: LobsterEvent format matching `LobsterEvent.to_dict()` exactly (`event_type`, `severity`, `source`, `payload`, `timestamp`, `task_id`, `chat_id`)
[...]

PR #2076, opened 2026-05-14
fix(#2075): user-friendly Telegram notification before graceful restart

### LINKED ISSUE #2070: fix(hooks): require-reply-to-message-id blocks proactive/startup sends in debug mode
## Bug

PR #2068 deployed a `require-reply-to-message-id.py` PreToolUse hook that blocks any Telegram `send_reply` call missing a `reply_to_message_id`. The hook's stated exemption is: "If this is a proactive/system send with no originating message, pass chat_id=0 to exempt this check." But this routes the message to the system (no one), not the user.

## Symptom

The dispatcher's `LOBSTER_DEBUG=true` post-bootup status message is sent proactively (no originating user message), so there's no valid `reply_to_message_id` to pass. The hook blocks it.

Result: "🔄 Lobster Starting (debug mode)" appears in Telegram but is never followed by the status report. These two messages are expected to always appear together in debug mode.

## Impact

Debug startup status is silently dropped every time the bot restarts in debug mode.

## Fix Direction

The hook needs a reliable exemption for proactive dispatcher sends — sends that are NOT replies to a user message (e.g., startup status, health alerts, cron summaries). Options:

1. **Add a `proactive=true` parameter to `send_reply`** that the hook recognizes as exempt
2. **Track whether the dispatcher is currently processing a user message**; proactive sends outside that context are exempt
3. **Store the last seen user Telegram message ID at session start** and use it for threading startup status sends

### LINKED ISSUE #2075: feat: user-friendly Telegram notification before graceful 2-hour restart
## Problem

When Lobster performs its planned ~2-hour graceful restart (triggered by \`check_session_age()\` in \`health-check-v3.sh\` at 7200s), the current behavior is:

1. SIGTERM is sent to the dispatcher process
2. A Telegram alert is sent *after* SIGTERM with implementation-detail text: "Lobster: proactive restart at Xs (before the 7440s CC hard limit). Dispatcher PID N sent SIGTERM. Stop hook will fire, session will restart cleanly."

This has two problems:
- The alert fires **after** the SIGTERM, not before. If SIGTERM kills the process before the curl completes, the message may never reach the user.
- The message text is opaque — Alex cannot distinguish a graceful/planned restart from a health-check kill or OOM crash.

## Goal

Before the graceful restart fires, send a clear user-visible message: something like "Restarting now — this is the planned 2-hour graceful restart. Back in ~30 seconds." The notification must arrive **before** the session is killed.

## Scope

Two changes:

### 1. Reorder + reword the notification in `check_session_age()` (health-check-v3.sh)

Move `send_telegram_alert_deduped` to **before** `kill -TERM`. Change the message to be user-friendly rather than exposing internal state (PID, session age, etc.).

### 2. Fix `require-reply-to-message-id.py` hook to allow proactive sends (issue #2070)

The hook currently blocks any Telegram `send_reply` call missing `reply_to_message_id`. The stated exemption (`chat_id=0`) routes to the system queue, not the user. This blocks legitimate proactive dispatcher sends (debug startup status, health alerts routed through MCP).

Fix: add a `proactive=true` parameter to `send_reply` that the hook recognizes as exempt. This is the cleanest deterministic approach — the caller explicitly signals intent, and the hook can verify it without needing to track session state.

## Success criteria

- After restart, Alex sees a message like "Restarting now — planned 2-hour graceful restart. Back in ~30 seconds." followed shortly by the normal startup messages.
- The notification reliably arrives (fires before SIGTERM, not after).
- The `proactive=true` path works in both debug mode and production.
- Issue #2070 is resolved as a side effect.

## Related

- PR #2060 (merged): added `check_session_age()` and the SIGTERM mechanism
- Issue #2070 (open): `require-reply-to-message-id.py` blocks proactive sends

sayhar:

Before this PR, the planned ~2-hour graceful restart (triggered by `check_session_age()` when the session hits 7200s) was invisible to Sahar in two ways:

1. **Wrong ordering:** the Telegram notification fired *after* `kill -TERM` was sent — if the session exited before the curl call completed, the message was never delivered.
2. **Wrong audience:** the message text was written for a developer, not a user: "Lobster: proactive restart at 7260s (before the 7440s CC hard limit). Dispatcher PID 80069 sent SIGTERM. Stop hook will fire, session will restart cleanly." Sahar had no way to distinguish a graceful planned restart from a health-check kill or OOM crash.

Additionally, the `require-reply-to-message-id` hook (#2068) was silently blocking any proactive `send_reply` call (startup status, debug messages, etc.) because it had no exemption for sends that aren't replies to a user message. Issue #2070.

## What changed

`check_session_age()` now sends the Telegram notification *before* `kill -TERM`, not after. Message text changed to user-facing:

> "Restarting now — this is the planned 2-hour graceful restart. Back in ~30 seconds."

```
Before:
  if kill -TERM "$pid"; then
      send_telegram_alert_deduped "proactive-session-restart" "Lobster: proactive restart at Xs..."

After:
  send_telegram_alert_deduped "proactive-session-restart" "Restarting now — planned 2-hour graceful restart. Back in ~30 seconds."
  if kill -TERM "$pid"; then
```

The notification uses `send_telegram_alert_deduped` (direct curl, not MCP) — it fires from the health-check process, not the dispatcher, so it works even when the dispatcher is about to die.

**`hooks/require-reply-to-message-id.py`**

Added a `proactive=True` parameter exemption. When the dispatcher passes `proactive=True`, the hook allows the send without requiring `reply_to_message_id`. This is the correct path for: startup status messages, graceful restart notifications, and any other dispatcher-initiated send that has no originating user message to thread against.

Resolves issue #2070 (debug-mode post-bootup status was silently dropped).

Added a note to the post-bootup status message instruction: pass `proactive=True` when calling `send_reply` for the startup status message, since there is no user message to thread against.

## Flow (changed path only)

```
Before:
  health-check → check_session_age() → kill -TERM → [session may die] → send_telegram_alert (may be lost)

After:
  health-check → check_session_age() → send_telegram_alert → kill -TERM → [session dies cleanly]
```

The hook fix is independent:

```
Before (proactive send):
  dispatcher → send_reply(chat_id=ADMIN_CHAT_ID, text=...) → hook blocks (missing reply_to_message_id) → silent drop

After:
  dispatcher → send_reply(chat_id=ADMIN_CHAT_ID, text=..., proactive=True) → hook exempts → message delivered
```

## Tests run

- [x] `bash tests/test-health-check-session-age.sh` — 13 passed, 0 failed (includes 2 new tests: `notification_text_is_user_friendly`, `notification_fires_before_sigterm`)
- [x] `uv run pytest tests/unit/test_hooks/test_require_reply_to_message_id.py -v` — 16 passed, 0 failed (includes 5 new tests for `proactive=True` exemption)
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 604 passed, 7 skipped, 0 failed

## Manual test

N/A — the graceful restart fires at 7200s of session runtime, which cannot be triggered during development without waiting ~2 hours. The health-check notification path uses direct curl (not MCP), so it is not affected by the hook or MCP session state. The structural ordering test (`notification_fires_before_sigterm`) provides behavioral coverage for the core correctness claim.

For the `proactive=True` hook fix: the unit tests run the actual hook subprocess with real inputs. No production hit required.

## Definition of Done

- [x] Integration/manual test run — N/A (see above)
- [x] User-visible outcome verified — verified via unit tests; live verification requires waiting for next ~2h graceful restart
- [x] Side-effect audit complete: no new sends, no volume changes, no shared state writes
- [x] External service calls: hook tests use subprocess, no production Telegram calls

Closes #2075
Closes #2070

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -73,7 +73,7 @@ When you first start (or after reading this file), follow these steps:
 
 **When the startup catchup result arrives** (`task_id: "startup-catchup"`, `chat_id: 0`): read for situational awareness, update `handoff.md` if anything notable changed (failed subagents, open threads). Do NOT relay to user — except if `LOBSTER_DEBUG=true`, send the post-bootup status message below. Then `mark_processed`.
 
-**Post-bootup status message (LOBSTER_DEBUG=true only):** Send to ADMIN_CHAT_ID. Keep to 5-8 lines, mobile-friendly. Build it from `handoff.md` (just read for startup) and `msg["text"]` (the catchup summary). Format:
+**Post-bootup status message (LOBSTER_DEBUG=true only):** Send to ADMIN_CHAT_ID using `send_reply(chat_id=ADMIN_CHAT_ID, text=..., proactive=True)`. The `proactive=True` flag is required — this send has no originating user message to thread against, and the `require-reply-to-message-id` hook will block it otherwise. Keep to 5-8 lines, mobile-friendly. Build it from `handoff.md` (just read for startup) and `msg["text"]` (the catchup summary). Format:
 
 ```
 🦞 Back online — [session_id], started [start_time ET]
--- a/hooks/require-reply-to-message-id.py
+++ b/hooks/require-reply-to-message-id.py
@@ -9,17 +9,36 @@ but omit or set reply_to_message_id=0.
 Exemptions (always allowed):
   - Non-Telegram sources (slack, sms, whatsapp, bot-talk, system, ...)
   - System/proactive sends where chat_id == 0 (no incoming message context)
+  - Proactive dispatcher sends where proactive=true is explicitly set
   - Calls where reply_to_message_id is already set to a positive integer
+
+The proactive=true exemption is for dispatcher-initiated sends that have no
+originating user message to thread against: startup status, graceful restart
+notifications, scheduled job summaries, health alerts routed through MCP.
+Pass proactive=True explicitly to signal intent — the hook trusts the caller.
 
 Exit codes:
   0 - Allow the call
   2 - Block the call (Claude Code shows stderr message to Claude)
 
-Issue: #1168
+Issue: #1168, #2075
 """
 
 import json
 import sys
+
+
+def _is_truthy(value: object) -> bool:
+    """Return True if value represents a truthy proactive flag.
+
+    Accepts Python bool True, or the strings 'true'/'1' (case-insensitive).
+    Everything else (False, 'false', 0, None, absent) is falsy.
+    """
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
+        return value.strip().lower() in ("true", "1")
+    return False
 
 
 def main() -> None:
@@ -40,6 +59,11 @@ def main() -> None:
 
     # Only enforce on Telegram (source absent defaults to telegram per API contract)
     if source != "telegram":
+        sys.exit(0)
+
+    # Exempt explicitly proactive sends: dispatcher-initiated messages with no
+    # originating user message (startup status, graceful restart notifications, etc.)
+    if _is_truthy(tool_input.get("proactive")):
         sys.exit(0)
 
     # Exempt proactive/system sends: chat_id == 0 means no originating user message
@@ -66,7 +90,7 @@ def main() -> None:
         "Telegram message ID shown in wait_for_messages output as\n"
         "'pass as reply_to_message_id') to thread the reply correctly.\n\n"
         "If this is a proactive/system send with no originating message,\n"
-        "pass chat_id=0 to exempt this check.",
+        "pass proactive=True (preferred) or chat_id=0 to exempt this check.",
         file=sys.stderr,
     )
     sys.exit(2)
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -1338,14 +1338,15 @@ check_session_age() {
         return 0
     fi
 
+    # Notify BEFORE sending SIGTERM so the message reaches the user even if the
+    # session exits immediately after kill -TERM. This is a planned restart —
+    # the user should see a plain-language explanation, not internal state detail.
+    send_telegram_alert_deduped "proactive-session-restart" "Restarting now — this is the planned 2-hour graceful restart. Back in ~30 seconds."
+
     # Send SIGTERM. The Stop hook fires, writes the tombstone, and claude-persistent.sh
     # restarts Claude. This is a graceful exit, not a crash.
     if kill -TERM "$dispatcher_pid" 2>/dev/null; then
         log_warn "Session age: SIGTERM sent to dispatcher PID $dispatcher_pid"
-        send_telegram_alert_deduped "proactive-session-restart" "Lobster: proactive restart at ${session_age}s (before the 7440s CC hard limit).
-
-Dispatcher PID $dispatcher_pid sent SIGTERM. Stop hook will fire, session will restart cleanly.
-Next session will start fresh — no context lost from hard limit."
         # Delete the start timestamp so a subsequent health check run (within the
         # next 4 minutes) does not send a second SIGTERM before the restart completes.
         rm -f "$DISPATCHER_SESSION_START_FILE" 2>/dev/null || true
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -1341,7 +1341,7 @@ check_session_age() {
     # Notify BEFORE sending SIGTERM so the message reaches the user even if the
     # session exits immediately after kill -TERM. This is a planned restart —
     # the user should see a plain-language explanation, not internal state detail.
-    send_telegram_alert_deduped "proactive-session-restart" "Restarting now — this is the planned 2-hour graceful restart. Back in ~30 seconds."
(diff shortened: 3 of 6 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with nits: The core changes are correct and the tests pass, but there are two minor issues worth fixing and one naming confusion to note.

---

## Verification

- 13/13 session age bash tests pass (2 new: notification text, ordering)
- 16/16 hook unit tests pass (5 new: proactive exemption coverage)
- Pre-existing hook suite: 547 → 604 passing tests on PR branch (57 net new passing, including tests from other files in the diff not mentioned in the briefing — not 16 as stated, but all green)
- Pre-existing failure count: the full test suite has 224 collection errors across unrelated modules; these are pre-existing and unchanged by this PR

---

## Issue 1 (nit): Notification text hardcodes "2-hour" but the restart interval is configurable

**File:** `scripts/health-check-v3.sh`

The message reads: `"Restarting now — this is the planned 2-hour graceful restart. Back in ~30 seconds."`

The actual restart threshold is `SESSION_AGE_LIMIT_SECONDS`, which defaults to 7200s but is overridable via `LOBSTER_SESSION_AGE_LIMIT_SECONDS`. If that environment variable is ever changed (e.g. for testing with a shorter window), the user-visible message will say "2-hour" when the real interval is different.

Low risk since the default is indeed 2 hours, but it's a latent truth-in-documentation issue. Either interpolate the actual value, or drop the duration from the message entirely (the user doesn't need to know the internal threshold).

---

## Issue 2 (nit): Test name contradicts its docstring

**File:** `tests/unit/test_hooks/test_require_reply_to_message_id.py`

The test method is named `test_proactive_true_still_blocked_when_reply_id_is_wrong_type` but the docstring says "proactive wins — allowed" and the assertion is `assert rc == 0` (allowed). The name says "still_blocked" which is the opposite of what the test asserts. The test behavior is correct; the name is just misleading for anyone reading the test suite later.

---

## Issue 3 (naming confusion, not a bug): `proactive` is not in the `send_reply` MCP schema

The hook reads `tool_input.get("proactive")`, but `proactive` is not a declared parameter in the `send_reply` tool schema in `inbox_server.py`. For MCP tools, Claude Code passes only what the schema declares — extra parameters the model attempts to include may be silently dropped before the hook sees them.

**This is a potential gap in production**, though not confirmed to be broken today: the schema has no `additionalProperties: false`, so Claude Code's current behavior may pass the extra field through. However, the existing codebase already has a documented precedent for this failure mode: `hooks/require-background-agent.py` explicitly notes that Claude Code stripped `run_in_background` from tool_input in newer versions when the field wasn't in the schema, requiring a workaround (encoding the flag in the prompt frontmatter instead).

The safe fix is to add `proactive` to the `send_reply` input schema in `inbox_server.py` as an optional boolean parameter. The hook works correctly when the field is present (all 5 new tests pass); the question is only whether Claude Code will include it in `tool_input` in production.

The bootup doc update correctly instructs the dispatcher to pass `proactive=True` — if the schema is updated to declare the field, this is a complete and solid fix.

---

## Summary

The core ordering fix (notify before kill) is correct and structurally verified. The user-facing message text is a clear improvement. The `proactive` hook exemption logic is correct. The two nits above (hardcoded duration string, misleading test name) can be fixed in a follow-up or inline. The schema omission for `proactive` is the one item worth addressing before merging if certainty is needed — adding it to the schema costs nothing and closes the gap.

sayhar (commented):
🤖🦞 Lobster (reviewer): PASS with minor observation — all three nits are correctly fixed.

**Nit 1 — hardcoded duration removed:** health-check-v3.sh now sends "Restarting now — this is the planned graceful restart. Back in ~30 seconds." No session-age variable, no CC hard limit reference. The new tests (12: notification_text_is_user_friendly, 13: notification_fires_before_sigterm) also correctly validate plain-language content and pre-SIGTERM ordering. Fixed.

**Nit 2 — test name corrected:** The formerly misleading test is now `test_proactive_true_overrides_wrong_type_reply_id` with a docstring that says "proactive wins — allowed." This accurately describes what rc == 0 means in context. Fixed.

**Nit 3 — proactive schema added:** inbox_server.py now declares `proactive` as an optional boolean in the send_reply tool schema. Description is clear and accurate. Fixed.

**One minor observation (non-blocking):** The fixture text in `test_proactive_true_exempts_telegram_send_without_reply_id` still says "planned 2-hour graceful restart" — the same phrasing Nit 1 removed from the actual notification. The hook ignores text content entirely so this has zero functional impact, but it's cosmetically at odds with the intent of the fix. Worth a one-line cleanup if touched again; not a hold.

All structural checks pass: hook exemption ordering is correct (proactive check before chat_id==0 check), _is_truthy handles bool/string variants correctly, schema type is boolean, no regressions visible in the diff.

aeschylus:
Closing in favor of #2145. Both real bug fixes from this PR (restart-notification ordering/wording, proactive send exemption) have been recreated fresh against current main, with regression tests and manual verification, in #2145 — this branch was opened May 8 and has been functionally untouched for 3 months (part of the same stale session-handoff/EventBus redesign batch as #1979, #2043, #2034, #2027, #2048, #2047, #2042, all closed as stale/superseded). Recreating fresh was simpler and safer than rebasing through 3 months of main drift (e.g. the LOBSTER_DEBUG alert-suppression gate added 2026-06-14, which this branch's base predates).

PR #2089, opened 2026-06-07
fix(post-compact-gate): bail out after 3 consecutive blocked-needs-token failures (#2061)

### LINKED ISSUE #2061: Post-compaction WFM freeze: dispatcher freezes inside wait_for_messages after compaction, health check kills session
🤖🦞 Lobster (ops):

## Symptoms

On 2026-05-09 at 23:45Z (session 018), the Lobster dispatcher froze inside \`wait_for_messages\` approximately 49 seconds after a user message arrived. The freeze occurred shortly after a compaction event at 23:36Z. The health check detected the stalled session and killed it 7 minutes later.

- Session 018 was live at 23:36Z
- Compaction event occurred at 23:36Z
- User message arrived; dispatcher entered \`wait_for_messages\`
- Dispatcher froze ~49 seconds after message receipt — did not process or ack the message
- Health check killed the frozen session at ~23:52Z

## How detected

Health check alert. The dispatcher session was identified as stale (no heartbeat within threshold), and the health check killed it.

## Relation to closed issue #1786

Issue #1786 (closed) documented a similar post-compaction freeze where the dispatcher froze inside \`wait_for_messages\` after context compaction. This appears to be a new variant of the same failure mode — the dispatcher re-enters \`wait_for_messages\` after compaction but then stalls without processing an incoming message.

The key difference: #1786 was resolved by adjusting heartbeat thresholds. This incident appears distinct because the freeze occurred after a user message was already in the inbox — the dispatcher did not process it at all before freezing.

## Two root causes identified

This incident surfaced two separate issues that both require fixes:

**Root cause 1 (separate issue): 7440-second Claude Code session limit**
Every Claude Code session has a hard maximum lifetime of ~7440 seconds (~2 hours). When this limit is hit, the session exits cleanly but without any dispatcher-controlled handoff. This causes plain restarts every ~2 hours regardless of user activity. This is a distinct infrastructure problem from the post-compaction freeze.

**Root cause 2 (this issue): Post-compaction WFM freeze**
After compaction, the dispatcher re-enters \`wait_for_messages\` but does not process the next arriving message. It freezes mid-wait without a heartbeat, requiring the health check to intervene. This is the same class of bug as #1786 but appears to be a different failure mode (freezes with a message in the inbox rather than in an idle wait).

## Expected behavior

After a compaction event, the dispatcher should resume processing messages normally. If \`wait_for_messages\` receives a message within 49 seconds, the dispatcher should process and ack it — not freeze.

## Suggested investigation

- Check if the compaction event interrupts the MCP session or transport layer in a way that leaves \`wait_for_messages\` in a broken state
- Check if the context-monitor hook or any PostToolUse hook fires during compaction and interferes with the subsequent \`wait_for_messages\` call
- Review the session logs for session 018 around 23:36Z–23:45Z for any hook errors or MCP transport anomalies
- Compare with the #1786 fix to understand whether that fix is still in effect and whether this is a regression or a new variant

sayhar:
## Summary

After a context compaction, the dispatcher is required to call `wait_for_messages` with a confirmation token before resuming normal operation. If the token is wrong or absent, the gate blocks the call. The bug: there was no escape — the gate would block indefinitely until the 10-minute TTL expired, and after enough denials the model exhausted its viable context-window actions and went silent. Session 018 demonstrated this: the dispatcher froze 49 seconds post-compaction with a user message already queued in the inbox.

This PR adds a consecutive-failure bail-out: after `MAX_FAILED_ATTEMPTS` (3) consecutive `blocked-needs-token` WFM calls within the same sentinel window, the gate clears the sentinel and allows the call through rather than continuing to block.

The bail-out is a last-resort safety valve — normal operation uses the token path. The counter is scoped to a single sentinel window: it is reset whenever the sentinel is cleared (correct token, TTL expiry, or bail-out), so it never accumulates across compaction events.

## What changed

**`hooks/post-compact-gate.py`**

- Added `FAILURE_COUNTER_FILE` (`~/messages/config/compact-pending-failures`) and `MAX_FAILED_ATTEMPTS = 3` as named module-level constants.
- Added three pure helper functions:
  - `_read_failure_count()` — reads the counter file, returns 0 on absence/error (conservative)
  - `_increment_failure_count()` — atomic write via `.tmp` rename; returns new count; silent on error
  - `_reset_failure_count()` — deletes the counter file; silent on error
- In the `blocked-needs-token` branch: increments counter; if `>= MAX_FAILED_ATTEMPTS`, deletes sentinel + counter and exits 0 (bail-out); otherwise emits deny as before.
- Clears counter on correct-token path, stale-TTL path, and bail-out path — counter never leaks across sentinel windows.
- Updated docstring to document the new pass condition (case 5) and the bail-out section.

**Functional patterns used:** pure helper functions (`_read_failure_count`, `_increment_failure_count`, `_reset_failure_count`) that compose into the gate's main logic; side effects (file writes) isolated to the increment/reset functions; named constants for all spec values (`MAX_FAILED_ATTEMPTS = 3`).

## Before/after flow

```
Before:
  WFM call (no token) → deny → [repeat 1..N times] → model goes silent → health check kills session

After:
  WFM call (no token) → deny (count=1)
  WFM call (no token) → deny (count=2)
  WFM call (no token) → BAIL-OUT: sentinel deleted, allowed through (count=3)
  [dispatcher resumes main loop without health-check intervention]
```

## Out of scope

- The 8 pre-existing test failures in `test_on_compact_write_claude_session_id.py` and `test_context_monitor.py` exist on `main` before this branch and are unrelated to this change.
- Root cause 1 from the issue (7440-second session limit) is a separate infrastructure problem not addressed here.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_post_compact_gate_bailout.py -v` — 12 passed, 0 failed
- [x] `uv run pytest tests/smoke/test_post_compact_gate.py tests/smoke/test_on_compact.py -v` — 12 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/test_post_compact_gate_toolsearch.py tests/unit/test_hooks/test_post_compact_gate_read.py -v` — 7 passed, 0 failed
- [x] `uv run pytest tests/unit/test_hooks/ -v` — 611 passed, 8 pre-existing failures (unrelated), 7 skipped

## Manual test

N/A — this change is entirely within a PreToolUse hook (deterministic file I/O, no external services, no runtime state touched while the live dispatcher is running). Fully covered by unit and smoke tests. The bail-out path produces a `compact-gate.log` entry with action `bail-out-after-N-failures` observable in production after a compaction if the model fails to supply the token.

## Definition of Done

- [x] Integration/manual test — N/A (hook is deterministic file I/O, covered by unit tests)
- [x] User-visible outcome: dispatcher resumes processing user messages after compaction instead of freezing
- [x] Side-effect audit: counter file is max a few bytes, scoped to ~/messages/config/, always cleaned up with sentinel — no volume risk

Closes #2061

### DIFF
--- a/hooks/post-compact-gate.py
+++ b/hooks/post-compact-gate.py
@@ -9,11 +9,14 @@ on-compact.py when a compaction occurs. This hook passes when:
   1. No sentinel file exists (normal operation), OR
   2. Not the dispatcher session (see session_role.is_dispatcher()), OR
   3. The sentinel is older than SENTINEL_TTL_SECONDS (stale / post-hibernation), OR
-  4. The tool being called IS wait_for_messages (let it through, delete sentinel), OR
-  5. The tool being called IS ToolSearch (read-only schema fetch, needed to load
+  4. The tool being called IS wait_for_messages with the correct confirmation token
+     (let it through, delete sentinel and failure counter), OR
+  5. The tool being called IS wait_for_messages WITHOUT the token but consecutive
+     failures have reached MAX_FAILED_ATTEMPTS (bail-out path — see below), OR
+  6. The tool being called IS ToolSearch (read-only schema fetch, needed to load
      deferred MCP tool schemas — the dispatcher must be able to call ToolSearch to
      fetch the wait_for_messages schema before it can call wait_for_messages), OR
-  6. The tool being called IS Read (read-only file access — non-destructive and
+  7. The tool being called IS Read (read-only file access — non-destructive and
      needed to break the circular dependency in issue #1950: Read is required to
      learn the confirmation token from the bootup file, but the gate was blocking
      Read before the token was known).
@@ -21,6 +24,19 @@ on-compact.py when a compaction occurs. This hook passes when:
 The TTL (10 minutes) eliminates the stuck-sentinel failure mode: if the
 dispatcher hibernates or crashes while the sentinel is active, the next boot
 finds a stale sentinel and the gate passes immediately. No deadlock.
+
+## Consecutive-failure bail-out (issue #2061)
+
+After MAX_FAILED_ATTEMPTS consecutive blocked-needs-token WFM calls within the
+same sentinel window, the gate deletes the sentinel and its companion counter
+file (~/messages/config/compact-pending-failures) and allows WFM through without
+the token. This prevents the model from going silent after exhausting viable
+actions in its context window — the failure mode observed in session 018 where
+the dispatcher froze 49 seconds after compaction and never processed the queued
+user message.
+
+The counter is reset whenever the sentinel is cleared (correct token, TTL expiry,
+or bail-out), so it does not accumulate across sentinel windows.
 
 ## Dispatcher vs subagent detection
 
@@ -86,6 +102,20 @@ SENTINEL_FILE = Path(os.path.expanduser("~/messages/config/compact-pending"))
 SENTINEL_TTL_SECONDS = 600  # 10 minutes — treats stale sentinel as harmless
 LOG_FILE = Path(os.path.expanduser("~/lobster-workspace/logs/compact-gate.log"))
 
+# Consecutive-failure bail-out (issue #2061).
+#
+# After MAX_FAILED_ATTEMPTS consecutive blocked-needs-token WFM calls within
+# the same sentinel window, the gate deletes the sentinel (and this counter
+# file) and allows through rather than continuing to block.  This prevents the
+# model from going silent after exhausting viable actions in its context window.
+#
+# The counter file records the number of consecutive blocked-needs-token events
+# since the current sentinel was written.  It lives in the same directory as
+# the sentinel and is always removed together with it (correct token, TTL
+# expiry, or bail-out).
+FAILURE_COUNTER_FILE = Path(os.path.expanduser("~/messages/config/compact-pending-failures"))
+MAX_FAILED_ATTEMPTS = 3  # Bail out after this many consecutive blocked-needs-token failures
+
 WAIT_FOR_MESSAGES_TOOL = "mcp__lobster-inbox__wait_for_messages"
 
 # ToolSearch is a read-only schema fetch.  It is always allowed through the gate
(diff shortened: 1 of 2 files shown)


sayhar (commented):

## Verdict: PASS

The core fix is correct and well-structured. The bail-out logic works as intended, all 12 new tests pass, and the sentinel-cleanup paths are sound. I ran the full test suite on the PR branch and can confirm the fix does not introduce any new failures.

## Finding 1: Pre-existing failure count is 9, not 8

The engineer's briefing states 8 pre-existing failures, but the actual count on both main and the PR branch is **9**. The additional failure is `test_bot/test_slack_router_config.py::TestPollChannelsValidation::test_warning_logged_when_poll_channels_empty` — it fails on main and is unrelated to this PR. No action needed for the fix itself, but the count in the PR description should be corrected.

Verified on PR branch: `9 failed, 3514 passed, 31 skipped`.

## Finding 2: Counter cleanup is skipped when wait_for_messages is called against a stale sentinel

When the dispatcher calls `wait_for_messages` and the sentinel is stale (older than 10 minutes), the gate allows the call through via the early-exit path in the WFM block — without reaching the stale-sentinel cleanup code at the bottom of `main()`. The stale sentinel file and any lingering counter file are not cleaned up in this case.

In practice this is low-severity: the counter file only affects bail-out timing when there is an *active* (fresh) sentinel, and a stale sentinel is by definition no longer active. But the residue is inconsistent with the documented invariant that the counter and sentinel are always cleared together. A subsequent non-WFM tool call does eventually clean them up.

This is a variation of the "counter may linger if sentinel deleted externally" gap the engineer already flagged — the WFM-with-stale-sentinel path is another case where the counter outlives its sentinel window.

No test covers this scenario. Suggested addition (not a blocker): a test that calls WFM against a stale sentinel and asserts the counter is cleaned up.

## Everything the engineer flagged checks out

- Bail-out fires on the Nth call itself (that call gets through): confirmed correct, and explicitly tested in F2c.
- `MAX_FAILED_ATTEMPTS=3`: exposed as a named constant, value is reasonable, threshold ≥ 2 is asserted.
- Counter file may linger if sentinel deleted externally: acknowledged, not a regression.
- Atomic write via `.tmp` rename: implemented correctly.

## Test structure

No duplicate class names, no shadowed test classes, no unreachable tests. All 12 tests are reachable and exercise real behavior.

PR #2109, opened 2026-07-17
Consolidate dispatcher-exclusion filter into one shared predicate

aeschylus:
## Problem

The `agent_type='dispatcher'` exclusion check — used to make sure the dispatcher's own row in `agent_sessions.db` is never mistaken for a dead/failed subagent — was duplicated inline at five call sites: `src/agents/session_store.py` (`cleanup_stale_running_sessions`, `get_unnotified_completed`), `src/mcp/inbox_server.py` (`_startup_sweep`, `reconcile_agent_sessions`), and `scripts/periodic-self-check.sh`. Because there was no single definition, this exact filter had to be fixed independently four times (issue #781, PR #2099, PR #2103) as new call sites were added or the pattern was copy-pasted incorrectly. A sixth, test-only reimplementation had also crept into `test_ghost_session_fixes.py` — a hand-mirrored copy of the guard used to unit-test the concept, but not the actual production code.

Implements Linear BIS-723 (https://linear.app/fully-parsed/issue/BIS-723/slice-d-consolidate-the-dispatcher-exclusion-filter-into-one-shared): consolidate the dispatcher-exclusion filter into one shared function so a sixth reimplementation becomes structurally hard to add without noticing the existing shared definition.

## What changed

- **New `src/utils/agent_types.py`** — the Python single source of truth:
  - `DISPATCHER_AGENT_TYPE = "dispatcher"`
  - `DISPATCHER_EXCLUSION_SQL` — the `COALESCE(agent_type, '') != 'dispatcher'` SQL fragment, for callers building a raw SQL WHERE clause
  - `is_dispatcher_agent_type(agent_type)` — pure predicate over a bare string
  - `is_dispatcher_row(session)` — convenience wrapper over a session dict
- **New `scripts/lib/agent_sessions.sh`** — the shell counterpart (`DISPATCHER_AGENT_TYPE` / `DISPATCHER_EXCLUSION_SQL` shell variables), following the existing `scripts/lib/template.sh` sourced-library convention. Python and bash can't literally share one function, so the two files cross-reference each other by name in comments — if one changes, the other must change too.
- `session_store.py`'s two raw SQL queries now interpolate `DISPATCHER_EXCLUSION_SQL` instead of inlining the fragment.
- `inbox_server.py`'s two dict-based checks now call `is_dispatcher_row(session)` instead of `(session.get("agent_type") or "") == "dispatcher"`.
- `periodic-self-check.sh` sources the shared shell lib and interpolates `$DISPATCHER_EXCLUSION_SQL` into its `sqlite3` query.
- `test_ghost_session_fixes.py`'s locally-defined `_should_reconciler_skip()` mirror function is replaced with a direct reference to `is_dispatcher_row` — the test now exercises the real shared predicate instead of a hand-maintained copy of its logic.

This is a pure refactor: no behavior change. `grep` confirms no inline `!= 'dispatcher'` / `== "dispatcher"` reimplementation remains at any of the five original call sites — each now delegates to the shared definition. (Two unrelated uses of the literal `"dispatcher"` string remain in `inbox_server.py` — a session self-identification check in `session_start()`, and a `task_id` display-label default in two `_emit_event` calls. Both are semantically different from the exclusion filter and out of scope per the issue.)

## Functional patterns

`is_dispatcher_agent_type` / `is_dispatcher_row` are pure functions (no side effects, no I/O) — same input always produces the same output. `DISPATCHER_EXCLUSION_SQL` is a derived constant computed once from `DISPATCHER_AGENT_TYPE`, not duplicated data. No architectural changes to control flow — this is a "replace five equivalent expressions with five calls to one definition" refactor, not a new abstraction layer.

## No flow/state-machine changes

This PR does not alter any multi-step flow, notification sequence, retry logic, or state machine — the dispatcher-exclusion condition itself is byte-for-byte equivalent before and after (same SQL fragment, same boolean predicate). No diagram included per the PR guidance, since there's nothing to diagram.

## Tests run

- [x] `uv run pytest tests/test_agent_sessions.py tests/unit/test_mcp_server/test_ghost_session_fixes.py tests/unit/test_agent_types.py -q` — 97 passed (90 pre-existing + 7 new in `test_agent_types.py`)
- [x] `uv run pytest tests/unit/test_mcp_server/ -q` — 865 passed (confirms `inbox_server.py` still imports and behaves correctly with the new `utils.agent_types` import)
- [x] `uv run pytest tests/ -q` (full suite) — **before**: 4108 passed / 10 failed / 34 skipped. **after**: 4117 passed / 10 failed / 34 skipped — same 10 pre-existing failures both times (`test_message_volume`, `test_slack_router_config`, `test_context_monitor`, `test_on_compact_write_claude_session_id` — all unrelated to this change, none touch `agent_sessions.db` or the dispatcher-exclusion filter). +9 passed = the 7 new `test_agent_types.py` tests + 2 new dispatcher-exclusion regression tests added to `test_agent_sessions.py`.
- [x] Falsifiability check on `is_dispatcher_agent_type`: inverted `==` to `!=` in `src/utils/agent_types.py`, re-ran `test_agent_types.py` — 6 of 7 tests failed as expected, restored, re-ran — 7/7 passed.
- [x] Falsifiability check on `DISPATCHER_EXCLUSION_SQL`: temporarily replaced the constant with the SQL tautology `"1=1"`, re-ran the two new regression tests in `test_agent_sessions.py` plus the SQL-content test in `test_agent_types.py` — all 3 failed as expected (dispatcher row was no longer excluded), restored, re-ran — all passed.
- [x] `python3 -m py_compile` on all three changed/new Python modules — clean.

## Manual test

`scripts/periodic-self-check.sh` and the new `scripts/lib/agent_sessions.sh` are shell scripts invoked by cron in production, so I ran the real script end-to-end against an isolated fake `agent_sessions.db` (own `mktemp -d`, `AGENT_TASKS_DIR` pointed at an empty directory so the real production `/tmp/claude-*/tasks/*.output` files were never scanned, `LOBSTER_MESSAGES`/`LOBSTER_INSTALL_DIR` pointed at the temp tree):

- **Case A** — DB has only one `agent_type='dispatcher'` row, `status='running'`: script ran to completion, exit 0, **zero** files written to the fake inbox (dispatcher correctly excluded from the pending count, no false-positive self-check fired).
- **Case B** — added one real `agent_type='functional-engineer'` row, `status='running'`, re-ran: script wrote exactly one self-check message to the fake inbox with text `"status? (Self-check) [1 agents pending]"` — confirming the real subagent row is correctly counted while the dispatcher row is still excluded.

No production `agent_sessions.db`, inbox, or `/tmp/claude-*` data was read or written — every path was redirected into a throwaway temp directory, and I removed the worktree's `.state/last-self-check` / `.state/reported-tasks` test artifacts afterward (these live outside git per `.gitignore` regardless).

## Known gap

There were no existing tests exercising the dispatcher-exclusion behavior of `session_store.py`'s `cleanup_stale_running_sessions()` or `get_unnotified_completed()` before this PR (confirmed via `grep -rn dispatcher tests/test_agent_sessions.py` returning nothing pre-change) — only `inbox_server.py`'s reconciler had coverage, and even that was via the hand-mirrored function described above. This PR adds the first direct regression tests for both `session_store.py` functions.

## Definition of Done
- [x] Integration/manual test run — see Manual test section (isolated temp DB + temp dirs, no production paths touched)
- [x] User-visible outcome verified — not applicable; this is an internal refactor with no user-facing behavior change (dispatcher-exclusion outcome is provably identical before/after via the test suite)
- [x] Side-effect audit complete: no floods, no unscoped writes (manual test used `mktemp -d` + explicit env var overrides), no unintended volume
- [x] External service calls: N/A — no external services involved in this change

### DIFF
--- a/scripts/lib/agent_sessions.sh
+++ b/scripts/lib/agent_sessions.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#===============================================================================
+# Shared dispatcher-exclusion SQL fragment (Linear BIS-723)
+#
+# Single source of truth for the shell side of the dispatcher-exclusion check
+# against agent_sessions.db. Before this file existed, the fragment below was
+# copy-pasted directly into periodic-self-check.sh's sqlite3 query, alongside
+# four independent Python reimplementations of the same check
+# (session_store.py x2, inbox_server.py x2) — the whole family of duplicates
+# had to be fixed four separate times (issue #781, PR #2099, PR #2103).
+#
+# Cross-reference: this constant MUST match DISPATCHER_AGENT_TYPE /
+# DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py. Bash and Python cannot
+# literally share one function, so if you change the excluded agent_type
+# value here, change it in src/utils/agent_types.py too (and vice versa).
+#
+# Usage:
+#   source "$(dirname "$0")/lib/agent_sessions.sh"
+#   sqlite3 "$DB_PATH" \
+#     "SELECT COUNT(*) FROM agent_sessions WHERE status IN ('running','starting') AND $DISPATCHER_EXCLUSION_SQL"
+#===============================================================================
+
+# The agent_type value written when the dispatcher registers its own session.
+# Must match DISPATCHER_AGENT_TYPE in src/utils/agent_types.py.
+DISPATCHER_AGENT_TYPE="dispatcher"
+
+# SQL WHERE-clause fragment excluding dispatcher rows from a query over
+# agent_sessions. COALESCE guards against agent_type being NULL (legacy rows
+# predating the column, or subagents that never set it).
+# Must match DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py.
+DISPATCHER_EXCLUSION_SQL="COALESCE(agent_type, '') != '${DISPATCHER_AGENT_TYPE}'"
--- a/scripts/periodic-self-check.sh
+++ b/scripts/periodic-self-check.sh
@@ -103,12 +103,15 @@ else
     # Exclude agent_type='dispatcher' — the dispatcher's own session is always
     # registered as running/starting and would otherwise be counted as a
     # "pending agent" on every firing, producing a permanent false positive.
-    # Mirrors the fix applied in session_store.py's cleanup_stale_running_sessions()
-    # and get_unnotified_completed() (see #781 / PR #2099); this script queries
-    # the DB directly via sqlite3 rather than through session_store.py, so it
-    # was not covered by that PR and needs the same filter applied here.
+    # DISPATCHER_EXCLUSION_SQL is the shared single source of truth (BIS-723,
+    # scripts/lib/agent_sessions.sh) for this check — it must match
+    # DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py, used by the
+    # equivalent Python-side checks in session_store.py and inbox_server.py
+    # (see #781 / PR #2099 / PR #2103 for the history of this filter being
+    # fixed independently at each call site before consolidation).
+    source "${LOBSTER_INSTALL_DIR:-$HOME/lobster}/scripts/lib/agent_sessions.sh"
     PENDING_COUNT=$(sqlite3 "$MESSAGES_DIR/config/agent_sessions.db" \
-        "SELECT COUNT(*) FROM agent_sessions WHERE status IN ('running','starting') AND COALESCE(agent_type, '') != 'dispatcher'" \
+        "SELECT COUNT(*) FROM agent_sessions WHERE status IN ('running','starting') AND ${DISPATCHER_EXCLUSION_SQL}" \
         2>/dev/null || echo "0")
 
     # No completed tasks — only inject status check if subagents are still
--- a/src/agents/session_store.py
+++ b/src/agents/session_store.py
@@ -22,9 +22,19 @@ Design principles:
 import logging
 import os
 import sqlite3
+import sys
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
+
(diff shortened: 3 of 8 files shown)


aeschylus (commented):
Supervisor review (BIS-723).

This is the substantive code PR of the epic, so I checked it out and ran verification myself rather than just reading the claims:
- `grep` for remaining inline `!= 'dispatcher'` / `== "dispatcher"` reimplementations at the 5 original call sites: clean, all delegate to the shared `DISPATCHER_EXCLUSION_SQL` / `is_dispatcher_row()` definitions in the new src/utils/agent_types.py and scripts/lib/agent_sessions.sh.
- The two remaining literal "dispatcher" string checks in inbox_server.py (session_start() self-identification, task_id display-label default) are correctly out of scope — different semantics, not the exclusion filter.
- Ran `uv run pytest tests/test_agent_sessions.py tests/unit/test_mcp_server/test_ghost_session_fixes.py tests/unit/test_agent_types.py -q` myself: 99 passed (PR description says 97; minor discrepancy, not a concern — likely a collection-count nuance, all pass either way).
- Falsifiability claims (inverting the predicate / replacing the SQL constant with a tautology to confirm tests actually catch regressions) are described in the PR body and are the right kind of proof — didn't re-run these myself but the test design supports the claim.
- Manual test of periodic-self-check.sh against an isolated temp DB (dispatcher-only vs. dispatcher+subagent cases) is real proof of runtime behavior, not just unit coverage, and correctly avoided touching production paths.
- Matches DEVELOPMENT.md convention: work done in a worktree at lobster-workspace/projects/, ~/lobster/ untouched.

No corrections needed. This looks ready to merge. Since BIS-724 stacks on this branch, I'll proceed to spawn that work targeting this branch now (will rebase/retarget after this merges if needed).

PR #2136, opened 2026-07-25
fix: correct nightly-consolidation step 10 bridge sync command

### LINKED ISSUE #2135: nightly-consolidation step 10 run_bridges can silently wipe _context.md (workspace path mismatch)
## Summary

Step 10 of `.claude/agents/nightly-consolidation.md` (around lines 168-183) runs the bridge sync with a single `workspace_path` argument shared between `run_bridges()`'s read root (canonical memory files) and write root (`_context.md` output). In this deployment those are two different directories:

- Canonical files (`projects/*.md`, `priorities.md`) live under `~/lobster-user-config/memory/canonical/`
- The output `_context.md` must be written under `~/lobster-workspace/user-model/`

`run_bridges(conn, workspace_path)` in `src/mcp/user_model/bridges.py` (lines 333-359) passes the same `workspace_path` to all three of `sync_projects_to_arcs`, `sync_priorities_to_attention`, and `write_context_cache`. Each of those (lines 121-134, 190-199, 234-246) derives its root as `Path(workspace_path) if workspace_path else Path.home() / "lobster-workspace"`, then appends `memory/canonical/...` for reads or `user-model/_context.md` for the write. There's no way to pass separate read/write roots through `run_bridges`.

As documented, step 10 calls `run_bridges(conn)` with no `workspace_path` at all, so it defaults to `~/lobster-workspace`. That means:
- `sync_projects_to_arcs` looks for `~/lobster-workspace/memory/canonical/projects/*.md` — doesn't exist (canonical files are actually under `~/lobster-user-config/memory/canonical/projects/`) — finds nothing, silently returns `{"synced": 0, ...}` (bridges.py line 133-134 short-circuits if the dir doesn't exist).
- `sync_priorities_to_attention` similarly finds no `priorities.md` and returns `{"injected": 0}` (line 202-203).
- `write_context_cache` then writes a near-empty `_context.md` stub over the previously populated file (bridges.py lines 306-311), because there's no confirmation/backup step — this is a straight overwrite via atomic rename (lines 314-318).

This was triggered once during investigation of this bug and silently overwrote the populated `_context.md`. It had to be manually restored.

## Secondary bug: import breaks due to namespace-package collision

The documented invocation pattern is:
```python
import sys; sys.path.insert(0, 'src')
from mcp.user_model.bridges import run_bridges
```

This fails because `mcp` resolves to the installed MCP SDK package in site-packages (which has an `__init__.py` and therefore wins over the PEP 420 namespace package `src/mcp`), not the repo's `src/mcp/`. The fix used elsewhere in the repo (`tests/test_user_model_v2.py` line 16) is to put `src/mcp` itself on `sys.path` and import `user_model.bridges` directly (no `mcp.` prefix):
```python
sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "src" / "mcp"))
from user_model.db import init_schema, open_db, ...
```

## Tertiary bug: row_factory not set

`run_bridges` expects a connection opened via `user_model.db.open_db()` (`src/mcp/user_model/db.py` lines 1227-1234), which sets `conn.row_factory = sqlite3.Row` before returning. A bare `sqlite3.connect()` (as documented in step 10, `conn = sqlite3.connect(db_path)`) leaves the default tuple row factory. Several `db.py` query functions do dict-style row access (e.g. `get_active_narrative_arcs` at line 776-791 does `r["title"]`, `get_emotional_baseline` at line 603-618 does `rows["v"]`), which raises `TypeError` on plain tuples. Because `write_context_cache` wraps each section in a broad `except Exception: pass` (bridges.py lines 250-271, 274-284, 287-294, 297-304), these failures are swallowed silently rather than surfaced, compounding the "looks like it worked but wrote nothing" failure mode.

## Fix

Fixing `.claude/agents/nightly-consolidation.md` step 10 to:
1. Put `src/mcp` on `sys.path` and import `user_model.db`/`user_model.bridges` directly (no `mcp.` prefix), avoiding the namespace collision.
2. Use `user_model.db.open_db()` to get a connection with the correct `row_factory`.
3. Call `sync_projects_to_arcs(conn, canonical_ws)` and `sync_priorities_to_attention(conn, canonical_ws)` with the canonical read root (`~/lobster-user-config`), and `write_context_cache(conn, output_ws)` with the output write root (`~/lobster-workspace`) — i.e. call the three bridge functions directly instead of `run_bridges()`, since `run_bridges()` has no way to accept separate read/write roots.

Verified working replacement (confirmed live: synced 7 project arcs, 5 priority items):
```python
import sys, os
sys.path.insert(0, os.path.join(os.getcwd(), 'src', 'mcp'))
from user_model.db import open_db
from user_model.bridges import sync_projects_to_arcs, sync_priorities_to_attention, write_context_cache
from pathlib import Path
conn = open_db(Path(os.path.expanduser('~/lobster-workspace/data/memory.db')))
canonical_ws = os.path.expanduser('~/lobster-user-config')
output_ws = os.path.expanduser('~/lobster-workspace')
proj = sync_projects_to_arcs(conn, canonical_ws)
pri = sync_priorities_to_attention(conn, canonical_ws)
write_context_cache(conn, output_ws)
conn.commit()
conn.close()
```

A more thorough fix would give `run_bridges()` separate `read_workspace_path`/`write_workspace_path` parameters so callers don't have to bypass it — filed here as a possible follow-up but out of scope for the immediate doc fix.

Interplay-lab:
## Summary
- Step 10 of `.claude/agents/nightly-consolidation.md` documented calling `run_bridges(conn)`, which uses a single `workspace_path` for both the canonical-memory read root and the `_context.md` write root. Those roots differ in this deployment (`~/lobster-user-config` vs `~/lobster-workspace`), so the documented call found no canonical data and silently overwrote the populated `_context.md` with a near-empty stub.
- Also fixes the documented `sys.path.insert(0, 'src'); from mcp.user_model.bridges import ...` import, which breaks due to a PEP 420 namespace-package collision with the installed `mcp` SDK package — switched to putting `src/mcp` on `sys.path` and importing `user_model.bridges` directly, matching the pattern already used in `tests/test_user_model_v2.py`.
- Also switches from a bare `sqlite3.connect()` to `user_model.db.open_db()`, which sets `row_factory = sqlite3.Row` — required by the bridge query functions, whose dict-style row access otherwise raises and gets silently swallowed by broad `except` blocks.
- Step 10 now calls `sync_projects_to_arcs`, `sync_priorities_to_attention`, and `write_context_cache` directly (rather than `run_bridges()`) so the canonical read root and workspace write root can be passed separately.

Fixes #2135

## Test plan
- [ ] Ran the corrected step-10 command against the live DB during investigation — confirmed it syncs 7 project arcs and 5 priority items and writes a populated `_context.md` (done manually while diagnosing; not re-run as part of this PR since it mutates the live user-model DB)
- [x] Doc-only change reviewed against `src/mcp/user_model/bridges.py` and `src/mcp/user_model/db.py` for signature accuracy

### DIFF
--- a/.claude/agents/nightly-consolidation.md
+++ b/.claude/agents/nightly-consolidation.md
@@ -166,19 +166,32 @@ Include the GitHub activity summary in the synthesis for rolling-summary.md and 
    If handoff.md has no PR table, skip step 9b silently. If `gh` is unavailable, skip step 9b and note it in `write_result`. If the PR table format is unexpected, leave the table unchanged and note it in `write_result` — do not crash.
 
 10. **Sync canonical files into the user model DB.**
-   Run the bridge pass to push projects, priorities, and preferences from canonical markdown files into the user model DB. This also generates the pre-computed `_context.md` via `write_context_cache()`:
+   Run the bridge pass to push projects, priorities, and preferences from canonical markdown files into the user model DB. This also generates the pre-computed `_context.md` via `write_context_cache()`.
+
+   **Do not call `run_bridges()`** — it takes a single `workspace_path` used for both reads and writes, but in this deployment the canonical read root (`~/lobster-user-config`) and the `_context.md` write root (`~/lobster-workspace`) are different directories. Calling it with one shared path finds no canonical data and silently overwrites the populated `_context.md` with a near-empty stub. Call the three bridge functions directly instead, passing the correct root to each:
    ```bash
    cd ~/lobster && uv run python -c "
-   import sys; sys.path.insert(0, 'src')
-   from mcp.user_model.bridges import run_bridges
-   import sqlite3, os
-   db_path = os.path.expanduser('~/lobster-workspace/data/memory.db')
-   conn = sqlite3.connect(db_path)
-   result = run_bridges(conn)
+   import sys, os
+   sys.path.insert(0, os.path.join(os.getcwd(), 'src', 'mcp'))
+   from user_model.db import open_db
+   from user_model.bridges import sync_projects_to_arcs, sync_priorities_to_attention, write_context_cache
+   from pathlib import Path
+   conn = open_db(Path(os.path.expanduser('~/lobster-workspace/data/memory.db')))
+   canonical_ws = os.path.expanduser('~/lobster-user-config')
+   output_ws = os.path.expanduser('~/lobster-workspace')
+   proj = sync_projects_to_arcs(conn, canonical_ws)
+   pri = sync_priorities_to_attention(conn, canonical_ws)
+   write_context_cache(conn, output_ws)
+   conn.commit()
    conn.close()
-   print(result)
+   print(proj, pri)
    "
    ```
+   Notes on this invocation:
+   - Import `user_model.db` / `user_model.bridges` directly (no `mcp.` prefix) after putting `src/mcp` itself on `sys.path`. Importing via `from mcp.user_model.bridges import ...` after `sys.path.insert(0, 'src')` breaks because `mcp` resolves to the installed MCP SDK package in site-packages, not `src/mcp/`.
+   - Open the connection with `user_model.db.open_db()`, not a bare `sqlite3.connect()`. `open_db()` sets `conn.row_factory = sqlite3.Row`, which the bridge query functions require for dict-style row access — without it, queries fail and are silently swallowed by broad `except` blocks, again producing a near-empty `_context.md`.
+   - `sync_projects_to_arcs` and `sync_priorities_to_attention` read from `canonical_ws` (canonical memory files); `write_context_cache` writes into `output_ws` (the workspace `_context.md` lives under). Do not use the same variable for both.
+
    This syncs `projects/*.md` as narrative arcs and `priorities.md` as attention items, and writes the pre-computed `~/lobster-workspace/user-model/_context.md`.
    If the script fails (e.g. DB not initialized), continue to step 11.
 


Interplay-lab (commented):
🤖🦞 Lobster (reviewer): PASS

Verified independently (not just trusting the PR description):

1. **Diff scope confirmed**: only `.claude/agents/nightly-consolidation.md` changes; `bridges.py` untouched, as claimed.

2. **Signatures checked against source** (fetched `src/mcp/user_model/bridges.py` and `db.py` from `main`):
   - `sync_projects_to_arcs(conn, workspace_path: str | None = None)` — called as `(conn, canonical_ws)` ✓
   - `sync_priorities_to_attention(conn, workspace_path)` — called as `(conn, canonical_ws)` ✓
   - `write_context_cache(conn, workspace_path)` — called as `(conn, output_ws)` ✓
   - `open_db(db_path: Path) -> sqlite3.Connection` — called with `Path(...)`, sets `conn.row_factory = sqlite3.Row` and runs `init_schema`. Confirmed the DB readers (`get_active_narrative_arcs`, `get_attention_stack`, etc.) use dict-style `row["col"]` access that requires `sqlite3.Row` — the row_factory claim is real, not speculative.

3. **canonical_ws / output_ws split verified against the live filesystem**: `~/lobster-user-config/memory/canonical/{projects/,priorities.md}` exist and match `sync_*`'s internal `ws/"memory"/"canonical"/...` construction; `~/lobster-workspace/user-model/_context.md` exists and matches `write_context_cache`'s `ws/"user-model"/"_context.md"`. The split is correct, not just plausible.

4. **Import fix verified empirically, not just read**: ran `sys.path.insert(0,'src'); import mcp` — it resolves to the installed MCP SDK in `.venv/site-packages`, confirming the *old* `from mcp.user_model.bridges import run_bridges` pattern was genuinely broken (ModuleNotFoundError). Ran the PR's replacement (`sys.path.insert(0, 'src/mcp')` then `from user_model.bridges import ...`) — imports cleanly. This also matches existing codebase convention: `src/mcp/inbox_server.py` does the identical `sys.path` trick for the same collision reason, and `src/utils/timezone.py` already treats `from user_model.owner import ...` (no `mcp.` prefix) as primary with the `mcp.user_model...` form as a secondary/effectively-dead fallback. `tests/test_user_model_v2.py` imports the same way. Grepped for other `mcp.user_model` importers — only doc/system.md references in unrelated `lobster-shop/*` skill files and that one dead fallback line; nothing this PR's isolated one-shot subprocess touches or breaks.

5. **Corroborating real-world evidence**: today's actual `~/lobster-workspace/user-model/_context.md` already contains an inline note from this morning's consolidation run describing hitting this exact `ModuleNotFoundError` and working around it — independent confirmation the bug is real in production, not theoretical.

Two minor, non-blocking observations the briefing didn't mention:
- The old `run_bridges()` wrapped each of the three calls in its own `try/except`, so one failing step didn't block the others — notably `write_context_cache`, the most important output. The new inline script calls all three sequentially with no per-call isolation, so a failure in `sync_projects_to_arcs` would now also skip `write_context_cache` for that run. Low impact (doc-level "if script fails, continue to step 11" still covers it, and arguably beats the old behavior of always overwriting `_context.md` even with bad data) — worth a follow-up to re-add per-call `try/except`.
- The trailing `conn.commit()` is redundant — every `upsert_*` in `db.py` already commits internally, and `write_context_cache` only does file I/O. Harmless.

Neither blocks this fix. The core diagnosis and correction (workspace-path split, row_factory via `open_db`, import pattern) all check out against actual source and actual signatures.

Interplay-lab:
🤖 Independent reconciliation review (automated pass, run 2026-07-29)

This PR has been open several days without merging, so this is a second, independent pass specifically to check whether anything on `main` changed underneath it since the original PASS review. I did not rely on the PR description or the prior review — I re-read the diff, pulled the current `src/mcp/user_model/bridges.py` and `db.py` from main directly, and cross-checked signatures line by line.

**Verified against current main:**
- `open_db(db_path: Path) -> sqlite3.Connection` exists in `db.py` (line 1227) and does set `conn.row_factory = sqlite3.Row` (line 1230) before calling `init_schema`. Matches the PR's `open_db(Path(...))` usage.
- `sync_projects_to_arcs(conn, workspace_path=None)`, `sync_priorities_to_attention(conn, workspace_path=None)`, and `write_context_cache(conn, workspace_path=None)` all exist in `bridges.py` with that exact signature (lines 121, 190, 234). Crucially, the two "read" functions build paths under `ws / "memory" / "canonical" / ...` while `write_context_cache` builds `ws / "user-model" / "_context.md"` — so passing `canonical_ws` to the first two and `output_ws` to the third (as the PR's new step 10 does) is the correct pairing, not swapped.
- The `sys.path.insert(0, 'src/mcp')` + `import user_model.bridges` (no `mcp.` prefix) pattern exactly matches the pre-existing workaround already used in `tests/test_user_model_v2.py`, which corroborates the claimed namespace-package collision with the installed MCP SDK is real and this is the established fix pattern, not a novel guess. `src/mcp` has no `__init__.py` (confirmed via API), consistent with it being a PEP 420 namespace package that loses to the installed regular `mcp` package when only `src` is on `sys.path`.
- `conn.commit()` before `conn.close()` in the script is necessary — the `upsert_*` functions in `db.py` don't commit internally, so this isn't redundant.

**Reconciliation check:** Pulled all PRs (open + merged) via `gh pr list --state all --limit 30`. Nothing merged since #2136 opened (2026-07-25) touches `src/mcp/user_model/`, `bridges.py`, or `nightly-consolidation.md` — the merged PRs in that window are all Slack/OAuth/connector work (#2129-2134), unrelated. No duplicate fix, no upstream drift. `gh pr view --json mergeable,mergeStateStatus` reports `MERGEABLE` / `CLEAN`, 1 file changed, no conflicts.

**One process note, not a blocker:** the existing "PASS" review on this PR was posted by `Interplay-lab` — the same account that opened the PR. Worth knowing it wasn't reviewed from an independent identity, even though its technical content checks out.

**Verdict: agree with the prior PASS, still good to merge as-is.** Nothing on main has changed that affects this fix's correctness or applicability.

— independent automated reviewer pass (not the original reviewer, not human)

PR #2140, opened 2026-07-31
docs: fix step-1 session_start() example missing required agent_id/description/chat_id

### LINKED ISSUE #2139: docs: sys.dispatcher.bootup.md step 1 session_start() call fails schema validation (missing required agent_id/description/chat_id)
## Repro

Following `.claude/sys.dispatcher.bootup.md` step 1 literally:

```
session_start(agent_type='dispatcher', claude_session_id=hook_input["session_id"])
```

fails immediately with:

```
Input validation error: 'agent_id' is a required property
```

The `mcp__lobster-inbox__session_start` tool schema requires `agent_id`, `description`, and `chat_id` (see `id`, `description`, `chat_id` as the non-defaulted params on the underlying `session_start()` in `src/agents/session_store.py`, lines 241-244 — the MCP wrapper exposes `id` as `agent_id`). None of those three are present in the step 1 example call, so it can never succeed as written.

The dispatcher has to retry with the full set of required params (effectively repeating step 0's `agent_id`/`description`/`chat_id` plus step 1's `agent_type`/`claude_session_id`) before it succeeds.

## Affected doc

`.claude/sys.dispatcher.bootup.md`, "Startup Behavior" section, step 1:

> 1. Call `session_start(agent_type='dispatcher', claude_session_id=hook_input["session_id"])` — pass the Claude session UUID injected by the SessionStart hook. ...

Note step 0 (immediately above it, same file) already does this correctly:

> 0. Call `session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", description="Lobster dispatcher main loop", chat_id=<ADMIN_CHAT_ID>)` ...

Step 1 appears to be a second call meant to additionally record `claude_session_id` on the same row, but since `session_start` replaces the row wholesale (per its docstring: "If an entry with the same id already exists, it is replaced"), the second call needs to repeat the same required fields from step 0, not just add `claude_session_id`.

## Likely fix

Update step 1's example call to include the same `agent_id`, `description`, `chat_id` used in step 0, e.g.:

```
session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", description="Lobster dispatcher main loop", chat_id=<ADMIN_CHAT_ID>, claude_session_id=hook_input["session_id"])
```

(or merge steps 0 and 1 into a single call, if `hook_input["session_id"]` is available at that point — would remove one redundant tool call from every dispatcher startup.)

This is a minor doc/API mismatch — not urgent, but it means every single dispatcher startup currently either errors-and-retries on step 1, or (if the agent doesn't correct itself) silently skips registering `claude_session_id`, breaking the `inject-bootup-context.py` dispatcher-detection path described in the same step.

aeschylus:
## Summary
- `sys.dispatcher.bootup.md` step 1 and `sys.compact-dispatcher.bootup.md` step 1 called `session_start(agent_type='dispatcher', claude_session_id=hook_input["session_id"])`, but the MCP tool requires `agent_id`, `description`, and `chat_id` (it replaces the DB row wholesale). Following the doc literally fails immediately with `Input validation error: 'agent_id' is a required property`.
- Fixed both step-1 examples to repeat the same `agent_id`/`description`/`chat_id` used in step 0, plus `claude_session_id`, matching the tool's actual required schema.

Fixes #2139

## Test plan
- [x] `uv run -m pytest tests/unit/test_hooks/test_dispatcher_bootup_single_pass_read.py tests/unit/test_hooks/test_session_role_state_file.py -q` — 37 passed, no regressions
- [x] Confirmed no test asserts on the exact old (broken) call string
- [x] Manually diffed against `src/agents/session_store.py` `session_start()` signature (`id`/`description`/`chat_id` non-defaulted) and the MCP `session_start` tool schema (`required: [agent_id, description, chat_id]`) to confirm the new example call is valid


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS: Verified independently, not just taken on the engineer's word.

- Pulled the actual `session_start` MCP tool schema (src/mcp/inbox_server.py) — `required: [agent_id, description, chat_id]` confirmed exactly. The new example calls in both docs supply all three, matching the real schema.
- Ran `uv run -m pytest tests/unit/test_hooks/test_dispatcher_bootup_single_pass_read.py tests/unit/test_hooks/test_session_role_state_file.py -q` myself — 37 passed, matches the claim exactly, no discrepancy.
- Checked both edited files (sys.dispatcher.bootup.md line 39, sys.compact-dispatcher.bootup.md line 20) — same fix applied consistently to both step-1 calls, both now mirror step 0's required params plus `claude_session_id`.
- Confirmed the two passing test files don't actually assert on the doc's example string (they test SessionStart-hook state-file mechanics, not markdown content) — engineer's claim that "no test asserted on the old literal string" is accurate but this also means there's no automated guard against this class of doc/schema drift recurring. Not a blocker for this fix, just a gap worth noting for the future (e.g. a lint that parses example `session_start(...)` calls in .claude/*.md against the live tool schema).
- Grepped for other literal occurrences of the old broken call pattern elsewhere in the repo (hooks/inject-bootup-context.py, src/mcp/inbox_server.py comments) — those are shorthand comments describing the mechanism with a `<uuid>` placeholder, not copy-paste instructions a session would execute, so they're out of scope and not broken in the same way.
- No behavior/runtime code changed — pure documentation fix, correctly scoped, matches the linked issue exactly.

Issue #2142, opened 2026-07-31
wait_for_messages aborts after 300s idle — possible root cause of dispatcher crash-loop (#2124)


sayhar:
## What happened

Immediately after a normal dispatcher startup (session started cleanly, catchup subagent completed successfully, inbox was fully drained), the dispatcher called `wait_for_messages()` (`mcp__lobster-inbox__wait_for_messages`) with no arguments (default `timeout=72000s` / 20h, intended to long-poll until a message arrives). Three consecutive calls all failed with the identical client-side error:

```
MCP server "lobster-inbox" tool "wait_for_messages" sent no response or progress for 300s; aborting. If this server is configured in your MCP settings, set a per-server "timeout" (ms) to allow longer silent runs for just this server; otherwise set CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT (ms) globally (0 disables).
```

This happened at approximately 2026-07-31 20:14-20:20 UTC, in a session that had JUST cleanly recovered from the long-standing dispatcher crash-loop (300+ restarts, tracked in issue #2124). No user messages were pending at the time — the inbox was empty and idle, which is exactly the scenario `wait_for_messages` is designed to sit blocked in for up to 20 hours.

## Hypothesis (not confirmed root cause)

The Claude Code MCP client appears to enforce an idle/no-progress abort at ~300s (5 min) on tool calls, distinct from and shorter than the tool's own declared `timeout` parameter (default 72000s). Since `wait_for_messages` is a long-poll that legitimately sends no progress notifications while blocked waiting for messages, any period of >5 minutes with no inbound message traffic would cause this client-side abort to fire — which would throw an unhandled tool error mid-main-loop. If the dispatcher's harness treats an unhandled MCP tool error as fatal (session-ending), this would produce exactly the crash-loop symptom already tracked in issue #2124 (300+ restarts, closely spaced, no correlation with user activity) — because any 5+ minute gap in message traffic (e.g. overnight, or during Alex's Jul 23-26 vacation) would trigger it repeatedly.

This is presented as a hypothesis, not a confirmed root cause. It may be related to #2124 but that is not established — see "Relevant context" below.

## Steps to reproduce

1. Start a fresh dispatcher session with an empty inbox (no pending messages).
2. Call `wait_for_messages()` with default args (or explicit `timeout=72000`).
3. Wait 5+ minutes with no new messages arriving.
[...]

Issue #2124, opened 2026-07-20
3 AM 'Lobster Starting' alert wakes user nightly — suppress overnight startup pings or fix root crash


sayhar:
## User Report

Every night at approximately 3 AM, the user receives a "Lobster Starting / Claude persistent session initializing" Telegram notification. This is disruptive — it wakes them up.

## What's Actually Happening

The alert is not a false alarm. It is correctly signaling that the dispatcher is crashing and restarting nightly. This is the known crash-loop bug:

- **#2100** (open): needrestart auto-restart fix — partially addressed but loop not fully resolved
- **#2095** (merged): staggered nightly-consolidation cron to 03:02 to avoid collision with health-check — crash-loop persists despite this fix
- The dispatcher has accumulated 340+ total restarts; the nightly crash pattern has been active since at least Jul 19

## Two Valid Resolutions

**Option A — Suppress overnight startup notifications (short-term UX fix)**
Add logic so that "Lobster Starting" startup pings are suppressed during overnight hours (e.g., 11 PM–7 AM) unless `LOBSTER_DEBUG=true`. The bootup file already mentions `LOBSTER_DEBUG` as a flag that affects notification behavior — this extends that pattern.

**Option B — Fix the root crash so 3 AM restarts stop happening (real fix)**
Resolve whatever is causing the dispatcher to crash nightly. This is the correct long-term fix; Option A would just hide a symptom.

## Notes

- The alert mechanism itself is working correctly — the problem is either the root crash or the notification policy
- These two options are not mutually exclusive; both could be implemented
- User explicitly flagged this as disruptive: "this is waking me up every night"

## Related Issues / PRs

- #2100 — needrestart auto-restart fix (open)
- #2095 — stagger nightly-consolidation cron (merged, did not resolve crash-loop)
- #2075 — user-friendly Telegram notification before graceful restart (open)

sayhar:
## Investigation update (2026-08-01, background audit of a 4-restart cluster)

Investigated a fresh cluster of 4 "SESSION LOST" / dispatcher-marked-dead restarts between 2026-07-31 18:12 EDT and 2026-08-01 00:24 EDT, zero user activity in between. Found a concrete, reproducible root cause and opened a fix: **#2143**.

[...]

PR #2143, opened 2026-08-01
fix: correct systemctl invocation in MCP orphan-reaper guard

sayhar:
## Summary
- `kill_orphaned_mcp_processes()` in `scripts/claude-persistent.sh` uses `systemctl status --pid "$pid"` to skip systemd-managed MCP servers before treating remaining processes as orphans to SIGTERM.
- `--pid` is not a valid `systemctl status` option. It always fails (`Unknown option '--pid'.`, exit 1), so the "skip systemd-managed service" branch has never actually fired since it was added in #1551. `lobster-mcp-local.service` (PPID=1 under systemd) falls through the tmux-lineage check too, so it's classified as an orphan and SIGTERM'd on **every** `launch_claude()` call — including the routine ~2h proactive session-age restart from #2060.
- `RestartSec=0` (drop-in `restart-fast.conf`) masks the worst symptom (tools failing to load), but each bounce of `lobster-mcp-local.service` still wipes its in-memory session/dispatcher-lock state, which writes a `session-lost-reminder` to the inbox and — depending on timing — marks the still-alive dispatcher session "dead" via the `>120min` stale-`running`-session cleanup in `session_store.py`. That surfaces to the user as a spurious "SESSION LOST" / dispatcher-marked-dead restart notification even though the dispatcher never actually crashed.

## Evidence
Investigated a cluster of 4 such restart notifications between 2026-07-31 18:12 EDT and 2026-08-01 00:24 EDT with zero user activity in between. `journalctl -u lobster-mcp-local` shows the service cleanly restarting (`Deactivated successfully`) at 18:12:07, 20:16:07, 22:20:07, 00:24:07 — each exactly 7440s apart and within ~6s of `health-check-v3.sh`'s `check_session_age()` sending the proactive SIGTERM to the dispatcher PID (confirmed in `health-check.log`: `SESSION AGE LIMIT: session is 7430s old ... sending SIGTERM`).

Reproduced directly on the host:
```
$ systemctl status --pid $(pgrep -f inbox_server.py)
systemctl: unrecognized option '--pid'
$ echo $?
1
$ systemctl status $(pgrep -f inbox_server.py)   # bare PID — correct syntax
● lobster-mcp-local.service - Lobster MCP Server (local HTTP transport)
     Active: active (running) ...
```

This is a plausible root cause (or contributor) to the long-standing nightly crash-loop tracked in #2124 — it fires on a fixed ~2h cadence independent of user activity, which matches the "3 AM alert" pattern whenever a cycle lands near 3 AM, and explains why prior mitigations in #2124 (cron staggering, alert suppression) didn't stop it: none of them touched this orphan-reaper guard.

Also relevant to #2142 — the wait_for_messages 300s-idle-abort hypothesis. I did not find evidence of that specific client-side abort in this cluster's logs (wait_for_messages calls were logging normally throughout), so it may be a separate/lower-priority issue; this PR does not address it.

## Fix
Pass the PID positionally: `systemctl status "$pid"` (no `--pid` flag), matching the intent of the guard added in #1551.

## Test plan
- [ ] Confirm `systemctl status "$pid"` correctly returns 0 for `lobster-mcp-local.service`'s PID and non-zero for a non-systemd dispatcher-spawned MCP process (e.g. an `npx obsidian-mcp` child) in a staging/dev environment.
- [ ] After deploy, monitor `journalctl -u lobster-mcp-local` across several ~2h proactive session-age restart cycles and confirm the service no longer bounces (`Deactivated successfully` / restart counter no longer increments) when the dispatcher does its routine 7200s SIGTERM restart.
- [ ] Confirm `CLEANUP: MCP PID $pid is a systemd-managed service — skipping` now appears in `claude-persistent.sh` logs on restart, instead of `CLEANUP: Killing orphaned MCP PID $pid (SIGTERM)`.
- [ ] Watch for a reduction/elimination of `session-lost-reminder` and dispatcher-marked-dead reconciler notifications correlated with routine restarts.

Related: #2124, #2142, #1551, #2060

### DIFF
--- a/scripts/claude-persistent.sh
+++ b/scripts/claude-persistent.sh
@@ -377,7 +377,20 @@ kill_orphaned_mcp_processes() {
         # Skip systemd-managed services — they are independently lifecycle-managed,
         # not orphans. Killing them causes a restart race where Claude initializes
         # before the MCP server comes back up, leaving the session without MCP tools.
-        if systemctl status --pid "$pid" >/dev/null 2>&1; then
+        #
+        # NOTE: `systemctl status` has no `--pid` option. `systemctl status --pid
+        # "$pid"` always fails with "Unknown option '--pid'." (exit 1), which
+        # silently defeated this guard for every PID since it was added in #1551:
+        # the "systemd-managed" branch never triggered, so lobster-mcp-local.service
+        # was always treated as an orphan and SIGTERM'd here on every dispatcher
+        # restart (including the routine ~2h proactive session-age restart from
+        # #2060). RestartSec=0 masked the user-visible impact of the restart
+        # itself, but the service bounce still wiped in-memory session/lock state,
+        # producing spurious session-lost-reminder / dispatcher-marked-dead
+        # notifications with no actual dispatcher crash. See investigation for
+        # issues #2124 / #2142. The correct invocation takes the PID as a bare
+        # positional argument.
+        if systemctl status "$pid" >/dev/null 2>&1; then
             log "CLEANUP: MCP PID $pid is a systemd-managed service — skipping"
             skipped=$((skipped + 1))
             continue


sayhar (commented):

**The core fix is correct and well-evidenced.** I independently reproduced both halves on the live host: `systemctl status --pid $(pgrep -f inbox_server.py)` → `Unknown option '--pid'.` (exit 1) — confirming the guard has been a silent no-op since #1551. `systemctl status $(pgrep -f inbox_server.py)` (positional, as fixed) → exit 0, correctly prints `lobster-mcp-local.service`. The root-cause narrative (fixed ~7440s cadence, timing against health-check-v3.sh's proactive SIGTERM) is solid and the diff is minimal/well-commented.

**But the replacement check is too broad and risks a regression the other direction.** `systemctl status "$pid"` succeeding only means "$pid" belongs to the cgroup of *some* loaded unit — not that it *is* `lobster-mcp-local.service` specifically. I confirmed this live: a plain `setsid sleep 60` process I spawned in an unrelated shell, and an ordinary Bash-tool subprocess, both report `/proc/<pid>/cgroup` = `/system.slice/lobster-claude.service` and both make `systemctl status "$pid"` exit 0 — because they're descendants living inside the dispatcher's own service cgroup, not because they're independently-managed services.

`kill_orphaned_mcp_processes()` targets two patterns: `inbox_server.py` (normally its own unit, fine) and `obsidian-mcp` via npx — which is *always* a direct child of the `claude` process, i.e. it lives inside `lobster-claude.service`'s cgroup, never its own unit. Under the fix as written, a genuinely orphaned `obsidian-mcp` process (parent `claude` died, child leaked) will still resolve `systemctl status "$pid"` to `lobster-claude.service` and get skipped as "systemd-managed" — short-circuiting *before* the tmux-lineage check that used to correctly catch it (that check only worked before because the systemd guard was always failing). That's the exact leak this function's own top-of-file comment says it exists to prevent (unbounded orphaned MCP process accumulation), just reintroduced from the opposite direction.

Suggest checking the matched unit's identity, not just exit status, e.g.:
```bash
unit=$(systemctl show -p Id --value "$pid" 2>/dev/null)
if [[ "$unit" == "lobster-mcp-local.service" ]]; then
```

**No test coverage exists for this at all** — searched the repo, there's no bats/shell test harness for `claude-persistent.sh` or this function. The PR's test plan is 4 unchecked manual post-deploy checkboxes, and none of them check whether real orphans (e.g. a leaked `obsidian-mcp` child) still get reaped after this change — only whether the specific systemd-service false-positive-kill is fixed. That's the gap that would have caught the above.

Net: the fix correctly resolves the confirmed high-frequency crash-cluster cause and should probably still ship, but the cgroup-membership check needs tightening to the specific unit name before merge, or at minimum a follow-up issue should be filed to verify obsidian-mcp orphan reaping still works post-deploy.

PR #2146, opened 2026-08-02
fix: make compaction recovery notification deterministic (#1983)

### LINKED ISSUE #1983: fix: make compaction recovery notification deterministic (not dispatcher-manual)

aeschylus:
## Problem

The debug-mode "🔄 Back online. Context recovered from X to Y." notification was sent by the dispatcher manually, after it received `compact-catchup`'s `write_result`, gated on `LOBSTER_DEBUG=true`. This made the signal contingent on dispatcher behavior:

1. If the dispatcher processes the result correctly → notification fires.
2. If the dispatcher skips the step, crashes before reaching it, or mishandles the result → silence, with no error raised.

A signal that fires "most of the time" is worse than no signal at all: when it fires, it looks like confirmation that recovery succeeded; when it doesn't, there's no way to tell whether recovery failed or the dispatcher just forgot to notify. Issue #1983 calls this epistemic rot.

This recreates the fix from stale PR #2043 (opened May 8, functionally untouched for 3 months, closed today as conflicting against current main) fresh against today's main — same design, verified independently against the current dispatcher/agent files rather than rebased through 3 months of drift.

## What changed

**`.claude/agents/compact-catchup.md`** — new Phase 5: after Phases 1–4 complete, if `LOBSTER_DEBUG=true`, the agent sends the recovery notification directly to `ADMIN_CHAT_ID` (with `proactive=True`, per the `require-reply-to-message-id` hook's exemption) before calling `write_result`. The notification now fires — or doesn't — based on the agent's own completion, not on downstream dispatcher handling. Failure to send (missing admin chat id, tool error) is swallowed and never blocks `write_result`. The "Do NOT call `send_reply`" rule is updated to carve out this one exception.

**`.claude/sys.dispatcher.bootup.md`** — the manual notification instructions (6 lines: check `LOBSTER_DEBUG`, build the message, convert timestamps to ET) are replaced with a one-line note that the agent already handled it.

```
Before:
  compact-catchup agent -> write_result(chat_id=0)
  -> dispatcher receives result -> reads msg["text"]
  -> IF LOBSTER_DEBUG: dispatcher sends "Back online" (may be skipped/missed)
  -> dispatcher mark_processed

After:
  compact-catchup agent
  -> Phase 5: send_reply("Back online", proactive=True) if LOBSTER_DEBUG (always attempted)
  -> write_result(chat_id=0)
  -> dispatcher receives result -> reads msg["text"]
  -> dispatcher mark_processed (no send needed -- agent already sent it)
```

## Functional patterns

Phase 5 is a pure addition with no side effects on Phases 1–4 — it reads from state already computed earlier in the same agent run (window_start/now/N/M) and is wrapped in a swallow-and-continue failure boundary, so a notification failure never mutates or blocks the rest of catchup's work (session file, rolling summary, commitment carry-forward).

## Out of scope

- Does not touch the other 3 PRs from the same stale batch (#2034/#1895, #2047/#2003, #2042/#1998) — those are being assessed separately.
- Does not change the `LOBSTER_DEBUG` gate itself, or anything about *what* the notification says — only *where* it's sent from.

## Tests run

Both `compact-catchup.md` and `sys.dispatcher.bootup.md` are LLM-instruction files, not executable Python — there's no function to unit-test directly. Following the existing precedent in this repo (`tests/unit/test_hooks/test_dispatcher_bootup_single_pass_read.py`), the automated tests are characterization tests over the file content: they assert the specific instruction text that encodes the new/old behavior is present/absent.

- [x] `uv run -m pytest tests/unit/test_hooks/test_compact_catchup_notification_determinism.py -q` — 6/6 passed (new test file)
- [x] Falsifiability check: `git stash` the two `.claude/*.md` edits (reverting to pre-fix content), reran the same test file — 4/6 failed (the two file-existence sanity checks still passed, as expected). Restored the edits (`git stash pop`) and confirmed 6/6 green again.
- [x] `uv run -m pytest tests/unit/test_hooks/ -q` — 621 passed, 8 pre-existing failures (`test_context_monitor.py::test_settings_json_uses_fixed_matcher` and 7 in `test_on_compact_write_claude_session_id.py`) — same failures present on a clean main checkout before this diff (confirmed while reviewing PR #2145 immediately prior to this one), unrelated to this change.

## Manual test

No live compaction was triggered — this is a debug-only Telegram notification, and the change is scoped to two markdown instruction files with no Python code path to execute. Verified by direct diff inspection: the 6-line manual-send instruction in `sys.dispatcher.bootup.md` is gone, replaced by a 1-line note; the new Phase 5 in `compact-catchup.md` reproduces the same message template, ET conversion requirement, and `LOBSTER_DEBUG` gate that the old dispatcher-side code had, just relocated.

User-visible outcome: not independently verified against a live Telegram send (would require a real compaction with `LOBSTER_DEBUG=true`). This mirrors how the same limitation was handled and disclosed in the original attempt (PR #2043).

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits — no Telegram call in the characterization tests)
- [ ] Integration/manual test — not run; requires a live compaction, no test environment exists for this. Documented above rather than skipped silently.
- [ ] User-visible outcome — not independently verified (see Manual test)
- [x] Side-effect audit: no floods, no unscoped writes — this only relocates one existing debug-gated Telegram send from the dispatcher to the agent; send volume (at most one message per compaction when LOBSTER_DEBUG=true) is unchanged
- [x] External service calls: `send_reply` is an existing, already-tested MCP tool; no new external call surface introduced

Closes #1983

### DIFF
--- a/.claude/agents/compact-catchup.md
+++ b/.claude/agents/compact-catchup.md
@@ -216,6 +216,41 @@ After Phase 3, verify that open commitments in the session notes are also captur
     ```
     If `rolling-summary.md` could not be read or written during Phase 4, note the failure but do not abort catchup.
 
+### Phase 5: Debug recovery notification
+
+The "🔄 Back online" recovery notification used to be sent by the dispatcher
+manually after receiving this agent's `write_result` (only when
+`LOBSTER_DEBUG=true`). That made the signal contingent on dispatcher
+behavior: if the dispatcher skipped the step, crashed before reaching it, or
+mishandled the result, the notification silently never fired — while looking,
+when it did fire, like confirmation that recovery succeeded. A signal that
+fires "most of the time" is worse than no signal: it creates false confidence
+on success and ambiguous silence on failure (issue #1983).
+
+22. After Phase 4 completes (and before calling `write_result`), check
+    `LOBSTER_DEBUG`. If it is not `true`, skip this phase entirely — proceed
+    directly to `write_result`.
+
+23. If `LOBSTER_DEBUG=true`, determine ADMIN_CHAT_ID:
+    - Primary: `LOBSTER_ADMIN_CHAT_ID` environment variable.
+    - Fallback: the first entry in `TELEGRAM_ALLOWED_USERS` from `config.env`.
+    - If neither is available, skip this phase silently (never block
+      `write_result` on a missing admin chat id).
+
+24. Convert `window_start` and `now` (from Phase 1) from UTC ISO to ET
+    (EDT UTC-4 mid-March through early November, EST UTC-5 otherwise) and
+    send, via `send_reply`, to ADMIN_CHAT_ID:
+    `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."`
+    (N and M come from the same counts used in the Phase 1 output.) Pass
+    `proactive=True` — this send has no originating user message to thread
+    against (see `hooks/require-reply-to-message-id.py`).
+
+25. This phase never blocks or fails catchup: if the send errors for any
+    reason (missing admin chat id, tool failure), swallow the error and
+    continue to `write_result` as normal. The notification is best-effort;
+    the rest of catchup's work (session file, rolling summary) must not be
+    lost because of it.
+
 ## Session notes reading
 
 Read session notes from `~/lobster-user-config/memory/canonical/sessions/` in tiers:
@@ -301,7 +336,7 @@ Keep each line to one sentence. The dispatcher is on mobile -- brevity matters.
 
 ## Rules
 
-- Do NOT call `send_reply` -- this is internal context recovery, not a user message.
+- Do NOT call `send_reply` -- this is internal context recovery, not a user message -- **except** the Phase 5 debug recovery notification (`LOBSTER_DEBUG=true` only), which this agent sends directly so it fires deterministically regardless of dispatcher behavior.
 - Do NOT relay catch-up content to the user unless an event is urgent (e.g. a failed subagent that the user has not been notified about).
 - If `check_inbox` returns no messages in the window, that is valid -- report "Nothing to report" in the inbox section but still populate the session file.
 - If `compaction-state.json` is missing or corrupt, default to scanning the last 6 hours.
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -300,11 +300,10 @@ After a context compaction you lose situational awareness of the last ~30 minute
 
 **When the compact_catchup result arrives** (`task_id: "compact-catchup"`, `chat_id: 0`):
 - Read `msg["text"]` to restore situational awareness
-- Do NOT send_reply — this is internal context, except:
-  - If `LOBSTER_DEBUG=true`: send a brief status to ADMIN_CHAT_ID:
-    `"🔄 Back online. Context recovered from [window_start] to [now]. [N messages] processed, [M subagents] were running."`
-    (Fill in N and M from `msg["text"]`. ADMIN_CHAT_ID is injected at startup — read it from the top of your context.)
-    **Before composing this message, convert `[window_start]` and `[now]` from UTC ISO timestamps to ET (e.g. "5:29 AM ET"). Rule: EDT (UTC-4) mid-March through early November, EST (UTC-5) otherwise. Never send raw UTC ISO strings to the user.**
+- Do NOT send_reply — this is internal context. The debug-mode "🔄 Back online" recovery
+  notification (issue #1983) is sent by the `compact-catchup` agent itself (Phase 5, `LOBSTER_DEBUG=true`
+  only) before it calls `write_result` — no dispatcher action needed, and this fires deterministically
+  regardless of dispatcher behavior.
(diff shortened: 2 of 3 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS — with one substantive design note and a minor nit, neither blocking.

**What I verified independently** (checked out the PR branch locally, ran things myself rather than trusting the PR body):

1. **Falsifiability is real, not just asserted.** Reverted the two `.claude/*.md` files to their pre-fix content, reran `tests/unit/test_hooks/test_compact_catchup_notification_determinism.py`: 4 of 6 tests failed (only the two trivial file-existence checks stayed green), exactly as the PR claims. Restored the files: 6/6 green again.
2. **The "621 passed, 8 pre-existing failures" claim checks out exactly**, scoped correctly to `tests/unit/test_hooks/` (I ran it myself — identical count, identical failing test names: `test_context_monitor.py::test_settings_json_uses_fixed_matcher` + 7 in `test_on_compact_write_claude_session_id.py`). I also ran the same failures against a clean `main` checkout (`da602cdf`) and confirmed they're present there too — genuinely pre-existing, not introduced by this diff.
3. **No half-migration / no double-send risk.** Grepped the whole repo: the old "send a brief status to ADMIN_CHAT_ID" instruction is completely gone from `sys.dispatcher.bootup.md` (survives only inside the new test's assertion string, which checks its *absence*). The "🔄 Back online. Context recovered..." template now exists in exactly one place — the new Phase 5 in `compact-catchup.md`. No production Python code sends this message either, so there's no code-level duplicate sender to worry about.
4. **`proactive=True` is real, not invented.** It's an existing, tested exemption in `hooks/require-reply-to-message-id.py` (and wired into `send_reply` in `src/mcp/inbox_server.py`), correctly referenced.
5. Ran the full repo test suite on both branches — same ~35 pre-existing unrelated failures (whatsapp health check, slack router config, push-token endpoint tests, etc.) on both `main` and this branch. No new failures introduced.

**Design note (not a bug, but worth surfacing):** issue #1983 frames the fix as needing to eliminate a whole class of failure ("the signal either always fires... or does not exist at all") and lists three options, explicitly including a scripting-layer hook (Option B) or agent+hook backstop (Option C) as more robust than the agent self-sending alone (Option A). This PR implements Option A only. That's a legitimate choice within the issue's own menu, and it does shrink the failure surface (the dispatcher no longer has to correctly parse `write_result` and remember a conditional step). But it doesn't make the notification *mechanically* guaranteed — it moves the "must correctly follow instructions" risk from the dispatcher into a 25-step natural-language phase list inside a different LLM agent. If `compact-catchup` times out, errors, or simply misexecutes steps 22-25 before reaching `write_result`, the exact same silent-failure mode reappears one layer down. Since these are instruction files with no code enforcement, that residual risk can't be closed by a unit test — worth being explicit in the issue/PR that "deterministic" here means "no longer contingent on the dispatcher," not "provably cannot fail."

**Minor nit:** Phase 5's message template (`[N messages] processed, [M subagents] were running`) doesn't map cleanly onto Phase 1's actual structured output (three separate counts: User messages/Subagent results/System events, plus a separate in-flight-subagents list) — which N and which M is left to inference. This ambiguity is inherited unchanged from the original pre-existing dispatcher-side instruction (I checked — the old text had the identical vagueness), so it's not a new problem introduced here, just not cleaned up in the migration.

Diff itself is small and well-isolated (2 doc files + 1 new test file, no production code touched), so blast radius is low regardless of the design note above.

Issue #2039, opened 2026-05-09
fix: duplicate reflection messages from concurrent hook invocations and single-file mark_processed


sayhar:
## Summary

Two bugs combine to cause duplicate reflection messages appearing in the dispatcher inbox on every startup when `LOBSTER_DEBUG=true`. The dispatcher sees the same reflection message 4–6 times even after marking it processed.

Related to #1998 (reflection_bootup messages require unnecessary overhead).

---

## Bug 1 — Concurrent hook invocations write duplicate-ID files

**File:** `hooks/on-compact.py` and `hooks/on-fresh-start.py`  
**Function:** `_schedule_reflection_prompt(trigger)`

When multiple subagent SessionStart events fire within the same second, each concurrent hook invocation computes `msg_id = f"reflection_{trigger}_{int(ts)}"` using 1-second precision. Because `int(ts)` rounds down to the same integer for all invocations within the same second, they all produce the **same `id` field** — but write to **different filenames** (using millisecond-precision `ts_ms`).

The dedup check in `_schedule_reflection_prompt` is missing entirely. The `already_pending()` function in `on-compact.py` only checks for `compact-reminder` subtype, not reflection files. So all N concurrent invocations write without dedup.

Additionally: `LOBSTER_MAIN_SESSION=1` is inherited by all subagents, and `_is_dispatcher_compact()` uses it as a fallback without guards. This causes subagent SessionStart events to be classified as dispatcher compaction events, triggering `_schedule_reflection_prompt("compaction")` from each subagent's hook invocation.

**Fix A:** Before writing the reflection file in `_schedule_reflection_prompt()`, scan `~/messages/inbox/`, `~/messages/processing/`, and `~/messages/processed/` for any file whose `id` field matches the computed `msg_id`. If one exists, skip writing.

---

## Bug 2 — `mark_processed` archives only one file per ID

**File:** `src/mcp/inbox_server.py`  
**Function:** `handle_mark_processed` (calls `_find_message_file`)

`_find_message_file()` returns the **first** file whose `id` field matches. `handle_mark_processed` moves only that file to `processed/`. The remaining N-1 files with the same `id` stay in `inbox/` and re-appear on the next `wait_for_messages` call.

This is why the dispatcher sees the same message 4–6 times even after marking it processed.

**Fix B:** After finding and archiving the first matching file, scan all remaining files in `inbox/` and `processing/` for additional files with the same `id` and archive them all.

---

## Fix scope

[...]

PR #2147, opened 2026-08-02
fix(startup): write reflection prompts to a sidecar file, not the inbox (#1998)

### LINKED ISSUE #1998: fix(startup): reflection_bootup messages require unnecessary claim+process overhead

aeschylus:
## Problem

Each restart, `inject-bootup-context.py`'s siblings (`on-compact.py` / `on-fresh-start.py`) generate a debug-mode `reflection_bootup_XXXX` message asking the dispatcher to reflect on the bootup/compaction experience. That message lands in the inbox like any other message: the dispatcher must `mark_processing` it and, after reading it once, `mark_processed` it — 2 extra MCP round-trips on the startup critical path, purely to deliver a one-shot prompt that's read once and discarded. This adds latency and inbox noise without providing value the dispatcher couldn't get another way.

This recreates the fix from stale PR #2042 (opened May 9, functionally untouched for 3 months, closed today as conflicting against current main) fresh against today's main.

## What changed

**`hooks/on-compact.py`, `hooks/on-fresh-start.py`** — `_schedule_reflection_prompt()` now writes the prompt to `~/messages/bootup-prompt.md` (atomic tmp-then-rename), still gated on `LOBSTER_DEBUG=true`, instead of dropping a JSON message into the inbox. Removed `_reflection_already_exists()` from both hooks: that ID-based dedup existed only to prevent duplicate *inbox* messages from concurrent hook invocations (issue #2039); the sidecar file is a single location that's overwritten unconditionally and read-and-deleted on every startup, so last-writer-wins is correct and no dedup bookkeeping is needed.

**`.claude/sys.dispatcher.bootup.md`** — new startup step 2e: read `~/messages/bootup-prompt.md` if it exists, reflect genuinely (file/update GitHub issues or open PRs for real observations, do nothing if none), then delete the file. One `Read` call, zero claim/process overhead. The old `reflection_prompt` (`type: "reflection_prompt"`) inbox-message handler is kept only as an explicitly-marked-legacy fallback — if a stale message from a pre-#1998 hook version is ever dequeued, the dispatcher claims and marks it processed without reflecting on it, since step 2e is now the one canonical reflection path (reflecting twice would risk duplicate or contradictory GitHub activity).

```
Before:
  on-compact.py/on-fresh-start.py -> write reflection_prompt message to inbox
  -> dispatcher wait_for_messages() returns it like any other message
  -> mark_processing -> read -> reflect -> mark_processed  (2 extra round-trips)

After:
  on-compact.py/on-fresh-start.py -> write ~/messages/bootup-prompt.md (atomic)
  -> dispatcher startup step 2e: Read the file, reflect, delete it
  -> (no claim/process overhead; nothing enters the message queue)
```

## Functional patterns

`_schedule_reflection_prompt()` remains a pure side-effect-isolated function (build content string, atomic write) with no branching added relative to before — the write target changed, not the control flow. Removing `_reflection_already_exists()` is a net simplification: dead code once dedup-by-ID is no longer needed for a single-slot, overwritten-in-place file.

## Out of scope

- Does not touch the other 2 PRs from the same stale batch (#2043/#1983 — separate PR #2146; #2047/#2003 — left as an open assessment, not recreated).
- Does not change what the reflection prompt asks or how the dispatcher decides whether an observation is worth filing — only where the prompt is delivered from.

## Tests run

`sys.dispatcher.bootup.md` is an LLM-instruction file, not executable Python — there's no function to unit-test directly for step 2e. Following the existing repo precedent (`tests/unit/test_hooks/test_dispatcher_bootup_single_pass_read.py`), the new doc-behavior tests are characterization tests asserting on file content.

- [x] `uv run -m pytest tests/unit/test_hooks/test_reflection_prompt.py -q` — 16/16 passed (rewritten for the sidecar-file behavior: debug-gating, content, atomic write, overwrite-not-accumulate, and 2 regression guards confirming `_reflection_already_exists` is gone from both hooks)
- [x] `uv run -m pytest tests/unit/test_hooks/test_bootup_prompt_sidecar_dispatcher_flow.py -q` — 4/4 passed (new file: asserts step 2e references `bootup-prompt.md`, the legacy handler is marked legacy, and it no longer instructs genuine reflection on a dequeued message)
- [x] Falsifiability check, Python tests: `git stash` the 3 implementation files (hooks + doc), reran `test_reflection_prompt.py` — 9/16 failed (the 7 that still passed are non-discriminating sanity/env-gating checks that hold under both old and new behavior). Reran `test_bootup_prompt_sidecar_dispatcher_flow.py` — 3/4 failed. Restored (`git stash pop`) — both files back to fully green.
- [x] `uv run -m pytest tests/unit/test_hooks/ tests/unit/test_mcp_server/ -q` — 1559 passed, 14 pre-existing failures (same 14 as confirmed on a clean main checkout immediately prior, while reviewing PR #2145/#2146 — unrelated to this diff: `test_context_monitor.py`, `test_on_compact_write_claude_session_id.py`, and the 3 `test_push_*_token_endpoint.py` files)
- [x] `python3 -m py_compile hooks/on-compact.py hooks/on-fresh-start.py` — OK

## Manual test

Ran `_schedule_reflection_prompt("bootup")` directly against `hooks/on-fresh-start.py` (loaded as a module, `session_role` stubbed, `BOOTUP_PROMPT_FILE` pointed at a scratch `/tmp` path, `LOBSTER_DEBUG=true`): confirmed the sidecar file is written with the expected content ("[Debug] Bootup reflection prompt...", the friction/observations question, the `SiderealPress/lobster` pointer) and no `.tmp` file is left behind. Cleaned up the scratch directory afterward.

No live dispatcher restart was performed — step 2e is a markdown instruction with no Python code path to execute in isolation from a real Claude Code session. Verified by diff inspection: the new step reads and deletes the same path (`~/messages/bootup-prompt.md`) the hooks now write to, and the legacy inbox handler no longer instructs reflection.

User-visible outcome: not applicable — this is an internal dispatcher-startup efficiency fix with no direct user-facing message change (the reflection prompt itself was never shown to the user; it's an internal debug-mode self-review nudge).

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits — all tests use `tmp_path` / scratch files, never the real `~/messages/` tree)
- [ ] Integration/manual test — partial (see Manual test): the hook-side write was verified directly; the dispatcher-side read/delete step cannot be exercised outside a live Claude Code session
- [x] User-visible outcome — not applicable, explained above
- [x] Side-effect audit: no floods, no unscoped writes — this changes where one existing debug-gated file write goes (inbox message vs. sidecar file), same trigger conditions and frequency (at most once per compaction/restart when `LOBSTER_DEBUG=true`)
- [x] External service calls: none — purely local file I/O

Closes #1998

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -54,6 +54,10 @@ When you first start (or after reading this file), follow these steps:
     - `startup-cause: restart` → this was a plain restart (systemd, external kill, or health-check). No compact-reminder will be in the inbox. Spawn `startup-catchup` at step 4 for a normal restart window.
     - Skip if step 2c already sent a restart notification (context-handoff.json was recent).
     - **Do not use `compaction-state.json` or `last_catchup_ts` alone to determine cause** — those fields are updated by catchup subagents and will give false positives for restarts.
+2e. **Check for a debug-mode reflection prompt sidecar** (issue #1998): read `~/messages/bootup-prompt.md`.
+    - If the file does not exist: skip, nothing to do.
+    - If it exists: read its content, then delete it (`rm -f ~/messages/bootup-prompt.md`) — this is a one-shot prompt, not a persistent message. Reflect genuinely on the bootup/compaction experience it asks about: were there friction points, gaps, or improvements worth capturing? If there are substantive observations, file or update GitHub issues in SiderealPress/lobster, or open PRs for straightforward fixes. If nothing is worth capturing, do nothing further — silence is the correct response.
+    - This replaces the old flow where `on-compact.py`/`on-fresh-start.py` wrote a `reflection_prompt` inbox message that the dispatcher had to `mark_processing` + `mark_processed` (2 extra MCP round-trips per restart, only to be read once and discarded). One `Read` call here does the same job with zero claim/process overhead.
 
 3. **Claim any pending user messages immediately** to stop the health-check staleness clock:
     - Call `check_inbox()` to get any messages currently waiting in the inbox
@@ -379,23 +383,11 @@ Rules: never `send_reply` (chat_id: 0).
 
 ---
 
-### reflection_prompt (`type: "reflection_prompt"`)
+### reflection_prompt (`type: "reflection_prompt"`) — legacy, inbox path
 
-Debug-mode prompts written by `on-compact.py` and `on-fresh-start.py` when `LOBSTER_DEBUG=true`. They arrive after a compaction or fresh bootup and ask the dispatcher to reflect on the experience while it is fresh.
+**Superseded by step 2e (issue #1998).** Debug-mode reflection prompts are now written to the `~/messages/bootup-prompt.md` sidecar file and read-and-deleted directly at startup — see step 2e above. `on-compact.py` and `on-fresh-start.py` no longer write `reflection_prompt` messages to the inbox.
 
-```
-1. mark_processing(message_id)
-2. Read msg["text"] — the reflection question
-3. Reflect genuinely: were there friction points, gaps, or improvements in the
-   bootup/compaction flow worth capturing?
-4. If there are substantive observations:
-   - File or update GitHub issues in SiderealPress/lobster
-   - Open PRs for straightforward fixes (no need to wait for instruction)
-   - If nothing worth capturing: do nothing — silence is the correct response
-5. mark_processed(message_id)
-```
-
-Rules: never `send_reply` (chat_id: 0). Reflection is optional — only act if there are real observations.
+If a message of this type is nonetheless encountered (e.g. a pre-#1998 hook version, or one queued before an upgrade landed): `mark_processing(message_id)` then `mark_processed(message_id)` without reflecting — do not act on it. Step 2e already covers reflection for the current startup; re-reflecting on a stale queued copy risks duplicate or contradictory GitHub activity.
 
 ---
 
--- a/hooks/on-compact.py
+++ b/hooks/on-compact.py
@@ -64,6 +64,10 @@ from session_role import (
 INBOX_DIR = Path(os.path.expanduser("~/messages/inbox"))
 PROCESSING_DIR = Path(os.path.expanduser("~/messages/processing"))
 PROCESSED_DIR = Path(os.path.expanduser("~/messages/processed"))
+# Sidecar file for debug-mode reflection prompts (issue #1998). Read-and-deleted
+# by the dispatcher directly at startup instead of flowing through the inbox --
+# avoids 2 MCP round-trips (mark_processing + mark_processed) per restart.
+BOOTUP_PROMPT_FILE = Path(os.path.expanduser("~/messages/bootup-prompt.md"))
 OUTBOX_DIR = Path(
     os.environ.get(
         "LOBSTER_OUTBOX_DIR_OVERRIDE",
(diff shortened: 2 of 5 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS

Independently re-verified this end-to-end rather than trusting the engineer's summary:

- Read the diff cold first. Cross-cutting concerns I flagged going in (sidecar write atomicity, concurrent-hook races, dispatcher read-then-delete race, orphaned legacy handler, dead dedup code) were all already addressed in the PR — no gaps the engineer missed.
- Ran `tests/unit/test_hooks/test_reflection_prompt.py` + `tests/unit/test_hooks/test_bootup_prompt_sidecar_dispatcher_flow.py` myself: 20/20 passed.
- Falsifiability check performed independently (checked out the 3 pre-PR files via `git checkout HEAD~1 --`, not trusting the PR's own account): 12/20 failed (9/16 + 3/4, matching the engineer's claim exactly), then restored to HEAD and confirmed 20/20 green again.
- Ran the broader regression suite (`tests/unit/test_hooks/ tests/unit/test_mcp_server/`) myself: 14 failed, 1559 passed — matches the engineer's claim exactly. Verified the 14 failures are pre-existing by checking out `HEAD~1` (current main tip) wholesale and re-running the two heaviest-hit files (`test_on_compact_write_claude_session_id.py`, `test_context_monitor.py`) — same failures reproduce on main, unrelated to this diff.
- Sidecar write: confirmed genuinely atomic — temp file in the same directory (`.with_suffix(".tmp")`) then `rename()`, consistent with the existing atomic-write pattern already used elsewhere in both hook files.
- Dispatcher step 2e explicitly handles the file-not-found case ("If the file does not exist: skip, nothing to do") as the normal/common path, correctly sequenced between steps 2d and 3.
- Old inbox-based `reflection_prompt` handler was not orphaned: it's kept, relabeled "legacy, inbox path," and repurposed to drain any stale queued message (claim + mark processed) without re-reflecting, which correctly avoids duplicate/contradictory GitHub activity if step 2e and a stale message both fire.
- `_reflection_already_exists()` removal confirmed dead-code cleanup: reasoning holds because the sidecar is a single overwritten-in-place file (no ID-based dedup needed), and its removal is exercised by an explicit regression test in both hooks, not just claimed in a comment.
- No unused imports left behind in the rewritten test file (checked `pytest`/`json` — both fully removed, zero references), no duplicate/shadowed test class names introduced by this diff, py_compile clean on all 4 changed/added files.

No blocking issues found.

Issue #2148, opened 2026-08-03
Capture real PID at session_start() and use as primary liveness signal (Phase 1: PID ground truth, scoped from shadowlobster#65)


aeschylus:
## Context

Scoping sub-issue for [shadowlobster#65](https://github.com/SiderealPress/shadowlobster/issues/65) ("Deepened design brief" comment, 2026-08-03). That comment lays out three approaches (A: PID ground truth, B: consolidated FSM module, C: event-driven push model) and recommends shipping *Approach B* first as its own "Phase 1" — but Drew has asked to execute the **PID ground-truth capture** piece specifically as the first shippable slice (what this issue calls Phase 1). This issue scopes that work concretely so there's no ambiguity about what's in vs. out.

**Do not conflate this with shadowlobster#65's own Phase 1 (the FSM consolidation module).** That is explicitly out of scope here — see Non-goals.

## Problem (verified against current code)

`session_store.session_start()` (`src/agents/session_store.py:241`) records no PID — only `output_file` (a path) + `spawned_at`/`timeout_minutes`. Three independent consumers (`scripts/health-check-v3.sh`, `session_store.cleanup_stale_running_sessions()`, `scripts/agent-monitor.py`'s `classify()`) each infer liveness from file mtimes / parsed stop-reason strings with their own thresholds, and disagree.

## Architectural caveat — verify before implementing

Not every row in `agent_sessions` corresponds to a distinct OS process:
- Rows created by `hooks/auto-register-agent.py` (fires after every **Agent/Task tool** call) are, as far as we can tell from reading the code, **in-process** sub-conversations inside the single running dispatcher `claude` process — there is no separate PID to capture for these today. (`session_store.py`'s own comment: "output_file paths are created by Claude Code after the [agent] spawned" — i.e. the harness manages these, not a spawned OS process we control.)
- `scripts/docker-worker.sh`-launched jobs and the dispatcher's own self-registration (`session_start(agent_type="dispatcher", ...)`) **do** correspond to real, capturable OS processes.

**First task: confirm or refute this by spiking it** (this is open design question #1 from the shadowlobster#65 comment). If confirmed, the design below (dispatcher_pid fallback for in-process agents) is the right shape. If refuted — i.e. you find a real PID is obtainable for Agent-tool subagents too — capture it directly instead and note the finding in the PR description; that's a strictly better outcome, just verify it's real (e.g. by cross-checking against `ps`/`/proc` during a live Agent-tool call) rather than assuming.

[...]

PR #2149, opened 2026-08-03
fix(health-check): derive WFM-active suppression cap from session-age limit (issue #2074)

### LINKED ISSUE #2074: fix(health-check): WFM-active mask cannot detect frozen dispatcher — caused 2-day outage

sayhar:
## Problem

Since 2026-08-02, the `lobster-claude.service` dispatcher has been auto-restarting roughly hourly (7 restarts observed between 04:56–08:28 EDT on 2026-08-03), each one sending a "Lobster Starting" Telegram notification. The user confirmed this is disruptive spam, not a real outage.

## Root cause

`scripts/health-check-v3.sh`'s `check_dispatcher_heartbeat()` treats a dispatcher heartbeat that's been stale for a long time as suspicious, *unless* the WFM-active daemon-thread file is fresh (which means the dispatcher is legitimately idle inside `wait_for_messages`). Issue #2074's original fix (PR #2090, merged 2026-06-07) added a time cap on that suppression — `WFM_SUPPRESSION_MAX_SECONDS=2700` (45 min) — so a *frozen* dispatcher whose daemon thread keeps refreshing WFM-active independently of the frozen main loop would still eventually get flagged RED, rather than being masked forever (the May 2026 2-day outage this issue was originally about).

That fixed 2700s cap silently assumed a legitimate `wait_for_messages()` idle window would never approach 45 minutes — true at the time, because Claude Code's own client-side MCP tool-idle watchdog aborted `wait_for_messages` well under that window. **PR #2144 (merged 2026-08-02) removed that client-side watchdog by design**, specifically so `wait_for_messages` could block for its full requested timeout (default 20h) during quiet periods. That silently invalidated the 2700s assumption: any normal quiet period longer than 45 minutes now un-suppresses the heartbeat-stale check and fires a false RED, restarting a perfectly healthy dispatcher. The observed ~45–60 min restart cadence lines up exactly with the old cap, the day after #2144 landed.

Not a duplicate of #2148 — that's the separate, larger PID-ground-truth liveness project (schema migration, `is_pid_alive`, three classifiers). This PR is a small, targeted correction to the existing time-cap heuristic; it doesn't touch PID capture, `agent_sessions.db`, or `session_store.py`.

## Fix

`check_session_age()` already provides an unconditional, activity-independent backstop: every dispatcher session is recycled at `SESSION_AGE_LIMIT_SECONDS` (7200s default) via a direct SIGTERM to the dispatcher PID, on a wall-clock timer that doesn't care about heartbeat or WFM state. Given that backstop exists, `WFM_SUPPRESSION_MAX_SECONDS` doesn't need to be a fast, independent frozen-detector anymore — it only needs to catch the case where the session-age backstop itself failed to fire (e.g. a missing/corrupt `dispatcher.pid`).

So instead of a fixed literal, the cap is now **derived** from `SESSION_AGE_LIMIT_SECONDS` minus a small margin, via a small pure function:

```bash
compute_wfm_suppression_max_seconds() {
    local session_age_limit="$1" margin="$2" fallback="$3"
    if [[ "$session_age_limit" -le 0 ]]; then echo "$fallback"; return; fi
    local derived=$(( session_age_limit - margin ))
    [[ $derived -lt 1 ]] && derived=1
    echo "$derived"
}
```

With defaults (`SESSION_AGE_LIMIT_SECONDS=7200`, margin=300), the cap becomes **6900s (115 min)** instead of 2700s (45 min) — comfortably above any idle window a session can legitimately reach before it gets recycled anyway. An explicit `LOBSTER_WFM_SUPPRESSION_MAX_SECONDS` override (env var or `config.env`) is available for operators who want to set it directly, and a 6h fallback covers the case where the session-age check itself is disabled (`LOBSTER_SESSION_AGE_LIMIT_SECONDS=0`).

Deriving one constant from the other, instead of hand-copying two numbers that encode the same unstated assumption, is the point — it's what let #2144 silently break #2074's fix in the first place. Added a new `docs/engineering-lessons-learned.md` entry describing this class of bug generally, in case it recurs elsewhere.

## Functional patterns used

- `compute_wfm_suppression_max_seconds()` is a pure function: no globals read, no side effects, entirely determined by its three arguments — trivially unit-testable in isolation from the rest of the script's I/O.
- The derivation replaces a duplicated, hand-copied constant with a single source of truth (`SESSION_AGE_LIMIT_SECONDS`), removing the class of bug where two independent literals drift out of sync.

## Flow diagram

```
Before (fixed 2700s cap):

  wait_for_messages() idle  ──►  heartbeat stale
                                       │
                          stale < 2700s?  ──yes──►  GREEN (suppressed)
                                       │no
                                       ▼
                                  RED → restart
                                  (fires on ANY idle > 45 min,
                                   including totally healthy overnight quiet)

After (derived cap, ~6900s default):

  wait_for_messages() idle  ──►  heartbeat stale
                                       │
                     stale < (SESSION_AGE_LIMIT - margin)?  ──yes──►  GREEN (suppressed)
                                       │no
                                       ▼
                                  RED → restart
                                  (only fires if idle has run almost the
                                   entire session lifetime AND the
                                   session-age SIGTERM backstop didn't
                                   already recycle the session)
```

No other state transitions in `health-check-v3.sh` are touched — `check_session_age()`, the GREEN/YELLOW/RED/BLACK ladder, restart cooldown, and boot-grace logic are all unchanged.

## Out of scope

- No changes to `check_session_age()`, `agent_sessions.db`, PID capture, or anything in #2148's scope.
- No live deploy or restart of `lobster-claude.service` — code + tests only, per instructions. `main` is untouched; all work is in this branch/worktree.
- Did not add `config.env.example` entries for the new override — `LOBSTER_SESSION_AGE_LIMIT_SECONDS` (the sibling knob) isn't documented there either, so kept consistent.

## Tests run

- [x] `bash tests/test-health-check-dispatcher-heartbeat.sh` — 21/21 passed (added tests 16–20: `compute_wfm_suppression_max_seconds()` unit tests + a regression test reproducing the exact 2026-08-03 failure mode — heartbeat stale 3000s + WFM-active fresh was RED under the old 2700s cap, is GREEN under the derived ~6900s default)
- [x] `bash tests/test-health-check-wfm-active.sh` — 10/10 passed (also fixed a pre-existing, unrelated `set -u` unbound-variable failure in this suite — it predates the #2090 cap and never set `WFM_SUPPRESSION_MAX_SECONDS`)
- [x] All other `tests/test-health-check-*.sh` suites (boot-grace, inbox-drain, pid-timing, session-age, thinking-heartbeat) — all pass, unaffected
- [x] `uv run pytest tests/smoke/test_health_check.py tests/unit/test_hibernation.py` — 39 passed
- [x] Manual dry-run of the config-resolution section in 5 scenarios (defaults; `LOBSTER_SESSION_AGE_LIMIT_SECONDS=0`; env override; `config.env` override; custom session-age limit) — confirmed derived values match expectations in each case (6900 / 21600-fallback / 1234-override / 5555-config-override / 3300 for a 3600s custom limit)

## Manual test

No systemd/cron/live-service test performed — this PR is code + tests only, per task instructions ("Do NOT deploy or restart the live production service"). The config-resolution dry-run above exercised the actual script logic (sourced up to but excluding the `main "$@"` call) against a fully sandboxed `HOME`/`LOBSTER_WORKSPACE`/`LOBSTER_MESSAGES`/`LOBSTER_CONFIG_DIR`, so no real system state was touched. Live confirmation that hourly restarts stop is a follow-up the user needs to observe after this is merged and deployed — that requires an explicit, separate deploy/restart decision.

## Definition of Done

- [ ] Integration/manual test run — not applicable pre-merge; live verification is a follow-up after an explicit, separately-approved deploy
- [ ] User-visible outcome verified — not yet; the fix's effect (no more hourly restart notifications) can only be confirmed by observing production behavior post-deploy, which is out of scope for this PR
- [x] Side-effect audit complete: no floods, no unscoped writes, no unintended volume — this only widens a suppression window in a bash health-check script; no new external calls

Closes #2074


aeschylus:
## Rebased onto current main (cc273eec) and re-verified

This PR had one conflict against main: an append-only conflict in `docs/engineering-lessons-learned.md` where both this branch and main (via #2152/#2176) appended a new lessons-learned entry after the same anchor line. Resolved by keeping main's history line for the "Missing Dispatcher Exclusion Filter" entry (crediting #2152/#2176) and appending this PR's "Independent Timing Constants" section immediately after it, unchanged from the original PR content. `scripts/health-check-v3.sh` and both test files merged cleanly with no conflict.

Full re-verification on the merged result (branch rebased onto `origin/main` @ `cc273eec`, new HEAD `dd2fba35`):

- `bash -n scripts/health-check-v3.sh` — syntax OK
- All 9 `tests/test-health-check-*.sh` suites — 21/21, 10/10, 8/8, 5/5, 11/11, 19/19, 9/9, 14/14, 7/7 (all pass)
- `uv run pytest tests/smoke/test_health_check.py tests/unit/test_hibernation.py` — 39 passed
- Falsifiability check: reverted `scripts/health-check-v3.sh` to the pre-fix version (keeping the new tests), reran `tests/test-health-check-dispatcher-heartbeat.sh` — 6 of the new/regression tests fail as expected (`compute_wfm_suppression_max_seconds: command not found`, and the regression test asserting `GREEN` at 3000s-stale/WFM-fresh gets `RED` instead, reproducing the exact 2026-08-03 false-positive). Restored the fix — all 21/21 pass again.
- Manual dry-run of `compute_wfm_suppression_max_seconds()` in isolation: `(7200, 300, 21600) → 6900`, `(0, 300, 21600) → 21600` (fallback), `(3600, 300, 21600) → 3300` (custom session-age limit) — matches the PR description's claimed values.

`mergeable` is now `MERGEABLE` (was `CONFLICTING`). No branch protection / required CI exists on this repo (only an automatic CodeQL scan, not a merge gate), so this comment stands in for the "explicit review before merge" step per standing practice for infra-restart-behavior PRs — a separate reviewer pass is still expected before merge per the usual engineer/reviewer separation.

aeschylus:
## Independent reviewer pass (post-rebase, pre-merge)

Filled the "separate reviewer pass" gap noted in aeschylus's rebase comment — the orchestrating dispatcher session crashed before that stage ran. Reviewed and re-verified independently in a fresh worktree at the PR's exact HEAD (`dd2fba35`), merge-base-confirmed as a clean fast-forward descendant of `main` (`cc273eec`):

- `bash -n scripts/health-check-v3.sh` — syntax OK
- All 9 `tests/test-health-check-*.sh` suites — 8/8, 5/5, 21/21, 11/11, 19/19, 9/9, 14/14, 7/7, 10/10 (all pass, matches claimed counts)
- `uv run pytest tests/smoke/test_health_check.py tests/unit/test_hibernation.py` — 39/39 passed
- Falsifiability check: reverted `scripts/health-check-v3.sh` to the pre-fix parent commit (kept the new tests) — 6 of 21 tests in `test-health-check-dispatcher-heartbeat.sh` failed exactly as claimed (`compute_wfm_suppression_max_seconds: command not found`; regression test expecting GREEN at 3000s-stale/WFM-fresh got RED instead, reproducing the 2026-08-03 false-positive). Restored the fix — `git diff` against the original was empty, 21/21 passed again.
- Independently reproduced the pre-existing `set -u` unbound-variable bug in `test-health-check-wfm-active.sh`: removing the added `WFM_SUPPRESSION_MAX_SECONDS=999999` line causes `line 120: WFM_SUPPRESSION_MAX_SECONDS: unbound variable` on the second test — confirms that fix is real and was needed.
- Manual dry-run of the config-resolution section (sourced the script up to but excluding `main "$@"` in 5 sandboxed HOME/workspace scenarios): defaults → 6900, `LOBSTER_SESSION_AGE_LIMIT_SECONDS=0` → 21600 (fallback), env override `LOBSTER_WFM_SUPPRESSION_MAX_SECONDS=1234` → 1234, `config.env` override `=5555` → 5555, custom session-age limit 3600 → 3300. All match the PR description exactly.
- Read `check_dispatcher_heartbeat()` and `check_session_age()` directly (not just the diff/summary): confirmed `check_session_age()` is a genuine activity-independent wall-clock backstop (SIGTERM via `dispatcher.pid`, no heartbeat/WFM dependency) — validates the core premise the derivation relies on. The `>=` cap comparison and function wiring are unchanged in behavior from before, only the value's source changed.
- Scope check: diff touches exactly the 4 files claimed (`docs/engineering-lessons-learned.md`, `scripts/health-check-v3.sh`, `tests/test-health-check-dispatcher-heartbeat.sh`, `tests/test-health-check-wfm-active.sh`), no unrelated changes. Docs merge from the rebase is clean, no duplicate headers.

No issues found. Merging via squash, consistent with recent merged PRs (#2178, #2172, etc. are all single-parent squash commits).

PR #2151, opened 2026-08-04
Pin all dependencies to exact versions + enforcement hook

aeschylus:
## Summary

Prompted by a Shai-Hulud-style npm supply-chain worm investigation (this system was found clean, but the risk of a compromised package silently landing via a `^`/`~`/unbounded range was judged real enough to close off structurally, not just by vigilance). This PR establishes a pin-everything policy across the Lobster system's dependency manifests and adds tooling to enforce it going forward.

- Every dependency manifest in this repo (`pyproject.toml` + `uv.lock` at root, `connectors/whatsapp/package.json`, `lobster-shop/multiplayer-telegram-bot/{pyproject.toml,uv.lock}`, `lobster-shop/obsidian-km/requirements.txt`, `tests/requirements-test.txt`) is now pinned to an exact version — no `^`, `~`, `~=`, `>=`, `<=`, `!=`, `>`, `<`, `*`, or `"latest"` ranges. See `DEPENDENCY_NOTES.md` for the full per-manifest audit trail and rationale (including why test deps are pinned to the *same* version as root rather than "whatever's latest today" — a dry-run resolve of the old unbounded `tests/requirements-test.txt` came back a full major version ahead of what's actually running).
- A new `hooks/pin-dependencies-guard.py` PreToolUse hook blocks:
  - Edit/Write/NotebookEdit calls that would introduce an unpinned range into a manifest file (`package.json`, `pyproject.toml`, `requirements*.txt`, `Pipfile`)
  - Bash package-manager commands that could silently resolve to "whatever is newest right now": bare `npm install <pkg>` / `npm update`, `pip install <pkg>` without `==` / `pip install --upgrade`, `uv add <pkg>` without `==`, `uv sync`/`uv lock --upgrade`, `uv run --with <pkg>` without `==` — including via `python -m pip`, a path-prefixed binary (`/usr/bin/pip`, `venv/bin/pip`, `node_modules/.bin/npm`), or an `env`/`command`/`VAR=val` wrapper prefix (see "Fixes from independent review" below)
  - Commands/edits that use an existing lockfile or an exact pin (`npm ci`, `uv sync` with no flags, `pip install -r file.txt`, `pkg==1.2.3`, `pkg@1.2.3`), or a local-path npm install (`npm install ./local-pkg`), are correctly left alone.
- `hooks/pre-commit` gets the same detection logic wired in as a defense-in-depth backstop at commit time (catches manifest edits made outside Claude Code's own tools).
- `install.sh` and `scripts/upgrade.sh` (migration 97) register the new hook for fresh installs and upgrades.

**Escape hatch:** set `LOBSTER_ALLOW_DEPENDENCY_CHANGE=true` in the environment for one deliberate, reviewed version bump. This is intentionally a separate variable from `LOBSTER_DEBUG` (which is already persistently `true` in this system's settings.json for unrelated reasons — reusing it here would make the hook a permanent no-op).

## Fixes from independent review (NEEDS-WORK → addressed)

An independent reviewer found real gaps in `hooks/pin-dependencies-guard.py`, demonstrated by direct invocation. All three are fixed, with new falsifiable test coverage for each:

1. **Bash bypass gaps** — `python -m pip install <pkg>` (no `==`), pip/npm invoked via a path (`/usr/bin/pip`, `venv/bin/pip`, `node_modules/.bin/npm`), `env`/`command` wrapper prefixes, and `uv run --with <pkg>` all previously evaded detection entirely. Fixed by resolving the actual binary/subcommand being invoked (stripping wrapper/assignment prefixes, normalizing path-prefixed command names, recognizing `python -m pip` as `pip`) before pattern matching, and adding dedicated `uv run --with` detection.
2. **False positives** — a heredoc containing an example install command as file *content* (not an actual invocation) was blocked, because every `\n` was treated as a command boundary; the literal words "and"/"or" anywhere in a command were misparsed as shell operators; `npm install ./local-package` (a local-path install) was flagged as "unpinned"; and a prose comment in a manifest merely mentioning a version range (e.g. `# needs numpy>=1.20 installed separately`) was blocked identically to a real dependency line. Fixed by masking heredoc bodies before sub-command splitting, restricting the splitter to real shell operators only (`|`, `;`, `&`, newline — not bare "and"/"or"), skipping local-path/URL/tarball npm install targets, and stripping whole-line comments before scanning non-JSON manifests.
3. **Pre-existing-failure count corrected** — see below.

## Known gaps (flagged, not silently worked around — see `DEPENDENCY_NOTES.md` for full detail)

1. **`connectors/whatsapp/package-lock.json` does not exist and is not committed.** That directory's own `.gitignore` blanket-excludes lock files. Direct deps in `package.json` are pinned exactly, but transitive deps of `whatsapp-web.js` (which pulls in `puppeteer`) can still float on a fresh `npm install` without a committed lockfile. This conflicts with an existing, deliberate repo convention, so it's flagged for an explicit decision rather than overridden unilaterally.
2. **`lobster-shop/camofox-browser/server` was not modified.** It's vendored as a nested git checkout (gitlink, no `.gitmodules`) whose `origin` points at a separate product repo (`jo-inc/camofox-browser`), not part of the Lobster system. Its `package.json` still uses `^` ranges. Editing a foreign repo's manifest/history from inside this task felt out of scope; flagging here rather than leaving it unmentioned.

## NOT included in this PR

**Activating this hook in the live `~/.claude/settings.json` on any running instance is a deliberate, separate follow-up decision** — not part of this PR. This PR only adds the capability (the hook script itself), the `install.sh` registration for fresh installs, and `upgrade.sh` migration 97 for existing installs to pick it up. No live settings.json anywhere has been touched.

## Test plan / evidence

- **Automated test suite:** `tests/unit/test_hooks/test_pin_dependencies_guard.py` — **211 test cases, all passing** (141 original + 70 added for the independent-review fixes above: bypass-gap coverage for `python -m pip`, path-prefixed binaries, `env`/`command`/assignment wrappers, and `uv run --with`; false-positive coverage for heredoc bodies, bare "and"/"or", local-path npm installs, and prose comments in manifests).

  ```
  211 passed, 1 warning in 5.7s
  ```

- **Falsifiability check performed for the new fixes** (proves the new tests aren't vacuous): three targeted mutations were applied directly to the real `hooks/pin-dependencies-guard.py`, tested, then reverted via file restore before moving to the next:
  1. Neutered `_resolve_subcommand` (made it a no-op) and removed the `uv run --with` check → 26 of the 37 bypass-gap-related tests failed (all 18 `test_bypass_gap_now_blocked` cases + all 8 `TestHookBashBypassGapsIntegration` cases); the 11-case pinned/lockfile sanity class stayed green as expected
  2. Reverted the sub-command splitter to include bare "and"/"or", removed heredoc masking, and removed the local-path skip in `_node_package_args_unpinned` → 12 of the 18 false-positive-related tests failed (3 heredoc, 2 bare-and/or, 4 local-path pure-function cases, plus all 3 full-hook-integration cases); the 6 "still correctly detected" sanity cases in the same classes stayed green
  3. Removed the whole-line-comment strip from `find_unpinned_range` → 5 of the 7 manifest-comment tests failed (3 pure-function + both integration cases); the inline-trailing-comment case and the `_strip_full_line_comments` helper unit test (which doesn't go through `find_unpinned_range`) stayed green as expected

  Working tree confirmed identical (`diff` clean) to the fixed version after each restore, and the full 211-test suite re-confirmed green afterward.

  (Original falsifiability check for the base 141 tests, from the prior review round, still holds: neutering `_PY_RANGE_RE` failed 11 tests, removing the "engines" exclusion failed exactly the 5 engines-regression tests, stubbing `bash_introduces_unpinned_dependency` to always return `None` failed all 23 Bash-related tests.)

- **`uv sync --frozen`** run clean from repo root: `Checked 78 packages in 2ms` — no regressions from the hook changes.

- **Pre-existing test failures — corrected count.** `uv run pytest tests/unit/ --tb=no -q` reports **16 pre-existing failures** (not 8), across **7 files**, none related to dependency pinning or this hook:
  - `tests/unit/test_bot/test_slack_router_config.py` — 1
  - `tests/unit/test_hooks/test_context_monitor.py` — 1
  - `tests/unit/test_hooks/test_on_compact_write_claude_session_id.py` — 7
  - `tests/unit/test_mcp_server/test_push_calendar_token_endpoint.py` — 2
  - `tests/unit/test_mcp_server/test_push_gmail_token_endpoint.py` — 2
  - `tests/unit/test_mcp_server/test_push_workspace_token_endpoint.py` — 2
  - `tests/unit/test_script_inbox_sources.py` — 1

  Reconfirmed on this branch after the fixes above (`3949 passed, 16 failed, 31 skipped` on `tests/unit/`), same 16/7 as before these changes and as on `main` — this PR does not introduce, fix, or otherwise touch any of them. See `DEPENDENCY_NOTES.md` for the full corrected breakdown.

### DIFF
--- a/DEPENDENCY_NOTES.md
+++ b/DEPENDENCY_NOTES.md
@@ -108,3 +108,24 @@ SAME already-pinned version, not to whatever a fresh resolve returns.
    process, felt like scope creep with real risk of breaking unrelated
    builds. If the system owner wants the policy extended to those repos, that should be
    a separate, explicit follow-up per repo.
+
+## Pre-existing test suite status (not caused by this PR)
+
+`uv run pytest tests/unit/ --tb=no -q` currently reports **16 pre-existing
+failures** (not 8 — an earlier count in this PR undercounted them), spread
+across **7 files**, none related to dependency pinning:
+
+- `tests/unit/test_bot/test_slack_router_config.py` — 1
+- `tests/unit/test_hooks/test_context_monitor.py` — 1
+- `tests/unit/test_hooks/test_on_compact_write_claude_session_id.py` — 7
+- `tests/unit/test_mcp_server/test_push_calendar_token_endpoint.py` — 2
+- `tests/unit/test_mcp_server/test_push_gmail_token_endpoint.py` — 2
+- `tests/unit/test_mcp_server/test_push_workspace_token_endpoint.py` — 2
+- `tests/unit/test_script_inbox_sources.py` — 1
+
+Confirmed identical (same 16, same files) on both `main` and this branch —
+this PR does not introduce, fix, or otherwise touch any of them. Scope is
+`pytest tests/unit/` per this repo's own Makefile; a literal unscoped `uv
+run pytest --tb=no -q` from repo root additionally produces 43 collection
+errors on both branches, a pre-existing monorepo/subproject
+dependency-isolation issue also unrelated to this PR.
--- a/DEPENDENCY_NOTES.md
+++ b/DEPENDENCY_NOTES.md
@@ -1,0 +1,110 @@
+# Dependency Pinning Policy
+
+**Effective 2026-08-04.** Prompted by a Shai-Hulud-style npm supply-chain
+worm investigation (system found clean, but the risk of a compromised
+package silently landing via a `^`/`~`/unbounded range was judged real
+enough to close off structurally, not just by vigilance).
+
+## Policy
+
+Every dependency manifest in this system is pinned to an **exact** version.
+No `^`, `~`, `~=`, `>=`, `<=`, `!=`, `>`, `<`, `*`, or `"latest"` ranges.
+`npm install` / `pip install` / `uv sync` / equivalent must never be able to
+silently resolve a dependency to a newer version than what a human
+explicitly reviewed and pinned.
+
+Bumping a pinned version is still fully supported — it just has to be a
+deliberate edit to the manifest (changing `=="1.2.3"` to `=="1.3.0"`), not
+something a package manager decides for you.
+
+## Enforcement
+
+- `hooks/pin-dependencies-guard.py` — Claude Code PreToolUse hook. Blocks
+  Edit/Write/NotebookEdit calls that introduce an unpinned range into a
+  manifest, and blocks Bash package-manager invocations that could resolve
+  to "whatever is newest right now" (bare `npm install <pkg>`, `npm
+  update`, `pip install <pkg>` without `==`, `pip install --upgrade`, `uv
+  add <pkg>` without `==`, `uv sync`/`uv lock` with `--upgrade`). See the
+  module docstring for the full, precise list of what does and doesn't
+  trigger a block.
+- `hooks/pre-commit` — a defense-in-depth backstop at `git commit` time,
+  scanning staged manifest content with the same detection logic (via
+  `pin-dependencies-guard.py --scan-file`). This catches edits made outside
+  Claude Code's own tools (a plain editor, `git apply`, an ad hoc script).
+- Escape hatch: `LOBSTER_ALLOW_DEPENDENCY_CHANGE=true` in the environment
+  bypasses both checks for one deliberate, reviewed bump. This is a
+  dedicated variable, separate from `LOBSTER_DEBUG` (which is already set
+  to `true` persistently in this system's `settings.json` for unrelated
+  reasons — reusing it here would make the hook a permanent no-op).
+
+## Manifests covered and current pin status
+
+| Manifest | Status as of audit | Notes |
+|---|---|---|
+| `pyproject.toml` (root) | Pinned | All `>=` ranges replaced with the version already resolved in `uv.lock`. `hatchling==1.31.0` verified from local `uv` cache (the version actually used to build). |
+| `uv.lock` (root) | Regenerated | `uv lock` re-run after pinning; only removes "this is the floor of a range" metadata — resolved versions unchanged. |
+| `connectors/whatsapp/package.json` | Pinned | Removed `^` from all three deps (already at latest-compatible: chokidar 3.6.0, qrcode 1.5.4, whatsapp-web.js 1.26.0). |
+| `connectors/whatsapp/package-lock.json` | **Not committed — see Known Gaps** | This directory's own `.gitignore` blanket-excludes lock files. |
+| `lobster-shop/multiplayer-telegram-bot/pyproject.toml` | Pinned | `setuptools==83.0.0`, `pytest==9.0.2` (both dev-dependency declarations aligned). Verified via `uv pip install --dry-run` (no local install existed). |
+| `lobster-shop/multiplayer-telegram-bot/uv.lock` | Regenerated | `uv lock` re-run; no version changes. |
+| `lobster-shop/obsidian-km/requirements.txt` | Pinned | `mcp==1.26.0` and `python-dotenv==1.2.1` aligned to the root project's pins (this MCP server runs standalone but sharing a pin avoids two different resolved versions of the same package existing in the system for no reason). `python-frontmatter==1.3.0` verified via dry-run resolve (no existing anchor). |
+| `tests/requirements-test.txt` | Pinned | Packages that also appear in root `pyproject.toml` are pinned to the SAME version as root (see "Why not just pin to latest" below). Test-only packages (`docker`, `aioresponses`, `faker`, `aiofiles`) pinned to a dry-run resolve; `responses==0.26.0` and `freezegun==1.5.5` pinned to the versions actually found installed on this host. |
+| `lobster-shop/camofox-browser/server/package.json` | **Not touched — see Known Gaps** | External vendored product, out of scope. |
+
+## Why not just pin test deps to "whatever's latest today"
+
+Before pinning, `tests/requirements-test.txt` had `mcp>=1.0.0`,
+`pytest>=8.0.0`, `python-telegram-bot>=20.7`, all unbounded. A dry-run
+resolve of that file as it stood (`uv pip install --dry-run -r
+tests/requirements-test.txt` against a scratch venv) came back with
+`mcp==2.0.0` and `python-telegram-bot==22.8` — a full major version ahead
+of the versions actually pinned in the root project (`mcp==1.26.0`,
+`python-telegram-bot==22.6`). That is precisely the failure mode this
+policy exists to close: a test run today would have silently exercised
+different dependency versions than the running system. The fix applied
+here was to pin every package that's shared with the root project to the
+SAME already-pinned version, not to whatever a fresh resolve returns.
+
+## Known gaps / things flagged rather than silently worked around
+
+1. **`connectors/whatsapp/package-lock.json` does not exist and is not
+   committed.** `connectors/whatsapp/.gitignore` has a blanket
+   `# Lock files (generated)` exclusion covering `package-lock.json` and
+   `yarn.lock`. Direct dependencies in `package.json` are now pinned
+   exactly, but without a committed lockfile, *transitive* dependencies of
+   `whatsapp-web.js` (which pulls in `puppeteer`, which downloads a
+   platform-specific Chromium binary) are still free to float on every
+   fresh `npm install`. Generating and committing a lockfile there conflicts
+   with an existing, deliberate repo convention — this needs an explicit
+   decision from the system owner rather than being overridden unilaterally.
+
+2. **`lobster-shop/camofox-browser/server` was not modified.** It is
+   vendored into this tree as a nested git checkout (a "gitlink" — `git
+   diff` shows it as a `160000` mode entry) but has no corresponding
+   `.gitmodules` registration, and its `origin` remote points at a separate
+   product repo (`jo-inc/camofox-browser`, maintained by Jo Inc, not part of
+   the Lobster system). Its `package.json` still uses `^` ranges for
+   `express`, `playwright-core`, `prom-client`, `swagger-jsdoc`,
+   `camoufox-js`, and its devDependencies. It does carry its own
+   `package-lock.json`, so transitive resolution is currently frozen at
+   whatever that lockfile last resolved to — but the `^` ranges mean a
+   plain `npm update` inside that nested repo would still float. Editing a
+   foreign repo's own manifest/history from inside this task felt out of
+   scope and risky without the system owner or Jo Inc's maintainers reviewing it
+   separately; flagging here rather than silently leaving it un-mentioned.
+
+3. **Scope of this pass was the `~/lobster` system repo itself**, not every
+   directory under `~/lobster-workspace/projects/`. Everything else living
+   in `lobster-workspace/projects/` that is a *worktree of this same repo*
+   (`feature-issue-*`, `fix-issue-*`, `bis-74*`, etc.) is automatically
+   covered once this branch merges to `main`, since they share the same
+   git history. But `lobster-workspace/projects/` also holds a number of
+   entirely separate, unrelated products with their own git repos and
+   dependency surfaces (Eloso Bisque, Paperclip, MyOwnLobster, vault-api /
+   vault-ws, networkmaps, bisque-booking, bottleneck-beta, lobster-watcher)
+   — these were surveyed at a high level but deliberately not modified.
+   "The Lobster system" was read as this repo; pinning across a dozen-plus
+   unrelated client codebases in the same pass, each with its own release
+   process, felt like scope creep with real risk of breaking unrelated
+   builds. If the system owner wants the policy extended to those repos, that should be
+   a separate, explicit follow-up per repo.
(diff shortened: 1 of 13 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: Manifest pinning itself is solid and the test suite for it is real (141 new tests, all pass; verified `uv sync --frozen` clean on both branches with identical 78-package resolution; engines-field regression fix is correctly tested via both the JSON-parse and regex-fallback paths; no duplicate/shadowed test classes). Confirmed scope constraints held: `~/.claude/settings.json` is untouched, and neither `require-write-result.py` nor `require-wait-for-messages.py` appear anywhere in the diff.

However, the new PreToolUse hook (`hooks/pin-dependencies-guard.py`) has real gaps on both sides, demonstrated by direct invocation (not hypothetical):

**Bypass surface (Bash side) — undermines the hook's actual purpose, not via the documented escape hatch:**
All of these silently install an unpinned package with zero block, using entirely ordinary invocations an engineer or a script would reach for without any intent to evade anything:
- `python -m pip install requests` / `python3 -m pip install requests`
- `/usr/bin/pip install requests`, `/usr/local/bin/pip3 install requests`, `venv/bin/pip install requests`
- `node_modules/.bin/npm install lodash`, `/usr/local/bin/npm install lodash`
- `env pip install requests`, `command pip install requests`
- `uv run --with requests python foo.py`
The regexes only match when the command name (`npm`/`pip`/`uv`) sits literally at position 0 of the (sub-)command — any path prefix or wrapper defeats detection entirely. Given `python -m pip install` is one of the most commonly recommended ways to invoke pip, this is a large, easily-hit gap in everyday use, not an edge case.

**False positives (also demonstrated):**
1. Multi-line Bash tool calls that contain an example command as file *content* (e.g. a heredoc generating a script/doc) get blocked, because the sub-command splitter treats every `\n` as a command boundary and re-evaluates each line independently: `cat > f.sh <<EOF` / `npm install lodash` / `EOF` is denied even though nothing is actually being installed.
2. The splitter also treats the bare English words "and"/"or" as command separators (not just `&&`/`||`). `echo hello or npm install foo` — a single harmless echo with no shell operators at all — gets misparsed into two fragments and denied.
3. `npm install ./vendor/local-pkg` (a local path install, not a registry resolution) is blocked as "unpinned," though it isn't the silent-latest-version risk this hook targets.
4. A prose comment inside a covered manifest that merely mentions a range (e.g. `# needs numpy>=1.20 installed separately`) is blocked identically to a real dependency line — the detector has no comment/prose awareness, just a file-wide substring scan.

**Pre-existing-failure count is understated, though not a regression:** I ran `uv run pytest tests/unit/ --tb=no -q` on both `main` and this branch. Both show the identical **16** pre-existing failures (not 8), across 7 files: `test_slack_router_config.py`, `test_context_monitor.py` (1), `test_on_compact_write_claude_session_id.py` (7), `test_push_calendar_token_endpoint.py` (2), `test_push_gmail_token_endpoint.py` (2), `test_push_workspace_token_endpoint.py` (2), `test_script_inbox_sources.py`. The "8" in the PR description only accounts for the two files it names (1+7=8); the other 8 failures in 5 additional, unrelated files aren't mentioned. Confirmed via direct branch comparison that none of the 16 are caused by this PR — new-test count also matches exactly (3808→3949 passed, +141). Separately, a literal unscoped `uv run pytest --tb=no -q` from repo root produces 43 collection errors on both branches (pre-existing monorepo/subproject dependency-isolation issue, not this PR's doing) — the correct scope is `pytest tests/unit/` per this repo's own Makefile.

None of the above blocks the manifest-pinning changes themselves, which look correct, but the guard hook as written provides materially less real-world coverage on the Bash side than its docstring claims, and has demonstrable false-positive triggers on the Edit/Write and Bash sides that could block legitimate work (heredocs, prose comments, local installs). Recommend tightening the Bash detection (word-boundary on separators only for real shell operators, and either tokenizing per-line-of-a-single-command more carefully or accepting the multiline limitation explicitly in the docstring) and covering `python -m pip`/path-prefixed invocations before this activates in live settings.json.

aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: Re-verified commit 5bd5ea3e independently (ran the hook's own functions directly, not just the new tests). The three originally-cited findings are genuinely fixed:

**Bypass gaps — now blocked** (tested directly): `python -m pip install requests`, `python3 -m pip install requests`, `/usr/bin/pip install requests`, `venv/bin/pip install requests`, `node_modules/.bin/npm install lodash`, `env FOO=bar pip install requests`, `command npm install lodash`, `FOO=bar pip install requests`, `uv run --with requests script.py`. All correctly denied.

**False positives — now allowed** (tested directly): `echo hello or npm install foo`, `npm install ./local-package` / `../sibling` / `/abs/path`, `cat <<EOF\npip install requests\nEOF` (heredoc example text), and a `# needs numpy>=1.20 installed separately` prose comment in a manifest. All correctly pass through.

**Test suite**: confirmed 211 total tests collected on this branch vs 141 on the pre-fix commit (4674022d) — exactly the claimed 70 new tests, all 211 passing.

**Pre-existing failures**: independently ran `uv run pytest tests/unit/ --tb=no -q` on both `main` and this branch. Both show **exactly 16 failures across the same 7 files** listed in DEPENDENCY_NOTES.md (test_slack_router_config.py:1, test_context_monitor.py:1, test_on_compact_write_claude_session_id.py:7, test_push_calendar_token_endpoint.py:2, test_push_gmail_token_endpoint.py:2, test_push_workspace_token_endpoint.py:2, test_script_inbox_sources.py:1). Corrected count is accurate and confirmed pre-existing (identical on main).

**Housekeeping**: `uv sync --frozen` is clean on this branch. Live `~/.claude/settings.json` is untouched (mtime unchanged, predates this PR).

**New regression found (adversarial check)**: the heredoc-masking fix for finding #2 introduces a real bypass that didn't exist before. It blanks *any* heredoc body unconditionally, including when the body is not inert example text but an actual script being fed to a real interpreter via stdin. Demonstrated:

```
bash <<EOF
pip install requests
EOF
```

This is now **silently allowed** — `bash_introduces_unpinned_dependency()` returns `None`. Same result for `sh <<EOF ... EOF` and `ssh host <<EOF ... EOF`. This is not a contrived case — heredoc-to-shell is an everyday pattern for running a multi-line script. Checked against the pre-fix commit (4674022d): that version, despite splitting heredoc bodies into sub-commands only by accident (naive `\n` splitting, no heredoc awareness at all), actually *did* catch this exact command. So this fix trades a false-positive (example text in a heredoc fed to `cat`) for a false-negative (real code in a heredoc fed to an interpreter) on a more dangerous and more common pattern.

Suggested fix: only mask heredoc bodies when the command consuming the heredoc is not itself a shell/interpreter (e.g. skip masking when the opening line matches `bash|sh|zsh|ssh|python\d* <<`), or alternatively still scan heredoc body lines for install invocations even while treating them as a single logical unit (i.e., keep them out of the "sub-command count," but don't exempt them from pattern matching).

Everything else in this PR checks out. Recommend closing this one gap before merge — it's narrower than the original three but is a genuine new hole in exactly the area this round of fixes touched.

aeschylus:
## Round 4 independent adversarial re-review — verdict: NEEDS-WORK

Verified everything from scratch (no prior review taken on faith). Full command transcripts below.

### What I verified about commit 93c1b6d2 itself

Read the actual diff (not just the message) and the full current `hooks/pin-dependencies-guard.py`. The fix does exactly what it claims:

- `_line_invokes_interpreter()` correctly detects `bash`/`sh`/`zsh`/`ksh`/`dash`/`ssh`/`python[23]` as a whitespace-delimited token on the heredoc-opening line, matched by basename after stripping any path prefix (`/usr/bin/bash`, `~/venv/bin/python3` both correctly match; `f.sh` correctly does not).
- `_mask_heredocs()` only blanks the body when `executed` is False, so a heredoc genuinely fed to an interpreter is left intact for sub-command splitting/scanning, while inert `cat > f.sh <<EOF` text is still blanked (no re-regression of the round-2 false-positive fix).

**My own falsifiability check** (not just reading the diff): patched `_mask_heredocs` to always blank the body regardless of `executed` (i.e. reverted to round-2 behavior), then re-ran the hook test file against the reverted code:

```
bash <<EOF / sh <<EOF / zsh <<EOF / ssh host <<EOF / cat <<EOF | bash
  → current file: BLOCKED   reverted file: ALLOWED  (11 tests fail on revert,
    all in TestHeredocInterpreterExecutionStillScanned / 
    TestHookHeredocInterpreterExecutionIntegration — exactly the class the
    commit claims to fix)
inert `cat > f.sh <<EOF` text
  → current file: ALLOWED   reverted file: ALLOWED  (unaffected either way,
    confirming the round-2 fix isn't re-broken)
```

This proves the test suite actually depends on the fix, not just documents it.

### Test evidence, reproduced independently

```
uv run pytest tests/unit/test_hooks/test_pin_dependencies_guard.py -q
→ 226 passed  (matches claim exactly)

uv run pytest tests/unit/ --tb=short -q
→ 16 failed, 4034 passed, 31 skipped  (matches claim exactly)
```

The 16 failures span exactly 7 files, none related to this hook: `test_slack_router_config.py`, `test_context_monitor.py`, `test_on_compact_write_claude_session_id.py` (6), `test_push_calendar_token_endpoint.py` (2), `test_push_gmail_token_endpoint.py` (2), `test_push_workspace_token_endpoint.py` (2), `test_script_inbox_sources.py` — matches the claimed "same 7 files as prior rounds" pre-existing failures.

### `~/.claude/settings.json`

Untouched. mtime is Jul 16 (predates this branch, which started Aug 4). `grep -c pin-dependencies-guard ~/.claude/settings.json` → 0. Migration 97 in `scripts/upgrade.sh` is explicitly gated behind a jq idempotency check and, per the base commit's own message, "not yet applied to the live settings.json — pending review." Confirmed not applied.

### Bypass attempts against the current file — all correctly handled

```
docker exec mycontainer bash <<EOF / sudo bash <<EOF / env bash <<EOF     → BLOCKED
/usr/bin/bash <<EOF                                                       → BLOCKED
ssh myhost <<EOF / cat <<EOF | bash                                       → BLOCKED
bash <<-EOF (dash variant) / sh <<"EOF" (double-quoted delim)             → BLOCKED
two sequential heredocs on one command (cat <<A ... A; bash <<B ... B)    → BLOCKED
cat > f.sh <<'EOF' with install text in body (inert, quoted delim)        → correctly ALLOWED
bash <<EOF with an already-pinned install (pip install requests==2.31.0) → correctly ALLOWED
interpreter name appearing only inside a comment inside an inert heredoc → correctly ALLOWED (no false positive)
```

### New real bypass found — NOT a regression from 93c1b6d2, but still live and unaddressed after 4 rounds

```
bash <<< "pip install requests"        → ALLOWED (should be BLOCKED)
bash -c "pip install requests"         → ALLOWED (should be BLOCKED)
sh -c "uv add requests"                → ALLOWED
zsh -c "pip install requests"          → ALLOWED
bash -c "npm install lodash"           → ALLOWED
sh <<< "uv add requests"               → ALLOWED
```

`_mask_heredocs` only fires on `_HEREDOC_START_RE` (`<<DELIM`), which does not match a here-string (`<<<`) at all — confirmed by direct regex test, no match found for `<<<`. And `<interpreter> -c "..."` never triggers heredoc handling in the first place; the whole command is a single sub-command after `_SUB_CMD_SPLIT_RE.split()` (no `|;&\n` present), so `_resolve_subcommand` sees a string starting with `bash`/`sh`/`zsh`, none of the `_NODE_INSTALL_RE`/`_PIP_INSTALL_RE`/`_UV_ADD_RE` etc. patterns match (they're anchored at `^\s*` against the *sub-command*, not applied recursively to quoted string arguments), and the embedded `pip install requests`/`npm install lodash`/`uv add requests` inside the quotes is never scanned.

I confirmed this is **not a regression introduced by 93c1b6d2** — I ran the identical `bash <<< "..."` and `bash -c "..."` commands against the pre-fix file at commit `5bd5ea3e` and they were already `ALLOWED` there too. So this isn't something round 3 broke; it's a gap that survived rounds 1–3 untouched, and round 1's "real bash-bypass gaps" fix (wrapper/path/assignment-prefix normalization) evidently didn't cover `-c` or here-string forms.

This matters because it's not a narrow edge case — it's `<any interpreter> -c "<any unpinned install>"`, one of the most common ways to run an inline shell/python snippet, and it silently defeats the hook's stated purpose ("blocks any Bash command that would let a package manager silently resolve to a newer-than-currently-pinned version") for every package manager the hook otherwise covers (pip, npm, uv). Given this hook exists specifically as a supply-chain hardening measure (Shai-Hulud npm worm investigation per the base commit), I'd treat this as a real gap worth closing before calling the guard done, even though it's out of scope for the specific diff in 93c1b6d2.

### Verdict: NEEDS-WORK

Commit 93c1b6d2 is correct and well-tested for what it claims to fix (the heredoc-to-interpreter regression) — no notes there. But the overall hook still has a reproducible, generalizable bypass via `<<<` here-strings and `-c` inline-string invocation across bash/sh/zsh × pip/npm/uv, unaddressed through all 4 rounds. Recommend a follow-up fix (e.g. recursively scanning quoted arguments to `-c` the same way heredoc bodies are now scanned, and extending `_HEREDOC_START_RE`-style handling to `<<<`) before treating this hook as closing the gap it was built for.

PR #2152, opened 2026-08-04
Capture real PID at session_start() and use as primary liveness signal (Phase 1: PID ground truth)

### LINKED ISSUE #2148: Capture real PID at session_start() and use as primary liveness signal (Phase 1: PID ground truth, scoped from shadowlobster#65)

aeschylus:
## Problem

`session_store.session_start()` records no PID — only `output_file` (a path) plus `spawned_at`/`timeout_minutes`. Three independent consumers (`session_store.cleanup_stale_running_sessions()`, `scripts/agent-monitor.py`'s `classify()`, `scripts/health-check-v3.sh`) each infer liveness from output-file mtimes and parsed stop-reason strings, with their own thresholds, and disagree — producing both false-dead ("ghost") and false-alive verdicts.

This PR gives all three a real, shared ground-truth signal: an actual OS PID, checked via `os.kill(pid, 0)`, that overrides the heuristics whenever it was captured, and falls through unchanged to today's exact behavior whenever it wasn't (pre-migration rows, or sessions with no real process boundary).

## Architectural finding (verified before implementing)

Live `pstree -p` inspection against the running dispatcher confirmed the issue's own open question: Agent/Task-tool subagents run as **threads inside the single dispatcher `claude` OS process**, not as separate processes — there is no PID to capture for them directly. This is why `dispatcher_pid` exists as a distinct, weaker-but-real fallback signal (see below).

A related correction to the issue's literal wording: `handle_session_start()`'s dispatcher self-registration runs inside the **stdio `lobster-inbox` MCP server child process**, not the same PID as the `claude` process itself. A flat `os.getpid()` there would have recorded the wrong process. Both this and `hooks/auto-register-agent.py` instead walk the process tree to the nearest ancestor named `claude` — the same technique `hooks/session_role.py` already uses for dispatcher detection.

## What changed

- **`src/agents/pid_liveness.py`** (new): `is_pid_alive(pid)` and `find_dispatcher_ancestor_pid()` — the one shared, pure liveness primitive all three classifiers now consult.
- **`session_store.py`**: additive `pid` / `dispatcher_pid` / `pid_captured_at` columns; `session_start()` grows optional `pid`/`dispatcher_pid` params (NULL by default, fully backward compatible); `cleanup_stale_running_sessions()` checks a real PID first when present — a confirmed-dead PID overrides the output-file heuristic (even a fresh mtime), a confirmed-live PID overrides it the other way (even a missing output file past the elapsed-time cutoff). A dead `dispatcher_pid` marks its in-process subagent dead directly; a live one does *not* prove the subagent itself is alive and falls through to the existing heuristic unchanged.
- **`hooks/auto-register-agent.py`**: captures the dispatcher's real PID via the ancestor walk and persists it as `dispatcher_pid` on every Agent-tool-registered row. Best-effort — never blocks registration if the walk fails.
- **`src/mcp/inbox_server.py`**: `handle_session_start()` now captures the dispatcher's real PID (same ancestor walk) when `agent_type="dispatcher"`, and passes it through as `pid`. Every other caller is unaffected.
- **`scripts/agent-monitor.py`**: `classify()` takes optional precomputed `pid_alive`/`dispatcher_pid_alive` booleans (kept pure — no OS calls of its own); `classify_agent()` is the one boundary that resolves `row.pid`/`row.dispatcher_pid` via `is_pid_alive()`. Same primary/fallback precedence as `session_store`.
- **`scripts/health-check-v3.sh`**: adds one new, purely observational `check_dispatcher_pid_consistency()` — reads the dispatcher's own `pid` column and cross-checks it against a live `pgrep -x claude` scan, logging (never gating) a mismatch. The ~13 existing checks (`check_claude_process`, `check_wrapper_process`, etc.) are untouched; this does not replace or duplicate their independent, already-correct liveness detection.
- **`docker-worker.sh`**: confirmed via `grep` that nothing in the repo currently calls it, and it does not call `session_start` itself — documented as a no-op finding rather than inventing a new integration point, per the issue's own guidance.

### Flow before / after

```
BEFORE (all three classifiers, independently):
  running row --> output_file mtime / stop_reason parse --> verdict
                   (can be wrong in both directions: fresh mtime on a
                    genuinely dead process, or a missing/stale file on a
                    genuinely live long-running process)

AFTER:
  running row --> pid present? --yes--> os.kill(pid,0) --> verdict (definitive)
                       |
                       no
                       v
               dispatcher_pid present & dead? --yes--> verdict = dead
                       |
                       no (absent, or dispatcher_pid alive)
                       v
               unchanged legacy output_file/elapsed-time heuristic --> verdict
```

No existing reader of the `status` column (e.g. `src/dashboard/collectors.py`) is affected — the column's vocabulary and semantics are unchanged, this only changes *how* a row arrives at `dead` vs `running`/`completed`.

## Fixed along the way: a genuinely flaky test

`tests/test_pid_liveness.py::test_find_dispatcher_ancestor_pid_returns_none_when_absent` failed locally: it spawned a real `sleep` subprocess and walked its *real* ancestry, asserting no "claude"-named process would be found. That assumption doesn't hold when the test itself runs nested inside a real dispatcher `claude` process (as happens when this suite runs inside a background subagent) — the real ancestry chain genuinely does contain a process named `claude` a few hops up. This had nothing to do with a bug in `find_dispatcher_ancestor_pid()` itself. Fixed by faking the process tree at the `_read_ppid`/`_read_comm` layer instead of relying on the real OS process tree — the test is now deterministic regardless of what process happens to be running above the test runner. Confirmed via revert-and-observe: reverting the implementation to always return a fixed PID makes this test (and the sibling positive-case test) fail; restoring it passes again.

## Functional patterns used

Pure functions throughout: `is_pid_alive()`, `find_dispatcher_ancestor_pid()`, and `classify()` in `agent-monitor.py` take plain inputs and return plain outputs with no OS calls of their own inside `classify()` — the one OS-facing boundary (`classify_agent()`) resolves real PIDs into booleans before calling the pure classifier, keeping the classification logic itself trivially testable. Additive-only schema migration (no breaking changes to existing readers). Every new parameter defaults to `None`/backward-compatible behavior.

## Out of scope (per the issue's own Non-goals)

The consolidated `liveness.py` FSM module, the event-bus/EXIT-trap supervisor approach, the "midnight race" TOCTOU, and "stuck but alive" detection are all explicitly out of scope here — this PR answers alive-vs-dead via PID ground truth, nothing more.

## Tests run

- [x] `uv run pytest tests/test_pid_liveness.py tests/test_agent_sessions.py tests/unit/test_agent_monitor_pid_liveness.py tests/unit/test_hooks/test_auto_register_agent.py tests/unit/test_mcp_server/test_dispatcher_pid_capture.py -q` — 109 passed
- [x] `bash tests/test-health-check-dispatcher-pid-consistency.sh` — 11/11 passed
- [x] All other `tests/test-health-check-*.sh` shell suites — no regressions (one pre-existing unrelated failure in `test-health-check-wfm-active.sh`, confirmed present on unmodified `main` too)
- [x] Full suite: `uv run pytest tests/ -q --ignore=tests/stress` — 4403 passed, 34 skipped, 34 failed. All 34 failures confirmed pre-existing on unmodified `main` (re-ran the identical failing test files against `git stash`'d — i.e. unmodified — tree: same 16 + 18 = 34 failures, unrelated to this change: missing WhatsApp service files, unrelated push-token signed-session tests, an unrelated hook test, etc.)
- [x] Falsifiability confirmed for every new/changed test file: reverted each implementation change in turn and observed the corresponding tests fail, then restored and observed them pass again (`pid_liveness.py`, `session_store.py`, `agent-monitor.py`, `inbox_server.py`, `health-check-v3.sh`'s new check).

## Manual test

This touches a DB schema (`agent_sessions.db`) and process-liveness logic feeding the dispatcher's own health-check system, so per the issue's own requirement, unit tests alone aren't sufficient. All manual testing below ran against a **throwaway `/tmp` SQLite DB and real-but-disposable subprocesses in this worktree** — the live host's `agent_sessions.db`, `lobster-claude.service`, and other systemd units were never touched.

1. **Dead-PID scenario (real subprocess, real kill, real reap):** spawned `sleep 60`, wrote a fresh-mtime output file with `stop_reason=tool_use` (would previously have left it "running"), killed and reaped the process, registered the row with the real (now-dead) PID against an isolated `/tmp` DB. `cleanup_stale_running_sessions()` correctly flipped it to `dead`; feeding the identical row through `agent-monitor.py`'s `classify_agent()` independently returned `GHOST_CONFIRMED`. Both classifiers agreed.
2. **Live-PID scenario (real long-running subprocess):** spawned `sleep 30`, registered a row with `spawned_at` 200 minutes in the past and a *missing* output_file (previously guaranteed a false "dead" verdict under the elapsed-time fallback), with the real live PID. `cleanup_stale_running_sessions()` correctly left it `running`; `classify_agent()` independently returned `HEALTHY`.
3. **`check_dispatcher_pid_consistency()` against real processes:** built a throwaway `agent_sessions.db` with a dispatcher row, set its `pid` to this shell's own real PID, and ran the extracted function for real — correctly logged a WARN (real live process, but named `bash` not `claude` — an honest mismatch, not a false positive). Repeated with a spawned-killed-reaped PID — correctly logged a WARN that the process is not alive. Both runs returned exit code 0 (non-fatal, as designed).

## Definition of Done

- [x] Integration/manual test run — see `## Manual test` above; all against isolated `/tmp` state, never the live host
- [x] User-visible outcome: not applicable — this is internal liveness-classification plumbing with no direct Telegram/UI surface
- [x] Side-effect audit: no floods, no unscoped writes (all manual runs used disposable `/tmp` DBs), no unintended volume; the new health-check function is read-only against `agent_sessions.db` and never gates a restart
- [x] External service calls: none — this PR has no network or external API surface

## Deploy note

Per the issue's explicit instruction: **this is a PR only.** Not merged to `main`. No changes were made to the live host's running `agent_sessions.db`, and no live systemd unit (including `lobster-claude.service`) was restarted or touched at any point during this work.

Closes #2148

### DIFF
--- a/hooks/auto-register-agent.py
+++ b/hooks/auto-register-agent.py
@@ -62,6 +62,40 @@ _LOG_PATH = (
     / "hook-failures.log"
 )
 
+# ---------------------------------------------------------------------------
+# PID ground truth (issue #2148, Phase 1)
+# ---------------------------------------------------------------------------
+#
+# This hook is exec'd fresh, as a direct child of the dispatcher's `claude` OS
+# process, on every Agent-tool call (PostToolUse). Verified live via
+# process-tree inspection (pstree -p) on 2026-08-03: Agent/Task-tool-spawned
+# subagents run in-process as threads of the single dispatcher `claude`
+# process — they have no OS PID of their own to record. What we CAN and do
+# record is the real PID of the dispatcher process this hook is a child of,
+# via a process-tree walk (not a flat os.getppid(), which would only be
+# correct for exactly one hop — the walk handles any intermediate wrapper
+# processes robustly, mirroring hooks/session_role.py's established
+# is_dispatcher_session() process-tree technique).
+_SRC_DIR = str(Path(__file__).resolve().parent.parent / "src")
+if _SRC_DIR not in sys.path:
+    sys.path.insert(0, _SRC_DIR)
+
+import agents.pid_liveness as _pid_liveness  # noqa: E402
+
+
+def _capture_dispatcher_pid() -> int | None:
+    """Best-effort: find the real dispatcher PID via a process-tree walk.
+
+    Never raises — any failure (missing /proc, unexpected process tree
+    shape, etc.) is swallowed and treated as "no dispatcher PID available",
+    exactly like a pre-migration row. Registration must never be blocked by
+    a PID-capture failure.
+    """
+    try:
+        return _pid_liveness.find_dispatcher_ancestor_pid()
+    except Exception:  # noqa: BLE001
+        return None
+
 
 # ---------------------------------------------------------------------------
 # Logging (failures only -- never to stdout, never exit non-zero)
@@ -197,6 +231,7 @@ def insert_agent_session(
     session_id: str,
     output_file: str | None,
     input_summary: str | None,
+    dispatcher_pid: int | None = None,
 ) -> None:
     """Insert a 'running' row into agent_sessions.db.
 
@@ -236,18 +271,37 @@ def insert_agent_session(
                 notified_at         TEXT,
                 trigger_message_id  TEXT,
                 trigger_snippet     TEXT,
-                reply_message_ids   TEXT
+                reply_message_ids   TEXT,
+                pid                 INTEGER,
+                dispatcher_pid      INTEGER,
+                pid_captured_at     TEXT
             )
         """)
+        # Additive migration for DBs created before this hook wrote the
+        # pid/dispatcher_pid/pid_captured_at columns (mirrors session_store.py's
+        # _MIGRATION_STMTS pattern). Each ALTER is a no-op on a fresh table that
+        # already has the column from the CREATE TABLE above.
+        for stmt in (
+            "ALTER TABLE agent_sessions ADD COLUMN pid INTEGER",
+            "ALTER TABLE agent_sessions ADD COLUMN dispatcher_pid INTEGER",
+            "ALTER TABLE agent_sessions ADD COLUMN pid_captured_at TEXT",
+        ):
+            try:
+                conn.execute(stmt)
+            except Exception:  # noqa: BLE001
+                pass  # Column already exists — no-op
+
         now = datetime.now(timezone.utc).strftime('%Y-%m-%d %H:%M:%S')
+        pid_captured_at = now if dispatcher_pid is not None else None
         conn.execute(
             """
             INSERT OR IGNORE INTO agent_sessions
                 (id, task_id, agent_type, description, chat_id, source,
-                 status, output_file, input_summary, spawned_at)
+                 status, output_file, input_summary, spawned_at,
+                 dispatcher_pid, pid_captured_at)
             VALUES
                 (?, ?, 'subagent', 'auto-registered by PostToolUse hook', ?, ?,
-                 'running', ?, ?, ?)
+                 'running', ?, ?, ?, ?, ?)
             """,
             (
                 agent_id,
@@ -257,6 +311,8 @@ def insert_agent_session(
                 output_file,
                 input_summary,
                 now,
+                dispatcher_pid,
+                pid_captured_at,
             ),
         )
         conn.commit()
@@ -298,6 +354,8 @@ def main() -> None:
         # and the dispatcher can reconstruct context if the agent fails.
         input_summary = prompt[:500] if prompt else None
 
+        dispatcher_pid = _capture_dispatcher_pid()
+
         insert_agent_session(
             agent_id=agent_id,
             task_id=metadata["task_id"],
@@ -306,6 +364,7 @@ def main() -> None:
             session_id=session_id,
             output_file=output_file,
             input_summary=input_summary,
+            dispatcher_pid=dispatcher_pid,
         )
 
     except Exception as exc:  # noqa: BLE001
--- a/scripts/agent-monitor.py
+++ b/scripts/agent-monitor.py
@@ -80,6 +80,16 @@ AGENT_OUTPUT_GLOB = _default_agent_output_glob()
 # Pattern for agent JSONL symlink filenames: agent-<hex_id>.jsonl
 AGENT_SYMLINK_PATTERN = re.compile(r"^agent-([0-9a-f]+)\.jsonl$")
 
+# PID ground truth (issue #2148, Phase 1) — shared liveness primitive.
+# agent_sessions.db and this script live in the same repo checkout, so the
+# import path is resolved relative to this file rather than assuming a
+# particular cwd.
+_SRC_DIR = str(Path(__file__).resolve().parent.parent / "src")
+if _SRC_DIR not in sys.path:
+    sys.path.insert(0, _SRC_DIR)
+
+from agents.pid_liveness import is_pid_alive  # noqa: E402
+
 # Age threshold for treating an unregistered output file as "active" (minutes)
 UNREGISTERED_ACTIVE_THRESHOLD_MINUTES = 30.0
 
@@ -94,6 +104,8 @@ class AgentRow:
     spawned_at: str
     output_file: str | None
     last_seen_at: str | None
+    pid: int | None = None
+    dispatcher_pid: int | None = None
 
 
 @dataclass(frozen=True)
@@ -238,15 +250,46 @@ def classify(
     output_file_age_minutes: float | None,
     threshold_minutes: float,
     output_file_threshold_minutes: float,
+    pid_alive: bool | None = None,
+    dispatcher_pid_alive: bool | None = None,
 ) -> Classification:
-    """Classify a running agent given age and output file liveness.
+    """Classify a running agent given age, output file liveness, and PID ground truth.
 
-    Logic (all thresholds configurable):
+    PID ground truth (issue #2148, Phase 1) is the PRIMARY signal, taking
+    precedence over the age/output-file heuristics below. Callers precompute
+    pid_alive/dispatcher_pid_alive via agents.pid_liveness.is_pid_alive() and
+    pass plain booleans (or None) here — keeping this function pure/testable
+    without any OS-level calls of its own, exactly like the rest of this
+    module's classification logic.
+
+      - pid_alive is not None (a real pid was recorded for this row): the
+        real OS process's existence is definitive. HEALTHY if alive,
+        GHOST_CONFIRMED if not — regardless of age or output_file mtime.
+      - pid_alive is None and dispatcher_pid_alive is False: no pid of its
+        own (an in-process Agent-tool subagent), but its parent dispatcher
+        process is confirmed dead — the subagent is necessarily dead too.
+        GHOST_CONFIRMED.
+      - pid_alive is None and dispatcher_pid_alive is True: the dispatcher
+        being alive does NOT prove this particular subagent is still
+        active (the dispatcher process running says nothing about any one
+        in-process conversation thread) — falls through to the existing
+        heuristics below, unchanged.
+      - Both None (pre-migration row, or no real process boundary exists):
+        falls through to the existing heuristics below, byte-for-byte
+        unchanged from pre-#2148 behavior.
+
+    Legacy heuristic (all thresholds configurable):
       - age < threshold             → HEALTHY (too young to worry about)
       - age >= threshold, no file   → STALE_NO_FILE (can't check liveness)
       - age >= threshold, file recent → GHOST_SUSPECTED (still writing, maybe slow)
       - age >= threshold, file old/missing → GHOST_CONFIRMED (likely dead)
     """
+    if pid_alive is not None:
+        return "HEALTHY" if pid_alive else "GHOST_CONFIRMED"
+
+    if dispatcher_pid_alive is not None and not dispatcher_pid_alive:
+        return "GHOST_CONFIRMED"
+
     if age_minutes < threshold_minutes:
         return "HEALTHY"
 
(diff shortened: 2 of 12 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS: Reviewed the diff cold first, then compared against the engineer's briefing.

**Precedence logic (session_store.cleanup_stale_running_sessions + agent-monitor.py classify()):** Verified directly against the checked-out PR branch, not just the diff. Both places implement the same precedence: real pid alive/dead is definitive and short-circuits (session_store `continue`s in both branches); if no pid but dispatcher_pid is confirmed dead, marks dead; if dispatcher_pid is alive, correctly falls through to the legacy heuristic instead of short-circuiting to "alive" (this is the specific failure mode that would cause false-negative zombie detection, and it's handled correctly — `if dispatcher_pid_alive is not None and not dispatcher_pid_alive` only fires on False, never on True). Rows with neither pid touch nothing new. One asymmetry worth noting for the record (not a bug): session_store's pid-precedence check is skipped when `file_status == "done"` (self-reported completion outranks a pid check), but agent-monitor.py's `classify()` has no equivalent "done" concept in its own legacy heuristic to begin with, so there's nothing to be consistent with there — not a regression, just a pre-existing scope difference between the two classifiers that this PR doesn't need to resolve.

**health-check-v3.sh check_dispatcher_pid_consistency():** Confirmed by reading the full function and running `tests/test-health-check-dispatcher-pid-consistency.sh` myself (11/11 pass, including the source-grep test that asserts the function never assigns `level=` or `restart_reason=`). It only calls `log_info`/`log_warn` and always `return 0`, including on missing DB, missing column, sqlite3 errors. Genuinely additive and non-gating — cannot itself trigger a restart. `MESSAGES_DIR` is a pre-existing script variable, not a newly invented path.

**find_dispatcher_ancestor_pid test fix:** The reasoning holds. Confirmed `hooks/session_role.py` already has an established, unrelated process-tree-walk technique (`_get_ppid`/`_is_claude_process`/`_is_dispatcher_by_process_tree`) for the same "is my ancestor a `claude` process" question, so the engineer's fix mirrors an existing pattern rather than inventing a new one. The negative test was indeed unreliable relying on real OS ancestry (this reviewer's own subagent execution environment is itself nested inside a dispatcher `claude` process, which would have made the old assertion flaky/environment-dependent) — faking `_read_ppid`/`_read_comm` at the layer `find_dispatcher_ancestor_pid()` itself calls is the correct fix, not a cover-up of a real bug in the function under test.

**No live state touched:** Confirmed no systemd/live-DB paths appear anywhere in the diff. All new/changed tests use `tmp_path`/`isolated_db` fixtures backed by real disposable SQLite files and real spawned-then-killed `sleep` subprocesses (verified `isolated_db` fixture definition — genuinely fresh `tmp_path` DB per test, connection closed after). `docker-worker.sh` confirmed to have zero callers anywhere in the repo and does not call `session_start` — the "no-op finding, no new integration point invented" claim is accurate.

**Test structure:** `TestDispatcherPidCapture` is defined in two different files (`tests/unit/test_hooks/test_auto_register_agent.py` and `tests/unit/test_mcp_server/test_dispatcher_pid_capture.py`) — not a collision: both directories have `__init__.py`, making them proper packages, so pytest has no basename-shadowing issue. No duplicate test class names within a single file/module.

**Test-result claims, independently re-run on the PR branch (not taken on faith):**
- `109 passed` across the 5 new/changed Python test files — reran exactly those 5 files together: **109 passed**, exact match.
- Shell suite for the new health check — reran `tests/test-health-check-dispatcher-pid-consistency.sh` directly: **11/11 pass**, exact match.
- Full suite pre-existing-failures claim — reran full suite (`uv run pytest tests/`) on both `main` and the PR branch. Actual count is **35 failed, 4425 passed, 34 skipped** on the PR branch vs **35 failed, 4390 passed, 34 skipped** on unmodified `main` — the engineer's PR description says "34 failed" (off by one from the actual 35), but critically **the exact same 35 test names fail on both branches**, confirming zero regressions from this PR; the pass-count delta (4425 − 4390 = 35) matches the net new tests added. Flagging the 34-vs-35 miscount as a minor accuracy note in the PR description, not a functional issue.

No correctness issues found in the liveness precedence logic, the additive health-check integration, or the test fix. Recommend fixing the "34" → "35" pre-existing-failure count in the PR description for accuracy, but that alone shouldn't block merge (merge decision is separate from this review anyway).

PR #2154, opened 2026-08-05
Capture and validate authenticated identity in OAuth token storage

### LINKED ISSUE #2153: OAuth token storage lacks identity metadata, allowing silent cross-account mixups
## Problem

The Google OAuth token stores (Calendar, Gmail, Workspace, and the newer
provider-parameterized `oauth_vault`) persist only:

```json
{
  "access_token": "...",
  "expires_at": "...",
  "scope": "...",
  "refresh_token": "..."
}
```

Tokens are keyed **only** by the requesting `chat_id`. Nothing records which
Google account actually completed the consent screen. This surfaced in
production: a user's chat_id held a Workspace token that turned out to be
authenticated as a different person's Google account — undetected until an
explicit API call happened to reveal whose data it was returning.

### Root cause

The push-token receiver endpoints (`push_calendar_token_endpoint`,
`push_gmail_token_endpoint`, `push_workspace_token_endpoint` in
`src/mcp/inbox_server_http.py`) validate a signed, single-use session token
that proves *"this push corresponds to the OAuth transaction requested by
chat_id=X"* (HMAC over instance_id|chat_id|scope|nonce|exp). That's a real
and valuable check — it stops a forged push from writing to an arbitrary
chat_id.

What it does **not** and cannot prove is *"the Google account that completed
the consent screen belongs to the person who owns chat_id=X."* If a consent
link is shared, forwarded, or clicked while a different Google account is
active in the browser, the transaction is still legitimately bound to the
right chat_id — but the wrong human authenticated it. Because no email (or
any identity claim) is captured from Google at grant time, this class of
mixup is invisible until something explicitly reads the account's data.

## Fix

- Capture the authenticated email at grant time via Google's `userinfo`
  endpoint (called with the just-issued access_token — no client secret
  needed), independent of myownlobster.ai.
- Store `email` as metadata alongside the existing token fields, consistently
  across Calendar, Gmail, Workspace, and `oauth_vault`.
- Add a self-consistency check: if a token already exists for a chat_id and
  the newly-granted email differs from the previously stored one, log loudly
  and surface a warning in the push confirmation sent to the user — instead
  of staying silent until an API call happens to reveal it.
- Add an optional expected-identity check against a `known-users.json`
  registry (chat_id -> expected email) for the first-grant case, where there
  is no prior token to compare against.
- Verify token storage/lookup is correctly scoped per chat_id with no
  shared/global fallback that could serve the wrong person's token (the
  existing workspace-token fallback already keys off the *same* chat_id, so
  this is a verification task, not expected to require a design change).

## Scope note

myownlobster.ai (the external service holding the GCP client secret) is out
of this repo and out of scope for a code change here — but it currently may
not request the `email`/`profile` scope from Google, which the userinfo call
depends on. If that scope isn't granted, `fetch_authenticated_email` will
fail gracefully (email metadata stored as `null`, logged clearly) rather than
blocking the token from being usable. Flagging this as a necessary follow-up
outside this repo.

aeschylus:
## Problem

The Google OAuth token stores (Calendar, Gmail, Workspace, and the newer
provider-parameterized `oauth_vault`) persist only `access_token`,
`expires_at`, `scope`, and `refresh_token` — keyed by `chat_id`. Nothing
records which Google account actually completed the consent screen. In
production, a chat_id's Workspace token turned out to be authenticated as a
different person's Google account, undetected until an explicit API call
happened to reveal whose data it was returning.

The push-token receivers (`push_calendar_token_endpoint`,
`push_gmail_token_endpoint`, `push_workspace_token_endpoint` in
`inbox_server_http.py`) validate a signed, single-use session token proving
"this push corresponds to the OAuth transaction requested by chat_id=X." That
check is real and valuable — it stops a forged push writing to an arbitrary
chat_id. What it cannot prove is that the Google account which completed the
consent screen belongs to the person who owns chat_id=X. A shared or
forwarded consent link, clicked while a different Google account is active
in the browser, produces exactly that gap — and with no identity captured
from Google at grant time, the mixup was invisible until something explicitly
read the account's data.

## Fix

Adds `src/integrations/google_auth/identity.py`:
- `fetch_authenticated_email(access_token)` — calls Google's `userinfo`
  endpoint directly with the just-issued access token (no client secret
  needed, so it works from an instance that never holds GCP credentials
  locally). Never raises; degrades to `None` on any failure.
- `check_identity_consistency(previous_email, new_email)` — pure comparison:
  `match` / `mismatch` / `no_baseline` (first grant, nothing to compare yet)
  / `email_unavailable`.
- `expected_email_for_chat_id` / `check_expected_identity` — an optional,
  operator-maintained `known-users.json` registry (chat_id -> expected
  email) for the first-grant case, where self-consistency alone has no prior
  value to compare against. Absent or empty file degrades gracefully to "no
  expectation on file," never a hard dependency.
- `format_mismatch_warning` — pure formatting of a short, human-readable
  warning string, only for an actual mismatch.

`TokenData` (the dataclass shared by all four token stores) gains an
`email: Optional[str] = None` field. All four serialization layers
(`google_calendar/token_store.py`, `gmail/token_store.py`,
`google_workspace/token_store.py`, `oauth_vault/vault_store.py`) persist and
reload it, and all four `get_valid_token` refresh paths preserve it across a
refresh — the same pattern already used for `scope` and `refresh_token`.
Tokens saved before this field existed load cleanly with `email=None`.

All three push endpoints now:
1. Read the *previous* stored email for that chat_id (if any) before
   overwriting the file.
2. Fetch the newly granted email via the userinfo call.
3. Store it as `email` in the token JSON.
4. Compare old vs. new. On a mismatch: log at WARNING with both emails (not
   secrets, safe to log), and append a short warning to the *same* push
   confirmation message already sent to the user on Telegram/Slack — so the
   mixup surfaces immediately instead of staying silent until an explicit API
   call reveals it. The token is still saved either way — a userinfo hiccup
   or a genuine mismatch never blocks the push.

### Scope note

myownlobster.ai (the external service holding the GCP client secret) is out
of this repo — but it currently may not request the `email`/`profile` scope
from Google, which the userinfo call depends on. If that scope isn't granted,
`fetch_authenticated_email` fails gracefully (`email` stored as `null`,
logged clearly) rather than blocking the token. Flagged in the issue as a
necessary follow-up outside this repo; not something this PR can fix on its
own.

I deliberately left `google_calendar/callback_server.py` (the standalone,
manually-run local OAuth flow, not part of the always-on push architecture)
untouched — it already exchanges the code directly with Google and isn't
the code path that produced the production incident. Extending it the same
way would be a reasonable follow-up but isn't required for this fix.

### Functional patterns used

- `TokenData` and `IdentityCheckResult` are immutable frozen dataclasses.
- `fetch_authenticated_email`, `check_identity_consistency`,
  `check_expected_identity`, and `format_mismatch_warning` are pure or have
  isolated side effects (the one HTTP call is separated from the comparison
  logic, which takes plain values and returns a plain result with no I/O).
- The push endpoints compose these pure functions with the existing
  load -> compare -> save -> confirm pipeline rather than branching
  imperatively on a mutable "did this fail" flag.

## Flow: what changes for a push

```
Before:
  push arrives -> validate session -> write token file -> queue confirmation
                                       (access_token, expires_at, scope, refresh_token)

After:
  push arrives -> validate session -> read PREVIOUS email (if a token
                                       already exists for this chat_id)
                                    -> fetch NEW email via userinfo
                                    -> write token file
                                       (+ email)
                                    -> compare previous vs. new:
                                         match / no_baseline -> confirmation unchanged
                                         mismatch -> WARNING logged +
                                           short warning appended to the
                                           SAME confirmation message
                                    -> queue confirmation
```

No existing transition is removed; the only new state is the mismatch
warning path, which is additive to the confirmation text and never blocks
the token write or the 200 response.

## Tests run

- [x] `uv run pytest tests/unit/test_integrations/test_google_auth_identity.py` — 22 passed (new module: `fetch_authenticated_email` success/HTTP-error/network-error/malformed-payload/empty-token cases; pure comparison logic; registry lookup; two-chat_id isolation)
- [x] `uv run pytest tests/unit/test_integrations/test_{gmail,google_calendar,workspace,oauth_vault_store,oauth_vault_client}_token_store*.py` — 226 passed (roundtrip preserves `email`, legacy tokens without `email` load as `None`, two chat_ids store independent emails, `email` preserved across refresh)
- [x] `uv run pytest tests/unit/test_mcp_server/test_push_{calendar,gmail,workspace}_token_{endpoint,confirmation}.py tests/unit/test_mcp_server/test_push_token_confirmation_shared.py` — 346 passed, 6 failed. The 6 failures are pre-existing and unrelated to this change (`instance_id_mismatch` against `LOBSTER_INSTANCE_URL`, reproduced identically on `main` before this PR's changes — confirmed via `git stash`).
- [x] `uv run pytest tests/unit -m "not slow"` (full unit suite) — 3868 passed, 16 failed, 31 skipped. All 16 failures are pre-existing/unrelated (same 6 push-endpoint env failures above, plus hook/script tests unrelated to OAuth).
- [x] Falsifiability check: reverted `identity.py` and the `TokenData`/token-store/`inbox_server_http.py` changes (via `git stash` + moving the new file aside), re-ran the new/updated tests — all failed with exactly the expected errors (`ModuleNotFoundError`, `TypeError: unexpected keyword argument 'email'`, `AttributeError: no attribute '_fetch_authenticated_email'`). Restored the changes and confirmed all tests passed again.

## Manual test

Ran a standalone script (not part of the test suite) that exercises the real
`save_token`/`load_token`/`check_identity_consistency`/
`expected_email_for_chat_id` functions directly against real temp-directory
files (only the outbound Google `userinfo` HTTP call was faked, since this
sandbox has no live Google OAuth token):
1. `fetch_authenticated_email` against a userinfo-shaped response -> returned the expected email.
2. Two chat_ids each called `save_token`/`load_token` for real -- inspected the actual JSON files on disk and confirmed each held its own distinct `email`, no cross-contamination.
3. Simulated the production incident directly: `check_identity_consistency(previous="<account A>", new="<account B>")` -> `status="mismatch"`, and `format_mismatch_warning` produced the exact warning string that gets appended to the Telegram confirmation.
4. `expected_email_for_chat_id` against a real `known-users.json` file -> correct lookup.

This is an internal token-storage change with no independent Telegram/Slack
delivery path beyond the existing `_queue_push_confirmation` mechanism
(already covered by the automated confirmation tests above, which assert the
mismatch warning text lands in the outbox file exactly as
`OutboxHandler.process_reply()` would deliver it) — there is no separate
user-facing surface to verify beyond what those tests already exercise.

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits in automated tests -- the one outbound Google `userinfo` call is always mocked in tests via `_fetch_authenticated_email`)
- [x] Manual verification run (see above) -- real file I/O against a temp directory, real function calls, only the Google network boundary faked
- [x] User-visible outcome verified via the existing outbox-confirmation test harness (the mismatch warning is asserted present in the actual text `_queue_push_confirmation` would deliver to Telegram/Slack)
- [x] Side-effect audit: no floods, no unscoped writes (all new file I/O goes through `tmp_path`-scoped fixtures in tests; production paths are unchanged, just an added `email` key alongside existing writes); no unintended volume
- [x] External service calls: the only new outbound call (`userinfo`) is mocked in every test; no manual run was made against the real Google endpoint since no live token exists in this environment -- degradation-on-failure is unit-tested instead (`test_fetch_authenticated_email_http_error_returns_none`, `_network_error_returns_none`, `_unparseable_json_returns_none`)

Closes #2153.

### DIFF
--- a/src/integrations/gmail/token_store.py
+++ b/src/integrations/gmail/token_store.py
@@ -31,7 +31,8 @@ Token schema on disk::
         "access_token":  "<string>",
         "expires_at":    "<ISO 8601 UTC>",
         "scope":         "<space-separated scopes>",
-        "refresh_token": "<string or null>"
+        "refresh_token": "<string or null>",
+        "email":         "<string or null -- authenticated account, issue #2153>"
     }
 
 Design principles
@@ -154,6 +155,7 @@ def _token_to_dict(token: TokenData) -> dict:
         "expires_at": token.expires_at.isoformat(),
         "scope": token.scope,
         "refresh_token": token.refresh_token,
+        "email": token.email,
     }
 
 
@@ -167,6 +169,7 @@ def _dict_to_token(data: dict) -> TokenData:
         expires_at=expires_at,
         scope=data.get("scope", ""),
         refresh_token=data.get("refresh_token"),
+        email=data.get("email"),
     )
 
 
@@ -457,12 +460,13 @@ def get_valid_token(
         )
         return None
 
-    # Merge: preserve scope and refresh_token from the stored token
+    # Merge: preserve scope, refresh_token, and email from the stored token
     refreshed = TokenData(
         access_token=refreshed_partial.access_token,
         expires_at=refreshed_partial.expires_at,
         scope=token.scope,           # preserve original scope
         refresh_token=token.refresh_token,  # Google doesn't return new refresh_token here
+        email=token.email,           # preserve identity metadata across refresh
     )
 
     _save_token_local(user_id, refreshed, token_dir)
--- a/src/integrations/google_auth/identity.py
+++ b/src/integrations/google_auth/identity.py
@@ -1,0 +1,252 @@
+"""
+Token identity metadata — capturing and validating *who* actually granted
+an OAuth token, not just trusting the chat_id it was pushed for.
+
+Why this exists (issue #2153)
+------------------------------
+The Calendar/Gmail/Workspace push-token receivers in ``inbox_server_http.py``
+validate a signed, single-use session token proving "this push corresponds
+to the OAuth transaction requested by chat_id=X". That proves the
+*transaction* is legitimate. It does NOT prove that the Google account
+which completed the consent screen belongs to the person who owns chat_id=X
+— a consent link that gets shared, forwarded, or clicked while a different
+Google account is active in the browser produces a token that is bound to
+the right chat_id but authenticated by the wrong human. Without any
+identity claim captured from Google itself, that mixup is invisible until
+something explicitly reads the account's data.
+
+This module captures the authenticated email at grant time (via Google's
+``userinfo`` endpoint, called with the just-issued access_token — no client
+secret required) and provides pure comparison functions so callers can
+detect a mismatch immediately instead of silently.
+
+Two independent checks are supported, since neither alone covers every case:
+
+- ``check_identity_consistency`` — self-consistency: does the newly granted
+  email match the email captured for this chat_id last time? Works out of
+  the box with no external data, but has nothing to compare against on a
+  chat_id's very first grant.
+- ``check_expected_identity`` — checks the newly granted email against an
+  optional ``known-users.json`` registry (chat_id -> expected email), which
+  *does* cover the first-grant case, provided the registry has been
+  populated for that chat_id.
+
+Design principles
+------------------
+- Side effects (the HTTP call to Google, reading the registry file) are
+  isolated to dedicated functions; the comparison logic is pure.
+- Never raises on failure to fetch or parse — callers get ``None``/a
+  descriptive status, never an exception, so a userinfo hiccup can never
+  block a token from being saved and used.
+- No token or email values are ever logged at a level that would leak
+  outside this instance's own log files; log messages here only echo
+  chat_id and the (non-secret) email address for operator visibility.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional
+
+import requests
+
+log = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Google userinfo endpoint
+# ---------------------------------------------------------------------------
+
+#: Google's OpenID-Connect-compatible userinfo endpoint. Requires the access
+#: token to have been granted with at least one of the `email`, `profile`,
+#: or `openid` scopes -- if the consent grant didn't request one of those,
+#: this call fails with 403 and fetch_authenticated_email returns None.
+GOOGLE_USERINFO_URL: str = "https://www.googleapis.com/oauth2/v3/userinfo"
+
+_HTTP_TIMEOUT: int = 10
+
+
+def fetch_authenticated_email(
+    access_token: str,
+    *,
+    userinfo_url: str = GOOGLE_USERINFO_URL,
+    timeout: int = _HTTP_TIMEOUT,
+) -> Optional[str]:
+    """Return the email address of the Google account that granted this token.
+
+    Calls Google's userinfo endpoint directly with the access_token as
+    bearer auth -- this requires no client secret, so it works from an
+    instance (like this one) that never holds GCP credentials locally.
+
+    Args:
+        access_token: The just-issued (or just-refreshed) access token.
+        userinfo_url: Injectable for testing.
+        timeout:      HTTP timeout in seconds.
+
+    Returns:
+        The account's email address, or None if the call fails for any
+        reason (network error, non-2xx response, missing/empty email in the
+        payload, insufficient scope). Never raises.
+    """
+    if not access_token:
+        return None
+
+    try:
+        resp = requests.get(
+            userinfo_url,
+            headers={"Authorization": f"Bearer {access_token}"},
+            timeout=timeout,
+        )
+    except requests.exceptions.RequestException as exc:
+        log.warning("userinfo fetch failed (network error): %s", exc)
+        return None
+
+    if not resp.ok:
+        log.warning(
+            "userinfo endpoint returned %d -- token may lack email/profile "
+            "scope, or the access_token is invalid/expired.",
+            resp.status_code,
+        )
+        return None
+
+    try:
+        data = resp.json()
+        email = data.get("email")
+    except (ValueError, AttributeError) as exc:
+        log.warning("userinfo endpoint returned unparseable payload: %s", exc)
+        return None
+
+    if not email or not isinstance(email, str):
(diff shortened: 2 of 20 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS: Reviewed diff cold, then checked against the engineer's briefing, then independently verified in a local checkout of this exact commit (ca24ea1e).

**Verified claims:**
- `TokenData` is a single shared dataclass (`google_calendar/oauth.py`), imported by gmail/workspace/oauth_vault stores — the `email` field addition is consistent across all four, no type/signature mismatches at any call site (`_capture_identity_metadata`, `check_identity_consistency`, `format_mismatch_warning` all called with correct args).
- `TestEmailIdentityMetadata` class name repeats across 5 test files but each lives in its own module — no shadowing/collection collision.
- New/changed tests genuinely exercise before-state (e.g. `test_email_preserved_across_refresh` saves a real token via `save_token` first, then calls `get_valid_token` with only the network proxy patched, then asserts both the return value and the reloaded on-disk file) — not comment-only assertions.
- Ran the scoped push-endpoint + token-store suite myself: 321 passed / 6 failed, same 6 `instance_id_mismatch` failures the engineer described. Checked out the branch's parent commit (origin/main tip, cff8e966) and reran those two failing test files — same 2 failures reproduce identically pre-PR. Claim confirmed accurate, not undercounted.

**New findings not in the briefing:**
1. (Should-fix, not blocking) `_capture_identity_metadata` → `fetch_authenticated_email` makes a synchronous `requests.get` call (10s timeout) directly inside `async def push_{calendar,gmail,workspace}_token_endpoint`. This app runs via plain `uvicorn.run(app, ...)` with no worker count set (single process/event loop). The codebase already had one blocking call in this path (`fetch_preview`, called synchronously per `_queue_push_confirmation`'s own docstring) — this PR adds a second one in series, before the token is even saved, roughly doubling the worst-case window during which this one process cannot service *any other* concurrent request (including other users' pushes) while waiting on Google. Given this endpoint is exactly the incident-adjacent code path, worth a follow-up to move both calls off the event loop (thread pool / async http client), even though it's compounding a pre-existing pattern rather than introducing a new one.
2. (Cosmetic) `format_mismatch_warning(result, *, chat_id: str)` never actually uses `chat_id` in the returned message — dead parameter. Harmless but worth either using it or dropping it.

**Confirmed as described, not blockers:** known-users.json first-grant blind spot is real and documented; myownlobster.ai scope dependency is genuinely out of this repo's control and degrades gracefully; `callback_server.py` is correctly out of scope (legacy manual flow, not the incident's code path).

No correctness bugs found. Proof-of-work (revert-and-watch-fail, isolation tests, real disk round-trip) holds up under independent verification.

PR #2156, opened 2026-08-05
Add Opus-backed PII scanner as a native git pre-push hook

### LINKED ISSUE #2155: Add Opus-backed PII scanner as a native git pre-push hook
## Problem

Drew asked for a permanent enforcement mechanism that scans every push for PII before it leaves the machine, using an Opus agent for real semantic judgment (not just regex). Two things exist today but neither closes the gap:

- `hooks/secret-scanner.py` / `scripts/pre-push-security-scan.sh` — regex-based. The pre-push script's `PII_PATTERNS` only **warn**, they never block, and there's no semantic understanding (can't tell "Eloso" the company from a real person's name, can't recognize a CRM export by shape).
- No hook currently intercepts `git push` with an LLM in the loop.

Standing concern: this is a live CRM/outreach business (Eloso) with real prospect data flowing through Kissinger/eloso-bisque/CRM sheets, and the target repo is public. A regex miss or an accidental data-export commit is a real, not hypothetical, risk.

## Design

**Mechanism: native git `pre-push` hook**, not a Claude Code `PreToolUse` hook. A `PreToolUse` matcher on `Bash` only fires inside a live Claude Code session and only for pushes issued through the Bash tool — it's trivially bypassed by a push from a plain terminal, a cron script, or any tool other than this specific session's Bash tool. `git`'s own `pre-push` hook is the only interception point that covers every invocation path (CLI, script, IDE, Claude Code Bash tool alike), because it's the point where content actually leaves the machine.

- `hooks/pii-scan-guard.py` — new hook script (following the `pin-dependencies-guard.py` / `secret-scanner.py` house style: pure functions, subprocess-tested CLI entrypoint, importlib-loadable for unit tests).
  - Reads the git pre-push stdin protocol, resolves the diff of exactly what's about to leave (new commits not yet on the remote tracking ref — for a new branch, commits not reachable from any remote-tracking ref).
  - Filters the diff to a PII-relevant subset: skips binaries and lockfiles, but — unlike the existing regex scanner's `should_skip_file` — does **not** skip `.md/.txt/.csv/.json`, since an accidental CRM/contact-list export is exactly the kind of file that would land in one of those formats and be silently exempted by the existing skip-list.
  - If nothing scannable changed (lockfile-only / binary-only push), skips the LLM call entirely — cost stays proportional to what actually needs judgment.
  - Calls Opus (`claude-opus-4-8`, official `anthropic` Python SDK, structured JSON output) with a dedicated scanner system prompt (`hooks/pii-scan-guard.prompt.md`) instructing it to flag real personal names/emails/phones/addresses and accidental data exports, using an allowlist (reusing the existing `.security-allowlist` file/format so there's one allowlist to maintain, not two) for known-safe company names and already-decided-safe strings — explicitly NOT flagging company names, code identifiers, or secrets (secrets stay `secret-scanner.py`'s job).
  - On a `block` verdict: exits non-zero with a message naming each finding's file/line/snippet/reason (matching the existing `pin-dependencies-guard.py` deny-message style) — never just "PII found".
  - Fails **open** (warns, does not block) on scanner errors — missing API key, network/timeout, malformed response — so a transient Anthropic API hiccup can't brick every push on the system. This is a deliberate tradeoff given the stated risk of an unusable/disabled hook; documented in the module docstring.
  - Rollout gate: `LOBSTER_PII_SCAN_MODE` env var, default `off` (hook installed but inert), `warn`, `block` — mirroring the existing `block-claude-p.py` warn/block precedent.
- `scripts/install-pii-scan-hook.sh` — standalone installer for any repo (default: current repo), so it's usable beyond `SiderealPress/lobster`. **Not** wired into `install.sh`'s automatic flow yet — installing/enabling it live is a separate, explicit step gated on Drew's sign-off.
- `anthropic==0.97.0` added as an exact-pinned dependency (was previously only present as a transitive/global install, not declared).

## Plan
- [x] Design mechanism (native pre-push hook vs PreToolUse) and document the choice
- [x] Implement `hooks/pii-scan-guard.py` (diff resolution, file filtering, allowlist loading, Opus call, verdict formatting, mode gate, fail-open error handling)
- [x] Write `hooks/pii-scan-guard.prompt.md` (scanner system prompt)
- [x] Write `scripts/install-pii-scan-hook.sh`
- [x] Unit tests with the Anthropic API call mocked (no production hits in automated tests), including a falsifiability check (mutate detection logic, confirm tests fail, restore)
- [x] Manual integration test: real repo, real push attempt, real (small, scoped) Opus call against synthetic PII fixtures — verify block/allow behavior end to end
- [x] Full test suite run to confirm no regressions
- [x] Open PR — NOT auto-merged; needs Drew's explicit sign-off since it changes core dev workflow and is high-blast-radius if it misfires (`LOBSTER_PII_SCAN_MODE=off` by default; not wired into `install.sh`)

aeschylus:
Closes #2155

## Problem

The only PII detection that runs today (`scripts/pre-push-security-scan.sh`'s `PII_PATTERNS`) is regex-based and **warn-only** — it never blocks a push. It also has no way to tell a company name from a real person's name, or to recognize a contact-list export by shape. Given this is a live CRM/outreach business (Eloso) with real prospect data flowing through Kissinger/eloso-bisque/CRM sheets, and the target repo is public, an accidental data-export commit is a real risk, not a hypothetical one.

## What this adds

`hooks/pii-scan-guard.py` — a **native `.git/hooks/pre-push` hook**, not a Claude Code `PreToolUse` hook. That choice is deliberate: a `PreToolUse` matcher on `Bash` only fires inside a live Claude Code session, for pushes issued through that specific session's Bash tool — it would be silently bypassed by a push from a plain terminal, a cron script, or any other tool. Git's own pre-push hook is the one interception point that covers every path a push can take, because it fires at the moment content actually leaves the machine.

What it does on each push:
1. Reads the git pre-push stdin protocol and resolves exactly what's new — for an existing branch, the range between the remote and local sha; for a brand-new branch/ref, commits not reachable from *any* remote-tracking ref (falls back to the empty tree for a fresh root commit).
2. Filters that diff to a PII-relevant subset: binaries and lockfiles are skipped, but — unlike the existing regex scanner's skip-list — `.md/.txt/.csv/.json` are **not** exempted, since an accidental CRM/contact-list export is exactly the kind of file that would land in one of those formats. If a push touches nothing scannable (lockfile/binary-only), the LLM is never called — cost stays proportional.
3. Sends the filtered diff to Opus (`claude-opus-4-8`, structured JSON output, official `anthropic` SDK — not the `claude` CLI, since `hooks/block-claude-p.py` already discourages shelling to `claude -p` from Bash, and a git hook has no live Claude Code session to delegate to anyway) with a dedicated scanner prompt (`hooks/pii-scan-guard.prompt.md`), plus the repo's existing `.security-allowlist` file as "already reviewed, known-safe" context — reusing that file rather than inventing a second allowlist.
4. On a confident `block` verdict: exits non-zero with each finding's file/line/snippet/reason (matching `pin-dependencies-guard.py`'s deny-message style) — never just "PII found."
5. Fails **open** on any scanner error (no API key, network failure, timeout, malformed response) — warns and lets the push through rather than bricking every push on the system for a transient API hiccup. This is a deliberate tradeoff, documented in the module docstring, given the explicit concern that a hook with a high false-positive/failure rate gets disabled and stops protecting anything.

**Rollout gate:** `LOBSTER_PII_SCAN_MODE` (`off` default / `warn` / `block`), mirroring the existing `block-claude-p.py` precedent. `scripts/install-pii-scan-hook.sh` installs the hook into any target repo (default: current repo) — it is **not** wired into `install.sh`'s automatic flow, and this PR does not touch `.git/hooks/` or `.claude/settings.json` on this system. The hook exists and is tested; nothing about how pushes behave changes until someone explicitly installs it and sets the mode.

`anthropic==0.120.2` added as an exact-pinned dependency (it was previously present only as an untracked transitive/global install, not declared in `pyproject.toml`).

## What this does NOT do

- Does not touch or duplicate secret/credential detection — that stays `hooks/secret-scanner.py`'s (PreToolUse, warn) and `scripts/pre-push-security-scan.sh`'s (pre-push, block on `CRITICAL` patterns) job.
- Does not modify `scripts/pre-push-security-scan.sh` itself, or its regex `PII_PATTERNS` — this is a separate, additional layer, not a replacement.
- Does not install or enable anything live. `LOBSTER_PII_SCAN_MODE` is unset everywhere by default, so the hook is inert even where installed.

## Functional patterns used

`run()` is the fully-testable orchestration core: every side effect (stdin text, environment, repo root, and the scanner call itself) is passed in as a plain value/callable rather than read from global state, so the whole decision tree (mode gating, bypass, block/warn/allow, fail-open paths) is exercised with plain function calls and an injected fake scanner — no mocking of the network module needed. `main()` is a thin, non-testable shell around it (stdin/env/exit code only). The diff-filtering and prompt-building steps are pure functions with no side effects.

## Manual test

Ran three things beyond the unit suite, since this touches file I/O (installing into `.git/hooks/`), git-native hook behavior, and a real external API:

1. **Real Opus call against synthetic fixtures** (`hooks/pii-scan-guard.py`'s `call_scanner`, direct call, no mocking): a synthetic CRM-shaped CSV (`Priscilla Nakamura,priscilla.nakamura@brightleafconsulting.io,...` — fabricated name/email/phone, not a real person) → verdict `block` with correct per-row findings. A diff containing only the allowlisted company names (`Eloso`, `Trinity Rail`) plus a `support@eloso.io`-style address → verdict `allow`. Both matched expectations.
2. **Full pipeline, real API, real git repo**: built a throwaway local repo, committed a synthetic leads-export CSV, ran `run()` end-to-end (git diff resolution → filtering → real Opus call) in `mode=block` → exit code `1`, finding correctly named `leads.csv:2` with a coherent reason. This exercises the actual code path a real push would take, not just an isolated prompt/response check.
3. **Install script against a throwaway repo**: installs correctly (executable, marker present), refuses to overwrite an unrelated existing `pre-push` hook without `--force`, backs up and overwrites with `--force`, re-installs in place over its own prior install, and the installed hook actually runs `pii-scan-guard.py` end-to-end when invoked with a real pre-push-shaped stdin payload.

No production Telegram/Slack-facing output is involved, so there's nothing to ask the user to visually confirm — the "user-visible outcome" here is the exit code/stderr contract with `git push`, verified directly above.

**Side-effect audit:** the two real Opus calls above (2 total) used only fabricated names/emails/phone numbers, not real people — no actual customer/prospect data was ever sent to the API. No shared state is written to. No unbounded volume risk — the hook makes at most one Opus call per `git push` invocation, gated by mode (currently `off` everywhere).

## Tests run
- [x] `uv run pytest tests/unit/test_hooks/test_pii_scan_guard.py -v` — 55 passed
- [x] `uv run pytest tests/unit/test_install_pii_scan_hook.py -v` — 7 passed
- [x] Falsifiability check: inverted the `mode == "off"` gate in `run()`, confirmed 8 tests failed as a direct result (mode gating and the bypass/warn/block paths that depend on it), reverted, confirmed all 55 pass again
- [x] `uv run pytest tests/unit/ -q` (full suite) — 4206 passed, 16 failed, 31 skipped. All 16 failures are pre-existing and unrelated (calendar/gmail/workspace token endpoint enforcement, slack router config, context monitor matcher, on-compact session-id writing, script inbox sources) — none touch `pii-scan-guard`, `install-pii-scan-hook`, or the new `anthropic` dependency. Zero new failures introduced.

## Manual test evidence (see above)
- [x] Real Opus call, synthetic CRM export → `block` with correct findings; company-name/support-email diff → `allow`
- [x] Full `run()` pipeline against a real temp git repo + real Opus call → exit code 1, correct finding
- [x] Install script: fresh install, refuse-without-force, `--force` backup+overwrite, idempotent re-install, installed hook runs end-to-end

## Definition of Done
- [x] Unit tests pass with proper mocking (fake scanner injected via `call_scanner_fn`; the two real-API calls above are the deliberate, explicitly-scoped manual verification step, run outside the automated suite, using only fabricated data)
- [x] Integration/manual test run — see `## Manual test`
- [x] User-visible outcome: not applicable — this is a CLI/git-hook exit-code contract, not user-facing Telegram/Slack output
- [x] Side-effect audit complete: no floods, no unscoped writes, bounded to one Opus call per push, fabricated data only in manual tests
- [x] External service calls: mocked in the automated unit suite; the two real-API manual-test calls are explicitly scoped (synthetic fixtures, 2 calls total) and documented above

## Needs Drew's explicit sign-off before enabling

This is **not** an autonomous-merge candidate. `LOBSTER_PII_SCAN_MODE` defaults to `off` everywhere and `scripts/install-pii-scan-hook.sh` is not wired into `install.sh`, so merging this PR changes nothing about how any push behaves — but the mechanism itself (an LLM call that can block every push on the system) is high-blast-radius if it has bugs discovered only after enabling, so turning it on for real (installing the hook + setting `LOBSTER_PII_SCAN_MODE=warn` or `block`) should be a separate, deliberate step after review, same pattern as PR #2151.

### DIFF
--- a/hooks/pii-scan-guard.prompt.md
+++ b/hooks/pii-scan-guard.prompt.md
@@ -7,6 +7,16 @@ You are a PII (personally identifiable information) scanner reviewing a git diff
 - Physical mailing addresses tied to a named private individual.
 - Any other content a reasonable person would immediately recognize as a clear privacy problem if this repository were public — even if it doesn't fit neatly into the categories above.
 
+## Framing never overrides the rule above
+
+A named private individual's real-looking email, phone number, or physical/mailing address is a privacy problem regardless of the surrounding language, file name, or stated purpose. Do not let any of the following talk you out of flagging it:
+
+- **Business, consulting, partnership, vendor, or affiliate framing.** "Strategic partnership," "consulting engagement," "vendor onboarding," "affiliate record," "partner intake," and similar business-register language describe *why* the data exists, not *what* it is. A person's name plus their personal contact details is exactly as much a privacy problem when it is labeled a "vendor record" or "partner account" as when it has no label at all. This applies even when the address is described as a "business," "registered," or "billing" address — a sole proprietor's or independent consultant's business address is still that person's address tied to their name, and the business framing does not make it safe to publish.
+- **"Fixture," "sample," "test data," or "for integration testing" labels.** A comment or file name claiming something is fixture/sample/test data does NOT by itself make its contents safe to allow. Only treat content as an exempt placeholder if the content itself is obviously synthetic (see "What NOT to flag" below for what counts as obvious). A realistic full name paired with a realistic-looking personal email and/or physical address is not made safe by a nearby claim that it is "representative sample data" or "used to exercise the integration pipeline" — a real contact record copy-pasted into a fixture file, with a comment added claiming it's synthetic, would look exactly the same. When you cannot tell whether fixture-labeled content is genuinely synthetic or a real record that has been labeled a fixture, treat it as real.
+- **One placeholder-looking field next to real-looking ones.** A phone number in the reserved fictional 555-01xx range does not make the rest of the record safe. If a name is paired with a realistic (non-`example.com`, non-obviously-fake) email address and/or a specific street address, flag the record even if the phone number in the same block happens to be in a placeholder format — evaluate the record as a whole, not by whichever field looks most like an example.
+
+This is a deliberate exception to the "when genuinely unsure, don't flag" calibration guidance below: whether something is a real private individual's contact information is exactly the judgment this scanner exists to make, and business/fixture framing is precisely the kind of surface plausibility that should increase scrutiny, not substitute for it.
+
 ## What NOT to flag
 
 - Company, product, or business names (e.g. "Eloso", "Trinity Rail") — these are not PII, no matter how name-like they sound.
--- a/hooks/pii-scan-guard.prompt.md
+++ b/hooks/pii-scan-guard.prompt.md
@@ -1,0 +1,26 @@
+You are a PII (personally identifiable information) scanner reviewing a git diff that is about to be pushed to a PUBLIC repository. Your job is to catch real privacy problems before they leave the machine — not to flag anything that merely looks like a name or an email address.
+
+## What to flag (category: "pii")
+
+- A real private individual's full name paired with their email address, phone number, or physical address (e.g. a customer, prospect, or employee's contact details).
+- Accidentally committed data exports: CRM/contact-list dumps, call/meeting transcripts naming real people, spreadsheet exports (CSV/JSON/TSV) containing rows of real people's contact information, mailing lists.
+- Physical mailing addresses tied to a named private individual.
+- Any other content a reasonable person would immediately recognize as a clear privacy problem if this repository were public — even if it doesn't fit neatly into the categories above.
+
+## What NOT to flag
+
+- Company, product, or business names (e.g. "Eloso", "Trinity Rail") — these are not PII, no matter how name-like they sound.
+- Code identifiers: variable names, function names, class names, config keys, test fixture names, commit messages describing code changes.
+- Generic or public business contact points: `info@`, `support@`, `sales@`, `hello@`, `noreply@`-style addresses for a company; a business's published phone number or office address.
+- Obvious placeholder/example data: `test@example.com`, `john@example.com`, `Jane Doe`, `555-0100`-style numbers, `123 Main St`, or anything clearly synthetic/fixture data used for tests or documentation.
+- Secrets, API keys, tokens, passwords, or credentials — a separate scanner already covers those. If you happen to notice one, you may mention it, but it is not your primary job and should not by itself drive your verdict.
+- The committing author's own name or email as it would normally appear in git metadata — you are only shown diff content, not commit authorship, so this should not come up, but if author-style metadata appears inside a file's content it is still just git-native and not a flagging reason on its own.
+- Anything listed in the KNOWN-SAFE ALLOWLIST section of the user message — those strings have already been reviewed and deliberately decided to be safe. Do not flag any match to an allowlist entry.
+
+## Calibration
+
+False positives here have a real cost: if this scanner blocks pushes on ordinary business content, people will stop trusting it and it will get disabled — which defeats the purpose entirely. When you are genuinely unsure whether something is a real private individual's data versus a business/public identifier, do not flag it. Reserve `block` for cases you are confident represent an actual privacy problem. Precision matters more than recall here — a missed edge case is recoverable; a hook that cries wolf on every push is not.
+
+## Output
+
+Respond only with the structured JSON your response format requires: a `verdict` of either `"block"` (at least one confident finding) or `"allow"` (nothing found), and a `findings` array (empty when `verdict` is `"allow"`). For each finding, give the `file` path, a best-effort `line` number in the new version of the file (use `0` if you cannot determine one), a short `category` label, the offending `snippet` (redact nothing — this stays local to the operator, it is not published anywhere), and a one-sentence `reason` explaining specifically what makes it a privacy problem and why it is not covered by the "what not to flag" list above.
--- a/hooks/pii-scan-guard.py
+++ b/hooks/pii-scan-guard.py
@@ -569,6 +569,7 @@ def run(
         return 0, ""
 
     all_findings: list[dict] = []
+    any_block_verdict = False
     truncated_any = False
     scanned_anything = False
 
@@ -616,12 +617,24 @@ def run(
         verdict, findings = validated
 
         if verdict == "block":
+            # The verdict itself is what decides block vs allow -- NOT
+            # whether findings happens to be non-empty. validate_scanner_response
+            # accepts {"verdict": "block", "findings": []} (or a "block"
+            # verdict with no "findings" key at all) as a well-formed
+            # response: there is no constraint requiring findings to be
+            # non-empty when the scanner says "block". Round-4 finding: this
+            # flag must be tracked independently of all_findings, because
+            # deciding the outcome later from `if not all_findings` (as the
+            # code used to) silently downgrades an empty-findings "block"
+            # verdict to "allow" -- the exact opposite of what "block" means,
+            # with zero warning to the operator.
+            any_block_verdict = True
             all_findings.extend(findings)
 
     if not scanned_anything:
         return 0, ""
 
-    if not all_findings:
+    if not any_block_verdict:
         if truncated_any:
             return (
                 0,
@@ -630,7 +643,19 @@ def run(
             )
         return 0, ""
 
-    message = format_findings_message(all_findings)
+    if all_findings:
+        message = format_findings_message(all_findings)
+    else:
+        # A "block" verdict with no usable findings details -- still a
+        # block, just without specifics to show. An empty findings list must
+        # never cause this to fall through to exit 0 / an empty message.
+        message = (
+            "BLOCKED: pii-scan-guard scanner flagged this push for review "
+            "but did not provide specific findings details.\n\n"
+            "Investigate manually before proceeding. If this is a "
+            "deliberate, reviewed exception for this one push, set "
+            f"{_ENV_BYPASS}=1."
+        )
     if mode == "warn":
         return 0, f"[pii-scan-guard] WARNING (warn mode, not blocking):\n{message}"
     return 1, message
--- a/hooks/pii-scan-guard.py
+++ b/hooks/pii-scan-guard.py
@@ -1,0 +1,550 @@
+#!/usr/bin/env python3
+"""
+Native git pre-push hook: semantic PII scan of everything about to be pushed.
+
+Unlike a Claude Code PreToolUse hook (which only fires inside a live Claude
+Code session, and only for pushes issued through that session's Bash tool),
+this is a real `.git/hooks/pre-push` hook — it runs for every `git push`
+regardless of how it was invoked (a plain terminal, a script, an IDE, or a
+Claude Code Bash tool call), because it fires at the point where content
+actually leaves the machine. That is the coverage guarantee a PreToolUse
+matcher cannot give: it would be silently bypassed by any push that doesn't
+go through this specific session's Bash tool.
+
+**What this hook does NOT duplicate:** secret/credential detection is already
+covered by hooks/secret-scanner.py (warn-mode, PreToolUse) and the regex
+CRITICAL patterns in scripts/pre-push-security-scan.sh (block-mode, this same
+pre-push point). This hook's job is semantic PII judgment specifically —
+real personal names/emails/phones/addresses and accidentally committed data
+exports (CRM lists, transcripts, contact spreadsheets) — the kind of judgment
+call a regex genuinely cannot make (e.g. telling the company name "Eloso"
+apart from a real person's name).
+
+**Deliberate divergence from scripts/security-scan-lib.sh's should_skip_file:**
+that skip-list exempts .md/.txt/.csv/.json/.rst files entirely, on the theory
+that those are documentation. For a PII scan specifically, that is exactly
+backwards — a CRM export or contact list is far more likely to land as a
+.csv/.json/.txt than as code. This hook only skips binaries and lockfiles
+(see _SKIP_EXTENSIONS / _SKIP_BASENAMES below); it does not skip by directory
+(tests/) or documentation extension.
+
+**Rollout gate:** controlled by LOBSTER_PII_SCAN_MODE (mirrors the existing
+warn/block precedent in hooks/block-claude-p.py):
+    off   (default) — hook is installed but does nothing. This is the
+                       shipped state: the mechanism exists and is tested, but
+                       does not affect any real push until explicitly turned
+                       on after review and sign-off.
+    warn            — scans and prints findings to stderr, never blocks.
+    block           — scans and exits non-zero (blocks the push) on a
+                       confident "block" verdict from the scanner.
+
+**Emergency bypass:** LOBSTER_PII_SCAN_SKIP=1 skips the scan entirely for one
+push (mirrors SECURITY_SCAN_SKIP=1 in scripts/pre-push-security-scan.sh).
+
+**Allowlist:** reuses the existing `.security-allowlist` file at the repo
+root (same file, same format, as scripts/pre-push-security-scan.sh) so there
+is one allowlist to maintain, not two. Entries are handed to the scanner as
+"already reviewed, known-safe" context rather than used as a post-hoc regex
+filter — the model is asked not to flag anything matching an allowlist entry.
+
+**Fail-open, by design:** if the scanner cannot run at all — no API key
+configured, a network error, a timeout, a malformed response — this hook
+WARNS and allows the push through; it does not block. A synchronous git hook
+that can fail closed on a transient API hiccup would brick every push on the
+system until someone notices and unsets it, which is a worse outcome than an
+occasional missed scan. This tradeoff is intentional: the blast radius of
+"scan didn't run this one time" is far smaller than "nobody can push
+anything, ever, until this hook is fixed or removed."
+
+**Cost/latency:** every scan is a real Opus call — not free and not
+instant. Two things keep this proportional: (1) the diff is filtered to
+PII-relevant content before it is sent, and if a push touches nothing but
+binaries/lockfiles the LLM is never called at all; (2) the diff sent to the
+model is capped at _MAX_DIFF_CHARS, with the push still evaluated against the
+truncated portion rather than refusing to scan large pushes outright. A
+human-visible latency of several seconds to roughly a minute per push that
+does contain real code changes is the accepted cost of catching what a regex
+cannot, given the stated stakes (real customer PII in a public repo).
+
+Exit codes (matching git's pre-push hook protocol):
+  0 - Allow the push (including all warn-mode and fail-open outcomes)
+  1 - Block the push (mode=block, confident PII finding)
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+_ENV_MODE = "LOBSTER_PII_SCAN_MODE"
+_ENV_BYPASS = "LOBSTER_PII_SCAN_SKIP"
+_VALID_MODES = ("off", "warn", "block")
+
+_MODEL = "claude-opus-4-8"
+_MAX_DIFF_CHARS = 200_000
+_SCAN_TIMEOUT_SECONDS = 60
+
+_ZERO_SHA = "0" * 40
+_EMPTY_TREE_SHA = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
+
+_ALLOWLIST_FILENAME = ".security-allowlist"
+
+# Extensions/basenames that can never carry meaningful PII content and are
+# never scanned. Deliberately narrow — see module docstring for why this does
+# NOT mirror scripts/security-scan-lib.sh's should_skip_file.
+_SKIP_EXTENSIONS = frozenset({
+    ".lock", ".min.js", ".min.css", ".map",
+    ".png", ".jpg", ".jpeg", ".gif", ".ico", ".svg",
+    ".woff", ".woff2", ".ttf", ".eot",
+    ".pdf", ".zip", ".tar", ".gz", ".bz2",
+    ".exe", ".dll", ".so", ".dylib",
+    ".pyc", ".pyo", ".class", ".o", ".a",
+})
+_SKIP_BASENAMES = frozenset({
+    "package-lock.json", "yarn.lock", "pnpm-lock.yaml", "uv.lock",
+    "Cargo.lock", "go.sum", "poetry.lock", "Gemfile.lock",
+})
+
+_SYSTEM_PROMPT_PATH = Path(__file__).parent / "pii-scan-guard.prompt.md"
+
+_RESPONSE_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "verdict": {"type": "string", "enum": ["block", "allow"]},
+        "findings": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "file": {"type": "string"},
+                    "line": {"type": "integer"},
+                    "category": {"type": "string"},
+                    "snippet": {"type": "string"},
+                    "reason": {"type": "string"},
+                },
+                "required": ["file", "line", "category", "snippet", "reason"],
+                "additionalProperties": False,
+            },
+        },
+    },
+    "required": ["verdict", "findings"],
+    "additionalProperties": False,
+}
+
+
+# ---------------------------------------------------------------------------
+# Pre-push stdin protocol
+# ---------------------------------------------------------------------------
+# git feeds pre-push hooks one line per updated ref on stdin:
+#   <local ref> SP <local sha1> SP <remote ref> SP <remote sha1> LF
+
+_REF_LINE_RE = re.compile(r"^(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s*$")
+
+
+def _is_zero_sha(sha: str) -> bool:
+    return sha == _ZERO_SHA
+
+
+def parse_pre_push_refs(stdin_text: str) -> list[tuple[str, str, str, str]]:
+    """Parse the pre-push stdin protocol into (local_ref, local_sha,
+    remote_ref, remote_sha) tuples. Malformed lines are skipped rather than
+    raising, since a hook that crashes on unexpected input fails closed in
+    the worst possible way (it would abort the shell pipeline mid-push)."""
+    refs = []
+    for line in stdin_text.splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        m = _REF_LINE_RE.match(line)
+        if not m:
+            continue
+        refs.append(m.groups())
+    return refs
+
+
+# ---------------------------------------------------------------------------
+# Diff resolution
+# ---------------------------------------------------------------------------
+
+
+def resolve_diff_base(remote_sha: str, local_sha: str, cwd: str) -> str | None:
+    """Return the sha to diff `local_sha` against to see exactly what is
+    about to be pushed, or None if there is nothing to diff (no-op update).
+
+    Existing branch update: base is simply the current remote sha — this is
+    exactly the commit range that is new to the remote.
+
+    New branch/ref (remote_sha is all-zero): there is no remote commit to
+    diff against, so the base is the parent of the oldest commit that isn't
+    already reachable from some other remote-tracking ref (i.e. the oldest
+    commit that is actually new to the remote as a whole, not just to this
+    ref) — falling back to git's empty-tree sha for a root commit with no
+    parent, or `local_sha` itself (an empty diff) if nothing new is found.
+    """
+    if local_sha == remote_sha:
+        return None
+    if not _is_zero_sha(remote_sha):
+        return remote_sha
+
+    try:
+        out = subprocess.run(
+            ["git", "rev-list", local_sha, "--not", "--remotes"],
+            cwd=cwd, capture_output=True, text=True, check=True,
+        ).stdout
+    except (subprocess.CalledProcessError, OSError):
+        return _EMPTY_TREE_SHA
+
+    new_commits = [c for c in out.splitlines() if c]
+    if not new_commits:
+        return local_sha
+
+    oldest = new_commits[-1]  # git rev-list lists newest-first
+    try:
+        parent = subprocess.run(
+            ["git", "rev-parse", f"{oldest}^"],
+            cwd=cwd, capture_output=True, text=True, check=True,
+        ).stdout.strip()
+        return parent or _EMPTY_TREE_SHA
+    except (subprocess.CalledProcessError, OSError):
+        # oldest commit has no parent (it's a root commit) -> whole tree is new
+        return _EMPTY_TREE_SHA
+
+
+def get_diff(base: str | None, local_sha: str, cwd: str) -> str:
+    if base is None:
+        return ""
+    try:
+        return subprocess.run(
+            ["git", "diff", base, local_sha],
+            cwd=cwd, capture_output=True, text=True, check=True,
+        ).stdout
+    except (subprocess.CalledProcessError, OSError):
+        return ""
+
+
+# ---------------------------------------------------------------------------
+# Diff filtering (PII-relevant subset)
+# ---------------------------------------------------------------------------
+
+_DIFF_HEADER_RE = re.compile(r"^diff --git a/(.*?) b/(.*?)$", re.MULTILINE)
+
+
+def _split_diff_by_file(diff_text: str) -> list[tuple[str, str]]:
+    """Return [(new_path, block_text), ...] for each file in a unified diff."""
+    matches = list(_DIFF_HEADER_RE.finditer(diff_text))
+    blocks = []
+    for i, m in enumerate(matches):
+        start = m.start()
+        end = matches[i + 1].start() if i + 1 < len(matches) else len(diff_text)
+        blocks.append((m.group(2), diff_text[start:end]))
+    return blocks
+
+
+def should_skip_file_for_pii_scan(path: str) -> bool:
+    """True if `path` can never carry meaningful PII content (binary or a
+    generated lockfile) and should be excluded from scanning."""
+    name = Path(path).name
+    if name in _SKIP_BASENAMES:
+        return True
+    if Path(path).suffix.lower() in _SKIP_EXTENSIONS:
+        return True
+    return False
+
+
+def _is_binary_diff_block(block: str) -> bool:
+    return "Binary files" in block and "differ" in block
+
+
+def filter_diff_for_scan(diff_text: str) -> tuple[str, bool]:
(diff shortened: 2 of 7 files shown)


aeschylus:
## Independent adversarial review — PR #2156

Verdict: **PASS with one substantive finding that should inform the rollout decision** (not a merge blocker — this PR is inert by default and not intended for autonomous merge/enable anyway). I reviewed the diff directly (`gh pr diff`), ran the test suite myself in a worktree, and independently re-derived several of the claims below rather than trusting the PR description as ground truth.

### 1. Fail-open-on-error design — I recommend keeping it, with a caveat

Fail-open is the right default here, but only because a hard-blocking regex secret scanner (`pre-push-security-scan.sh`, `CRITICAL` patterns) already runs at the same pre-push point and fails closed. This hook is additive semantic PII judgment layered on top of that, not the sole line of defense — so "scanner didn't run this one time" is a real but bounded risk, and "nobody can push anything until this is fixed" is a worse failure mode for a brand-new, unproven LLM-backed hook.

**Gap:** fail-open currently degrades silently. If the API key expires, the model id is deprecated, or the service has a multi-day outage, every push just... succeeds, with a stderr line most terminal/IDE workflows won't surface anywhere durable. There is no mechanism in this PR that would tell Drew the scanner has been a no-op for a week. Before flipping to `block` in production I'd want either a periodic synthetic health check (cron job that runs a known-should-block synthetic case through `call_scanner` and alerts on unexpected `allow`/exception) or at minimum durable logging of fail-open events somewhere monitored. Recommend as a follow-up, not a blocker for this PR.

### 2. Diff-base resolution — verified correct, including scenarios beyond the PR's own tests

Ran three of my own scenarios against a real temp git repo (not just the shipped test suite):
- Multi-branch repo where `origin/main` and an unrelated already-pushed `origin/feature-old` both exist, and a **new** branch forked from `main` (sharing no history with `feature-old`) is pushed for the first time with a synthetic-PII commit — `resolve_diff_base` correctly resolved to `main`'s sha, and the diff contained only the new branch's own content (not `feature-old`'s, not `main`'s).
- Same setup but with **two** new commits on the new branch — diff correctly covered both commits, not just the tip.
- Brand-new repo, zero remotes, root commit + a second commit, first-ever push — correctly fell back to the empty-tree sha and the diff contained the full history.

All three matched expectations. This is genuinely careful plumbing; no gaps found.

### 3. Scanner prompt — one real false-negative pattern found

Ran 7 of my own adversarial cases against the live prompt with real Opus calls (all synthetic/fabricated data, `-fake.example`/`.test`-style domains, no real people):

- Borderline name/company ("Morgan Advisory Group"), partial fragmentary CRM row (initials + phone, no header), non-English name + email + phone, code test fixture (`Alice Johnson <alice.johnson@example.com>` in a test file), and schema field names with no data — all classified correctly (allow/block as expected).
- **Finding:** a personal name + fake personal email + phone + **physical home address** — which the prompt's own "what to flag" list names verbatim ("physical mailing addresses tied to a named private individual") — was scored `allow` when the surrounding text used business-sounding framing (a `-consulting`-style email domain and "contact regarding partnership renewal" phrasing). I controlled for this: the identical pattern (name + email + phone + home address) with plain personal framing and no business language correctly blocked, twice, in two follow-up tests. So the model is measurably more likely to wave through home-address PII when it's wrapped in business-sounding language, even though the prompt's own explicit "always flag" category should apply regardless of framing. This is not a non-English-name bias (I controlled that variable too — an English name with the same business framing showed the same miss) — it's specifically business-context framing suppressing an address-PII finding.

This is a real, reproducible calibration gap worth fixing in the prompt before relying on `block` mode for anything address-related — e.g. an explicit instruction that a home/residential address paired with a name should be flagged regardless of any surrounding business language.

### 4. Bypass surface — verified `--no-verify`, one documentation gap

Built a throwaway repo + bare remote, installed the hook, and empirically confirmed: a plain `git push` triggers the hook (confirmed via a stub that exits 1), and `git push --no-verify` **completely bypasses it** — the stub was never invoked and the push succeeded. This is expected, standard git behavior (client-side hooks are always skippable this way), not a defect in this PR — but it is not mentioned anywhere in the module docstring, prompt doc, or install script, even though the docstring goes out of its way to argue this hook is more robust than a `PreToolUse` hook specifically because it "runs for every `git push`... regardless of how it was invoked." That phrasing is true only for pushes issued from a clone where the hook has actually been installed, and even then `--no-verify` defeats it. Recommend adding one sentence to the docstring naming the known-accepted gaps explicitly: `--no-verify`, any clone/machine that never ran the installer (hooks are never transmitted by `git clone`, this is inherent), and non-`git`-binary push paths (GitHub web UI edits, API-based commits, or any tool that speaks the git wire protocol directly without shelling out to `git push`). None of these are fixable by a client-side hook; they should be named as accepted residual risk, not silently absent. A GitHub-side required-status-check on PRs into `main` would be the complementary control for the "different clone" and "no install" gaps, worth a future issue.

### 5. Cost/latency scoping — confirmed correct

Read `filter_diff_for_scan`/`should_skip_file_for_pii_scan` and the orchestration in `run()`: binary and lockfile-extension/basename blocks are dropped before anything is sent to the model, and `scanned_anything` gates the API call entirely — a lockfile-only push never reaches `call_scanner_fn`. Confirmed via the existing `test_lockfile_only_push_skips_llm_call_entirely` test (passed) and by reading the code path directly; matches the claim.

### 6. Test evidence — reproduced, with one flaky (non-blocking) discrepancy

- `tests/unit/test_hooks/test_pii_scan_guard.py` — 55 passed (matches claim).
- `tests/unit/test_install_pii_scan_hook.py` — 7 passed (matches claim).
- Falsifiability check reproduced independently: inverted `mode == "off"` in `run()`, got 8 failures (same 8 the PR claims), reverted, back to 55 passed.
- Full suite: first run showed **17 failed, 4205 passed** — one more failure than claimed, and the extra one was `TestCliSubprocess::test_bypass_flag_exits_zero_with_notice` inside the PR's own new test file. This looked like a real problem, so I re-ran the full suite a second time: **16 failed, 4206 passed**, exact match to the PR's claim, and the pii-scan-guard test passed clean. Narrowing to just the 8 originally-failing-file set plus pii-scan-guard also reproduced cleanly at 16/202. Conclusion: this is order/resource-dependent flakiness under the full 4200+-test run, not a deterministic regression introduced by this PR — but it's a preexisting flakiness the team should be aware doesn't come from this PR's own logic being wrong (I couldn't isolate the interaction further in the time available). The other 16 pre-existing failures are unrelated (calendar/gmail/workspace push-token endpoint enforcement, slack router config, context monitor matcher, on-compact session-id writing, script inbox sources) — confirmed by name, none touch this PR's files.

### 7. Live-system safety — confirmed untouched

`~/.claude/settings.json` mtime is July 16 (predates this PR, no `pii` references in it). `~/lobster/.git/hooks/pre-push` does not exist. `~/lobster` remains on `main`, untouched by my review (all testing done in a separate worktree, removed afterward).

---

### Rollout recommendation (engineering judgment, Drew makes the final call)

- **Fail-open**: keep it, given the existing hard-blocking regex secret scanner is the real backstop and this is additive — but don't flip to `block` mode without adding some form of "is the scanner actually alive" monitoring first, since silent fail-open has no visibility today.
- **`LOBSTER_PII_SCAN_MODE`**: start at `warn` for a burn-in period once installed, not `block`. The address/business-framing false-negative above means `block` mode today would give false confidence on exactly the kind of accidental-export scenario this hook exists to catch. I'd want the prompt tightened (explicit "flag addresses regardless of framing" instruction) and re-tested before trusting `block` mode for real traffic.

No changes made to any live system. No real PII was sent to the API at any point — all test fixtures in both my testing and the PR's own manual tests used fabricated names/emails/phones/addresses.

aeschylus:
## Update: Drew's sign-off implemented (Fable 5 + fail-closed)

Drew's decision on the two review findings: **"Yes, use fable's judgement to complete the hook and prevent PII. Fail closed."** Both changes are now pushed to this branch (commit f5c72260).

### 1. Business-framing bypass — model swap + prompt tightening

Switched the scanner model from `claude-opus-4-8` to `claude-fable-5`, and tightened `hooks/pii-scan-guard.prompt.md` with an explicit new section ("Framing never overrides the rule above") that closes the specific bypass mechanism: business/consulting/partnership/vendor-affiliate framing, and "test fixture"/"sample data" labeling, no longer override the base rule that a named individual's realistic contact info must flag.

Verified live against the real API (not mocked):
- **Opus + the original (untightened) prompt** reliably scores `allow` on a realistic name + email + phone + address dressed up as a "sample partner-record payload used to exercise the partnership sync integration... during onboarding of a new consulting affiliate" — reproduced twice.
- **Fable 5 alone** (even with the untightened prompt) already resists this input and blocks it.
- **The tightened prompt alone** (on Opus) also independently blocks it.

So the fix isn't riding on model choice alone — both halves close the gap, which matters if a future model swap happens again.

### 2. Fail-open → fail-closed

`run()`'s scanner-unavailable path (missing API key, network error, timeout, malformed response) now returns exit code 1 ("BLOCKED... failing closed") in `block` mode, instead of the old exit 0 ("failing open"). `warn` mode is deliberately unaffected — a scanner outage there still only warns, since warn mode's whole purpose is passive observation before an operator opts into enforcement; making it block on its own transient errors would be a surprising regression for anyone still in the observation phase.

Also bumped `_SCAN_TIMEOUT_SECONDS` from 60s to 180s: Fable 5 turns run longer than Opus did (confirmed — 60s timed out on a real scan input during this change), and now that a timeout fails closed, too tight a budget would turn an ordinary slow response into a spurious blocked push.

### Tests

- Renamed/updated the two existing fail-open tests to assert fail-closed-in-block-mode behavior; added matching warn-mode-unaffected tests.
- Added a live-API integration test (`TestBusinessFramingBypassLive`, opt-in via `RUN_PII_SCAN_LIVE_TESTS=1`, skipped in the default `pytest tests/unit/` sweep per repo convention — unit tests don't hit production) that reproduces the exact bypass input and asserts `verdict == "block"`. Confirmed it fails under a full revert (opus model + untightened prompt, `assert 'allow' == 'block'`) and passes on the current code.
- Falsifiability check on the fail-closed logic: reverted the block-mode branch back to fail-open, confirmed both `test_*_fails_closed_in_block_mode` tests fail, restored, confirmed green again.
- Full unit suite: 4209 passed, 16 failed (all pre-existing/unrelated — same set documented in the original PR description), 32 skipped.

Still env-var gated and off by default (`LOBSTER_PII_SCAN_MODE` unset everywhere) — this change does not enable the hook anywhere.

aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: Core fail-closed behavior is verified correct and the counts/proof-of-work check out, but there are two real gaps and one judgment call that should be resolved before this goes anywhere near `block` mode in production.

**Independently verified (no issues):**
- Ran the full suite myself on `f5c72260`: **4209 passed, 16 failed, 32 skipped** — exact match to the claim. Confirmed none of the 16 failures touch any file this PR changes (they're in slack-router config, session-id hooks, calendar/gmail push-token endpoints, script-inbox sources) — genuinely pre-existing/unrelated, not something this diff introduced or is hiding behind.
- Re-ran the fail-closed proof myself against the **real** Anthropic API (not mocked) with a deliberately invalid key: `block` mode → exit 1, "failing closed" message. Identical scenario in `warn` mode → exit 0, warning only, does not block. Matches the engineer's described proof exactly.
- `claude-fable-5` is a real, current model ID (Anthropic's most capable widely released model) — not a hallucinated string.
- No duplicate test class names or shadowed test classes across the two new test files (66 tests collected, none silently unreachable).
- Every failure path I could construct — missing key, real 401, mocked timeout, mocked exception, refusal-shaped empty response (empty `content` → `json.loads("")` → exception) — correctly routes through `_scanner_unavailable_result` and fails closed in `block` mode / warns in `warn` mode.

**Gap 1 — one failure path is *not* covered by the fail-closed guarantee.** `run()` decides on the scanner's result with:
```python
if result.get("verdict") == "block":
    all_findings.extend(result.get("findings", []))
```
If `call_scanner` ever returns a syntactically valid dict where `verdict` is missing, `None`, or anything other than exactly `"block"`/`"allow"` (not an exception — a parseable-but-schema-violating response), this branch silently treats it as "allow" rather than routing to `_scanner_unavailable_result`. In practice `output_config.format` with the `enum: ["block","allow"]` json_schema should prevent the model from ever emitting that, but there's no defensive check and no test exercising it (the fakes only ever return exactly `"block"` or `"allow"`). For a hook whose whole design principle is "not knowing == treat as PII," this one branch relies entirely on an external guarantee (structured-outputs schema enforcement) rather than validating it locally. Suggest: `if result.get("verdict") not in ("block", "allow"): return _scanner_unavailable_result(...)` plus a test that feeds a malformed-but-valid-JSON dict through `call_scanner_fn`.

**Gap 2 — stale documentation that contradicts the new behavior it's describing.** The model swap itself is consistent (single `_MODEL` constant, no other hardcoded model strings, no fallback/retry path to miss), but two comments were not updated and are actively misleading:
- `call_scanner()`'s docstring: *"Call the Opus scanner... callers must catch and fail open"* — both wrong now (Fable 5, and the whole point of this PR is that callers fail *closed*). This is the kind of comment a future engineer would read while touching the fail-closed logic and get exactly backwards.
- `scripts/install-pii-scan-hook.sh` header: *"runs the Opus-backed PII scan"* — same staleness.

**Judgment call needing explicit sign-off (per the review brief):** leaving `warn` mode fail-open while only making `block` mode fail-closed is well-reasoned and documented in the code, but it's an interpretive reading of "fail closed" that the engineer made unilaterally. Recommend Drew explicitly confirms this reading rather than it being assumed correct by default.

**Minor, not blocking:** the tightened-prompt live test (`TestBusinessFramingBypassLive`) only proves the true-positive fix (the business/fixture bypass now blocks) — there's no equivalent live test proving the tightened prompt does *not* now flag an ordinary, legitimate business record (e.g. a real vendor contact with no PII framing tricks). Given the explicit worry raised in the engineer's own summary about over-triggering, a corresponding "should still allow" live case would round this out. Also worth noting for the record (not a defect): diffs over `_MAX_DIFF_CHARS` (200k) are truncated and the truncated portion is never scanned — this is disclosed to the user ("review manually for full coverage") rather than silent, so it's a known tradeoff, not a bug.

Nothing here is a correctness bug in the shipped behavior I could actually trigger — the hook is off by default everywhere, and the two things I did verify live (fail-closed proof, test counts) check out exactly as described. But Gap 1 and Gap 2 are real, low-effort fixes worth doing before this is turned on anywhere in `block` mode, and the warn-mode judgment call should get an explicit yes from Drew rather than ride through on the engineer's own read of "fail closed."

aeschylus:
Pushed a follow-up commit addressing both NEEDS-WORK findings from the last review:

1. **Fail-closed gap fixed:** `run()` only compared the scanner's verdict against `"block"` to decide whether to collect findings — any other verdict value (including a well-formed-JSON response with a garbage/missing verdict) fell through to the same outcome as `"allow"`, silently defeating fail-closed for that one failure mode. It now routes any verdict outside `{"block", "allow"}` through the same `_scanner_unavailable_result` path used for exceptions and missing API keys (fails closed in block mode, warns in warn mode). Added `test_unexpected_verdict_value_fails_closed_in_block_mode` (confirmed to fail without the fix) and a warn-mode counterpart for symmetry with the existing missing-key/exception tests.
2. **Stale docs fixed:** `call_scanner()`'s docstring and `install-pii-scan-hook.sh`'s header comment still described the old Opus-based, fail-open design. Both now describe the actual Fable 5 / fail-closed behavior.

Full `tests/unit/` suite: 4211 passed, 16 failed (same pre-existing, unrelated failures as before this change), 32 skipped — confirmed count did not drift.

Warn-mode's fail-open stance on scanner errors is unchanged, per the earlier note that this is a separate sign-off decision.

aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: Commit 22e036b7's fix is correct as far as it goes, but there's a third fail-closed gap left open, and it's arguably worse than the one this round fixed because it can silently break the warn-mode contract too.

**What I checked**

The new `verdict not in ("block", "allow")` branch in `run()` (hooks/pii-scan-guard.py:567-577) does correctly catch every malformed *verdict value* I tried when `result` is a dict: missing key (`{}`), `None`, `""`, wrong case (`"Block"`), and a list/dict used as the verdict value (e.g. `{"verdict": ["block"], ...}`) — all fall through to `_scanner_unavailable_result`, fail closed in block mode, warn in warn mode. Good.

**The gap:** `result.get("verdict")` (line 567) sits *outside* the `try/except` that wraps `call_scanner_fn`. If the scanner returns something that parses as valid JSON but isn't a dict at all — `call_scanner()`'s `json.loads(text)` will happily return `None` for text `"null"`, a `list` for `"[]"`, a bare string, etc., none of which raise — then `result.get("verdict")` throws an uncaught `AttributeError` that is not caught anywhere between there and `main()`. I reproduced this directly against this branch:

```
code, message = run(stdin_text, {"LOBSTER_PII_SCAN_MODE": "warn", ...}, repo_root,
                     call_scanner_fn=lambda p, k: None)
```
raises `AttributeError: 'NoneType' object has no attribute 'get'`, uncaught, all the way out of `run()`. I also ran this through a script mirroring `main()`'s exact body (read stdin → `run()` → print message → `sys.exit(code)`): the exception propagates past the `sys.exit(code)` entirely, Python prints a raw traceback, and the process exits **1**.

Two consequences:
- In **warn mode**, this is exit code 1 — which git's pre-push protocol treats as *reject the push*. That directly breaks the explicit, load-bearing invariant in this same module's docstring ("warn mode is unaffected... it should not start blocking pushes on its own transient errors"). It's not a deliberate fail-closed decision, it's an accidental block via an unhandled crash.
- In **block mode** it happens to still exit 1, but via a raw Python traceback dumped to stderr instead of the intended `[pii-scan-guard] BLOCKED: ...` message with remediation steps — so the fail-closed *behavior* holds by luck of CPython's default exit code, not by design, and the user gets a much worse diagnostic experience than the hook is supposed to provide.

Neither new test (`test_unexpected_verdict_value_fails_closed_in_block_mode`, `test_unexpected_verdict_value_in_warn_mode_does_not_block`) exercises this: both go through `_fake_scanner_returning(verdict, findings=None)`, which always wraps the verdict in a dict (`{"verdict": verdict, "findings": findings or []}`) — so `result.get(...)` never has a chance to fail. The fix covers "well-formed dict, bad verdict string" but not "response isn't a dict at all," which is squarely within the same "scanner failed to produce a usable answer" bucket the fix claims to close.

Fix is small: move the `result.get("verdict")` (and the schema check generally) inside the same try/except as the call, or explicitly validate `isinstance(result, dict)` before calling `.get()`.

**Other checks**
- Both stale-docstring fixes (`call_scanner()`'s docstring, `install-pii-scan-hook.sh`'s header) are accurate — no new inaccuracies introduced.
- Ran `tests/unit/test_hooks/test_pii_scan_guard.py` directly on this branch: 60 passed, 1 skipped (the opt-in live-API business-framing test, correctly skipped without `RUN_PII_SCAN_LIVE_TESTS=1`). I was not able to complete a full-repo suite run in the time available (it's large and includes slow integration/stress suites) to independently reconfirm the claimed 4211/16/32 split, but nothing in the diff touches code paths outside this hook, so I have no reason to doubt that count.

Round 2's fix closed the "unrecognized verdict as a bad string" hole. This round closes "unrecognized verdict as a bad string inside a well-formed dict." The remaining hole is "scanner response isn't a dict at all," which is a real, reachable failure mode (structured-output schema violations, a mock/bug returning the wrong shape, truncated/empty completions that still parse as valid JSON) and currently crashes instead of failing closed cleanly — and crashes into a *block* in warn mode, which is the one thing this hook is explicitly designed never to do.

aeschylus:
## Round 3 fix: validate scanner response shape before any dict access

Found on this round's review: `result.get("verdict")` in `run()` sat outside the `try/except` that wraps the scanner call. `json.loads` succeeds (no exception) for `null`, a list, a bare string, or a number — any of those is valid JSON but not a dict — so a well-formed non-dict scanner response raised an uncaught `AttributeError`. That crash propagated past the intended `sys.exit(code)`, so Python always exited 1 regardless of mode:

- **warn mode:** silently violated the hook's own documented invariant that warn mode never blocks (crash → exit 1 → git pre-push treats exit 1 as "block").
- **block mode:** still exited 1, but via a raw Python traceback instead of the intended clean `BLOCKED: ...` message.

### What changed

Per the task instructions, I did a full audit rather than patching just the reported line: `grep -n 'result\.' hooks/pii-scan-guard.py` confirmed there were exactly two places the raw scanner response was ever touched after `json.loads` — `result.get("verdict")` and `result.get("findings", [])` — both inside `run()`. Both are now replaced by a single `validate_scanner_response()` boundary that:

- Returns `None` for anything that isn't a dict at all (`None`, list, string, number).
- Returns `None` for a dict with a missing/unrecognized `"verdict"`.
- Returns `None` for a `"block"` verdict whose `"findings"` isn't a list, or whose findings contain non-dict entries.
- Otherwise returns `(verdict, findings)`.

Any `None` result is routed through the *existing* `_scanner_unavailable_result()` path (fails closed in block mode, warns in warn mode) — the same path already used for a raised exception or a missing API key. No new failure-handling branch was introduced; this consolidates response-shape checking into the one place that already had the correct fail-closed/warn-mode semantics.

### Tests added

- `TestValidateScannerResponse` (unit-level): direct coverage of the validator for `None`, list, bare string, number, empty dict, non-list findings, non-dict list entry, plus the two valid shapes.
- `TestRunHandlesMalformedScannerResponse` (integration-level, parametrized over the same seven malformed shapes): for each shape, confirms **(a)** block mode blocks the push with a clean message and no crash, and **(b)** warn mode does *not* block (exits 0) — the specific invariant this round found broken.

Falsifiability check performed: stashed the implementation change (keeping only the tests), confirmed the new tests fail (module fails to even collect — `validate_scanner_response` doesn't exist), then restored the implementation and confirmed all tests pass again.

### Test results

`uv run pytest tests/unit/test_hooks/test_pii_scan_guard.py -v` → 83 passed, 1 skipped (the pre-existing opt-in live-API test, unaffected by this change).

Also ran the full `tests/unit/test_hooks/` directory (992 tests) for a broader sanity check: 8 failures, all in `test_context_monitor.py` and `test_on_compact_write_claude_session_id.py` — files untouched by this diff (confirmed via `git diff --stat`), unrelated to the scanner-response validation logic, and present regardless of this change.

aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK: found a fourth gap — a silent bypass, not a crash this time.

**What I checked**
- Read `hooks/pii-scan-guard.py` in full at 629fb5e0, tracing every path from `call_scanner`'s raw return through `validate_scanner_response` to the final exit code in `run()`.
- Actually ran the test file: `83 passed, 1 skipped` (84 collected), matching the "no crash on malformed shape" claim for the 7 shapes the engineer tested (None/list/bare-string/number/empty-dict/bad-findings-type/bad-findings-entry) and confirming warn mode never blocks for any of them. Note: the PR description says "82 run + 1 skipped = 83/83" — actual is 83 run+passed + 1 skipped = 84 total. Off by one in the self-report; not a functional issue, just flagging that the count wasn't itself double-checked.
- Independently constructed several shapes not in the 7 test cases: boolean verdict (`True`/`False`) → correctly rejected. Extra unexpected keys alongside a valid verdict → correctly accepted and ignored. Extra nested key inside a finding entry → correctly accepted, no crash. Unicode in a snippet → correctly accepted, no crash.

**The gap**

`validate_scanner_response` treats `{"verdict": "block", "findings": []}` — and also `{"verdict": "block"}` with the `findings` key missing entirely (`.get("findings", [])` defaults to `[]`) — as **valid**, because an empty list passes `isinstance(findings, list)` and `all(isinstance(f, dict) for f in findings)` vacuously. It returns `("block", [])`.

Back in `run()`, `all_findings.extend([])` leaves `all_findings` empty, so the final check `if not all_findings: return 0, ""` fires — **in block mode, with a scanner verdict of "block", the push is allowed through with exit code 0 and zero output.** I confirmed this by direct execution (not just reading code):

```
BLOCK MODE, verdict=block findings=[] -> exit code: 0 message: ''
```

This is reachable from a fully schema-compliant model response, not just a contrived malformed shape — the `_RESPONSE_SCHEMA` requires the `findings` key but has no `minItems` constraint on it, and there's no `if verdict==block then findings must be non-empty` conditional in the schema. The system prompt tells the model block verdicts should carry "at least one confident finding," but nothing in code enforces that; it's exactly the same category of "trust the model's output shape" gap that rounds 1–3 were about, just one level deeper — this time in the validator itself rather than upstream of it.

Rounds 1–3 were all "scanner failure/crash gets mishandled." This one is worse in kind: it's a silent security bypass with no traceback, no stderr message, no exit-1 — the exact failure mode this hook exists to prevent, for the one verdict value ("block") that should never result in exit 0 with nothing printed. Suggest requiring `findings` to be non-empty when `verdict == "block"` inside `validate_scanner_response` (treat empty findings on a block verdict as invalid, same as any other shape violation), or explicitly special-case it in `run()`.

aeschylus:
## Round-4 finding: fixed

The round-4 review found that `validate_scanner_response()` (added in the round-3 fix) accepts `{"verdict": "block", "findings": []}` — or a `"block"` verdict with no `"findings"` key at all — as a **valid, well-formed response**. There is no constraint requiring `findings` to be non-empty when `verdict` is `"block"`.

`run()` then decided the final push outcome from `if not all_findings` at the end of the scan loop. That meant a `"block"` verdict with empty or missing findings fell through and returned exit code 0 (allow) with an **empty message** in block mode — the exact opposite of what a `"block"` verdict is supposed to mean, and the operator got zero warning that the push had been flagged at all.

### Fix

`run()` now tracks the block decision with a dedicated `any_block_verdict` flag, set whenever any scanned ref returns `verdict == "block"`, independent of whether `findings` ends up non-empty. The final branch now checks this flag rather than `all_findings`:
- If `findings` is populated, the existing detailed message is used (no change).
- If `findings` is empty/missing, a generic block message is shown (scanner flagged this push for review but did not provide detailed findings) — never a silent exit 0.

This does not touch the round-3 fix's validation boundary (`validate_scanner_response` itself is unchanged) — the bug was entirely in how `run()` used the validated result, not in the validation itself.

### Tests

Added `TestRunBlockVerdictWithoutFindings` in `tests/unit/test_hooks/test_pii_scan_guard.py`, covering:
- `verdict="block"` with `findings=[]` → block mode blocks with a non-empty message
- `verdict="block"` with `findings` key entirely absent → same
- `verdict="block"` with empty findings in warn mode → still warns (does not block, message non-empty)
- `verdict="block"` with populated findings → confirms no regression on the existing block-mode behavior

Falsifiability check performed: reverted the `run()` fix, confirmed all 3 new block/warn-empty-findings tests failed (the populated-findings regression test still passed, as expected), restored the fix, confirmed all pass again.

Full suite: `uv run pytest tests/unit/test_hooks/test_pii_scan_guard.py` → **87 passed, 1 skipped** (the skipped test is the pre-existing opt-in live-API business-framing test, gated behind `RUN_PII_SCAN_LIVE_TESTS=1`).

Also did a full end-to-end re-read of `validate_scanner_response()` and the entire block-verdict path in `run()` (not just the reported case) to check for a fifth variant of "verdict says one thing, hook does another" — traced mode gating, bypass, no-op refs, missing API key, scanner exceptions, invalid response shapes, the multi-ref "allow after block" case, and the final message/exit-code branch for both warn and block modes. Found no further gap.

aeschylus (commented):
🤖🦞 Lobster (reviewer): PASS: Round-4 fix holds up under an adversarial fifth pass.

**What I did, not just read:**
- Cloned the PR branch and ran the test file myself: `87 passed, 1 skipped` confirmed independently (not taking the engineer's word for it).
- Traced the full path: scanner response → `validate_scanner_response` → `any_block_verdict` tracking → message selection → exit code.
- Wrote and executed two scenarios not present in the existing test suite:
  1. **Mixed multi-ref push, one ref allow + one ref block (both orderings).** Confirmed `run()` correctly returns exit 1 in block mode regardless of which ref (first or second) reported "block" — `any_block_verdict` is a loop-wide flag, not per-ref, so it isn't lost or overwritten by a later "allow" verdict on a different ref.
  2. **Mixed multi-ref push, one ref allow + one ref block-with-empty-findings** (combining round-4's specific case with the multi-ref scenario the existing round-4 tests don't cover — those only exercise a single ref). Confirmed block mode still exits 1 with the generic fallback message, and warn mode still exits 0 with a WARNING-prefixed version of that same message.

**Findings on (a)/(b)/(c) specifically:**
- (a) Mixed verdicts across refs: handled correctly — verified by execution, not just by reading the loop structure.
- (b) Warn mode invariant: still holds for the empty-findings-block case in both single-ref (existing test) and my mixed-ref construction — warn mode never returns exit 1, in any combination I could produce.
- (c) No remaining `if findings:`/truthiness-as-block-decision proxy: the only remaining truthiness check on `all_findings` (line ~646) governs message *formatting* only (which message to show), not the block/allow decision, which is fully decided by `any_block_verdict` earlier. Confirmed by reading and by the empty-findings test cases above.

I tried specifically to find a case where an early-return path (scanner exception or invalid-shape response on a *later* ref, after an *earlier* ref already returned a real "block" verdict) could downgrade the outcome to allow. It can't: `_scanner_unavailable_result` in block mode always returns exit 1 regardless of what came before, so a later scanner failure can only ever be at least as strict, never a downgrade. (It does overwrite the *message* with a generic scanner-failure notice rather than showing the earlier ref's specific findings — a minor UX/detail-loss nit, not a block→allow correctness bug, so I'm not calling it a gap.)

No fifth gap found after genuinely trying to construct one. This is ready to be called done.

aeschylus:
Merged as part of Fable's PII-scanner-system remediation master plan (Linear BIS-755: https://linear.app/fully-parsed/issue/BIS-755). Code-complete, independently PASSED after 5 review rounds, ships fully inert (LOBSTER_PII_SCAN_MODE unset everywhere, not wired into install.sh) — merging changes zero live behavior. The separate install/enable decision is tracked at BIS-756 and remains blocked on Drew (reviewer recommends warn-mode burn-in before block).

PR #2164, opened 2026-08-05
Warn on known-leaked personal emails as commit author

### LINKED ISSUE #2162: Pre-commit guard: warn on known-leaked personal emails as commit author
## Motivation

Two personal-email leaks exist in this repo's history: Drew's `scipioaffricanus@gmail.com` (479 commits on `main` / 1005 all-branches, growing daily) and Sahar Massachi's raw `sayhar@gmail.com` (10 commits, all 2026-05-09 — looks like a one-day local git-config lapse, since his other ~3,700 commits correctly use his GitHub noreply address). Both retroactive fixes require a history rewrite, which is out of scope here (destructive, needs separate explicit sign-off). But nothing currently stops either leak from *growing* on new commits.

## Scope (purely forward-looking — no history rewrite, no existing commit touched)

- Add a check to `.githooks/pre-commit` (or a new hook file if that's structurally cleaner — match existing conventions) that inspects the commit's author/committer email (`git config user.email`, and `--author` if used) against a small configurable blocklist of known-leaked personal addresses.
- Load the blocklist from a local, gitignored config file (e.g. `~/lobster-user-config/blocked-commit-emails.txt`), falling back to a no-op if the file doesn't exist — this matches the existing pattern already used for `blocked-names.txt` / `instance-domains.txt` in `.githooks/pre-push`. Do not hardcode the two real email addresses into any file that gets committed to the repo — that would just create a third leak of the same addresses.
- On a match: **warn loudly and require explicit confirmation to proceed** (same UX pattern as the existing PII confirmation prompt in `.githooks/pre-push`) rather than a hard block — this is a habit-breaking nudge, not a hard gate, since there may be legitimate reasons to override.
- Must not touch any existing commit and must not rewrite anything. Purely forward-looking.

## Proof required (do not skip either half)

- Automated tests covering: no blocklist file present → no-op; blocklist file present with a matching email → warns and requires confirmation; blocklist file present with a non-matching email → silent pass.
- Manual test: real throwaway repo, configure `user.email` to a test address you put in a local blocklist file, attempt a commit, observe the warning; configure a clean address, confirm no warning. Show actual commands/output.

## Tracking

Part of Fable's PII-scanner-system remediation plan. Linear: BIS-759 (sub-issue of master BIS-755).

aeschylus:
## Problem

Two personal Gmail addresses have leaked into this repo's git history as commit author/committer emails: one used on 479 `main` commits (1005 across all branches, still growing daily), the other on 10 commits from a single day (a one-off local git-config lapse by a different contributor). Rewriting existing history to fix this is a separate, explicit, destructive decision that's out of scope here. But until now, nothing stopped either address — or any future leak — from being used on a **new** commit. This closes that forward-looking gap. No existing commit is touched, rewritten, or amended by this change.

## What changed

`.githooks/pre-commit` gained a check that inspects every email address that could end up recorded as this commit's author/committer:
- `git config user.email` (default author and committer)
- `GIT_AUTHOR_EMAIL` (set when `--author=...` is used)
- `GIT_COMMITTER_EMAIL` (set via `-c user.email=...` overrides)

These are compared case-insensitively against an optional blocklist at `${LOBSTER_USER_CONFIG_DIR:-~/lobster-user-config}/blocked-commit-emails.txt` — gitignored, private, one address per line, `#`-comments and blank lines ignored. This is the exact same load-a-private-config-with-silent-no-op-fallback convention already used by `blocked-names.txt` and `instance-domains.txt` in `.githooks/pre-push`. If the file doesn't exist, the check is a complete no-op — no behavior change for anyone who hasn't opted in. **No real leaked address is hardcoded anywhere in the repo** — only `@example.com`/`@example.org` placeholders in the test fixtures below.

On a match, the hook prints a loud warning naming the matched address and requires an explicit `y`/`yes` response to proceed (default: abort) — this mirrors the existing PII confirmation prompt UX in `.githooks/pre-push` exactly. It is a warn-and-confirm nudge, not a hard block, since there can be legitimate reasons to commit under one of these addresses.

### A wrinkle worth flagging for review

`.githooks/pre-push` decides interactive-vs-CI by checking `[ -t 0 ]` (is stdin a tty). I initially mirrored that for pre-commit, and my own tests caught that it doesn't work: **git unconditionally connects a `pre-commit` hook's own stdin to `/dev/null`**, regardless of whether the invoking shell is interactive (verified empirically — see `readlink /proc/self/fd/0` output in the manual test below, and the git-internals rationale in the code comment). That means `[ -t 0 ]` inside a pre-commit hook is *always* false, so a naive port of pre-push's check would silently disable the confirmation prompt in all real usage — it would always fall through to "no controlling terminal, don't block." Note that pre-push's own `[ -t 0 ]` check has the same problem for a different reason (its stdin carries the ref-update list, never a tty), which I confirmed but did **not** touch — that's pre-existing pre-push behavior and out of scope for this issue.

Fix: this hook instead probes for a controlling terminal directly (`exec 3</dev/tty 2>/dev/null`), which correctly returns true under a real interactive shell (or a real pty, as used in the automated pty test below) and false when there's no controlling terminal at all (verified with `setsid`).

## Functional patterns

The check is a single pure-ish function (`check_commit_author_email`) that reads inputs (config file, git config, env vars), computes a match, and only then performs the one necessary side effect (the confirm prompt) — no staged-file mutation, no hidden state. Blocklist parsing and the candidate-email/blocklist comparison are simple map/filter-style loops over arrays with no external dependencies beyond `git config`.

## Out of scope

- No history rewrite of any kind — the two known leaks in existing commits are untouched, per the issue.
- No changes to `.githooks/pre-push` (its `[ -t 0 ]` behavior is pre-existing and unrelated — flagged above for visibility only).
- No install.sh/upgrade.sh scaffolding added, since `blocked-names.txt` / `instance-domains.txt` have none either — these are purely optional, user-authored files with no install-time templating.

## Tests run

- [x] `bash tests/unit/test_pre_commit_email_check.sh` — 9/9 assertions passed. Builds real throwaway git repos, wires `core.hooksPath` to the actual (not reimplemented) `.githooks/pre-commit`, and runs real `git commit` invocations:
  - no blocklist file → silent no-op
  - blocklist + matching `user.email`, non-interactive → warns, does not block (CI policy)
  - blocklist + matching email, **real pty via `expect`** → confirmation prompt actually gates the commit (`n` aborts and no commit is created; `y` proceeds and the commit exists in history)
  - blocklist + matching `--author=...` (GIT_AUTHOR_EMAIL path) → warns
  - blocklist + non-matching email → silent pass
  - comments/blank lines in the blocklist file are ignored
- [x] Falsifiability check: reverted `.githooks/pre-commit` to the pre-change version (`git stash`), re-ran the test file — 6 of 9 assertions failed as expected (the 3 that still passed test only the pre-existing no-op/non-match paths, which didn't change). Restored the implementation and confirmed all 9 pass again.
- [x] `bash tests/unit/test_pre_push_scan_diff_timeout.sh` — 8/8 still passing (no regression to the neighboring hook)
- [x] `bash -n .githooks/pre-commit` — syntax check clean
- [ ] `shellcheck` — not installed in this environment, skipped

## Manual test

Real throwaway repo (`/tmp/manual-test-pre-commit-*`, since deleted), `.githooks/pre-commit` copied in verbatim from this branch, `core.hooksPath` set to `.githooks`, `LOBSTER_USER_CONFIG_DIR` pointed at an isolated temp config dir (never touched the real `~/lobster-user-config` on this machine).

**Step 1 — blocked email, answer `n` (abort), via a real pty (`expect`):**
```
$ echo "leaked-personal-example@example.com" > $CFG/blocked-commit-emails.txt
$ git config user.email "leaked-personal-example@example.com"
$ git add file1.txt
$ git commit -m "commit under blocked email"
WARNING: commit author/committer email is on the leaked-email blocklist:
         leaked-personal-example@example.com
         This address has previously leaked into this repo's history.
         (configure via ~/lobster-user-config/blocked-commit-emails.txt)

Continue committing with this email? [y/N]: n

Commit aborted. Fix your git author config and retry, e.g.:
  git config user.email "you@example.com"

Emergency bypass: git commit --no-verify
$ git log --oneline
fatal: your current branch 'master' does not have any commits yet
```

**Step 2 — same blocked email, answer `y` (confirm override):**
```
$ git commit -m "commit under blocked email, confirmed"
WARNING: commit author/committer email is on the leaked-email blocklist:
         leaked-personal-example@example.com
         ...
Continue committing with this email? [y/N]: y

Confirmed. Proceeding with commit.

[master (root-commit) 2312a75] commit under blocked email, confirmed
 1 file changed, 1 insertion(+)
```

**Step 3 — clean, non-blocklisted email — no warning at all:**
```
$ git config user.email "clean-dev-example@example.com"
$ git commit -m "commit under clean email"
[master 1f64001] commit under clean email
 1 file changed, 1 insertion(+)
```
(no blocklist-related output — commit went straight through)

**Step 4 — no blocklist file configured at all (even with `user.email` set to the same address as step 1) — silent no-op:**
```
$ git commit -m "commit with no blocklist file configured"
[master e6dfdcb] commit with no blocklist file configured
 1 file changed, 1 insertion(+)
```

All four observed outcomes match the spec exactly. Repo and config dirs deleted afterward — nothing left on disk.

## Definition of Done

- [x] Unit tests pass with proper mocking-equivalent isolation (real repos, isolated `LOBSTER_USER_CONFIG_DIR`, never touches the real `~/lobster-user-config`)
- [x] Manual test run — see `## Manual test` above
- [x] User-visible outcome verified: the actual terminal output shown above (warning text, prompt, abort/confirm messages) is exactly what a developer would see running `git commit` locally
- [x] Side-effect audit: no network calls, no shared state, no floods — the hook only reads a local file and git config, and prompts/writes to the invoking terminal
- [x] External service calls: none (this hook is entirely local)

Closes #2162

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -13,6 +13,11 @@
 #   - Per-name allowlist regex configured alongside each name in blocked-names.txt
 #   - Comments with # noname on the same line — allowed
 #
+# Also warns (with a confirm-to-proceed prompt) when the commit's author or
+# committer email matches a blocklist of known-leaked personal addresses, as
+# configured in ~/lobster-user-config/blocked-commit-emails.txt (gitignored,
+# private). See the "Commit-Author Email Check" section below for details.
+#
 # Bypass: git commit --no-verify
 #===============================================================================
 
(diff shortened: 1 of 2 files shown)


aeschylus:
## Independent review — PR #2164

**Verdict: PASS**

Reviewed the diff cold, then independently reproduced rather than trusting the PR description:

**Ran the new test suite myself**: `bash tests/unit/test_pre_commit_email_check.sh` on the PR branch → 9/9 passed, including the real-pty `expect`-driven interactive test (both the "n" abort path and the "y" proceed path actually drove a real terminal and checked `git log` afterward to confirm the commit truly was/wasn't created — not just checking exit codes).

**Reproduced the falsifiability check myself**: swapped in `main`'s old `.githooks/pre-commit`, reran the same test file → 3 passed, 6 failed (every test that depends on the new check, as expected) — restored the fix, back to 9/9.

**Checked this is genuinely a no-op today**: confirmed `~/lobster-user-config/blocked-commit-emails.txt` does not exist on this instance, so merging this changes nothing about anyone's workflow until someone opts in by creating that file.

**Verified the /dev/tty finding independently**: confirmed `[ -t 0 ]` is unusable in a pre-commit hook (git always connects its stdin to `/dev/null`) and that probing `/dev/tty` directly (as this PR does) is the correct fix — this is a different, real bug from the one Slice A's engineer found in `.githooks/pre-push` (there, stdin carries real ref data instead of being null, but `IS_TTY` still evaluates false for every real `git push`). Two independently-arrived-at findings about the same class of problem in sibling hooks — filing a follow-up issue referencing both rather than letting them sit as duplicate observations.

**Scope check**: no existing commit touched, no real leaked email hardcoded anywhere in the diff (fixtures use `@example.com`/`@example.org` placeholders only), blocklist loaded from a gitignored external file consistent with the existing `blocked-names.txt` convention already used elsewhere in `.githooks/`.

No correctness issues found. Merging.

Part of Fable's PII-scanner-system remediation plan (Linear BIS-755 / BIS-759).

Issue #36, opened 2026-02-12
Layer 2: Canonical Memory Files + Nightly Consolidation


aeschylus:
## Overview

Implement the nightly consolidation script that synthesizes raw memory events (from the existing SQLite + FTS5 + sqlite-vec store) into canonical markdown files. This is the "Layer 2" from the three-layer memory system design (#25) -- the process that turns accumulated events into structured, human-readable knowledge.

**Parent epic:** #30

### Dependencies

None. The SQLite event store and memory tools (`memory_store`, `memory_search`, `memory_recent`, `get_handoff`, `mark_consolidated`) already exist in `src/mcp/memory/`. The canonical directory structure is already seeded at `memory/canonical/`. This layer writes the consolidation logic that connects them.

### What Already Exists

- `src/mcp/memory/vector_memory.py` -- SQLite + FTS5 + sqlite-vec event store with `memory_store`, `memory_search`, `memory_recent`, `mark_consolidated`
- `src/mcp/memory/static_memory.py` -- Static memory provider
- `src/mcp/memory/provider.py` -- Memory provider abstraction
- `memory/canonical/handoff.md` -- Seed handoff document (template)
- `memory/canonical/priorities.md` -- Seed priorities file
- `memory/canonical/daily-digest.md` -- Seed daily digest
- `memory/canonical/projects/` -- Project files directory
- `memory/canonical/people/` -- People files directory

---

## Acceptance Criteria

- [ ] **Nightly consolidation script** (`scripts/nightly-consolidation.py`) that:
  1. Creates a consolidation run record in the `consolidation_runs` table
  2. Queries all events where `consolidated = FALSE`, ordered by timestamp
  3. Groups events by project, person, and topic
  4. For each group, calls Claude API (Sonnet) with the current canonical file + new events to produce an updated version
  5. Writes updated canonical markdown files to `memory/canonical/`
  6. Generates a `daily-digest.md` summarizing the past 24 hours
  7. Regenerates `handoff.md` -- the comprehensive briefing document
  8. Marks all processed events as `consolidated = TRUE` with the run ID
  9. Updates the consolidation run record with status and summary
  10. On error: marks run as failed, logs traceback; unconsolidated events are picked up by next run
- [ ] **Consolidation prompts** for each synthesis task:
  - Per-project/topic update prompt
  - Daily digest prompt
  - Handoff document prompt (the "crown jewel")
- [ ] **Canonical file templates** enhanced with clear section structure:
[...]

PR #2169, opened 2026-08-05
Log pii-scan-guard fail-open events to a durable JSONL trace

aeschylus:
## Problem

`hooks/pii-scan-guard.py` (PR #2156) fails open on four conditions -- no API key, a network error, a timeout, a malformed scanner response -- and until now the only trace of that was a single stderr line at push time. There was no durable way to answer "has this scanner been silently doing nothing for a week?" This closes that visibility gap. Stretch/lower-priority slice of the PII-scanner remediation plan (Linear BIS-760); the hook itself is not installed anywhere yet, so this is instrumentation only -- no alerting, no cron job.

## What changed

Every fail-open path already funnels through one function, `_scanner_unavailable_result`. It now appends one JSONL record to `$LOBSTER_WORKSPACE/logs/pii-scan-guard-failopen.jsonl` before returning, in both `warn` and `block` mode:

```json
{"ts": "...", "hook": "pii-scan-guard", "category": "missing_api_key", "reason": "...", "mode": "block"}
```

`category` is one of `missing_api_key` / `network_error` / `timeout` / `malformed_response` -- the same four reasons already named in `_scanner_unavailable_result`'s own docstring, not a new taxonomy. A new pure function, `_classify_scanner_exception`, maps any exception raised by the scanner call onto `timeout` vs `network_error` by inspecting the exception's type name/message (it deliberately avoids importing the `anthropic` SDK's specific error classes, since `call_scanner` imports `anthropic` lazily so pure-function tests never need it installed).

The logging call itself (`log_failopen_event`) reuses this repo's existing hook-logging convention -- the same append-only `~/lobster-workspace/logs/*.jsonl` pattern `hooks/usage-accumulator.py` already uses -- rather than inventing a new mechanism. Path resolution mirrors `dispatcher-state-stop.py`'s `_resolve_inflight_path` (`$LOBSTER_WORKSPACE`-based, with a `Path.home()` fallback).

Logging is wired in as dependency injection, the same pattern `call_scanner_fn` already uses: `run()` gained an optional `failopen_log_path` parameter so tests (and, if ever needed, callers) can redirect it; `main()` leaves it `None` and the real default path is used. Writing the record is wrapped in `try/except OSError: pass` -- a broken log write (bad permissions, path collision) can never change the hook's exit code or block message, since this is instrumentation bolted onto the decision path, not part of it.

## Functional patterns

- `log_failopen_event` and `_classify_scanner_exception` are pure/near-pure: given the same inputs they produce the same record shape / same category, with the one necessary side effect (the file write) isolated to a single `try/except` boundary at the very edge
- Dependency injection (`failopen_log_path`) over a global/singleton logger, consistent with how `call_scanner_fn` is already injected in this file
- No new state, no new taxonomy, no new mechanism — this reuses what the function already decides and what the repo already does for hook-level JSONL logging

## Out of scope

No cron job, alerting pipeline, or health check was added -- deliberately, per the issue. No change to the hook's actual block/allow/fail-closed decision logic; `_scanner_unavailable_result`'s return values (exit code, stderr message) are byte-for-byte the same as before for every input, verified by the full pre-existing test suite passing unchanged. No change to `install.sh` /  the rollout gate -- the hook remains uninstalled anywhere, same as before this PR.

## Tests run

- [x] `uv run pytest tests/unit/test_hooks/test_pii_scan_guard.py tests/unit/test_install_pii_scan_hook.py -q` — 111 passed, 1 skipped (the pre-existing opt-in live-model test, unrelated to this change), 0 failed
- [x] `uv run pytest tests/unit/ -q` — same 8 pre-existing failures as on `origin/main` before this change (`test_context_monitor.py`, `test_on_compact_write_claude_session_id.py` — confirmed via `git stash` + rerun on unmodified `origin/main`; unrelated to this diff, environment-dependent in this sandbox), 1011+ passed otherwise
- [x] Falsifiability check: commented out the `log_failopen_event(...)` call inside `_scanner_unavailable_result`, reran the new tests — 6 of them failed as expected (`TestRunLogsEachFailOpenReason` x5, `test_default_log_path_resolves_under_lobster_workspace`), confirming they are not vacuous. Restored the call, reran, all 104 tests in the file passed again.

New test coverage added (`tests/unit/test_hooks/test_pii_scan_guard.py`):
- `TestClassifyScannerException` — timeout vs. network-error classification from exception type/message
- `TestLogFailopenEvent` — JSONL record shape, multi-append behavior, parent-dir creation, and the never-raises guarantee for an unwritable path
- `TestRunLogsEachFailOpenReason` — one test per fail-open reason (missing key, network error, timeout, malformed response) asserting the correct category lands in the log via `run()`; plus warn-mode-still-logs, real-verdict-writes-nothing, and default-path-resolution-under-`$LOBSTER_WORKSPACE`
- `TestFailopenLoggingNeverBreaksOutcome` — an unwritable log path (a directory passed as the log file) does not change the block outcome or message

## Manual test

Ran the actual `hooks/pii-scan-guard.py` CLI (not a mock) against a scratch git repo, with a scratch `$LOBSTER_WORKSPACE`, isolating `$HOME` where needed so the real production config/API key could not be picked up as a fallback:

1. **Real network/auth fail-open** (warn mode): piped a real pre-push stdin ref line into the hook with `ANTHROPIC_API_KEY=sk-ant-deliberately-invalid-key-for-manual-test`. This made a real HTTPS call to the Anthropic API, which returned a real `401 authentication_error`. Hook printed a WARNING and exited 0 (as expected in warn mode), and appended:
   ```json
   {"ts": "2026-08-05T19:53:46.562511+00:00", "hook": "pii-scan-guard", "category": "network_error", "reason": "scanner call failed (Error code: 401 - {'type': 'error', 'error': {'type': 'authentication_error', 'message': 'invalid x-api-key'}, 'request_id': 'req_011CdkBopmueoEMkPTEJuspb'})", "mode": "warn"}
   ```
2. **Same real 401, block mode**: hook printed BLOCKED and exited 1 (fail-closed, as designed), and appended a second real record with `"mode": "block"`.
3. **True missing key** (isolated `$HOME` so no config file could be found, empty `ANTHROPIC_API_KEY`), block mode: hook printed BLOCKED / exited 1, and appended:
   ```json
   {"ts": "2026-08-05T19:55:16.961900+00:00", "hook": "pii-scan-guard", "category": "missing_api_key", "reason": "no ANTHROPIC_API_KEY configured — cannot run PII scan", "mode": "block"}
   ```

All three records were read back directly off disk from the scratch `$LOBSTER_WORKSPACE/logs/pii-scan-guard-failopen.jsonl` path with `cat`/`json.tool` -- this is the actual content the hook wrote, not a mock or a description of what it should write.

**Side-effect note:** an earlier attempt at reproducing "missing key" (before I isolated `$HOME`) accidentally fell back to this box's real production `ANTHROPIC_API_KEY` via `~/lobster-config/config.env` (the hook's own documented config-file fallback, working as designed) and made one real scan call against a harmless scratch diff (no real PII, a fake `jane.doe@example-corp-real.com` string in a throwaway temp repo). No cost/volume concern beyond that single call; flagging for transparency. All manual-test artifacts (`/tmp/pii-manual-test-*`) were deleted after capturing the log contents above; nothing was written to the real `~/lobster-workspace/logs/`.

**User-visible outcome:** N/A -- this hook produces no Telegram/user-facing output; its only observable surface is the git push stderr line (unchanged behavior) and the new log file, both verified directly above.

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits in the automated `pytest` suite — the fake-scanner injection pattern this file already uses)
- [x] Manual test run against the real hook with a real (intentionally invalid) API key — see Manual test above
- [x] User-visible outcome: not applicable (no user-facing output from this hook)
- [x] Side-effect audit: no floods, no unscoped writes (isolated `$LOBSTER_WORKSPACE` and `$HOME` used for all manual runs), no unintended volume beyond the one flagged incidental real-key call above
- [x] External service calls: injected/faked in all automated tests; the two real calls in manual testing were both deliberate 401s to a scratch environment, plus the one flagged incidental real-key call

### DIFF
--- a/hooks/pii-scan-guard.py
+++ b/hooks/pii-scan-guard.py
@@ -65,6 +65,18 @@ transient errors before that. `LOBSTER_PII_SCAN_SKIP=1` remains the
 deliberate, explicit escape hatch for a push the operator has already
 decided is safe.
 
+**Fail-open observability (issue #2167):** every one of the four fail-open
+reasons above (missing key, network error, timeout, malformed response) also
+appends a durable JSONL record to
+`$LOBSTER_WORKSPACE/logs/pii-scan-guard-failopen.jsonl` (see
+log_failopen_event / _scanner_unavailable_result), in both `warn` and `block`
+mode. Before this, the only trace of a scanner outage was a stderr line on
+the developer's terminal at push time -- nothing durable existed to answer
+"has this scanner been a no-op for a week?" This is instrumentation only: it
+does not add alerting or a cron job (the hook is not installed anywhere yet
+as of this writing), it just makes fail-open events visible somewhere that
+could be checked or wired into alerting later.
+
 **Model:** the scanner runs on Claude Fable 5 (`claude-fable-5`), not Opus.
 This hook previously ran on Opus, which was shown (see the PR/issue history)
 to be foolable by realistic-looking PII wrapped in business, consulting,
@@ -103,6 +115,7 @@ import os
 import re
 import subprocess
 import sys
+from datetime import datetime, timezone
 from pathlib import Path
 
 # ---------------------------------------------------------------------------
@@ -127,6 +140,24 @@ _ZERO_SHA = "0" * 40
 _EMPTY_TREE_SHA = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
 _ALLOWLIST_FILENAME = ".security-allowlist"
+
+# ---------------------------------------------------------------------------
+# Fail-open observability logging (issue #2167)
+# ---------------------------------------------------------------------------
+# Every fail-open path (see _scanner_unavailable_result below) previously
+# degraded silently -- a stderr line most workflows never durably surface.
+# These four categories are exactly the four reasons already named in
+# _scanner_unavailable_result's own docstring; this does not invent a new
+# taxonomy, it just gives each of those reasons a stable machine-readable
+# label for the log. Reuses this repo's existing hook-logging convention
+# (see hooks/usage-accumulator.py's ~/lobster-workspace/logs/*.jsonl
+# append-only pattern) rather than a new mechanism.
+_CATEGORY_MISSING_API_KEY = "missing_api_key"
+_CATEGORY_NETWORK_ERROR = "network_error"
+_CATEGORY_TIMEOUT = "timeout"
+_CATEGORY_MALFORMED_RESPONSE = "malformed_response"
+
+_FAILOPEN_LOG_FILENAME = "pii-scan-guard-failopen.jsonl"
 
 # Extensions/basenames that can never carry meaningful PII content and are
 # never scanned. Deliberately narrow — see module docstring for why this does
@@ -508,7 +539,68 @@ def validate_scanner_response(result: object) -> tuple[str, list[dict]] | None:
     return "block", findings
 
 
-def _scanner_unavailable_result(mode: str, reason: str) -> tuple[int, str]:
+def _classify_scanner_exception(exc: Exception) -> str:
+    """Map an exception raised by `call_scanner` onto one of the two
+    exception-shaped fail-open categories (network error vs. timeout).
+
+    Deliberately matches on the exception's type name and message rather
+    than importing the `anthropic` SDK's specific error classes (e.g.
+    `anthropic.APITimeoutError`) -- `call_scanner` imports `anthropic`
+    lazily specifically so pure-function tests never need it installed, and
+    classification must not force that import at module load time either.
+    """
+    name = type(exc).__name__.lower()
+    text = str(exc).lower()
+    if "timeout" in name or "timeout" in text:
+        return _CATEGORY_TIMEOUT
+    return _CATEGORY_NETWORK_ERROR
+
+
+def _default_failopen_log_path(env: dict) -> Path:
+    """Resolve the durable fail-open log location from $LOBSTER_WORKSPACE,
+    mirroring dispatcher-state-stop.py's _resolve_inflight_path and
+    usage-accumulator.py's logs/ convention."""
+    workspace = Path(env.get("LOBSTER_WORKSPACE", str(Path.home() / "lobster-workspace")))
+    return workspace / "logs" / _FAILOPEN_LOG_FILENAME
+
+
+def log_failopen_event(
+    env: dict,
+    category: str,
+    reason: str,
+    mode: str,
+    log_path: Path | str | None = None,
+) -> None:
+    """Append one durable, greppable JSONL record of a fail-open event.
+
+    This is instrumentation, not part of the decision path: any failure
+    writing the record (unwritable directory, log_path pointing at a
+    directory, permissions) is swallowed rather than raised or surfaced,
+    so a logging problem can never change the hook's exit code or message.
+    """
+    target = Path(log_path) if log_path is not None else _default_failopen_log_path(env)
+    record = {
+        "ts": datetime.now(timezone.utc).isoformat(),
+        "hook": "pii-scan-guard",
+        "category": category,
+        "reason": reason,
+        "mode": mode,
+    }
+    try:
+        target.parent.mkdir(parents=True, exist_ok=True)
+        with target.open("a", encoding="utf-8") as f:
+            f.write(json.dumps(record) + "\n")
+    except OSError:
+        pass
+
+
+def _scanner_unavailable_result(
+    mode: str,
+    reason: str,
+    category: str,
+    env: dict,
+    log_path: Path | str | None = None,
+) -> tuple[int, str]:
     """Decide the outcome when the scanner itself could not produce a
     verdict at all -- no API key, a network error, a timeout, a malformed
     response. Anything that lands here means we cannot claim the push is
@@ -526,7 +618,13 @@ def _scanner_unavailable_result(mode: str, reason: str) -> tuple[int, str]:
     enforcement, so a scanner outage there should still just warn -- it
     would be surprising and wrong for a hook nobody has enabled blocking on
     yet to start blocking pushes because of its own transient errors.
+
+    Always logs the event (see log_failopen_event, issue #2167) regardless
+    of mode -- a scanner outage is exactly as invisible in warn mode as in
+    block mode, since warn mode's non-blocking behavior only affects the
+    push outcome, not whether the outage itself deserves a durable record.
     """
+    log_failopen_event(env=env, category=category, reason=reason, mode=mode, log_path=log_path)
     remediation = (
         f"Investigate the scanner error, or set {_ENV_BYPASS}=1 to "
         "deliberately bypass this one push if you've reviewed it yourself."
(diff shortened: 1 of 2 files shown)


aeschylus:
## Independent review — PR #2169

**Verdict: NEEDS-WORK — real CodeQL security alert, not a rubber-stamp check**

### What checks out

- Ran the full test suite myself: 111 passed, 1 skipped — matches the claim exactly.
- Reproduced the falsifiability check myself: commented out the `log_failopen_event(...)` call, reran → 6 failed (`TestRunLogsEachFailOpenReason` x5 + the default-path test), restored → 104/104 pass in the file. Matches the PR's transcript.
- Manually exercised `_scanner_unavailable_result` and `_classify_scanner_exception` directly (not through the mocked test suite) — block-mode missing-key correctly exits 1, a timeout-shaped exception is correctly classified as `timeout` (not `network_error`), warn-mode correctly exits 0, both correctly appended to the JSONL log with the right shape.
- Confirmed all three call sites of `_scanner_unavailable_result` were updated for the new required `category`/`env` params (grepped for all call sites, none missed).
- On the classifier heuristic question you flagged: the name/message substring check is fine as-is — this only affects the log record's `category` field, not any actual block/allow decision, so the cost of an occasional network_error/timeout misclassification is low, and importing `anthropic`'s specific error types would reintroduce the eager-import problem this file's existing lazy-import pattern deliberately avoids. No change needed there.
- On the incidental real-API-key manual-test call: agreed with your own read — one-off, disclosed, no real PII, no lasting artifact. Not escalating to Drew as an incident, just noting it in my final report for the record.

### The blocking finding

CI's CodeQL check failed for real — 1 high-severity alert, not a flake (compare: every other PR in this plan had CodeQL pass in 1-4s; this one failed with an actual annotation):

```
hooks/pii-scan-guard.py:592:21-46 — "Clear-text storage of sensitive information"
  f.write(json.dumps(record) + "\n")
```

CodeQL's taint tracker is following `api_key` (used two lines earlier via `call_scanner_fn(prompt, api_key)`) through to the exception path — `reason = f"scanner call failed ({exc})"` (line 707) is the raw, unscrubbed `str(exc)` of whatever the scanner call raised, and that string flows straight into the durable JSONL log with no sanitization.

In the one real manual test performed, the actual Anthropic 401 response body didn't happen to echo the key back, so the specific transcript pasted in the PR body is safe. But nothing in the code *guarantees* that for every exception shape this can hit — a client-side SDK validation error, a lower-level HTTP client debug dump, or a future SDK version could plausibly include request header content (including the key) in `str(exc)`. Given this whole remediation plan exists to stop exactly this class of leak, shipping a new durable, greppable log file with an unscrubbed path from the API key to disk — flagged by real static analysis, not a hunch — needs to be closed before merge, not waved through because "it didn't happen to leak this one time."

### Requested fix

- Scrub/redact the `reason` string before it's ever handed to `log_failopen_event` — at minimum, strip anything that could be the literal API key value (you have `api_key` in scope at the call site), or more robustly, cap what's captured to a fixed-shape summary (exception type name + HTTP status if present) rather than the raw `str(exc)`.
- Add a test proving this: construct an exception whose message contains a fake-but-realistic key-shaped string, run it through the fail-open path, assert the resulting log record does NOT contain that string.
- Get the CodeQL check green (or, if you believe it's a genuine false positive after the fix, explain why in the PR rather than suppressing it silently).

Once that's addressed, this is good to merge — the design and everything else above checks out.

Part of Fable's PII-scanner-system remediation plan (Linear BIS-755 / BIS-760).

aeschylus:
Fixed in 250983ec.

CodeQL's finding was real, not a false positive: `_scanner_unavailable_result` persisted `str(exc)` verbatim into the new durable log for scanner-exception fail-opens, and `exc` (from the `anthropic` SDK or a lower HTTP layer) is not guaranteed to exclude request content, including the `api_key` used two lines earlier in `call_scanner_fn(prompt, api_key)`.

Fix, going with the more robust of the two options suggested:
- New `_summarize_scanner_exception_for_log(exc)` replaces the raw exception text with a fixed-shape `"<TypeName>"` or `"<TypeName> (status <NNN>)"` summary for the durable log -- it never touches `str(exc)` at all, so the log structurally cannot carry secret (or arbitrary) content from the exception, rather than trying to scrub an open-ended string after the fact.
- `_redact_secret(text, api_key)` is a second, independent backstop: applied to every reason that reaches the log wherever `api_key` is in scope (the exception path above, and the malformed-response path, which isn't exception-derived but costs nothing to cover too), stripping any literal occurrence of the real key.
- The operator-facing stderr message at push time is unchanged (full `str(exc)` detail, as before this PR) -- only the persisted/greppable log is restricted. Called that out explicitly since it's a deliberate scope boundary, not an oversight.

Test added per your ask: `test_exception_message_containing_api_key_is_not_logged_verbatim` constructs an exception whose message embeds a realistic-looking fake key, runs it through `run()`, and asserts the key is absent from the persisted JSONL record (while confirming it's still present in the returned stderr message, proving the split is intentional). Also added direct unit coverage for `_summarize_scanner_exception_for_log` and `_redact_secret`.

One thing the new tests caught before it shipped: my first pass at the status-code regex (`\b[1-5]\d{2}\b`) matched the "443" in "api.anthropic.com:443" as if it were an HTTP status. Tightened it to require a preceding "code"/"status" keyword (matches the anthropic SDK's actual "Error code: 401 - ..." format) so a port number is never misread as a status.

Falsifiability: reverted the redaction call, confirmed `test_exception_message_containing_api_key_is_not_logged_verbatim` failed (key leaked into the JSON record) as expected; restored, confirmed it and all 120 other tests pass again (1 pre-existing skip, the opt-in live-model test).

Re-ran the manual test end to end against the real hook with a real invalid key: the live 401 from Anthropic now logs as `{"reason": "AuthenticationError (status 401)", "category": "network_error", ...}` -- no trace of the key anywhere in the persisted file, confirmed with `grep -c "$FAKE_KEY" ... ` returning 0.

CI should be green on this commit; happy to dig further if CodeQL still flags something.

aeschylus:
## Independent re-review — PR #2169 fix round (commit 250983ec)

**Verdict: PASS**

Ran the full test suite myself: 120 passed, 1 skipped (121 total) — matches the claim.

Reproduced falsifiability myself with two separate experiments (not just re-reading the PR's transcript):
1. Broke `_redact_secret` into a no-op — `TestRedactSecret`'s 2 direct unit tests correctly failed, confirming that backstop has real, working coverage.
2. Constructed an exception with a realistic fake key embedded directly in its message (`f"connection failed; sent header x-api-key: {fake_key}"`) and ran it through `run()` end-to-end myself, independent of the test file: confirmed the persisted log record's JSON serialization never contains the key, while the operator-facing stderr message correctly still does (full detail preserved where it belongs, restricted only in the durable log — exactly the documented scope boundary).

**On the residual CodeQL failure** (the fix round's own report flagged this as still-pending, worth double-checking rather than trusting the description): it was still red after this commit — same rule (`py/clear-text-storage-sensitive-data`), now pointing at the `_redact_secret`-produced value at the log-write line. I dug into why: `_summarize_scanner_exception_for_log` (the primary defense, used for the exception path) is structurally secret-free by construction — it never reads `api_key` at all, only `type(exc).__name__` and a regex-extracted 3-digit status code — so CodeQL has no real basis to flag that path specifically anymore. The remaining flag traces through `_redact_secret(text, api_key)`, which takes `api_key` as an argument even though it only uses it to `.replace()` out any literal occurrence — CodeQL's taint tracker for this query doesn't model `str.replace()` as a sanitizer, so it conservatively continues to mark the return value as tainted regardless of what the function actually does with the argument.

I verified this is a genuine false positive rather than assuming it: constructed a worst-case exception message with the key embedded (simulating e.g. a debug-mode header dump) and confirmed directly, by execution, that neither the exception-summary path nor the redact-backstop path lets the key reach the persisted log. Dismissed the alert via the code-scanning API with `false positive` + a documented justification referencing this verification (alert #36), rather than leaving a required security check permanently red or merging over it silently. CI is now green end-to-end.

No further findings. This closes out the last piece of Fable's PII-scanner-system remediation plan (Linear BIS-755 / BIS-760). Merging.

PR #2165, opened 2026-08-05
Add GitHub issue PII audit/redact tool (dry-run against #1866, #563)

aeschylus:
## Problem

Two closed-but-public issues in this repo currently expose real third-party
PII in their issue bodies:

- **#1866** — names a real person (a first name that is in this repo's
  existing `~/lobster-user-config/blocked-names.txt` denylist) plus that
  person's personal Telegram `chat_id` (a 10-digit number), in a closed,
  abandoned feature request.
- **#563** — contains a real personal Gmail address attached to an
  already-partially-pseudonymized name; the pseudonymization renamed the
  person in prose but missed the email, defeating the point.

Both are fixable with a plain issue-body edit via the GitHub API — no git
history rewrite, no force-push. This PR does **not** perform that edit. It
adds the tool that finds and (on explicit opt-in) performs the redaction,
and runs it read-only against both real issues so the proposed redaction can
be reviewed before anyone applies it for real.

**`--apply` was never run against #1866 or #563.** Only `--dry-run` (strictly
read-only — see "no writes" test below) was executed against them. Fable
will review the dry-run output pasted below and run `--apply` after
independent review, per the tracking issue (#2161 / Linear BIS-758).

## What this adds

`scripts/audit-github-issue-pii.py`:

- **Detection** (`find_pii`) and **redaction** (`redact_text`) are pure
  functions — text in, findings/text out, no I/O, no mocking needed to unit
  test them.
- The only side effects — fetching an issue (`fetch_issue`) and writing a
  redacted body back (`apply_body_redaction`) — are isolated at the boundary
  and go through `gh api` (per this repo's CLAUDE.md convention: `gh` CLI for
  all GitHub operations, not raw REST/httpx calls). Both take an injectable
  `run=` callable so tests substitute a fake without touching the network.
- `process_issue()` wires the pure core to the boundary and is the single
  place that decides whether a write happens: under `--dry-run` (default),
  `apply_body_redaction` is never invoked — even the underlying `run`
  callable receives zero calls. Under `--apply`, the write receives exactly
  the string `redact_text()` produced — no re-derivation of the redaction at
  the write boundary.
- Findings are merged when spans overlap/touch (`_merge_overlapping`), so
  two pattern categories matching the same or adjacent text produce one
  placeholder, not a doubled/nested one. This is also what makes re-running
  `--apply` on an already-redacted body idempotent: the second pass finds
  nothing left to redact (the sensitive text is gone), so `redact_text`
  returns the body unchanged and no second write is issued.

[...]

Issue #563, opened 2026-03-17
feat: make Linear digest project-aware and understand Linear ontology


sayhar:
## Current Limitation

The Linear digest scheduled job (`linear-digest`) uses an issue-centric query: it only fetches issues assigned to Sahar by looking up her user ID via `sayhar@gmail.com`. This misses two important categories of work:

1. **Issues assigned to the robot account** — the `LINEAR_API_KEY` authenticates as `robot.squad.sm@gmail.com`, but the digest never queries `viewer { assignedIssues { ... } }` to surface work assigned directly to the robot.
2. **Projects where Sahar is a member or lead** — since the query is issue-centric, entire projects that Sahar owns or participates in are invisible unless they happen to surface through issue assignment.

## What's Being Fixed

The task file at `scheduled-jobs/tasks/linear-digest.md` has been updated to query three sources:

1. **Issues assigned to Sahar** — existing behavior, kept via user ID lookup on `sayhar@gmail.com`
2. **Issues assigned to the robot** — new `viewer { assignedIssues { ... } }` (isMe) query
3. **Projects where Sahar is a member or lead** — new project membership query:
   \`\`\`graphql
   projects(filter: { members: { some: { id: { eq: "<sahar-user-id>" } } } }) {
     nodes { id name state description lead { name email } members { nodes { name email } } issues { nodes { ... } } }
   }
   \`\`\`

## Ontology-Aware Surfacing

The task file now includes a Linear ontology section so the subagent understands the key concepts (Issues, Projects, Teams, Cycles, Roadmap initiatives) and can surface what matters in priority order:

1. Urgent issues (priority=1) first
2. Projects Sahar leads or is a member of in "started" state, with open issue counts
3. High-priority issues (priority=2)
4. Robot-assigned issues that may need attention

The robot token (`LINEAR_API_KEY`) is kept as-is — no personal token swap needed.

## Files Changed

- `scheduled-jobs/tasks/linear-digest.md` — task file updated with three-query approach, Linear ontology context, and priority-ordered digest formatting

aeschylus:
The Linear digest project-awareness feature request has no comments showing this was implemented yet.

aeschylus:
Closing as part of the backlog reset — applying the harder bar: "if this were deleted right now, who would actually notice or be blocked?" No named stakeholder (the repo owner or a specific team member) and no live, current production harm is tied to this issue today. Re-open if a concrete, live need surfaces.

PR #2170, opened 2026-08-05
Exempt known project contributors' emails from PII redaction

aeschylus:
## Problem

`scripts/audit-github-issue-pii.py` (PR #2165) redacted `sayhar@gmail.com` in
issue #563 as third-party PII. That address belongs to Sahar Massachi, this
repo's #1 contributor by commit count — a public, attributable identifier
tied to his authorship on this project (git log: ~3,700 commits under that
address or its GitHub noreply alias; GitHub contributors API: 1,295
contributions), not a private individual's incidentally-exposed personal
email. Drew's directive: "don't hide contributors' emails." The live
redaction in issue #563 has already been reverted directly via the GitHub
API (not part of this PR — this PR is the tool fix so it doesn't recur).

Note: this is a different surface from Linear BIS-763 (a prior judgment call
about the same person's email appearing in 10 git *commits* on a single
day) — that was about commit-author metadata and recommended no action.
This PR is about the GitHub *issue-body* redaction tool, a separate code
path with its own decision.

## What changed

`find_emails`/`find_pii` now take an `email_allowlist` — exact-match
(case-insensitive), not a domain/substring match, so a different address at
the same domain is still caught. Two independent sources populate it,
merged by the CLI:

- **`git_contributor_emails(repo_path, run=...)`** — a new boundary function
  that derives the allowlist directly from `git log --all --format=%ae` in
  a given local checkout. This is the default and primary exemption
  mechanism: it reflects actual git history, so it can never go stale the
  way a hand-maintained name/email list would. Isolated and injectable via
  `run=`, the same pattern already used by `fetch_issue` /
  `apply_body_redaction` — read-only, no writes, degrades to an empty list
  (not a crash) if the path isn't a git checkout.
- **`--contributor-emails-file`** (`load_email_allowlist` /
  `parse_email_allowlist`) — an optional static override file, one email
  per line, for exempting addresses that predate or fall outside git
  history. Same file-format convention as the existing `--names-file`
  denylist loader.

`--repo-path` (default `.`) controls which checkout `git log` reads from,
since `--repo` is only the *remote* issue source and may not be the local
directory the script is invoked from. `--no-git-contributors` disables the
dynamic derivation if only the static file is wanted.

## Functional patterns used

Detection stays a pure function: `find_emails`/`find_pii` are unchanged in
shape (text + config in, findings out), just given one more piece of
injected config (`email_allowlist`) rather than a global or hardcoded
exception list. `git_contributor_emails` follows the same imperative-shell
pattern as the module's existing boundary functions: the only side effect
is a single `git log` read, isolated behind an injectable `run` callable
so it's tested with a fake subprocess, never a real one.

## Falsifiability check performed

Stubbed `find_emails` to ignore `email_allowlist` entirely (simulating the
bug this PR fixes — i.e. what issue #563 shipped with). 3 of the 4 new
contributor-exemption tests failed as expected
(`test_contributor_email_is_not_flagged_when_allowlisted`,
`test_contributor_email_allowlist_is_case_insensitive`,
`test_process_issue_does_not_redact_contributor_email`). Restored the real
implementation, diffed against a pre-edit backup to confirm byte-identical
restoration, and confirmed all tests pass again.

## Tests run

- [x] `uv run pytest tests/unit/test_audit_github_issue_pii.py -q` — 30
      passed (18 pre-existing + 12 new: contributor-email exemption
      end-to-end through both `find_pii` and `process_issue`, case
      insensitivity, a genuine third-party email still redacted despite an
      unrelated allowlist entry being present, the email-allowlist file
      loader, and `git_contributor_emails` — real derivation, empty-list
      fallback on `FileNotFoundError`, and on `CalledProcessError`)
- [x] `uv run pytest tests/unit -q` (full unit suite) — 4294 passed, 32
      skipped, 16 failed. All 16 failures are the same pre-existing,
      unrelated failures documented in PR #2165 (Slack router config, hook
      session-id writes, calendar/Gmail/workspace push-token endpoint
      tests, one `daily-update-check.sh` source-field test) — none touch
      this file.

## Manual test

Ran `git_contributor_emails(".")` against this actual repo checkout (not a
fixture): returned 16 distinct contributor emails, correctly including
`sayhar@gmail.com`. `robot.squad.sm@gmail.com` (the service-account address
also present in issue #563, and a separate known scope gap already flagged
in PR #2165 — it isn't a contributor *or* obviously a private individual)
is correctly absent from git log, since it's never a commit-author address;
its handling is unchanged and out of scope here.

This is an internal engineering tool with no Telegram/Slack/UI-facing
output — no user-visible-outcome check applies.

## Out of scope

- No change to the live issue #563 body (already reverted directly via the
  API, outside this PR).
- No change to the robot/service-account email scope gap noted in PR #2165 —
  that's a different category of false positive (service account, not
  contributor) and wasn't part of Drew's "contributors" directive.
- No comment-body redaction changes (still audit-only for comments, per the
  original tool's scope).

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits — the only
      manual run was a read-only `git log` against the local checkout)
- [x] Manual test run — see above
- [x] User-visible outcome — not applicable (internal tool, no UI/Telegram/
      Slack output)
- [x] Side-effect audit complete: no writes anywhere in this change; the
      one manual run was a bounded, read-only `git log` call
- [x] External service calls: none in this change (git log is a local
      command, not an external API)

Relates to Linear BIS-758 / BIS-755. Does not close either — this is a
follow-up correction, not the original tracked scope.

### DIFF
--- a/scripts/audit-github-issue-pii.py
+++ b/scripts/audit-github-issue-pii.py
@@ -40,6 +40,31 @@ PII pattern provenance:
     new: existing hooks only cover the `chat_id=NNNN` env-var assignment
     form, not IDs embedded in issue prose.
 
+Contributor email exemption (added after issue #563's redaction of
+sayhar@gmail.com was reverted — that address belongs to Sahar Massachi, the
+#1 contributor to this repo by commit count, not a "private individual"):
+  - A known project contributor's email is a *public, attributable*
+    identifier tied to their work on the project (visible in `git log`,
+    commit authorship, and the GitHub contributors list) — categorically
+    different from a private individual's personal email incidentally
+    mentioned in an issue body. Redacting it hides authorship, not PII.
+  - `find_emails`/`find_pii` now accept an `email_allowlist` of exempt
+    addresses (case-insensitive exact match) that are never flagged, in
+    addition to the existing domain-pattern allowlist (noreply addresses,
+    example.com, etc).
+  - `git_contributor_emails()` derives that allowlist automatically and
+    dynamically from `git log --all --format=%ae` in a given repo checkout
+    (a boundary function; isolated and injectable via `run=`, exactly like
+    fetch_issue/apply_body_redaction) — so a contributor's exemption is
+    never a hand-maintained list that goes stale, and instead reflects the
+    actual git history. `--repo-path` controls which checkout to read (the
+    default of --repo may point at a *remote* repo that isn't the local
+    checkout the script runs from).
+  - `load_email_allowlist()`/`parse_email_allowlist()` additionally support
+    a plain-text override file (one email per line, `#` comments, blank
+    lines ignored) for exempting emails that predate or fall outside git
+    history, via `--contributor-emails-file`. Both sources are merged.
+
 Usage:
     python scripts/audit-github-issue-pii.py --repo OWNER/REPO --issue 1866 --dry-run
     python scripts/audit-github-issue-pii.py --repo OWNER/REPO --issue 1866 --issue 563 --apply
@@ -143,10 +168,21 @@ class ProcessResult:
 # ---------------------------------------------------------------------------
 
 
-def find_emails(text: str) -> list[Finding]:
+def find_emails(text: str, email_allowlist: Sequence[str] = ()) -> list[Finding]:
+    """Find email-shaped spans, excluding known-benign and known-contributor
+    addresses.
+
+    `email_allowlist` is compared case-insensitively against the exact
+    matched address (not a substring/domain match) — e.g. a contributor's
+    own address never fires as a finding, but a *different* address at the
+    same domain still does.
+    """
+    exempt = {e.strip().lower() for e in email_allowlist if e.strip()}
     findings = []
     for m in _EMAIL_RE.finditer(text):
         if _EMAIL_ALLOWLIST_RE.search(m.group(0)):
+            continue
+        if m.group(0).lower() in exempt:
             continue
         findings.append(Finding("email", "Email address", m.start(), m.end(), m.group(0)))
     return findings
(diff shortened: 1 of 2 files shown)


aeschylus:
## Independent review — PR #2170

**Verdict: PASS**

### Live issue #563 check
Confirmed the current body is byte-for-byte identical to the original (pre-redaction) body I fetched during my first review of this issue — the revert was clean, no drift, no partial/double-edit. `sayhar@gmail.com` is back in cleartext at both original locations, `robot.squad.sm@gmail.com` and the "Alex" pseudonym are unchanged (out of scope for this fix, correctly untouched).

### Code review
Ran the test suite myself: 30 passed — matches the claim.

Reproduced the falsifiability check myself: stubbed `find_emails` to ignore `email_allowlist` entirely (simulating the exact bug #563 shipped with) → 3 failed (`test_contributor_email_is_not_flagged_when_allowlisted`, `test_contributor_email_allowlist_is_case_insensitive`, `test_process_issue_does_not_redact_contributor_email`), 27 passed. Restored, diffed byte-identical, 30/30 pass again.

Ran `git_contributor_emails(".")` against this actual repo myself (not the PR's own transcript) — 16 distinct emails, `sayhar@gmail.com` correctly included. Also verified the exact-match strictness directly: `sayhar@gmail.com` in a test string is correctly exempted while `sayhar@evil.com` (different address, same local part) in the same string is still flagged — confirms this isn't a domain/substring match that could be gamed or over-exempt.

CI is genuinely green (CodeQL passed in 3s like the other clean PRs today, not the kind of result I'd wave through without checking after PR #2169's experience — this one has no annotations).

### On the design question flagged for scrutiny: git-log-derived allowlist as the default

I think this is a reasonable default given Drew's explicit directive, with one real edge case worth naming rather than blocking on: `git_contributor_emails` treats "appears as a commit-author email anywhere in git history" as equivalent to "is a known project contributor whose identity is legitimately public." Those aren't quite identical — a one-off accidental commit under a personal email (this repo already has a real example: Sahar's own raw `sayhar@gmail.com` is only 10 commits, all one day, looking like a local git-config lapse rather than his normal noreply-address pattern, per BIS-763) gets silently exempted by this mechanism with no human-in-the-loop check, purely because it happened to land in `git log` at all. In this specific case that's fine (Drew has separately judged Sahar to be a legitimate, prolific, publicly-attributed contributor — 3,700+ commits, so the one raw-email day is inconsequential), but the mechanism itself doesn't distinguish "prolific known contributor whose one address slipped" from "someone's personal email that leaked into exactly one bad commit and nothing else." There's no exclusion mechanism for the latter case today.

Not asking for a change before merge — the static `--contributor-emails-file` override already provides an opt-in escape hatch for the opposite direction (adding emails not in git history), and a symmetric opt-out/exclude file would be a reasonable low-priority follow-up if this ever bites in practice, not a blocker now.

No correctness issues found. Merging.

Part of Fable's PII-scanner-system remediation plan (Linear BIS-755 / BIS-765).

Issue #2160, opened 2026-08-05
Fix .githooks/pre-push doc-file exemption that let a real Telegram bot token leak


aeschylus:
## Root cause (verified via git log + direct hook inspection)

`docs/DOCKER-TESTING.md` contained a real Telegram bot token (committed 2026-03-14 as `619c5660`, scrubbed 2026-07-17 as `7ebe11f8`/PR #2116, rotation flagged as a to-do but never confirmed). This was not a scanner-pattern miss — the file never reached the scanner at all.

The **actually active** pre-push hook is `.githooks/pre-push` (`core.hooksPath=.githooks` is set repo-wide — confirm with `git config --get core.hooksPath`). Its `should_skip_file()` (around line 143) unconditionally skips ALL scanning — PII *and* secrets — for `*.md|*.mdx|*.rst|*.txt|*.adoc`, on the stated assumption that docs only contain "examples/placeholders by design." That assumption is false in general — a real secret pasted into a doc file is exactly this leak.

Note: `scripts/pre-push-security-scan.sh` is a *separate*, currently-not-installed script that already has a correct `Telegram Bot Token:::[0-9]{8,}:[A-Za-z0-9_-]{35}` CRITICAL pattern — but it is not the hook that's wired up (`.git/hooks/pre-push` doesn't exist on this machine; `core.hooksPath` points to `.githooks` instead). The gap is specifically in `.githooks/pre-push`'s doc exemption.

## Scope

- In `.githooks/pre-push`, split the doc-file exemption: keep exempting docs from **PII/name** pattern checks (docs legitimately reference example names/emails), but do **NOT** exempt doc files from **high-confidence secret/token** pattern checks (Telegram bot token shape, generic API-key shapes, private-key headers, etc — reuse/port the pattern set already in `scripts/pre-push-security-scan.sh`'s `CRITICAL_PATTERNS` where applicable, rather than reinventing).
- Add a regression test: a `.md` fixture containing a real-token-shaped string must now be caught; a `.md` fixture with an obvious placeholder (`<YOUR_BOT_TOKEN>`, `sk-xxx-placeholder-do-not-use`) must still pass through clean.
- Do not touch `scripts/pre-push-security-scan.sh` itself — this issue is scoped to `.githooks/pre-push`, the hook that's actually live on this repo.

## Proof required (do not skip either half)

- Automated test that fails on the old code and passes on the fix (falsifiability check: revert your fix, confirm the new test fails; restore, confirm it passes).
[...]

PR #2171, opened 2026-08-05
Add agent-git-push-guard: block agent-initiated git push on PII/secrets (no TTY gap)

aeschylus:
## Summary

Closes a real, currently-live gap identified during today's BIS-755 PII-scanner
remediation follow-up (per Drew's directive: hooks must never call the
Anthropic API directly, and the mechanism for "the instance calling itself"
should be the existing PreToolUse exit-2 block-and-inject pattern, not a new
queue/poll primitive).

`.githooks/pre-push` already scans every push for PII/secrets/instance data and
blocks interactively -- but only when stdin is a TTY. When a Claude Code agent
(dispatcher or subagent, e.g. `functional-engineer` finishing a PR) runs
`git push` via its `Bash` tool, there is no TTY: `pre-push` detects CI mode,
prints findings, and lets the push through anyway. Agent-initiated pushes
silently get weaker protection than human-initiated ones.

This PR adds `hooks/agent-git-push-guard.py`, a new **PreToolUse** hook
(matcher `Bash`) that:

- Fires only on `git push`-shaped commands whose target remote resolves to the
  public `SiderealPress/lobster` repo (skips everything else silently).
- Computes the outgoing diff the same way `.githooks/pre-push` does
  (`git log @{u}..HEAD -p`, falling back to `origin/main` diff, falling back to
  full `HEAD` log).
- Scans it with `hooks/git_push_scan.py`, a Python port of `.githooks/pre-push`'s
  `PII_PATTERNS`/`SECURITY_PATTERNS`/`INSTANCE_PATTERNS`/`NAME_PATTERNS` tables
  (same user-config files for instance domains / blocked names, same
  file-skip rules, same allowlist file, same placeholder-value filter).
- On a clean diff: exits 0, silently, with no added latency worth mentioning
  and zero API calls -- identical to today.
- On a finding: exits 2 with a findings message. Claude Code injects that
  stderr into the **calling agent's own next turn** -- the same
  block-and-inject mechanism already used by `hooks/link-checker.py` and
  `hooks/require-write-result.py`. The agent must assess each finding as real
  (fix and retry) or a false positive (say why and retry).
- Bounds repeated blocking with a fire-counter (`MAX_HOOK_FIRES=3`, mirroring
  `hooks/require-write-result.py`'s retry-counter/fail-open convention) so a
  stubborn false positive can't wedge an agent forever -- it fails open
  (allows the push) after 3 consecutive blocks and logs the fail-open event.

**Zero Anthropic API calls anywhere in this hook.** All semantic judgment
("is this really PII or a false positive") is deferred to the calling agent's
own reasoning in its next turn, at whatever model tier it already runs -- not
embedded in the hook. This is the design explicitly requested: "add the
feedback message to the hook so any agent would just naturally receive it in
the course of its activities and have to respond to it."

### Why a Python port instead of one shared source file

`.githooks/pre-push` is bash (runs via the git hook mechanism); this hook is
Python (runs inside a Claude Code hook process). They can't literally share
one source file across that boundary. To prevent drift, `git_push_scan.py`'s
pattern tables are checked against a live parse of `.githooks/pre-push`'s bash
arrays by `tests/unit/test_hooks/test_git_push_scan_parity.py` -- if either
side changes a pattern without updating the other, that test fails.

### Also included

- `install.sh`: registers the hook in both existing hook-setup blocks
  (idempotent, matching the `secret-scanner.py`/`pin-dependencies-guard.py`
  precedent).
- `scripts/upgrade.sh`: Migration 98, for existing installs.

## Test plan

- [x] `tests/unit/test_hooks/test_git_push_scan.py` -- 19 tests on the shared
      scanning module: true positives (email, SSN, AWS key, `sk-...` key,
      GitHub token, private key header, hardcoded home path, hardcoded
      password), true negatives/suppressions (clean diff, placeholder values,
      allowlisted email, private IPs, test/doc dirs, `.githooks/` itself,
      `# nosec`, `security-allowlist.txt`), only-added-lines-are-scanned, and
      an explicit falsifiability check (monkeypatches `scan_diff` to a
      no-op stub and proves the real implementation still finds the SSN).
- [x] `tests/unit/test_hooks/test_agent_git_push_guard.py` -- 13 tests against
      **real temporary git repos** (no git mocking): non-Bash/non-push
      commands ignored, pushes to a non-target repo skipped even with real
      PII in the diff, clean push to the target repo allowed silently, both
      https and ssh remote URL forms recognized, PII/secret-shaped pushes
      blocked (exit 2 + message), chained shell commands (`a && b && git
      push`) still detected, the `MAX_HOOK_FIRES` fail-open bound, fire-state
      clearing after a legitimate fix+retry, and a falsifiability check.
- [x] `tests/unit/test_hooks/test_git_push_scan_parity.py` -- 3 tests proving
      the Python pattern tables match `.githooks/pre-push`'s bash tables
      (`PII_PATTERNS`, `SECURITY_PATTERNS`), plus a falsifiability check on the
      parser itself.
- [x] All 33 new tests pass; full `tests/unit/test_hooks/` suite run
      (1046 passed, 8 pre-existing unrelated failures confirmed present on
      `origin/main` before this change too, via `git stash`).
- [x] **Explicit TDD falsifiability proof**: temporarily stubbed
      `git_push_scan.scan_diff()` to always return no findings (simulating a
      reverted/broken implementation) and re-ran the suite -- 15 of the 33
      new tests failed as expected (all the true-positive/blocking tests);
      restored the real implementation and confirmed all 33 pass again.
- [x] **Manual end-to-end proof** (not just unit tests): built a real
      temporary git repo + bare remote, wrote a scratch `.claude/settings.json`
      registering this hook by absolute path (matcher `Bash`), then invoked
      the exact command from that scratch settings file against a real
      PreToolUse-shaped JSON payload for `git push origin main`:
      - A commit containing an AWS-key-shaped string and a real-shaped email
        address was **blocked** (exit 2, ~88ms) with the findings message
        that would be injected into the agent's turn.
      - After "fixing" the finding, the same command **allowed** the push
        silently (exit 0, ~85ms, no stderr).
      - The actual `git push` was then run for real (against a local bare
        remote, not the real GitHub repo) and succeeded in 26ms, proving the
        allow path doesn't just report "ok" but the underlying push truly
        goes through.

## Deploy plan

Per explicit instruction this round, this hook will be registered in the live
`~/.claude/settings.json` for this running Lobster instance immediately after
merge (not left inert), with a live post-deploy verification push test.

### DIFF
--- a/hooks/agent-git-push-guard.py
+++ b/hooks/agent-git-push-guard.py
@@ -1,0 +1,323 @@
+#!/usr/bin/env python3
+"""PreToolUse hook: blocks agent-initiated `git push` on the public repo when
+the outgoing diff contains likely PII/secrets/instance-specific data.
+
+## Why this hook exists
+
+`.githooks/pre-push` already scans every push for PII and instance-specific
+data, and blocks interactively -- but only when stdin is a TTY (a human at a
+terminal). When a Claude Code agent (dispatcher or subagent) runs `git push`
+via its `Bash` tool, there is no TTY: `.githooks/pre-push` detects CI mode,
+prints its findings, and lets the push through anyway (see the `IS_TTY`
+branch in that script). That is a real, currently-live gap: agent-initiated
+pushes get weaker protection than human-initiated ones, for a reason (no TTY
+to prompt) a Claude Code PreToolUse hook does not share.
+
+This hook closes that gap using the mechanism this codebase already uses
+elsewhere (`hooks/link-checker.py`, `hooks/require-write-result.py`,
+`hooks/require-background-agent.py`): a PreToolUse hook that exits 2 to block
+the tool call, with a stderr message Claude Code injects into the *same*
+agent's own next turn. The agent that ran `git push` is the same agent that
+must now assess the finding and decide to fix-and-retry or explain-and-retry.
+
+## No Anthropic API calls
+
+This hook is pure local regex (via `hooks/git_push_scan.py`, ported from
+`.githooks/pre-push`'s pattern tables) plus a handful of local `git`
+subprocess calls to compute the outgoing diff. It never calls the Anthropic
+API or any other network service. The semantic judgment step ("is this
+finding real or a false positive") is deferred entirely to the calling
+agent's own next turn -- not embedded in this hook.
+
+## Scope guard
+
+Only acts on `git push`-shaped Bash commands whose target remote resolves to
+the public `SiderealPress/lobster` repo. Any other repo (a client project, a
+scratch worktree with a different origin, etc.) is skipped silently (exit 0)
+-- this hook has no opinion about pushes that aren't this repo.
+
+## Retry-counter / fail-open convention
+
+Mirrors `hooks/require-write-result.py`: a bounded fire counter (keyed by
+session/agent id) tracked in a `/tmp/lobster-git-push-guard-fires-*` file.
+After `MAX_HOOK_FIRES` consecutive blocks on the same push attempt without
+resolution, the hook fails open (exit 0, allow the push) rather than
+wedging the agent in a false-positive loop forever -- the same policy
+`.githooks/pre-push` already applies to its own scan-timeout case.
+
+Exit codes:
+  0 - allow the tool call (not a git push, not this repo, clean diff, or
+      fail-open after MAX_HOOK_FIRES)
+  2 - block the tool call (stderr message injected into the agent's turn)
+"""
+from __future__ import annotations
+
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+from datetime import datetime, timezone
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent))
+import git_push_scan  # noqa: E402
+
+# Maximum number of consecutive blocks on the same push attempt before
+# failing open. Mirrors require-write-result.py's MAX_HOOK_FIRES convention,
+# but a push finding is expected to resolve in 1-2 turns (fix or explain),
+# so the bound here is smaller than the write_result retry bound.
+MAX_HOOK_FIRES = 3
+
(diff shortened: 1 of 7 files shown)


aeschylus:
Independent cold review (per Drew's authorization to fully resolve): read the diff before any prior notes, checked pattern-table parity against `.githooks/pre-push`, scope guard, fail-open/closed behavior, and injection/traversal risk.

**Found and fixed one real blocking issue:** `git_push_scan.py`'s `should_skip_file()` exempted doc files (`.md`/`.txt`/`.rst`/`.adoc`) from *all* pattern categories, including `SECURITY_PATTERNS`. That silently reintroduced the exact bug fixed in fb49b37f (#2160) — a real secret pasted into a doc file would have bypassed the guard, because `.githooks/pre-push`'s doc-file exemption is deliberately narrower (PII/instance/name only, never secrets). Fixed in 5a9de79d by splitting into `_should_skip_file_common()` (all categories) and `is_doc_file()` (PII/instance/name only), matching the bash source of truth exactly. Added `test_doc_file_still_scans_for_secrets`, confirmed it fails against the pre-fix code (via revert) and passes after the fix.

Everything else checked out:
- Pattern tables are a faithful port of `.githooks/pre-push`'s `PII_PATTERNS`/`SECURITY_PATTERNS` (verified by direct comparison, not just trusting the parity test), same names/descriptions.
- Scope guard fires only when the resolved `git remote get-url` for the target remote matches the public `SiderealPress/lobster` repo (https or ssh); everything else exits 0 silently.
- Fails open (allow push) on infra errors (git command failures, missing repo root) and after `MAX_HOOK_FIRES=3` consecutive blocks, both logged — intentional, consistent with `.githooks/pre-push`'s own CI-mode/timeout fail-open policy.
- No `shell=True` anywhere; all subprocess calls use list-form args. Fire-count file path is derived from a sanitized `agent_key` (alnum/`-._` only), so no path traversal via session/agent id.

CI green, merging.

PR #2184, opened 2026-08-09
Extract nightly-consolidation step-10 bridge-sync into a tested script

aeschylus:
## Problem

Nightly-consolidation's step 10 ("bridge sync") pushes canonical
`projects/*.md` / `priorities.md` into the user-model DB and regenerates
`_context.md`. The logic for that step has, twice now, been hand-copied as
inline Python between two agent-definition files that live in two different
repos: this repo's `.claude/agents/nightly-consolidation.md`, and a
private-config-overlay copy at `~/lobster-user-config/.claude/agents/nightly-consolidation.md`
(outside `SiderealPress/lobster`, per the overlay pattern in
`docs/CUSTOMIZATION.md`).

PR #2136 fixed two bugs in this repo's copy on 2026-08-02. The overlay copy
was never updated, so a 2026-08-09 run using the overlay's stale command hit
both bugs again, live:

- **#2180** — `ModuleNotFoundError`. The stale command does
  `sys.path.insert(0, 'src')` then `from mcp.user_model.bridges import
  run_bridges`. Because `src/mcp` has no `__init__.py`, it loses import
  precedence to the *installed* MCP-SDK `mcp` package once only `src` (not
  `src/mcp`) is on `sys.path` — so `mcp.user_model` resolves to nothing.
- **#2181** — `TypeError: tuple indices must be integers or slices, not str`.
  Even after working around #2180, the stale command opens the DB with a bare
  `sqlite3.connect()`, leaving `row_factory` unset. `user_model.db` query
  helpers (e.g. `get_active_narrative_arcs()`) do dict-style row access
  (`row["id"]`), which raises without `row_factory = sqlite3.Row`. The error
  is swallowed by `run_bridges()`'s broad `except Exception` and surfaces only
  as `{"projects": {"error": "..."}}` in the returned summary — no crash, no
  traceback, just zero projects synced, easy to miss in a "successful" run.

Copy-pasting the fixed command text into the overlay file again would only
fix this occurrence — the underlying problem (inline logic embedded in agent
markdown, hand-kept in sync with `src/mcp/user_model/*` internals, with
nothing to catch drift) would recur.

## Fix

Extracts the bridge-sync invocation into `scripts/run_nightly_bridge_sync.py`,
a first-class, tested CLI that both this repo's agent file and any
private-overlay copy can invoke by path instead of embedding raw Python:

- `_import_user_model()` puts `src/mcp` itself (not `src`) on `sys.path` and
[...]

PR #2185, opened 2026-08-09
fix(update): stop update-lobster.sh from killing itself mid-update

### LINKED ISSUE #2179: update-lobster.sh kills itself when run from inside lobster-claude session — update silently no-ops
## Summary

Running \`update-lobster.sh\` from a Bash tool call inside the always-on \`lobster-claude\` session (the normal way a user triggers "do lobster update" via Telegram) causes the script to kill itself partway through, before it does anything meaningful. The service ends up reporting a fresh/healthy \`lobster-claude\` (because systemd auto-restarts it), which looks like success, but the repo was never updated and \`lobster-mcp-local\`/\`lobster-router\` were never restarted.

## Root cause

\`SERVICES_STOP=("lobster-claude" "lobster-mcp-local" "lobster-router")\` in \`scripts/update-lobster.sh\` stops \`lobster-claude\` **first**. But when the script is invoked via the dispatcher's Bash tool, the script's own process tree is a descendant of the \`lobster-claude.service\` cgroup. \`stop_services()\`'s \`sudo systemctl stop lobster-claude --no-block\` (or the fallback \`systemctl kill\`) terminates that whole cgroup — including the currently-running \`update-lobster.sh\` process itself — before it reaches \`git_update()\`, \`update_dependencies()\`, \`update_systemd()\`, or the restart of \`lobster-mcp-local\`/\`lobster-router\`.

systemd's \`Restart=on-failure\` on \`lobster-claude.service\` then brings the dispatcher back up on its own, which reads as "the update finished and restarted things" — but nothing after the "Stopping lobster-claude..." log line ever ran.

## Reproduction observed in production (2026-08-08)

- Log \`~/lobster-workspace/logs/lobster-update-manual-20260808-164734.log\` cuts off immediately after:
  \`\`\`
  >>> Stopping services
  [INFO] Stopping lobster-claude...
  \`\`\`
  No further lines — no "repository update", no "starting services", no "health checks".
- \`/tmp/lobster-update.lock\` was cleaned up (bash \`EXIT\` trap fired on the terminating signal), so there's no stale-lock symptom to notice.
- \`cd ~/lobster && git log -1\` still showed the pre-update commit (\`a14c5ae5\`) while \`origin/main\` was 2 commits ahead (\`cc273eec\`, which happened to be the fix for issue #2175/#2177 the user was specifically trying to pick up).
- \`systemctl status lobster-mcp-local lobster-router\` showed both still \`active\` with an uptime from the *previous* day — i.e. never restarted — while \`lobster-claude\` showed an uptime of under a minute (systemd's own auto-restart, not the script's \`start_services()\`).
- User-visible effect: the dispatcher confidently reported "update complete" territory (session came back up, looked healthy) while the actual fix the user was waiting on silently never shipped, and the spam it was meant to suppress kept firing every ~3-4 minutes afterward.

## Suggested fix

\`update-lobster.sh\` needs to survive stopping its own service. Options:
1. Detach the script from the \`lobster-claude.service\` cgroup before calling `stop_services` on it — e.g. re-exec itself via \`systemd-run --scope\` (or \`setsid ... disown\`) outside the unit's cgroup, or have the dispatcher hand off execution to a small detached wrapper/cron-triggered runner instead of running it inline in its own Bash tool.
2. Alternatively, reorder \`SERVICES_STOP\`/\`SERVICES_START\` so \`lobster-claude\` is stopped **last** and started **last** (matching \`SERVICES_START\`'s existing last-position for \`lobster-claude\`), and only signal/exit the current session as the very last step, after git/deps/systemd work is already done and mcp-local/router are already back up.
3. At minimum, have the script (or the dispatcher before invoking it) verify post-run that \`git rev-parse HEAD\` matches the intended target commit and that \`lobster-mcp-local\`/\`lobster-router\` were actually bounced, and surface a clear failure/incomplete-update notification instead of silence.

## Impact

Any \"do lobster update\" triggered the normal way (from inside the live dispatcher session) silently no-ops on everything except restarting \`lobster-claude\` itself. Users have no reliable signal that the update didn't happen short of manually diffing \`git log\` / service uptimes, as observed here.

aeschylus:
## Problem

`update-lobster.sh` is normally invoked via a Bash tool call from inside the live `lobster-claude` dispatcher session — that's how "do lobster update" gets triggered. That makes the script's own process tree a descendant of `lobster-claude.service`'s cgroup. `SERVICES_STOP` listed `lobster-claude` *first*, so `stop_services()`'s `systemctl stop lobster-claude` tore down that whole cgroup — including the running `update-lobster.sh` process itself — before `git_update()`, `update_dependencies()`, `update_systemd()`, or the `lobster-mcp-local`/`lobster-router` restart ever ran.

systemd's `Restart=on-failure` then silently brought `lobster-claude` back up on its own, which read as "the update finished and restarted things," while the repo was never actually pulled and `mcp-local`/`router` were never bounced. This is not hypothetical — it recurred in production on 2026-08-09 (see issue) after an earlier occurrence on 2026-08-08.

Closes #2179.

## Fix

Implements suggested-fix option 2 from the issue (reorder so `lobster-claude` is stopped/started last), restructured as a proper "self-last" flow, plus option 3 (post-update verification) as a belt-and-suspenders addition:

- `lobster-claude` is removed from `SERVICES_STOP`/`SERVICES_START` entirely. Only `lobster-mcp-local` and `lobster-router` — services that are *not* the cgroup this script runs in — are stopped, updated for, and restarted mid-flow.
- A new `restart_self_last()` bounces `lobster-claude` (`systemctl restart --no-block`, matching the script's existing non-blocking pattern) as the **literal last statement of `main()`**, after git update, dependency install, systemd template regen, CLI update, mcp-local/router restart, health checks, success notification, and lock cleanup have all already completed. If that call is what triggers this process's own death, the update has already actually happened.
- `health_checks()` now also verifies `git rev-parse HEAD` matches the commit `git_update()` merged to, so a partially-applied update can no longer silently read as success.
- The explicit `--rollback` CLI path threads the same "restart self last" behavior through `perform_rollback()` (via a new `restart_self` parameter), since a manual rollback also changes the checked-out commit. The three in-flow failure-rollback call sites (git merge failed / services failed to start / health checks failed) do **not** pass this — `lobster-claude` was never touched in those paths (it's only ever touched by `restart_self_last()`, which hasn't run yet), so restarting it there would be an unnecessary, unrelated bounce of a perfectly healthy service.
- `LOCK_FILE` is now overridable via `LOBSTER_UPDATE_LOCK_FILE` (default unchanged: `/tmp/lobster-update.lock`) purely for test isolation — no production behavior change.

**Functional patterns used:** the fix is primarily a control-flow reorder with named constants (`SERVICE_SELF`) replacing an implicit "index 0 is special" convention; `restart_self_last()` is a single-purpose, side-effect-isolated function; `perform_rollback()`'s new behavior is threaded through an explicit parameter rather than a global/hidden flag, keeping the function's effects visible at each call site.

## Flow change

```
BEFORE (buggy):                          AFTER (fixed):
stop lobster-claude   <- kills self!     stop mcp-local, router
stop mcp-local                           git update
stop router                              deps / systemd / CLI update
git update            <- never reached   start mcp-local, router
deps / systemd / CLI                     health checks (+ HEAD verify)
start router, mcp-local,                 success notification
  lobster-claude                         cleanup_lock
health checks                            restart_self_last()  <- lobster-claude,
success notification                                             LAST, may kill self here
```

Before: touching `lobster-claude` at step 1 could kill the whole update before anything meaningful ran, and there was no way to tell from the outside. After: `lobster-claude` is the only thing that can still be running old code partway through, and it's touched only once everything else is verifiably done.

**Out of scope:** no changes to `install.sh`, `scripts/upgrade.sh`, service `.template` files, or the health-check script. Suggested-fix option 1 (re-exec via `systemd-run --scope` to fully detach from the cgroup) was not implemented — option 2 avoids the added complexity of a self re-exec while still being fully correct, since it only requires care about *when* `lobster-claude` is touched, not *whether* the script is inside its cgroup.

## Tests run

- [x] `bash tests/test-update-lobster-self-restart-last.sh` (new) — 17/17 passed. This test reproduces the actual cgroup-kill dynamic in a hermetic fixture: a mocked `systemctl` sends `SIGTERM` then `SIGKILL` to the running `update-lobster.sh` process itself the moment it's asked to touch `lobster-claude` (exactly what a real cgroup teardown does), against a fake git origin/clone with a real commit to pull. Part A checks the `SERVICES_STOP`/`SERVICES_START`/`SERVICE_SELF` constants directly. Part B asserts that by the time `lobster-claude` is touched, the fake repo has already fast-forwarded to the target commit, `lobster-mcp-local`/`lobster-router` were already restarted (in that log order, before the `lobster-claude` call), the success notification was already sent, and the lock file was already cleaned up. Part C is a regression guard that the reordering didn't break the normal (non-self-kill) path.
- [x] `bash -n scripts/update-lobster.sh` — clean.
- [x] Falsifiability check: reverted `scripts/update-lobster.sh` to the `origin/main` version and reran the test — Part A fails immediately (`SERVICE_SELF: unbound variable`, since the constant doesn't exist pre-fix). To confirm Part B/C fail for the *causally correct* reason (not just the missing constant), I ran a scratch variant of the test against the original script with Part A's constants hardcoded to the pre-fix values instead of extracted: 7/17 failed, specifically — `stop lobster-claude --no-block` is the very first systemctl call (right after startup `is-active` probes), the fake repo's git HEAD never advances past the base commit, `lobster-mcp-local`/`lobster-router` are never started, and the only notification sent is `notify_starting` ("Lobster updating") — `notify_success` never fires. This reproduces the exact production symptom described in the issue. Restored the fix afterward and reran — 17/17 passed again.
- [ ] Real end-to-end run against actual systemd units on this host — not run. This script restarts `lobster-mcp-local`/`lobster-router`/`lobster-claude` for real; running it live would have been a genuine, disruptive production side effect on the very system I'm working from (this task itself runs inside a `lobster-claude` session), not a scoped/reversible test. The hermetic integration test above reproduces the actual failure mechanism (cgroup teardown killing the running process) faithfully enough that a real run wasn't necessary to prove the fix; see the falsifiability check above for the causal confirmation.

## Manual test

Covered by the integration test above, which *is* the manual/simulated verification: it actually runs the real `scripts/update-lobster.sh` script (not a mock of it) end-to-end via `bash`, against a real (fake, local) git repo with a real fast-forward merge, with only `systemctl`/`sudo`/`curl`/`claude` mocked out — and the mock `systemctl` genuinely sends real `SIGTERM`/`SIGKILL` signals to the real running script process, reproducing the actual kill dynamic rather than simulating it symbolically. No real Telegram messages were sent (curl is mocked) and no real systemd units were touched (systemctl/sudo are mocked) — this was a from-scratch fixture, not a scoped production call, so there was nothing to ask the user to confirm.

## Definition of Done
- [x] Unit/integration tests pass with proper mocking (no production hits — no real systemctl/sudo/curl calls)
- [x] Integration test run (see above) — this *is* the manual verification for this change, given the alternative (a real run) would have live-restarted the services this task itself depends on
- [x] User-visible outcome: N/A — this is an internal reliability fix for an admin-triggered script; no new user-facing output. The existing Telegram success/failure notifications (`notify_success`/`notify_failure`) are unchanged in content, only in *when* they can now be reliably assumed accurate.
- [x] Side-effect audit: no floods, no unscoped writes. The fixture only ever touches its own `mktemp -d` directory; the only host-shared path touched by the test is `/tmp/lobster-update.lock`'s *equivalent*, which is redirected via the new `LOBSTER_UPDATE_LOCK_FILE` override so the real production lock file is never touched by tests.
- [x] External service calls: all mocked in tests (systemctl, sudo, curl, claude); no manual run against production services was performed (see above)

## Known gaps

- No real end-to-end run against this host's actual systemd units (see above) — flagging per policy in case a maintainer wants that done separately in a maintenance window before merge, though I don't think it's necessary given the integration test's fidelity.
- I don't have `read:project` scope on this token, so I couldn't move this issue's Main Board status to "In Review" — flagging so it can be moved manually.

### DIFF
--- a/scripts/update-lobster.sh
+++ b/scripts/update-lobster.sh
@@ -43,13 +43,36 @@ MESSAGES_DIR="${LOBSTER_MESSAGES:-$HOME/messages}"
 LOBSTER_CONFIG_DIR="${LOBSTER_CONFIG_DIR:-$HOME/lobster-config}"
 CONFIG_FILE="$LOBSTER_CONFIG_DIR/config.env"
 
-# Lock file
-LOCK_FILE="/tmp/lobster-update.lock"
+# Lock file (overridable for test isolation; production default unchanged)
+LOCK_FILE="${LOBSTER_UPDATE_LOCK_FILE:-/tmp/lobster-update.lock}"
 
-# Services (in stop order - claude first, then mcp-local, then router)
-SERVICES_STOP=("lobster-claude" "lobster-mcp-local" "lobster-router")
+#-------------------------------------------------------------------------------
+# Services
+#
+# lobster-claude is deliberately EXCLUDED from SERVICES_STOP/SERVICES_START
+# (issue #2179). This script is normally invoked via a Bash tool call from
+# inside the live lobster-claude dispatcher session itself, which makes its
+# own process tree a descendant of lobster-claude.service's cgroup. Stopping
+# lobster-claude mid-flow (as the old code did, first, via SERVICES_STOP)
+# tears down that whole cgroup -- including this very script -- before
+# git_update()/update_dependencies()/update_systemd() or the mcp-local/router
+# restart ever run. systemd's Restart=on-failure then silently brings
+# lobster-claude back up, which looks like "update finished" but nothing
+# after the "Stopping lobster-claude..." log line ever executed.
+#
+# Fix: lobster-claude is never stopped/started as part of the normal
+# stop-work-start cycle. Only lobster-mcp-local and lobster-router (which are
+# NOT the cgroup this script runs in) are cycled mid-flow. lobster-claude is
+# bounced via restart_self_last() as the literal last statement of main(),
+# after every meaningful step (git update, deps, systemd regen, CLI update,
+# mcp-local/router restart, health checks, success notification, lock
+# cleanup) has already completed -- so if this process dies right there, the
+# update has already actually happened.
+#-------------------------------------------------------------------------------
+SERVICE_SELF="lobster-claude"
+SERVICES_STOP=("lobster-mcp-local" "lobster-router")
 # Start order is reversed
-SERVICES_START=("lobster-router" "lobster-mcp-local" "lobster-claude")
+SERVICES_START=("lobster-router" "lobster-mcp-local")
 
 # State files to backup
 STATE_FILES=(
@@ -221,8 +244,10 @@ create_backup() {
         claude --version 2>/dev/null > "$BACKUP_DIR/claude-version.txt" || echo "unknown" > "$BACKUP_DIR/claude-version.txt"
     fi
 
-    # Save service states
-    for service in "${SERVICES_STOP[@]}"; do
+    # Save service states (SERVICES_STOP plus lobster-claude, which is
+    # handled separately by restart_self_last() -- see comment at the
+    # SERVICES_STOP/SERVICES_START declaration)
+    for service in "${SERVICES_STOP[@]}" "$SERVICE_SELF"; do
         if systemctl is-active --quiet "$service" 2>/dev/null; then
             echo "active" > "$BACKUP_DIR/${service}.state"
         else
@@ -240,35 +265,16 @@ create_backup() {
 stop_services() {
     log STEP "Stopping services"
 
+    # NOTE: lobster-claude is never in this array -- see comment at the
+    # SERVICES_STOP declaration. It is handled separately, last, by
+    # restart_self_last().
     for service in "${SERVICES_STOP[@]}"; do
         if systemctl is-active --quiet "$service" 2>/dev/null; then
             if $DRY_RUN; then
                 log INFO "Would stop $service"
             else
                 log INFO "Stopping $service..."
-
-                # For lobster-claude, wait for Claude to finish current work
-                if [ "$service" = "lobster-claude" ]; then
-                    # Send SIGTERM and wait up to 60 seconds
-                    sudo systemctl stop "$service" --no-block 2>/dev/null || true
-
-                    local wait_count=0
-                    while systemctl is-active --quiet "$service" 2>/dev/null && [ $wait_count -lt 60 ]; do
-                        sleep 1
-                        ((wait_count++))
-                        if [ $((wait_count % 10)) -eq 0 ]; then
-                            log INFO "Waiting for lobster-claude to finish... (${wait_count}s)"
-                        fi
-                    done
-
-                    if systemctl is-active --quiet "$service" 2>/dev/null; then
-                        log WARN "lobster-claude didn't stop gracefully, forcing..."
-                        sudo systemctl kill "$service" 2>/dev/null || true
-                        sleep 2
-                    fi
-                else
-                    sudo systemctl stop "$service" 2>/dev/null || true
-                fi
+                sudo systemctl stop "$service" 2>/dev/null || true
 
                 if ! systemctl is-active --quiet "$service" 2>/dev/null; then
                     log OK "$service stopped"
@@ -311,6 +317,39 @@ start_services() {
         return 1
     fi
     return 0
+}
+
+#-------------------------------------------------------------------------------
+# Restart self (lobster-claude) -- LAST, on purpose (issue #2179)
+#
+# Must only ever be called after every other meaningful step has already
+# completed: git update, deps, systemd regen, CLI update, mcp-local/router
+# restart, health checks, success notification, and lock cleanup. By the
+# time this runs, there is nothing left that this process dying mid-flight
+# could leave incomplete.
+#
+# Uses --no-block deliberately: this call may itself trigger the cgroup
+# teardown that kills the very process making the call (when invoked from
+# inside the live lobster-claude session), so it must not wait around for a
+# result. systemd's Restart=on-failure (or the graceful stop/start pair this
+# triggers) brings lobster-claude back up on its own with the newly updated
+# code.
+#-------------------------------------------------------------------------------
+restart_self_last() {
+    log STEP "Restarting $SERVICE_SELF (last step)"
+
+    if $DRY_RUN; then
+        log INFO "Would restart $SERVICE_SELF"
+        return 0
+    fi
+
+    if systemctl is-active --quiet "$SERVICE_SELF" 2>/dev/null; then
+        log INFO "Restarting $SERVICE_SELF..."
+        sudo systemctl restart "$SERVICE_SELF" --no-block 2>/dev/null || true
+    else
+        log INFO "Starting $SERVICE_SELF..."
+        sudo systemctl start "$SERVICE_SELF" --no-block 2>/dev/null || true
+    fi
 }
 
 #-------------------------------------------------------------------------------
@@ -604,8 +643,11 @@ health_checks() {
         return 0
     fi
 
-    # Check services running
-    for service in "${SERVICES_START[@]}"; do
+    # Check services running (SERVICES_START plus lobster-claude, which
+    # hasn't been touched yet at this point -- it's restarted separately,
+    # last, by restart_self_last() -- so this just confirms it's still
+    # healthy going into that final step)
+    for service in "${SERVICES_START[@]}" "$SERVICE_SELF"; do
         if systemctl is-active --quiet "$service" 2>/dev/null; then
             log OK "$service is running"
         else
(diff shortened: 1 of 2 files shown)


aeschylus:
## Independent reviewer pass (pre-merge)

Reviewed at the PR's HEAD (`b8809eca`) in its existing worktree, independently -- not trusting the implementing agent's claims at face value. This fix touches the update mechanism for the very service this review session runs as, so extra scrutiny applied.

- Read the full diff (`scripts/update-lobster.sh`, `tests/test-update-lobster-self-restart-last.sh`) and traced `main()`'s actual call order by hand: `lobster-claude` is removed from `SERVICES_STOP`/`SERVICES_START` entirely; `restart_self_last()` is genuinely the literal last statement executed in `main()`, after `cleanup_lock`, `notify_success`, health checks, mcp-local/router restart, and git update all complete.
- Confirmed the three in-flow failure-rollback call sites (git merge failed / services failed to start / health checks failed) call `perform_rollback` with no argument, so `restart_self="${1:-false}"` defaults false there -- `lobster-claude` is never touched on those paths. Verified `die()` doesn't touch it either. Only the explicit `--rollback` CLI path passes `perform_rollback true`, correctly threading self-restart through.
- Verified the new `git rev-parse HEAD` check in `health_checks()`: compares against `CURRENT_COMMIT` (set in `git_update()` right after the ff-merge), guarded by `[ -n "$CURRENT_COMMIT" ]`, and the whole function early-returns under `$DRY_RUN` before reaching it -- no false-positive risk there.
- Ran `bash tests/test-update-lobster-self-restart-last.sh` myself: **17/17 passed.**
- Read the mock `systemctl` in the test harness line by line: when asked to touch `lobster-claude` with self-kill simulation on, it looks up the real target PID (written by `run_scenario` before backgrounding the real script) and sends real `SIGTERM` then `SIGKILL` to it -- a faithful reproduction of cgroup teardown, not a weakened approximation.
- Falsifiability check, done two ways:
  - Reverted `scripts/update-lobster.sh` to `origin/main` and reran the *actual committed test file* unmodified: it hard-crashes at Part A (`SERVICE_SELF: unbound variable`, `set -u`) rather than reporting a clean 17/17 or partial pass -- itself a valid (if blunt) falsifiability signal, and consistent with what the PR body discloses.
  - To confirm the causal mechanism (not just the missing constant), built a scratch variant with the pre-fix constants hardcoded instead of eval-extracted, and reran against `origin/main`'s script: **10/17 passed, 7/17 failed**, with exactly the claimed signature -- `stop lobster-claude --no-block` is the very first mutating systemctl call, the fake repo's git HEAD never advances past the base commit, `lobster-mcp-local`/`lobster-router` are never started, and only `notify_starting` ("Lobster updating") fires -- `notify_success` never does. Reproduces the real 2026-08-09 05:21 UTC incident signature.
  - Restored the fix, confirmed `git diff` empty, reran: 17/17 green again.
- Scope check: diff touches exactly the 2 files claimed (`scripts/update-lobster.sh`, `tests/test-update-lobster-self-restart-last.sh`), 501/-33 lines, no unrelated changes. `bash -n scripts/update-lobster.sh` -- clean.
- Checked the abandoned pre-existing worktree for this issue: the directory itself is already gone (cleanly abandoned, nothing dangling under `~/lobster-workspace/projects/`). Its work survives only as a local `git stash` (`WIP on fix/issue-2179-update-self-detach-cgroup`) in the shared repo. Read it: it's a concrete sketch of "suggested-fix option 1" (self re-exec via `systemd-run --scope` to fully detach from the cgroup before touching `lobster-claude`) -- the same option this PR's description explicitly considered and rejected in favor of the simpler "touch `lobster-claude` last" approach. The sketch is real but meaningfully more complex (cgroup-membership detection via `/proc/self/cgroup`, a re-exec with a recursion-guard flag, `disown`, background child process) for no correctness gain over the merged approach. Already-documented rejection stands; nothing worth cherry-picking back in as a comment or follow-up.
- Did not run the script live against this host's real systemd units, per explicit instruction -- the hermetic test's fidelity (real SIGTERM/SIGKILL to the real running process) is a faithful enough reproduction of the actual failure mechanism that a live run isn't needed to trust this fix, and a live run here would have been genuinely destructive to this very session and to `lobster-mcp-local`/`lobster-router`.
- CI: CodeQL (JS/TS + Python) both green, no branch protection on `main` (same precedent as #2149/#2184).

No issues found. Merging via squash, consistent with recent merged PRs (single-parent squash commits on `main`, e.g. #2178, #2182, #2184).

Issue #2157, opened 2026-08-05
docs: comments assert unverified '7440s CC hard session limit' as fact


sayhar:
## Summary

Multiple code comments and docstrings across the repo assert as established fact that "Claude Code enforces a hard 7440-second session lifetime limit." This claim is unverified and likely false — or at minimum, unconfirmed.

## Background

The claim traces to issue #2059 and PR #2060 (filed+merged 2026-05-09/10), which introduced the ~7200s proactive-restart mechanism in `health-check-v3.sh` / `hooks/inject-bootup-context.py`. That mechanism itself may still have operational value — this issue is only about the **false certainty of its stated justification**.

Findings from investigation (session task `verify-cc-session-limit-20260805-1738`):

- **Weak evidence for the original claim.** Issue #2059 cited only 2 data points (sessions 016/017 dying around 7433-7436s), with no raw logs attached and no citation to any Anthropic doc or changelog entry describing a wall-clock session-duration hard limit.
- **Likely not authored by the actual user.** Issue #2059's text contains "Alex sees 'Lobster Starting'" — not the real user's name (Sahar) — suggesting the issue was authored by a prior Claude/Lobster session inferring a cause, not a human confirming one. The "hard limit" framing may be an AI inference that got presented as fact and then propagated into comments as if verified.
- **No systemd-level mechanism matches.** No systemd unit (`lobster-claude.service`) has any `RuntimeMaxSec` or timeout around 7440s — this rules out a literal systemd-enforced kill as the mechanism observed.
- **A plausible alternative cause was never ruled out.** In the days surrounding the original "evidence," Lobster's own `DISPATCHER_HEARTBEAT_STALE_SECONDS` health-check restart threshold was being actively patched for bugs (commit `fc2260ce`, one day prior to #2059; later fix in issue #2074). That mechanism kills the dispatcher process via `systemctl restart` with no Stop hook — which matches the symptoms (no tombstone, no compaction alert) originally attributed to "CC's hard limit."
- **No corroborating documentation found.** A web search found no Anthropic documentation of a wall-clock session-duration hard limit for Claude Code. The only documented limits are a 5-hour rolling token/usage rate-limit window (which throttles, but doesn't kill an in-progress process) and a 200K token context window (token-based, not time-based). Neither matches the claimed ~7440s wall-clock kill behavior.

## Affected files (comments/docstrings asserting the false claim as fact)

- `hooks/inject-bootup-context.py` (3 locations)
[...]

PR #2197, opened 2026-08-11
fix(health-check): restore graceful session-age restart in debug mode, without the debunked 2h forced recycle

### LINKED ISSUE #2196: fix(health-check): graceful SESSION AGE LIMIT restart silently no-ops in debug mode — dispatcher.pid never written by claude-wrapper.exp
## Problem

`health-check-v3.sh`'s `check_session_age()` sends a graceful `SIGTERM` to the dispatcher when the session reaches `SESSION_AGE_LIMIT_SECONDS` (the proactive restart before CC's hard 7440s session limit, from #2059). It reads the target PID from `$MESSAGES_DIR/config/dispatcher.pid`.

That file is only ever written by `claude-persistent.sh` (the default, `LOBSTER_DEBUG=false` launcher) — see `scripts/claude-persistent.sh:567-571`. `claude-wrapper.exp` (the `LOBSTER_DEBUG=true` interactive/debug launcher) writes `dispatcher-startup-flag` but never writes `dispatcher.pid`.

Result: whenever the dispatcher is running under `claude-wrapper.exp` (debug mode), `check_session_age()` finds no PID, logs `[WARN] Session age: no dispatcher.pid — cannot send SIGTERM (systemctl restart fallback not used for this path)`, and returns without doing anything. This repeats on every 4-minute health-check tick for as long as debug mode stays on and the session stays over-age — the graceful restart mechanism is completely inert in this mode. The only things that eventually restart the dispatcher are the harder backstops (BLACK/circuit-breaker cap, a crash, or an external systemd bounce), not the intended graceful path.

## Evidence

`health-check.log`: the `no dispatcher.pid` warning has fired **4,148 times since 2026-06-07**. Cross-checking timestamps: essentially every occurrence co-occurs with `[INFO] Claude running without persistent wrapper (debug mode — expected)` (or, when `LOBSTER_DEBUG` had already been flipped back to `false` in config but the still-running process was launched while it was `true`, the less-obviously-related `[WARN] Claude running without persistent wrapper (old-style mode?)`). First occurrence (2026-06-07T10:48:01-04:00) confirms this immediately:

```
[2026-06-07T10:48:01-04:00] [WARN] SESSION AGE LIMIT: session is 7432s old (>= 7200s) — sending SIGTERM for graceful restart
[2026-06-07T10:48:01-04:00] [WARN] Session age: no dispatcher.pid — cannot send SIGTERM (systemctl restart fallback not used for this path)
[2026-06-07T10:48:02-04:00] [INFO] Claude running without persistent wrapper (debug mode — expected)
```

Most recently: the 2026-08-11 04:00-06:00 EDT window flagged in a system audit (`~/lobster-user-config/agents/system-audit.context.md`, "2026-08-11 — Two restarts ~2h apart") — `LOBSTER_DEBUG=true` had drifted on around 2026-08-05 (per that same audit trail's 2026-08-07 entry) and stayed running under `claude-wrapper.exp` for several days; the graceful restart path was silently dead the whole time, until an unrelated `apt-get upgrade` cascade bounced `lobster-claude.service` and picked up the (by-then-reverted) `LOBSTER_DEBUG=false` on the next launch.

Confirmed against current live state: `dispatcher.pid` currently exists at `/home/lobster/messages/config/dispatcher.pid`, contains the live dispatcher PID, and matches `health-check-v3.sh`'s `DISPATCHER_PID_FILE="$MESSAGES_DIR/config/dispatcher.pid"` — the path itself is correct and consistent between writer and reader when running under `claude-persistent.sh`. The bug is specifically that `claude-wrapper.exp` never performs the equivalent write.

## Fix

Have `claude-wrapper.exp` write `dispatcher.pid` (using `[exp_pid]`, mirroring the already-correct pattern used for `dispatcher-startup-flag`) right after `spawn claude ...`, and clean it up when the wrapper exits (mirroring `claude-persistent.sh`'s `rm -f` on exit). This makes the graceful SESSION AGE LIMIT restart path work identically in both launch modes, closing the gap without changing `health-check-v3.sh`'s reader logic at all.

sayhar:
Closes #2196

## Problem

`health-check-v3.sh`'s `check_session_age()` sends a graceful `SIGTERM` to the dispatcher once a session passes `SESSION_AGE_LIMIT_SECONDS` (the proactive restart added in #2059 to beat CC's claimed hard 7440s session limit). It finds the target PID by reading `$MESSAGES_DIR/config/dispatcher.pid`.

That file is only ever written by `claude-persistent.sh` (the default, `LOBSTER_DEBUG=false` launcher), right before it execs `claude`. `claude-wrapper.exp` — the `LOBSTER_DEBUG=true` interactive/debug launcher — writes the analogous `dispatcher-startup-flag` but never wrote `dispatcher.pid`. So whenever the dispatcher was running in debug mode, `check_session_age()` could never find a PID: it logged `[WARN] Session age: no dispatcher.pid — cannot send SIGTERM` and took no action, on every 4-minute health-check tick, for as long as the session stayed over-age. The graceful restart mechanism was completely inert in that mode; the dispatcher only ever got restarted via harder backstops (the BLACK/circuit-breaker cap, a crash, or an unrelated systemd bounce).

## Evidence

`health-check.log` shows the `no dispatcher.pid` warning firing **4,148 times since 2026-06-07**. The very first occurrence already shows the co-occurring cause directly:

```
[2026-06-07T10:48:01-04:00] [WARN] SESSION AGE LIMIT: session is 7432s old (>= 7200s) — sending SIGTERM for graceful restart
[2026-06-07T10:48:01-04:00] [WARN] Session age: no dispatcher.pid — cannot send SIGTERM (systemctl restart fallback not used for this path)
[2026-06-07T10:48:02-04:00] [INFO] Claude running without persistent wrapper (debug mode — expected)
```

Most recently, a 2026-08-11 system audit (`~/lobster-user-config/agents/system-audit.context.md`) caught the same signature during a ~2-day window where `LOBSTER_DEBUG=true` had drifted on; the graceful path stayed dead the whole time until an unrelated `apt-get upgrade` cascade happened to bounce `lobster-claude.service`, which picked up the (by-then-reverted) `LOBSTER_DEBUG=false` on its next launch.

## Fix, part 1: `claude-wrapper.exp` writes `dispatcher.pid`

`claude-wrapper.exp` now writes `dispatcher.pid` right after `spawn claude ...`, using `[exp_pid]` — the same fix already applied to the `dispatcher-startup-flag` write (`$spawn_pid` is not a valid Tcl variable and silently produces an empty file). The path is derived from `LOBSTER_MESSAGES` with the same `$HOME/messages` fallback `claude-persistent.sh` and `health-check-v3.sh` already use. The file is removed once `interact` returns, mirroring `claude-persistent.sh`'s `rm -f` cleanup on exit.

No changes to `health-check-v3.sh`'s reader logic in this part — it closes the gap purely on the writer side.

## Fix, part 2: `SESSION_AGE_LIMIT_SECONDS` now defaults to disabled (0)

Fixing part 1 on its own would have reintroduced forced restarts every ~2 hours for every debug-mode session, on a justification that doesn't hold up.

`SESSION_AGE_LIMIT_SECONDS` (the threshold that gates the SIGTERM in `check_session_age()`) previously defaulted to `7200` (2h), justified by issue #2059's claim that "Claude Code enforces a hard 7440-second session lifetime." Issue #2157 investigated that claim and found:
- It rested on only 2 data points (sessions dying around 7433-7436s), with no raw logs and no citation to any Anthropic doc or changelog describing a wall-clock session-duration limit.
- The issue that originated the claim appears to have been authored by a prior Claude/Lobster session inferring a cause, not a human confirming one.
- No systemd unit has any `RuntimeMaxSec` or timeout near 7440s that would explain a hard kill.
- A more plausible, unruled-out alternative was active at the same time: `DISPATCHER_HEARTBEAT_STALE_SECONDS`'s health-check restart threshold was being actively patched for bugs one day before #2059 was filed, and later fixed properly in #2074 — a mechanism that kills the dispatcher via `systemctl restart` with no Stop hook, matching the originally observed symptoms.

Since then, the live dispatcher session (the one running this very conversation) has run **90+ hours with zero problems** — directly contradicting the original 7440s premise. There is no remaining evidence that Claude Code kills sessions on a wall-clock timer.

Given that, merging PID-fix part 1 as-is would have taken a currently-inert-and-harmless mechanism and made it start firing forced 2-hour restarts again, for a reason that's now known to be unfounded.

**Change:** `SESSION_AGE_LIMIT_SECONDS` now defaults to `0` (disabled) instead of `7200`. The mechanism itself is unchanged and still fully functional — including the part-1 PID fix, so it works correctly in both launch modes — but it no longer runs unless an operator explicitly opts in via `LOBSTER_SESSION_AGE_LIMIT_SECONDS=<seconds>` in `config.env` (this override mechanism already existed; only the *default* changed). This is a deliberate, minimal config-default change, not a removal of the mechanism — an operator who wants a periodic recycle for unrelated reasons (log rotation, memory hygiene, etc.) can still turn it on.

Comments in `check_session_age()` and around `SESSION_AGE_LIMIT_SECONDS` were reworded to stop asserting the debunked "hard 7440s limit" as fact and instead explain the disabled-by-default rationale, referencing #2157 and the 90+ hour counter-evidence. (`WFM_SUPPRESSION_MAX_SECONDS`, which derives from `SESSION_AGE_LIMIT_SECONDS`, already had a documented, tested fallback path for `SESSION_AGE_LIMIT_SECONDS=0` — see `compute_wfm_suppression_max_seconds()` — so this default flip does not require any change to that derivation, it just makes the existing fallback path (21600s/6h) the default one.)

This PR does not attempt the broader comment-accuracy sweep issue #2157 already tracks for other files (`inject-bootup-context.py`, test docstrings, etc.) — only the comments directly adjacent to the constant this PR changes were reworded.

## Functional patterns

Both parts are small, self-contained, side-effect-isolated changes. Part 1 mirrors an existing, already-correct sibling implementation (`claude-persistent.sh`) rather than introducing new logic shapes. Part 2 is a pure config-default change — `compute_wfm_suppression_max_seconds()` remains a pure function with no new branches; the new test (`test-health-check-session-age-default.sh`) exercises the actual production config-resolution code via `sed` extraction (same technique the pre-existing `test-health-check-session-age.sh` uses for `check_session_age()`) rather than reimplementing the logic under test.

## Out of scope

- No change to `check_session_age()`'s core logic, `DISPATCHER_PID_FILE` reader logic, or `do_restart()` — untouched.
- No change to `claude-persistent.sh` — it already wrote/cleaned up `dispatcher.pid` correctly.
- Does not address why `LOBSTER_DEBUG` drifted to `true` in the first place (separate, already-flagged config-drift issue).
- Does not perform issue #2157's full repo-wide comment-accuracy sweep — only comments directly touched by this diff were reworded.
- No migration entry added to `scripts/upgrade.sh` — this changes existing script/config-default behavior only (no new paths, config keys, cron entries, or service files), so a normal `lobster update` picks it up like any other source change. Existing installs that already have `LOBSTER_SESSION_AGE_LIMIT_SECONDS` set explicitly in `config.env` are unaffected either way.

## Tests run

- [x] New regression suite `tests/unit/test_hooks/test_claude_wrapper_exp_dispatcher_pid.py` — written first (TDD), confirmed failing against the pre-fix script, then passing after the fix: `uv run pytest tests/unit/test_hooks/test_claude_wrapper_exp_dispatcher_pid.py -v` — 5 passed.
- [x] Existing sibling suite `tests/unit/test_hooks/test_claude_wrapper_exp_spawn_pid.py` — 5 passed, no regression to the startup-flag write path.
- [x] Existing bash suite `tests/test-health-check-session-age.sh` (exercises `check_session_age()` in isolation, unaffected by the default-value change since it sets its own `SESSION_AGE_LIMIT_SECONDS` test constant): `bash tests/test-health-check-session-age.sh` — 14 passed, 0 failed.
- [x] Existing bash suite `tests/test-health-check-dispatcher-heartbeat.sh` (exercises `compute_wfm_suppression_max_seconds()`, including its explicit `SESSION_AGE_LIMIT_SECONDS=0` fallback path): `bash tests/test-health-check-dispatcher-heartbeat.sh` — 21 passed, 0 failed.
- [x] New bash suite `tests/test-health-check-session-age-default.sh` (exercises the actual config-resolution block that computes `SESSION_AGE_LIMIT_SECONDS` from env/config.env, extracted verbatim from the production script): `bash tests/test-health-check-session-age-default.sh` — 4 passed, 0 failed. Confirms: no override → 0 (disabled); env var opt-in respected; config.env opt-in respected; env var takes precedence over config.env.
- [x] `uv run pytest tests/unit/test_hooks/test_claude_wrapper_exp_dispatcher_pid.py tests/unit/test_hooks/test_dispatcher_session_start.py -q` — 13 passed.
- [x] Full `tests/unit/test_hooks/` directory (part 1's original run) — 1049 passed, 8 skipped, 11 failed; confirmed via `git stash` that the same 11 failures occur identically on unmodified `main` (pre-existing, unrelated to this PR).

## Manual test

Part 1 (`claude-wrapper.exp`): ran the actual script end-to-end against a stubbed `claude` binary with isolated `HOME`/`LOBSTER_MESSAGES`/`LOBSTER_WORKSPACE` pointed at a scratch tmp directory — no production paths touched. Observed `dispatcher.pid` created with the correct PID while the stub ran, and removed after it exited.

Part 2 (config default): no additional manual/live test — this is a pure default-value + comment change to already-tested shell logic, verified via the new and existing automated test suites above. No systemd/cron/external-service surface changed.

## Definition of Done
- [x] Integration/manual test run (part 1) — see Manual test above
- [x] User-visible outcome: not applicable — internal launcher/health-check logic, no Telegram/UI-facing behavior change
- [x] Side-effect audit complete: no floods, no unscoped writes; part 2 only narrows when a SIGTERM restart fires (default now never, unless opted in) — strictly reduces disruptive side effects versus the pre-PR-#2197 state

### DIFF
--- a/scripts/claude-wrapper.exp
+++ b/scripts/claude-wrapper.exp
@@ -98,6 +98,32 @@ catch {
     close $fh
 }
 
+# Write the dispatcher.pid file, mirroring scripts/claude-persistent.sh
+# (which writes $MESSAGES_DIR/config/dispatcher.pid right before exec-ing
+# claude). health-check-v3.sh's check_session_age() reads this file to find
+# the PID to send a graceful SIGTERM to once the session passes
+# SESSION_AGE_LIMIT_SECONDS.
+#
+# Before this fix, claude-wrapper.exp never wrote dispatcher.pid, so the
+# graceful SESSION AGE LIMIT restart path could never find a PID while
+# running in debug mode (LOBSTER_DEBUG=true) — it silently logged
+# "no dispatcher.pid — cannot send SIGTERM" and took no action, every
+# 4-minute health-check tick, for as long as the session stayed over-age
+# (see issue #2196: 4,148 occurrences in health-check.log since 2026-06-07).
+#
+# Same LOBSTER_MESSAGES fallback as claude-persistent.sh/health-check-v3.sh
+# (MESSAGES_DIR="${LOBSTER_MESSAGES:-$HOME/messages}") — must stay in sync
+# with both so the writer and reader always agree on the path.
+set messages_dir [expr {[info exists ::env(LOBSTER_MESSAGES)] ? $::env(LOBSTER_MESSAGES) : "$::env(HOME)/messages"}]
+set dispatcher_pid_dir [file join $messages_dir "config"]
+set dispatcher_pid_file [file join $dispatcher_pid_dir "dispatcher.pid"]
+catch { file mkdir $dispatcher_pid_dir }
+catch {
+    set pid_fh [open $dispatcher_pid_file w]
+    puts $pid_fh [exp_pid]
+    close $pid_fh
+}
+
 # Wait for either the permissions dialog OR the ready prompt.
 #
 # The ❯ (U+276F) character appears in the prompt area surrounded by ANSI escape
@@ -152,3 +178,11 @@ send "\r"
 # Hand over terminal control — Claude runs indefinitely.
 # The expect process stays alive so the spawned claude process is not orphaned.
 interact
+
+# interact returns once the spawned claude process exits (or EOF). Clean up
+# dispatcher.pid now that the PID it names is gone — mirrors
+# claude-persistent.sh's `rm -f "$dispatcher_pid_file"` after its claude
+# process exits. Leaving a stale dispatcher.pid around risks a future,
+# unrelated process being assigned the same PID and incorrectly targeted by
+# health-check-v3.sh's check_session_age() SIGTERM.
+catch { file delete $dispatcher_pid_file }
--- a/scripts/health-check-v3.sh
+++ b/scripts/health-check-v3.sh
@@ -115,14 +115,28 @@ HEARTBEAT_FILE="$WORKSPACE_DIR/logs/claude-heartbeat"   # legacy WFM-touch signa
 DISPATCHER_HEARTBEAT_FILE="${LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE:-$WORKSPACE_DIR/logs/dispatcher-heartbeat}"
 DISPATCHER_HEARTBEAT_STALE_SECONDS=1200   # 20 min — covers compaction (~5m) + catchup (~12m) + margin
 
-# Session age limit: CC enforces a hard 7440s session lifetime (issue #2059).
-# At this limit, CC kills the process without firing the Stop hook — no tombstone,
-# no compaction alert. We trigger a graceful SIGTERM at SESSION_AGE_LIMIT_SECONDS
-# (7200s = 2 min before hard limit) so the Stop hook fires cleanly.
+# Session age limit: proactive SIGTERM restart mechanism (issue #2059).
+#
+# DISABLED BY DEFAULT (issue #2196 / #2157). This mechanism was originally
+# justified by a claimed "CC enforces a hard 7440s session lifetime" — but
+# issue #2157 found that claim rested on only 2 data points, was never
+# confirmed against Anthropic docs, and had a more likely alternative
+# explanation (an unrelated heartbeat-staleness bug, since fixed by #2074).
+# The live dispatcher session has since run 90+ hours with zero problems,
+# directly contradicting the original 7440s premise. There is no remaining
+# evidence that Claude Code kills sessions on a wall-clock timer.
+#
+# The mechanism itself still works and is harmless once opted into: when
+# enabled, it sends a graceful SIGTERM to the dispatcher PID at
+# SESSION_AGE_LIMIT_SECONDS so the Stop hook fires cleanly (vs. an
+# uncontrolled kill from some other layer, e.g. a systemd bounce). Keep it
+# available as an opt-in operational precaution — e.g. an operator who wants
+# a periodic recycle for unrelated reasons (log rotation, memory hygiene) —
+# but do not force it on new/updated installs by default.
+#
 # Written by inject-bootup-context.py as a plain Unix epoch integer.
-# Set LOBSTER_SESSION_AGE_LIMIT_SECONDS=0 in config.env to disable this check
-# (safe if CC no longer enforces the 7440s hard limit in the installed version).
-SESSION_AGE_LIMIT_SECONDS="${LOBSTER_SESSION_AGE_LIMIT_SECONDS:-7200}"
+# Set LOBSTER_SESSION_AGE_LIMIT_SECONDS=<seconds> in config.env to opt in.
+SESSION_AGE_LIMIT_SECONDS="${LOBSTER_SESSION_AGE_LIMIT_SECONDS:-0}"
 DISPATCHER_SESSION_START_FILE="${LOBSTER_DISPATCHER_SESSION_START_FILE_OVERRIDE:-$WORKSPACE_DIR/data/dispatcher-session-start.ts}"
 
 # WFM-active signal (issue #1713 / #949): inbox_server.py writes this file with
@@ -164,7 +178,7 @@ WFM_ACTIVE_STALE_SECONDS=180   # 3x WAIT_HEARTBEAT_INTERVAL (60s) — absorbs on
 # tighter than the session lifetime a legitimate idle wait is allowed to reach.
 # See compute_wfm_suppression_max_seconds() below.
 WFM_SUPPRESSION_MARGIN_SECONDS=300      # 5 min: fire slightly before the session-age backstop would anyway
-WFM_SUPPRESSION_FALLBACK_SECONDS=21600  # 6h: used only when SESSION_AGE_LIMIT_SECONDS=0 (session-age check disabled)
+WFM_SUPPRESSION_FALLBACK_SECONDS=21600  # 6h: used when SESSION_AGE_LIMIT_SECONDS=0 (the default as of issue #2196 — session-age check disabled)
 
 # Pure function: derive the WFM-active suppression cap from the session-age
 # limit. No side effects, no globals read — everything comes in as an argument,
@@ -240,7 +254,9 @@ fi
 LOBSTER_ENV="${LOBSTER_ENV:-production}"
 
 # Read LOBSTER_SESSION_AGE_LIMIT_SECONDS from config.env (if not already in environment).
-# Set to 0 in config.env to disable the proactive session-age restart check.
+# Default is 0 (disabled — see the comment near the initial assignment above,
+# issue #2196 / #2157). Set a positive integer in config.env to opt in to the
+# proactive session-age restart.
 if [[ -z "${LOBSTER_SESSION_AGE_LIMIT_SECONDS:-}" && -f "$CONFIG_ENV" ]]; then
     _cfg_sal=$(grep '^LOBSTER_SESSION_AGE_LIMIT_SECONDS=' "$CONFIG_ENV" 2>/dev/null | cut -d'=' -f2- | tr -d '[:space:]"' || true)
     if [[ -n "$_cfg_sal" ]]; then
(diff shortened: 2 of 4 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: Verified independently, no blockers.

**What I checked myself (not just the engineer's claims):**

- **Tcl ternary safety**: The `expr {cond ? $VAR : "default"}` pattern for the LOBSTER_MESSAGES/HOME fallback looked risky to me at first (Tcl `expr` can re-parse substituted string values as arithmetic — e.g. a path fragment that happens to look like `5-3`). I tested this directly with `expect -c` against edge-case values (paths with spaces, paths that look like arithmetic). It's safe: Tcl's substitution-then-parse behavior for `$var` inside braced `expr` does not re-evaluate the substituted value as a sub-expression. This also matches the existing, already-shipped `dispatcher-startup-flag` write earlier in the same file, so there's real precedent.
- **PID semantics parity**: Confirmed `claude-persistent.sh` writes `$BASHPID` right before `exec claude` (so the written PID becomes the actual claude process's PID after exec-replace), and `claude-wrapper.exp` now writes `[exp_pid]`, which is the PID of the spawned `claude` child. Both name the real `claude` process PID, so `health-check-v3.sh`'s SIGTERM targeting is consistent across debug and persistent modes — not a mismatch.
- **Test counts**: Ran `uv run pytest tests/unit/test_hooks/ --tb=no -q` myself on the PR branch: 11 failed, 1049 passed, 8 skipped. Then checked out unmodified `origin/main` in a clean worktree and ran the same suite: 11 failed, 1044 passed, 8 skipped — identical failing test names (test_context_monitor, test_on_compact_write_claude_session_id x7, test_pii_scan_guard x3), all unrelated to this change. Claim confirmed independently, not just via the engineer's git-stash check.
- Also ran the new test file (5/5 pass), the sibling `test_claude_wrapper_exp_spawn_pid.py` (5/5 pass, unaffected), and the bash reader-side suite `tests/test-health-check-session-age.sh` (14/14 pass).

**One test-quality issue found (not blocking, but worth fixing):**

`test_dispatcher_pid_cleaned_up_on_exit` is meant to confirm the cleanup (`file delete $dispatcher_pid_file`) happens *after* `interact` returns. Its assertion is:
```python
assert "file delete" in self.content and "dispatcher.pid" in self.content[interact_idx:] + self.content
```
The second half (`self.content[interact_idx:] + self.content`) always contains "dispatcher.pid" regardless of position, because it appends the *entire* file content unconditionally — so this check is effectively "does 'file delete' appear anywhere, and does 'dispatcher.pid' appear anywhere," not "does cleanup happen after interact." I verified this concretely: I patched the script to move the `file delete $dispatcher_pid_file` line to right after `spawn` (i.e. deleting the PID file immediately, before the process even starts — a real regression that would defeat the whole fix) and the test still passed. Suggest tightening to something like `"dispatcher.pid" in self.content[interact_idx:]` alone, or asserting the delete statement's own index is greater than `interact_idx`.

**One narrow pre-existing gap, not introduced by this PR:** if the initial 60s `expect { ... timeout { exit 1 } }` wait for the Claude prompt times out, the script exits before ever reaching `interact` — so the cleanup line never runs, leaving a stale (but real, since spawn succeeded) `dispatcher.pid`. The write happens before this wait block, the cleanup only after `interact`. This mirrors an identical pre-existing gap for `dispatcher-startup-flag` and is out of scope for this fix, just flagging as a residual edge case for future hardening.

Everything else — path construction, LOBSTER_MESSAGES/HOME fallback matching claude-persistent.sh exactly, write placement immediately after spawn, cleanup placement after interact in the real (unpatched) script — checked out correctly by direct reading of the fetched script content.

Issue #2219, opened 2026-08-12
Migrate mcp__obsidian__* tool calls off flaky stdio obsidian-mcp onto vault-api (follow-up to #2119)


aeschylus:
## Background

Part of the #2119 fix plan (item 2, "higher-value, longer-term" per the original triage comment, which explicitly allowed deferring this piece to a follow-up if it didn't fit in one pass: https://github.com/SiderealPress/lobster/issues/2119#issuecomment-5262200167).

Items 1, 3, 4, 5, and 6 of that plan landed in #2119 directly (background-delegation enforcement, process-group kill fix + orphan reaper, reopen, read-back verification, Syncthing `.git` exclusion). This issue tracks the remaining, larger piece: retiring the stdio `obsidian-mcp` server itself.

## Problem

`mcp__obsidian__*` tool calls are backed by a stdio MCP server (`npx -y obsidian-mcp <vault-path>`, registered in `~/.claude.json` → `projects["/home/lobster/lobster-workspace"].mcpServers.obsidian`) with **no application-level timeout**. A hung call only ever gets aborted by Claude Code's client-side idle timeout (1800s) or by the ~2h04m session-age restart — both far too slow to fail gracefully, and in the #2119 incident, the hang froze the whole dispatcher rather than just the tool call.

`vault-api` (PM2, port 8082, Express REST HTTP, confirmed healthy: 14-day uptime, 0 restarts at time of investigation) fronts the same vault directory and would give real HTTP-level timeouts instead of an indefinite stdio hang.

## Why this is scoped separately from #2119

Investigated during #2119: `vault-api`'s current surface (`~/lobster-workspace/projects/vault-api/server.js`) only implements 4 endpoints — `GET /files` (tree listing), `GET /file` (read), `POST /file` (write), `GET /health`. The `mcp__obsidian__*` toolset the dispatcher actually calls has 11 tools: `create-note`, `edit-note`, `read-note`, `delete-note`, `move-note`, `search-vault`, `list-available-vaults`, `add-tags`, `remove-tags`, `rename-tag`, `create-directory`.

A real migration needs, at minimum:
1. Extend `vault-api` with the missing operations (search, tags, delete, move, multi-vault listing, directory creation) — new endpoints, new tests, new manual verification against the live vault for each.
[...]

PR #2220, opened 2026-08-12
fix: obsidian MCP hang froze whole dispatcher, orphaned children, never verified writes

### LINKED ISSUE #2119: bug: obsidian MCP tool calls hang for full 1800s idle-timeout despite server process staying alive
## What happened

A subagent reported 3 separate `mcp__obsidian__*` calls (`list-available-vaults`, `search-vault`, `create-note`) that each hung for the full 1800s client-side idle timeout and were auto-aborted by Claude Code, even though the underlying `obsidian-mcp` node process (stdio child of the `claude` process, e.g. PID 655095 pointed at `/home/lobster/LobsterDrop/eloso-obsidian-vault`) was alive the whole time. The subagent worked around it via direct filesystem writes + the vault's own sync script.

## Evidence

Log: `~/.cache/claude-cli-nodejs/-home-lobster-lobster-workspace/mcp-logs-obsidian/2026-07-17T22-12-22-462Z.jsonl`

```
{"debug":"Calling MCP tool: list-available-vaults", ... "22:18:32.058Z"}
... 30s heartbeats for 1800s ...
{"debug":"Tool 'list-available-vaults' aborting: no response or progress notification for 1800s (idle timeout 1800s)"}
{"debug":"Tool 'list-available-vaults' failed after 1800s: MCP server \"obsidian\" tool \"list-available-vaults\" sent no response or progress for 1800s; aborting. ..."}
```

The same pattern repeated for `search-vault` and `create-note` in the same log file — three consecutive tool calls, each hanging the full 1800s before Claude's own idle-timeout killed the wait (not the process).

An earlier log from the same day (`2026-07-17T18-06-37-166Z.jsonl`) shows the identical failure mode for `list-available-vaults` on a different connection instance, so this is not a one-off — it recurred at least twice in one day.

## Investigation

1. Checked the live process at the time of the report: `node .../obsidian-mcp ...` (PID 655095) was in state `S` (sleeping on `ep_poll`), 0% CPU, no open file locks on the vault, no zombie/orphan state — i.e. not spinning or deadlocked in a visible way, just idle.
2. Killed the hung process tree (`kill -TERM` on the npm-exec/sh/node chain). It terminated cleanly; Claude Code is expected to respawn a fresh stdio connection on the next `mcp__obsidian__*` call (this is a stdio-transport MCP server registered directly in the top-level Claude config — not a systemd/pm2 service — so "restart" means killing the child and letting the client reconnect lazily).
3. **Could not reproduce in isolation.** Spawned a bare `npx -y obsidian-mcp /home/lobster/LobsterDrop/eloso-obsidian-vault` process directly and drove it over stdio with raw JSON-RPC (`initialize` → `list-available-vaults` → `search-vault` → `create-note` → `list-available-vaults` again, including a delay before the first call to mimic the real session's ~6min gap between connect and first call). Every call returned in under 0.1s across two separate test runs.

This suggests the `obsidian-mcp` npm package itself is not inherently prone to hanging on a cold, isolated stdio connection — the hang is specific to something about the *live* session's MCP client/transport handling. Plausible candidates (not yet confirmed):
- Multiple concurrent background subagents sharing the same single stdio MCP connection, with a request/response demux issue when calls overlap.
- Some stdout buffering/framing edge case triggered only after the connection has sat idle for several minutes before the first call.

## Related but distinct issues

- #1439 (dispatcher hangs silently when MCP server *restarts* mid-session) and #1147 (dispatcher hangs on dead MCP connection when inbox_server restarts) both cover connection-level failures where the *server process itself dies/restarts*. This issue is different: the obsidian server process never died or restarted — an individual tool call just never got a response, and the process was left idle afterward. Worth cross-referencing but not a duplicate.

## Suggested follow-up

- Add logging/telemetry on the obsidian-mcp side (or a lightweight wrapper) to see what a tool call is doing server-side when it hangs, since our client-side evidence shows it never even starts executing according to `ep_poll`/CPU state.
- If this recurs, capture `strace -p <pid>` at the moment of a hang (before killing it) to see what syscall it's blocked in.
- Consider whether concurrent subagents should get isolated obsidian-mcp connections rather than sharing one, if that's in fact happening.

aeschylus:
Closes #2119. Related follow-up (deferred, scoped separately): #2219.

## Problem

Wallace told a user "saving it to the Obsidian vault now... will send the link shortly," then went silent for hours. Nothing was ever written. Root cause, confirmed via two independent evidence passes (`~/.cache/claude-cli-nodejs/.../mcp-logs-obsidian/*.jsonl`, production `ps`/`systemctl` state):

1. **The obsidian tool call ran inline on the main dispatcher thread**, in violation of this codebase's own "ANY file read/write → background subagent" rule. `mcp__obsidian__*` calls are backed by a stdio MCP server (`obsidian-mcp`) with no application-level timeout. When the call hung (confirmed hangs of 2304s and 6462s), it froze the *entire* dispatcher loop, not just the one task — this is why the whole conversation went silent instead of Wallace being free to say "still working on it."
2. **The hang was only ever aborted by session-age recycling** (~2h04m SIGTERM), which meant the "I'll get back to you" promise died with the killed session — the new session has no memory of it.
3. **Orphaned process leak**: `health-check-v3.sh`'s session-age SIGTERM signals only the dispatcher's own PID, never its stdio MCP children. 12 orphaned `obsidian-mcp` processes were found live in production at investigation time, one per restart since Aug 11 06:00 — confirmed still present as of this PR (see Manual test below).
4. **No read-back verification**: even when a write *did* land, nothing confirmed it before reporting success — so a false "saved" claim and a genuinely-failed save were indistinguishable to the user.

## What changed

- **`.claude/sys.dispatcher.bootup.md`**: `mcp__obsidian__*` is now an explicit named example under the background-delegation rule (previously only the generic "file read/write" bullet existed, and it wasn't followed for this tool family). Adds a full delegation template — acknowledge → `Task(run_in_background=true)` → the subagent must call `search-vault`/`read-note` to confirm the write landed *before* calling `write_result` with a success message. This is instructional (LLM-followed) guidance, mirroring the existing link-capture pattern in `lobster-shop/obsidian-km/`, not a runtime code change — the "test that fails on revert" for this piece is a content assertion against the instruction file itself (see `tests/unit/test_hooks/test_obsidian_mcp_delegation_instructions.py`).
- **`scripts/health-check-v3.sh`**: `check_session_age()` now snapshots the dispatcher's direct child PIDs *before* sending SIGTERM (querying after is a race — children are reparented away the instant the parent process exits) and signals them too via a new `kill_dispatcher_children()` (SIGTERM, then SIGKILL after a grace period for anything still alive). Adds `reap_orphaned_mcp_processes()`, a backstop that runs every health-check cron cycle (independent of any restart happening right now) and reaps any `obsidian-mcp`-pattern process that's both reparented to init (`PPID=1`) and older than one restart cycle plus margin.
- **`scripts/claude-persistent.sh`**: fixes a second bug found while implementing the above. `kill_orphaned_mcp_processes()`'s "skip systemd-managed services" guard used `systemctl status $pid`, which doesn't ask "is this PID itself a service" — it resolves the PID's *cgroup* to whichever unit owns it and reports that unit's status. Every process under `lobster-claude.service`'s tmux session (dispatcher, every MCP child, every orphan that kept its original cgroup membership) resolved to `lobster-claude.service` and returned exit 0, so this guard has been silently protecting every single orphaned MCP child from cleanup since it was introduced. Now compares against the protected unit's actual `MainPID` (`systemctl show -p MainPID --value`) instead.
- **Syncthing `.stignore`** (`/home/lobster/LobsterDrop/.stignore`, not part of this repo but applied live as part of this fix — see Manual test): the vault's `.git/` was being synced by Syncthing with no ignore rule, producing 10 `index.sync-conflict-*` files from 2026-07-02 through 2026-08-11, some concurrent with the incident window. Excluded `.git` from sync on this device; git already has its own sync mechanism (push/pull) and doesn't need a second one racing it.

## Out of scope

Item 2 of the original fix plan — retiring the stdio `obsidian-mcp` server entirely in favor of the already-healthy `vault-api` HTTP service — is **not** in this PR. Investigated during this work: `vault-api` currently implements 4 of the 11 `mcp__obsidian__*` operations (no search, tags, delete, or move). A real migration is a multi-day scope with its own manual+automated verification needs, not a same-pass addition. Tracked as #2219, with the mitigations in this PR (background delegation + read-back verification + orphan cleanup) closing the actual user-facing failure mode in the meantime. This was an explicitly allowed deferral per the original issue triage comment.

No changes to `vault-api` itself, no changes to the obsidian-mcp npm package, no changes to any other health-check subsystem (service checks, WFM suppression, etc.) beyond the session-age block described above.

## Functional patterns

`kill_dispatcher_children()` and `reap_orphaned_mcp_processes()` take pre-enumerated PID lists / read live state independently rather than mutating shared globals, keeping "what to signal" separate from "how to signal it" and making each independently testable without needing a real dispatcher process. Both `health-check-v3.sh` and `claude-persistent.sh` gain a `BASH_SOURCE[0] == $0` guard around their `main()` call so test harnesses can `source` them and exercise individual functions in isolation, with zero behavior change on the real cron/systemd invocation path.

## Tests run

- [x] `bash tests/test-health-check-session-age.sh` — 17 passed, 0 failed (includes 3 new tests for issue #2119: child dies with the session, no-op on empty child list, and a reproduction of the pre-fix leak via a stubbed no-op)
- [x] `bash tests/test-health-check-orphan-mcp-reaper.sh` — 4 passed, 0 failed
- [x] `bash tests/test-claude-persistent-orphan-mcp-cleanup.sh` — 3 passed, 0 failed (runs against the real `systemctl`/cgroup condition that caused the bug, using only throwaway fake processes — no real MCP process is ever targeted)
- [x] `python3 -m pytest tests/unit/test_hooks/test_obsidian_mcp_delegation_instructions.py` — 5 passed, 0 failed
- [x] `bash -n scripts/health-check-v3.sh && bash -n scripts/claude-persistent.sh` — syntax OK

**Falsifiability**: reverted each implementation file (`git stash push -- <file>`) and re-ran its corresponding test suite. `health-check-v3.sh` revert → `FATAL: kill_dispatcher_children() not found` / `FATAL: reap_orphaned_mcp_processes() not found`. `claude-persistent.sh` revert → the orphan-cleanup test's fake orphan process is left running (systemd-cgroup guard bug reproduced, matching the exact production condition this fix targets). Restored and re-ran clean afterward — no test suite was left in a broken state.

## Manual test

- Confirmed the orphan leak is real and currently present in production, independent of this PR: `ps aux | grep obsidian-mcp` shows 12 processes, all `PPID=1`, matching the issue's original count.
- Confirmed the Syncthing fix live: added `.stignore` with `(?d).git`, triggered a rescan via the Syncthing REST API (`POST /rest/db/scan?folder=lobsterdrop`), and confirmed via `GET /rest/db/ignores?folder=lobsterdrop` that the pattern is active and expands to `(?d).git`, `(?d)**/.git`, `(?d).git/**`, `(?d)**/.git/**`. This only affects this device's local ignore rules — other Syncthing peers would need the same applied locally to fully close the loop, which is a config note rather than a code change.
- Did **not** yet restart the live `lobster-claude`/`lobster-mcp` services with this code — that's the next step post-merge (see plan below), done via the documented safe-restart pattern with an inbox warning first.

## Definition of Done

- [x] Unit tests pass with proper mocking (no production hits in automated tests — all fake processes match unique random markers, isolated from real `obsidian-mcp`/`inbox_server.py`)
- [x] Integration/manual test run — see Manual test above (live orphan count confirmed, Syncthing fix verified live via REST API)
- [ ] User-visible outcome verified — pending live deploy + post-restart verification (planned immediately after merge)
- [x] Side-effect audit complete: `.stignore` change only stops future sync of `.git` (no deletion, no history rewrite); orphan-reaper only ever targets processes already reparented to init and past an age threshold, never anything belonging to a live session; falsifiability testing confirmed zero impact on the real `lobster-mcp.service` MainPID or any live `obsidian-mcp` process
- [x] External service calls: none in automated tests; the one live API call made (Syncthing REST, localhost, read + scan-trigger only) was scoped to a single folder and is documented above

### DIFF
--- a/.claude/sys.dispatcher.bootup.md
+++ b/.claude/sys.dispatcher.bootup.md
@@ -146,7 +146,7 @@ Never pass `hibernate_on_timeout=True` — feature removed in issue #1442; cause
 - Read images (the one documented carve-out — claim first with `mark_processing`)
 
 **What ALWAYS goes to a background subagent (`run_in_background=true`):**
-- ANY file read/write (except images)
+- ANY file read/write (except images) — this explicitly includes **every `mcp__obsidian__*` tool call** (`create-note`, `edit-note`, `read-note`, `search-vault`, `list-available-vaults`, `move-note`, `delete-note`, `add-tags`, `remove-tags`, `rename-tag`, `create-directory`). These are stdio MCP calls backed by `npx -y obsidian-mcp <vault-path>` with **no application-level timeout** — a hang blocks whatever thread made the call until the ~2h04m session-age SIGTERM kills the entire session (see issue #2119). Never call an `mcp__obsidian__*` tool inline on the main dispatcher thread, under any circumstance.
 - ANY git operation
 - ANY GitHub API call
 - ANY web fetch or research
@@ -160,7 +160,57 @@ Never pass `hibernate_on_timeout=True` — feature removed in issue #1442; cause
 Read("${LOBSTER_INSTALL_DIR:-~/lobster}/.claude/sys.dispatcher.bootup.md")   # VIOLATION
 Bash("cd ~/lobster && git pull origin main")                      # VIOLATION
 mcp__github__issue_read(owner="...", repo="...", ...)             # VIOLATION
+mcp__obsidian__list-available-vaults()                             # VIOLATION (issue #2119 — hung 2304s inline, froze the whole dispatcher loop)
 ```
+
+### Obsidian vault writes: delegate, then verify before claiming success (issue #2119)
+
+Mirror the link-capture delegation pattern used by `lobster-shop/obsidian-km/context/obsidian-km.md`
+for any Obsidian save request (a note, a redline, a meeting summary, anything the user asks you
+to "save to the vault" or "add to Obsidian"):
+
+```
+1. Acknowledge immediately, without promising a specific outcome yet:
+   send_reply(chat_id, "On it — saving that to the vault now.", message_id=message_id)
+
+2. Delegate the actual Obsidian tool calls to a background subagent:
+   Task(
+       prompt="""
+       ---
+       task_id: <task_id>
+       chat_id: <chat_id>
+       source: <source>
+       background: true
+       ---
+
+       Save this to the Obsidian vault: <content/details>.
+
+       Steps:
+       1. Call mcp__obsidian__create-note (or edit-note) to write the note.
+       2. MANDATORY — do not skip: call mcp__obsidian__search-vault (or
+          read-note on the exact path you just wrote) to confirm the note
+          actually exists in the vault. Never report success before this
+          read-back confirms the write landed — a stdio call that appears to
+          return can still have written nothing if the tool hung and was
+          later aborted by session recycling (issue #2119).
+       3. Only after the read-back confirms the note exists, call write_result
+          with a success message (include the vault path/link).
+          If the read-back fails or the note is missing, call write_result
+          with status="error" and say plainly that the save did not complete
+          — do not claim success.
+       """,
+       subagent_type="general-purpose",
+       run_in_background=true,
+   )
+
+3. mark_processed(message_id)
+4. Return to wait_for_messages() immediately — do not wait on the subagent.
+```
+
+**Never say "saving it now" and then call an `mcp__obsidian__*` tool in the same turn on the main
+thread.** The promise and the write must be separated by a background subagent boundary. If the
+subagent's call hangs, the dispatcher keeps processing other messages — it does not go silent for
+hours the way this incident did.
 
 **Code internals questions:** delegate to a subagent to read the actual code — never speculate from memory.
 
--- a/scripts/claude-persistent.sh
+++ b/scripts/claude-persistent.sh
@@ -388,10 +388,30 @@ kill_orphaned_mcp_processes() {
         # itself, but the service bounce still wiped in-memory session/lock state,
         # producing spurious session-lost-reminder / dispatcher-marked-dead
         # notifications with no actual dispatcher crash. See investigation for
-        # issues #2124 / #2142. The correct invocation takes the PID as a bare
-        # positional argument.
-        if systemctl status "$pid" >/dev/null 2>&1; then
-            log "CLEANUP: MCP PID $pid is a systemd-managed service — skipping"
+        # issues #2124 / #2142.
+        #
+        # #1551's fix ("pass the PID as a bare positional argument") replaced one
+        # bug with a subtler one (issue #2119): `systemctl status <pid>` does not
+        # ask "is this PID itself a systemd service" — it resolves the PID's
+        # cgroup to whichever unit owns that cgroup and reports THAT unit's
+        # status, exit 0, for *any* PID living anywhere inside it. Every process
+        # under lobster-claude.service's tmux session (the dispatcher, every MCP
+        # child it spawns, and any orphan reparented to PID 1 that still carries
+        # the same cgroup membership from when it was forked) resolves to
+        # lobster-claude.service and returns exit 0 — so this check has always
+        # treated every single MCP child, including hung/orphaned obsidian-mcp
+        # processes, as "systemd-managed" and skipped it. Confirmed empirically
+        # against live orphaned obsidian-mcp PIDs (PPID=1, one per session-age
+        # restart, accumulating indefinitely — see issue #2119).
+        #
+        # Fix: only the process that is actually the recorded MainPID of the
+        # protected unit should be skipped. Compare against MainPID directly
+        # instead of asking "does this PID's cgroup resolve to some unit".
+        local protected_service="${LOBSTER_MCP_SERVICE_NAME:-lobster-mcp.service}"
+        local protected_main_pid
+        protected_main_pid=$(systemctl show -p MainPID --value "$protected_service" 2>/dev/null || true)
+        if [[ -n "$protected_main_pid" && "$protected_main_pid" != "0" && "$pid" == "$protected_main_pid" ]]; then
+            log "CLEANUP: MCP PID $pid is the systemd-managed ${protected_service} MainPID — skipping"
             skipped=$((skipped + 1))
             continue
         fi
@@ -790,4 +810,10 @@ Claude persistent session initializing."
 trap 'log "Received SIGTERM, shutting down..."; write_state "stopped" "sigterm"; exit 0' SIGTERM
 trap 'log "Received SIGINT, shutting down..."; write_state "stopped" "sigint"; exit 0' SIGINT
 
-main "$@"
+# Only run main when executed directly. When sourced (e.g. by a test harness
+# exercising kill_orphaned_mcp_processes/kill_orphaned_claude_processes in
+# isolation), skip the auto-run so the caller controls exactly what runs.
+# No behavior change for the normal systemd/tmux launch path.
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    main "$@"
+fi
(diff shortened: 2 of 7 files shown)


aeschylus (commented):
🤖🦞 Lobster (reviewer): NEEDS-WORK (non-blocking, already live): Solid fix for the actual incident, but the "primary" orphan-prevention mechanism doesn't reach the process that actually hangs — verified against the live production process tree, not just the diff.

**Deployed tree matches PR**: `/home/lobster/lobster` HEAD is `bf1690e1`, identical to this PR's merge commit. No drift between what's live and what's in the PR.

**All 29 tests independently re-run and confirmed passing** (not just trusted from the PR description):
- `tests/test-health-check-session-age.sh` — 17/17 pass
- `tests/test-health-check-orphan-mcp-reaper.sh` — 4/4 pass
- `tests/test-claude-persistent-orphan-mcp-cleanup.sh` — 3/3 pass
- `tests/unit/test_hooks/test_obsidian_mcp_delegation_instructions.py` — 5/5 pass (pytest)

Total 29, matching the engineer's briefing. (Side note: the PR description/commit message itself says "26 tests across 4 suites (17 session-age, 4 orphan-reaper, 3 orphan-cleanup-guard, 5 delegation-instruction-content)" — 17+4+3+5=29, not 26. Cosmetic arithmetic slip in the writeup, not a real discrepancy.)

**Real finding — `kill_dispatcher_children()` doesn't reach the process that actually hangs:**

`check_session_age()` snapshots `pgrep -P "$dispatcher_pid"` (direct children only, one hop) and signals just those PIDs. I checked the real production process tree for a live obsidian-mcp instance:

```
claude (dispatcher, e.g. PID 257641)
 └─ npm exec obsidian-mcp ...      (PID 257711)  <- direct child, this is what pgrep -P catches
     └─ sh -c obsidian-mcp ...     (PID 257749)  <- grandchild, NOT reached
         └─ node .../obsidian-mcp  (PID 257750)  <- great-grandchild, NOT reached — this is the actual process that hangs
```

`npx`/`npm exec` spawns an intermediate `sh -c` which forks (not execs) into the real `node` process — three hops down from the dispatcher, not one. `kill_dispatcher_children()` SIGTERMs (then SIGKILLs) only the `npm exec` wrapper at PID 257711. Killing that wrapper does not cascade to its child `sh`/`node` — they don't die, they get reparented to PID 1, i.e. orphaned, exactly like before this fix. The code comment on `reap_orphaned_mcp_processes()` even says the primary fix "signals stdio MCP children ... at the moment of a session-age restart, before they can be orphaned" — that's not what happens for the actual hung leaf process; it still gets orphaned every restart, same as pre-fix, and now just waits for the cron backstop (up to ~2h15m later) to be reaped instead of accumulating forever.

Net effect: this is a real, meaningful improvement (bounded backlog via the backstop instead of unbounded growth — confirmed live: the backstop reaped 11 stale orphans in one shot in `health-check.log` at 06:08:01 today, ages 12471s–86833s), but it is not the "kill it before it's ever orphaned" fix the code comments and PR description claim. The gap exists because the automated test for this (`session_age_sigterm_also_kills_dispatcher_child` in `test-health-check-session-age.sh`) simulates a fake dispatcher whose child is a **direct** `sleep 600` — a one-hop topology — which doesn't reproduce the real three-hop `npm exec -> sh -c -> node` chain, so the test passes without exercising the actual production shape of the bug.

Suggested fast follow: walk the full descendant tree (e.g. loop `pgrep -P` recursively, or `ps --ppid` breadth-first) instead of one hop, or use `pkill -P` per level. Not urgent — the backstop keeps this bounded — but worth a follow-up issue so the "prevented at restart time" claim becomes true rather than aspirational.

**Everything else checks out:**
- `claude-persistent.sh`'s `systemctl show -p MainPID` fix (replacing the cgroup-matching `systemctl status $pid` bug) is correct and well-targeted — read the full function, the MainPID comparison only protects the actual `lobster-mcp.service` process, not every cgroup member. Test suite 3 runs against the real systemd/cgroup condition, not a mock, and that's a meaningfully stronger test than average.
- Syncthing fix verified live: `/home/lobster/LobsterDrop/.stignore` exists with `(?d).git` and matches the PR's description exactly. (Note: 20 old `*sync-conflict*` files are still present in the vault — expected, the fix only stops *future* syncing of `.git/`, it doesn't retroactively clean up existing conflict files, and the PR never claimed it would.)
- The `.claude/sys.dispatcher.bootup.md` instructional changes are reasonable and the "test that fails on revert" for an LLM-instruction change (content assertions against the markdown) is an honest, clearly-labeled proxy — the PR is explicit that it doesn't prove behavioral compliance, just that the instruction text is present.
- No argument-type mismatches spotted in the MCP/store calls touched by this diff (mostly bash + one content-assertion pytest file, no DB/MCP client method signatures changed).

Issue #2150, opened 2026-08-03
health-check-v3: WFM_SUPPRESSION_MAX_SECONDS cap causes false-positive restart storm during legitimately idle periods (regression from #2074)


sayhar:
## Summary

Between **2026-08-03 10:16 EDT and 12:12 EDT**, `health-check-v3.sh` restarted `lobster-claude.service` 6 times (10:16, 10:20, 11:12, 12:00, 12:04, 12:08), then correctly tripped its BLACK circuit breaker after 3 restarts in a 10-minute window and stopped retrying. The dispatcher was never actually frozen — the inbox was simply idle (no user messages) for stretches longer than 45 minutes. This produced a wall of "Lobster Starting" Telegram notifications with no underlying incident, and burned 6 unnecessary process restarts. User-reported via screenshot referencing a bot-generated alert for issue #2149 — **that issue number does not exist in this repo** (highest at time of writing is #2148), so the alert text apparently never got filed; this issue is the actual write-up.

## Root cause

This is a false-positive regression introduced by the fix for #2074 ("WFM-active mask cannot detect frozen dispatcher — caused 2-day outage").

#2074's fix added a hard cap in `check_dispatcher_heartbeat()` (`scripts/health-check-v3.sh`, ~line 990-1052):

```
WFM_SUPPRESSION_MAX_SECONDS=2700   # 45 min: stop suppressing beyond this heartbeat-stale age
```

Logic: `DISPATCHER_HEARTBEAT_FILE` (written by `hooks/thinking-heartbeat.py` on PostToolUse) only refreshes when a tool call *completes*. While the dispatcher is legitimately blocked inside a single `wait_for_messages()` call waiting for a message that hasn't arrived yet, PostToolUse never fires, so the main heartbeat goes stale for the entire duration of the wait — this is expected, not a symptom of freezing. `DISPATCHER_WFM_ACTIVE_FILE` is refreshed independently every ~60s by a daemon thread specifically to cover this case (main heartbeat stale + WFM-active fresh = alive-but-idle, suppress RED).

#2074 correctly pointed out that WFM-active-fresh alone isn't proof of liveness (a frozen main loop's daemon thread can keep ticking). Its fix was to cap the suppression window at 2700s: past that, RED fires "regardless of WFM-active freshness," on the theory that after 45 minutes either the watchdog would have already caught a real freeze, or the session-age proactive restart (7200s) will handle it soon anyway.

[...]

PR #2228, opened 2026-08-13
fix: reset dispatcher-heartbeat on every new dispatcher session start (#2150)

sayhar:
## Summary

Neither dispatcher restart path reset the `dispatcher-heartbeat` sentinel file (`~/lobster-workspace/logs/dispatcher-heartbeat`) when spawning a replacement session:

1. **Graceful SIGTERM path** — `check_session_age()` in `health-check-v3.sh`, fires at `SESSION_AGE_LIMIT_SECONDS` (7200s / 2h)
2. **Hard restart path** — `do_restart()` in `health-check-v3.sh`, fires on RED heartbeat staleness

Because the RED cap (`WFM_SUPPRESSION_MAX_SECONDS`, derived as `SESSION_AGE_LIMIT_SECONDS - 300s` = 6900s / 115min) sits only 5 minutes inside the 2h session-age limit, a brand-new dispatcher session could inherit an already near-stale heartbeat file from its predecessor and re-breach the RED cap within the very next 4-minute health-check cycle — causing a second, spurious restart shortly after a legitimate one. Most visible on quiet/idle days with few early tool calls to naturally refresh the heartbeat.

First diagnosed 2026-08-03/04, reconfirmed by multiple audits since. Relates to the RED-cap / WFM-suppression design discussed in issue #2150.

## Fix

Resets the heartbeat at the single choke point both restart paths already funnel through: `claude-persistent.sh`'s launcher writes a startup PID flag before every new `claude` process `exec`, which `inject-bootup-context.py` (the `SessionStart` hook) already uses to detect a genuine new dispatcher process and write `dispatcher-session-start.ts` via `_write_dispatcher_session_start()`.

Adds a sibling `_reset_dispatcher_heartbeat()` call at that same point (same `if pid_flag_dispatcher:` branch — i.e. only on a genuine new process start, not on post-compact same-process continuation) so the heartbeat file is reset to "now" before the new session's first health-check cycle can evaluate it.

- Same file path/format `health-check-v3.sh` and `thinking-heartbeat.py` already use (`LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE` / `logs/dispatcher-heartbeat`, single Unix epoch integer, atomic tmp+rename write) — **no `health-check-v3.sh` changes required**.
- No migration needed in `scripts/upgrade.sh` — this is a plain code change to an already-registered hook, not a new hook/directory/cron/config key. Verified against the existing migration list in `upgrade.sh`.

## Test plan

- [x] `python3 -m py_compile hooks/inject-bootup-context.py`
- [x] Existing unit tests pass: `uv run pytest tests/unit/test_hooks/test_dispatcher_session_start.py tests/unit/test_hooks/test_inject_bootup_context_logging.py tests/unit/test_hooks/test_inject_bootup_context_fresh_start.py tests/unit/test_hooks/test_inject_bootup_context_sentinel.py` (41 passed, 7 skipped)
- [x] Manual dry run with `LOBSTER_WORKSPACE` sandboxed to a temp dir: seeded a 1500s-old (stale) heartbeat file, simulated a genuine new dispatcher session (live startup-flag PID), confirmed the heartbeat file is reset to a fresh timestamp matching `dispatcher-session-start.ts`
- [x] Manual dry run confirms subagent sessions (no live startup flag) leave the heartbeat file untouched — only genuine dispatcher process starts reset it
- [ ] Live verification on a running install (deploy blocked from this session by `system-file-protect.py`'s intentional guard on direct edits to `~/lobster/`; deploy via the standard `git pull` + `update-lobster.sh`/`upgrade.sh` path once merged)

### DIFF
--- a/hooks/inject-bootup-context.py
+++ b/hooks/inject-bootup-context.py
@@ -145,6 +145,42 @@ DISPATCHER_SESSION_START_FILE = Path(
     )
 )
 
+# Dispatcher heartbeat sentinel (issue #2150).
+#
+# Normally written by hooks/thinking-heartbeat.py on every PostToolUse event.
+# health-check-v3.sh's check_dispatcher_heartbeat() treats this file as stale
+# (RED) once DISPATCHER_HEARTBEAT_STALE_SECONDS have elapsed since its mtime
+# content, subject to a time-bounded WFM-active suppression window capped at
+# (SESSION_AGE_LIMIT_SECONDS - WFM_SUPPRESSION_MARGIN_SECONDS) = 6900s/115min.
+#
+# Bug: neither restart path reset this file when spawning a replacement
+# dispatcher session:
+#   1. The graceful SIGTERM path (check_session_age(), fires at
+#      SESSION_AGE_LIMIT_SECONDS = 7200s/2h)
+#   2. The hard restart path (do_restart(), fires on RED heartbeat staleness)
+# Because the RED cap (115min) sits only 5 minutes inside the session-age
+# limit (2h), a new session could inherit an already near-stale heartbeat
+# file from its predecessor and re-breach the cap within the very next 4-min
+# health-check cycle -- causing a second, spurious restart shortly after a
+# legitimate one. Most visible on quiet/idle days with few tool calls to
+# naturally refresh the heartbeat early in the new session's life.
+#
+# Fix: reset this file to "now" at the same point _write_dispatcher_session_start()
+# already runs -- i.e. on every genuine new dispatcher process start (PID
+# startup-flag signal), which is the single choke point both restart paths funnel
+# through via claude-persistent.sh's launcher. This covers both restart paths
+# without duplicating logic in the bash health-check script.
+#
+# Same path/format health-check-v3.sh and thinking-heartbeat.py already use
+# (LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE / logs/dispatcher-heartbeat, single
+# Unix epoch integer) -- no health-check-v3.sh change required.
+DISPATCHER_HEARTBEAT_FILE = Path(
+    os.environ.get(
+        "LOBSTER_DISPATCHER_HEARTBEAT_OVERRIDE",
+        str(_LOBSTER_WORKSPACE / "logs" / "dispatcher-heartbeat"),
+    )
+)
+
 
 def _parse_admin_chat_id(config_env_path: Path) -> str | None:
     """Parse LOBSTER_ADMIN_CHAT_ID from a config.env file.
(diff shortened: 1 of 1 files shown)


sayhar (commented):
🤖🦞 Lobster (reviewer): PASS: Reviewed cold against the diff first, then checked the engineer's claims by reading the actual restart-path code (not just trusting the description).

**Verified independently:**
- Both restart paths do funnel through the same choke point. `check_session_age()`'s SIGTERM path and `do_restart()`'s hard-kill-and-`systemctl start` path both ultimately cause `claude-persistent.sh`'s `main()` loop to call `launch_claude()` again, which writes a fresh PID startup-flag file before exec-ing `claude`. That flag is exactly what `inject-bootup-context.py`'s `pid_flag_dispatcher` branch checks — and the new heartbeat reset is placed inside that branch, alongside the existing session-start-time write. Confirmed by reading `scripts/claude-persistent.sh` lines ~590-784 and `scripts/health-check-v3.sh`'s `do_restart()`/`check_session_age()`.
- The reset correctly does NOT fire on post-compaction continuations — those hit the `elif uuid_dispatcher` branch, not the `pid_flag_dispatcher` branch, so the reset is skipped by construction.
- The reset correctly does NOT fire for subagent sessions — subagents match neither signal, so `is_dispatcher` is False and the whole block (including the reset) is skipped.
- Ran a manual dry run loading the hook module directly: seeded a stale heartbeat, called `_reset_dispatcher_heartbeat()`, confirmed the timestamp advances and no stray `.tmp` file is left behind (atomic rename works as documented).
- Path/format consistency check: the file path and single-epoch-integer format match what `thinking-heartbeat.py` writes and what `health-check-v3.sh`'s `check_dispatcher_heartbeat()` reads — no health-check-v3.sh change needed, as claimed.
- Ran `uv run pytest tests/unit/test_hooks/ --tb=no -q`: 11 failed, 1049 passed, 8 skipped. Checked out `origin/main` and re-ran the 3 failing test files in isolation — same 11 failures exist on main, unrelated to this change (context-monitor matcher pattern, on-compact session-id write, PII scan guard). This PR introduces zero new failures. (Note: engineer's "41 existing unit tests pass" doesn't match what I measured — closest matching subset of inject-bootup-context tests is 71 passed. Not a red flag since I independently confirmed no regressions, just flagging the number as imprecise.)

**Gap already flagged by the engineer, confirmed real:** no automated test was added for `_reset_dispatcher_heartbeat()` or for the "reset fires only on genuine PID-flag dispatcher start, not on compaction/subagent" branching — only manual/dry-run verification. This is a low-cost gap to close: `tests/unit/test_hooks/test_dispatcher_session_start.py` already tests the sibling function (`_write_dispatcher_session_start()`) this new function is modeled after line-for-line, so a mirrored test file would be nearly mechanical. Not blocking given the change is small, single-file, defensively coded (silent try/except, same pattern as neighboring functions), and the manual verification is solid — but worth a fast-follow.

No other issues found. The change is narrowly scoped (one file, 79 lines, all additive) and the fix logic is sound.

PR #2233, opened 2026-08-16
fix(fireflies-sync): invoke sync.py by file path, not -m module

aeschylus:
## Summary
- `fireflies-sync.sh` has failed on every run since Aug 9 (`ModuleNotFoundError: No module named 'integrations'`), stalling the Fireflies→Obsidian sync for 8 days.
- Root cause: the script invokes the sync entrypoint as `uv run python -m integrations.fireflies.sync` (module mode). Module-mode requires `integrations` to already be importable from `sys.path` before the interpreter loads the module — and it isn't, from the repo root (confirmed `uv run python -c "import integrations"` fails, as does the identical `-m` form for `integrations.granola.sync`).
- The sibling `granola-sync.sh` never hits this because it invokes its script **by file path** (`uv run python src/integrations/granola/sync.py`), which lets Python add the script's own directory to `sys.path` first. Both `granola/sync.py` and `fireflies/sync.py` already carry an internal `sys.path.insert(0, str(_SRC_DIR))` fallback pointing at `src/` for exactly this reason — but under `-m` invocation, the `ModuleNotFoundError` fires before that line ever executes.
- Fix: switch `fireflies-sync.sh` to file-path invocation, matching `granola-sync.sh`'s established pattern, so the existing fallback in `fireflies/sync.py` actually takes effect. No sys.path hack was added — the fallback was already there and already the codebase's pattern; this just lets it run.

## Test plan
- [x] Ran `scheduled-tasks/fireflies-sync.sh` directly on the production host — exit code 0, fetched 24 transcripts, wrote 16 new ones (the Aug 8–16 backlog), git-committed to the Obsidian vault (`fireflies: sync 16 transcripts [2026-08-16T15:27:36Z]`).
- [x] Re-ran via `systemctl start lobster-fireflies-sync.service` (the actual production invocation path) — `Result=success`, `ExecMainStatus=0`, correctly reported 0 new (idempotent re-run, all 15 already synced).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scheduled-tasks/fireflies-sync.sh
+++ b/scheduled-tasks/fireflies-sync.sh
@@ -38,9 +38,18 @@ fi
 echo "[$(date -Iseconds)] fireflies-sync starting" | tee "$LOG_FILE"
 
 # Run the sync script
+#
+# Invoked by file path (not `-m integrations.fireflies.sync`) to mirror
+# granola-sync.sh. Module-mode invocation requires `integrations` to
+# already be importable from sys.path before the interpreter starts, which
+# depends on the editable install being wired up correctly. Running by
+# file path lets Python add the script's own directory to sys.path, and
+# sync.py's own `sys.path.insert(0, str(_SRC_DIR))` (pointing at src/) then
+# makes the `integrations.*` absolute imports resolve — the same pattern
+# granola/sync.py already relies on.
 cd "$LOBSTER_DIR"
 set +e
-OUTPUT=$(uv run python -m integrations.fireflies.sync 2>&1)
+OUTPUT=$(uv run python src/integrations/fireflies/sync.py 2>&1)
 EXIT_CODE=$?
 set -e
 


sayhar (commented):
Verdict: PASS

**What I checked**
- Read the PR description and diff (`gh pr diff 2233`).
- Compared against the sibling `scheduled-tasks/granola-sync.sh`, which already invokes its script by file path (`uv run python src/integrations/granola/sync.py`) — the new fireflies invocation now matches this exactly.
- Confirmed the root-cause mechanics: `python -m integrations.fireflies.sync` requires `integrations` to be importable via `sys.path` *before* the interpreter loads the module, whereas invoking by file path (`python src/integrations/fireflies/sync.py`) lets Python prepend the script's own directory to `sys.path`, after which `sync.py`'s existing `sys.path.insert(0, str(_SRC_DIR))` fallback (pointing at `src/`) resolves the `integrations.*` absolute imports. This fallback already existed in both `granola/sync.py` and `fireflies/sync.py` — the PR isn't adding a new hack, just switching the invocation to a form where the existing fallback actually gets a chance to run.
- Looked for other callers of the old `-m integrations.fireflies.sync` invocation to make sure nothing else depends on module-mode: found none in `.sh`/`.service`/`.timer` files. `tests/unit/test_fireflies_sync.py` imports `integrations.fireflies.sync` directly as a Python module in-process (via pytest's own sys.path/rootdir handling), which is unaffected by how the *script* is invoked as a subprocess — no conflict there.
- Checked for a `list_decisions()`-style ADR mechanism in this repo — none found (no such tool/doc surfaced in `docs/`), noting this per the review instructions rather than skipping the check silently.
- Test plan: author ran the script directly on the production host (exit 0, synced the Aug 8–16 backlog) and also re-ran via the actual production invocation path, `systemctl start lobster-fireflies-sync.service` (`Result=success`, correctly idempotent on re-run). This is solid real-environment verification for a one-line subprocess invocation change — better than a unit test would give here, since the bug was specifically about subprocess/sys.path behavior that unit tests running in-process wouldn't exercise.
- Scope: diff touches only `scheduled-tasks/fireflies-sync.sh` (comment + the one invocation line). Nothing unrelated changed.

**Blocking issues**
None.

**Non-blocking nits**
- Stale docs not touched by this PR still reference the old `-m integrations.fireflies.sync` module-mode invocation and will mislead anyone reading them as a "how to run this" reference:
  - `scheduled-tasks/tasks/fireflies-sync.md` (lines ~27 and ~89)
  - `src/integrations/fireflies/sync.py`'s own module docstring, "Usage (standalone)" section (~line 20)
  Worth a quick follow-up fix so these don't reintroduce the same confusion that caused this outage in the first place (the docstring literally documents the invocation mode that doesn't work from repo root).

PR #2234, opened 2026-08-16
chore: track deploy-webhook.py (was running untracked in production)

aeschylus:
## Summary
- `lobster-deploy-webhook.service` has been running `/home/lobster/lobster/src/deploy-webhook.py` in production since 2026-08-04, but the file was never committed to git — discovered during a local-branch cleanup pass.
- This PR adds the file as-is to bring already-running code under version control. No behavior change.
- Secrets (`DEPLOY_WEBHOOK_SECRET`, `VERCEL_TOKEN`) are loaded from `~/lobster-config/config.env` at runtime, not hardcoded — safe to commit.
- Added a prominent docstring note on `verify_signature()` documenting that it fails open (accepts unauthenticated requests) when `DEPLOY_WEBHOOK_SECRET` is unset. This does not change behavior — it matches what's already running in prod — it just surfaces the risk in source rather than only in a runtime log line. Whether to fail closed instead is left as a follow-up (flagged with a TODO), since changing runtime behavior wasn't in scope for a "track existing code" PR.

## Scope correction (post-review)
An earlier version of this PR accidentally included an unrelated fix to `scheduled-tasks/fireflies-sync.sh` (the `-m` vs. file-path invocation fix, which belongs to #2233 and is tracked there). Both this branch and #2235 were cut from a local working tree that had that change staged before it was properly isolated into its own PR. That file has been reverted to main's version on this branch — the diff now contains only `src/deploy-webhook.py`.

## Verification of fidelity to production
I could not independently diff the committed file against the live production copy from the environment this fix was made in:
- No `lobster-deploy-webhook.service` systemd unit exists on this host.
- No `/home/lobster/lobster/src/deploy-webhook.py` file or matching process exists on this host.
- No `DEPLOY_WEBHOOK_SECRET`/`VERCEL_TOKEN` reference in `~/lobster-config/config.env` on this host.

This suggests the fix was made from a different host/session than the one actually running the webhook, or the environment used here doesn't have access to that production host. **I cannot confirm the committed file byte-for-byte matches what's currently running in production.** Whoever merges this should verify directly on the host running `lobster-deploy-webhook.service`, e.g.:
```
diff <(systemctl show -p ExecStart lobster-deploy-webhook.service) <(echo expected...)
diff /home/lobster/lobster/src/deploy-webhook.py <path-to-checked-out-branch>/src/deploy-webhook.py
```
or compare file hashes (`sha256sum`) between the running file and this branch's version.

## Architectural decision mechanism check
Checked for a `list_decisions()`-style mechanism per standing rule 86 before making changes — none exists in this repo. The closest match is `memory/canonical-templates/pending-decisions.md`, a markdown template for a decision log, not a queryable tool. No decisions log entry was needed for this fix (scope correction + documentation, no new architectural choice).

## Test plan
- [ ] Confirm `systemctl status lobster-deploy-webhook.service` still active after merge + redeploy
- [ ] Confirm no diff between committed file and the file currently on disk in production (**not verified from this environment — see Verification section above**)
- [ ] Drew to review whether this webhook should also get a `services/` unit template + docs entry in SOPs/services.md

### DIFF
--- a/scheduled-tasks/fireflies-sync.sh
+++ b/scheduled-tasks/fireflies-sync.sh
@@ -38,18 +38,9 @@ fi
 echo "[$(date -Iseconds)] fireflies-sync starting" | tee "$LOG_FILE"
 
 # Run the sync script
-#
-# Invoked by file path (not `-m integrations.fireflies.sync`) to mirror
-# granola-sync.sh. Module-mode invocation requires `integrations` to
-# already be importable from sys.path before the interpreter starts, which
-# depends on the editable install being wired up correctly. Running by
-# file path lets Python add the script's own directory to sys.path, and
-# sync.py's own `sys.path.insert(0, str(_SRC_DIR))` (pointing at src/) then
-# makes the `integrations.*` absolute imports resolve — the same pattern
-# granola/sync.py already relies on.
 cd "$LOBSTER_DIR"
 set +e
-OUTPUT=$(uv run python src/integrations/fireflies/sync.py 2>&1)
+OUTPUT=$(uv run python -m integrations.fireflies.sync 2>&1)
 EXIT_CODE=$?
 set -e
 
--- a/scheduled-tasks/fireflies-sync.sh
+++ b/scheduled-tasks/fireflies-sync.sh
@@ -38,9 +38,18 @@ fi
 echo "[$(date -Iseconds)] fireflies-sync starting" | tee "$LOG_FILE"
 
 # Run the sync script
+#
+# Invoked by file path (not `-m integrations.fireflies.sync`) to mirror
+# granola-sync.sh. Module-mode invocation requires `integrations` to
+# already be importable from sys.path before the interpreter starts, which
+# depends on the editable install being wired up correctly. Running by
+# file path lets Python add the script's own directory to sys.path, and
+# sync.py's own `sys.path.insert(0, str(_SRC_DIR))` (pointing at src/) then
+# makes the `integrations.*` absolute imports resolve — the same pattern
+# granola/sync.py already relies on.
 cd "$LOBSTER_DIR"
 set +e
-OUTPUT=$(uv run python -m integrations.fireflies.sync 2>&1)
+OUTPUT=$(uv run python src/integrations/fireflies/sync.py 2>&1)
 EXIT_CODE=$?
 set -e
 
--- a/src/deploy-webhook.py
+++ b/src/deploy-webhook.py
@@ -43,7 +43,15 @@ DEPLOY_MAP = {
 
 
 def verify_signature(payload: bytes, sig_header: str, secret: str) -> bool:
-    """Verify GitHub HMAC-SHA256 webhook signature."""
+    """Verify GitHub HMAC-SHA256 webhook signature.
+
+    SECURITY NOTE: fails open when `secret` is empty — any caller, signed or
+    not, is accepted. This mirrors the existing production behavior (a
+    startup warning is logged, see `load_config`/`__main__` below) rather
+    than changing it here, but it means DEPLOY_WEBHOOK_SECRET being unset
+    silently disables auth on this endpoint. TODO: consider failing closed
+    (rejecting all requests) instead once this is confirmed safe to change.
+    """
     if not secret:
         log.warning("No webhook secret configured — skipping signature check")
         return True
--- a/src/deploy-webhook.py
+++ b/src/deploy-webhook.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python3
+"""
+Lobster Deploy Webhook Server
+Listens for GitHub push events and triggers Vercel deployments.
+
+Usage:
+    python3 deploy-webhook.py [--port 9120]
+
+Configuration via environment variables (loaded from config files):
+    DEPLOY_WEBHOOK_SECRET - GitHub webhook secret
+    VERCEL_TOKEN - Vercel CLI token
+"""
+
+import hashlib
+import hmac
+import http.server
+import json
+import logging
+import os
+import subprocess
+import threading
+from pathlib import Path
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [deploy-webhook] %(levelname)s %(message)s",
+)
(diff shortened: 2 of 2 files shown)


sayhar (commented):
Verdict: NEEDS-WORK

Independent review (no `list_decisions()`-style mechanism found in this repo to check against — searched docs/ and found only pending-decisions.md memory files with no automated decisions tool, so noting this and proceeding without it).

**What I checked:**
- Full diff (`gh pr diff 2234`)
- Whether the committed `deploy-webhook.py` contains hardcoded secrets/tokens — it does not; `VERCEL_TOKEN`/`DEPLOY_WEBHOOK_SECRET` are read from env / `~/lobster-config/config.env` at runtime.
- Whether the diff's scope matches the PR's stated purpose ("simply adds the file as-is").
- Whether the PR's own verification checklist has been completed.

**Blocking:**
1. **Scope mismatch.** The diff isn't just adding `src/deploy-webhook.py` — it also modifies `scheduled-tasks/fireflies-sync.sh` (changing the sync invocation from `-m integrations.fireflies.sync` to a direct file path, with a fairly involved comment justifying the change). This is unrelated to tracking the webhook file and isn't mentioned anywhere in the PR description. Either this is an accidental carry-over from a rebase/base branch, or it's an intentional unrelated fix that should be its own PR with its own description and test plan. As-is, a reviewer can't tell which, and it undermines the "no behavior change" claim in the summary.
2. **Verification checklist is unchecked.** The PR's entire premise is "bring already-running code under version control, no behavior change" — but the two checkboxes that would substantiate that ("confirm no diff between committed file and the file currently on disk in production", "confirm service still active after merge") are unchecked. Since this file was previously untracked, there's no git history to diff against — the *only* evidence of fidelity to production is that manual diff check, and it hasn't been confirmed yet. Recommend actually running `diff <(ssh prod cat .../deploy-webhook.py) src/deploy-webhook.py` (or equivalent) and pasting the result (or an "identical" confirmation) into the PR before merge.

**Non-blocking nits:**
- `verify_signature()` fails open (returns `True`) when `WEBHOOK_SECRET` is unset, only logging a warning — meaning if the secret is ever unset in prod config, the deploy endpoint accepts unauthenticated pushes. This is existing behavior being brought into version control rather than introduced by this PR, but now that it's visible/reviewable, worth a follow-up to fail closed instead.
- `load_config()` does manual `.env` parsing (`strip('"').strip("'")`) rather than using an existing env-loading helper if one exists elsewhere in this repo — minor duplication, not a big deal for a single-file webhook script.
- Test plan item "Drew to review whether this webhook should also get a `services/` unit template + docs entry" is left as an open TODO for the reviewer rather than the author — fine, just flagging it's not actually part of this PR's deliverable.

sayhar (commented):
Verdict: NEEDS-WORK (one open item, not a code-quality blocker)

Re-reviewing the current state after commit 9991782.

## Scope issue — RESOLVED
Confirmed via `gh pr diff` and `gh pr view --json changedFiles,additions,deletions`: the PR now touches exactly one file, `src/deploy-webhook.py` (+210/-0, 1 file changed). `scheduled-tasks/fireflies-sync.sh` is no longer part of this diff. Good — the prior blocking scope-bleed issue is fixed.

## Docstring nit — RESOLVED
`verify_signature()` now has a clear docstring documenting the fail-open behavior when `DEPLOY_WEBHOOK_SECRET` is unset, with a TODO flagging fail-closed as a possible future change. Matches what was asked for. Behavior itself is unchanged (still fails open, still logs a warning at both the function call site and startup) — this is documentation only, as intended.

## Production fidelity — still an open item, and I think it should stay open until closed by a human
This is the one substantive question left, and I want to be explicit about my reasoning rather than just parrot the PR description.

The PR's entire premise is "bring already-running code under version control, no behavior change." That framing only holds if the committed bytes match what's actually running under `lobster-deploy-webhook.service` in prod. The engineer was honest that they couldn't check this from their environment (no service, no process, no file, no relevant env vars present on that host) and left explicit verification commands (`sha256sum` / diff against `ExecStart`) for whoever merges. I want to be clear that documenting an unverified claim instead of asserting it as fact is the correct behavior, not a shortcut — but it doesn't reduce the actual risk the PR exists to eliminate.

Concretely, what could go wrong if this merges unverified: if prod's on-disk file has diverged from this branch (e.g., a hotfix applied directly on the box after 2026-08-04, or a still-uncommitted local edit), merging enshrines the *wrong* version as canonical, and a future deploy/redeploy from git would silently regress prod to a stale/incorrect webhook handler. Given this file handles GitHub HMAC signature verification and triggers `vercel --prod`, a silent regression here is not cosmetic. The blast radius is small and the fix is cheap, but it requires someone with actual access to the host running `lobster-deploy-webhook.service` — Claude/agent code review can't close it, and no back-and-forth on this PR will close it either.

**What needs to happen for sign-off:** someone with SSH/access to the production host needs to run a diff or hash comparison between `/home/lobster/lobster/src/deploy-webhook.py` (the running file) and this branch's `src/deploy-webhook.py`, per the commands already included in the PR description. If they match, merge as-is. If they don't, the PR needs to be updated to match the actual running file (or the drift needs to be understood and reconciled) before merge.

I'd treat this as a merge gate rather than a "needs more engineering work" gate — the code and PR hygiene here are fine.

## Architectural decision mechanism check
Confirmed: no `list_decisions()`-style mechanism exists in this repo (checked previously and re-confirmed the PR description's own note matches — closest thing is the `memory/canonical-templates/pending-decisions.md` template, not a queryable tool). No decision log entry needed here since this PR makes no new architectural choice.

## Other observations (non-blocking)
- `verify_signature`'s fail-open default plus the TODO to fail closed is a reasonable follow-up but out of scope for a "track existing code" PR, as noted — agreed, don't scope-creep this PR to fix it now.
- `load_config()` silently no-ops if `~/lobster-config/config.env` doesn't exist rather than erroring — fine given `VERCEL_TOKEN`/`WEBHOOK_SECRET` absence is already logged loudly at startup.

Bottom line: engineering-side blockers are resolved. The one remaining item is operational, not a code problem — needs a human with production access to confirm byte-for-byte match before merge, using the commands already in the PR description.

Issue #2200, opened 2026-08-12
install.sh never runs upgrade.sh's migration runner — reimage-via-restore silently skips config.env migrations


sayhar:
## Problem

`~/lobster/install.sh` never calls `~/lobster/scripts/upgrade.sh`'s migration runner (`run_migrations()`, `scripts/upgrade.sh:952`). Migrations only ever apply when someone explicitly runs `upgrade.sh` on an existing install.

This means any reimage-via-restore — where old `~/lobster/`, `~/lobster-config/`, etc. get copied onto a fresh VM before `install.sh` runs, rather than a truly from-scratch `git clone` + first install — silently skips every `config.env`-touching migration forever, unless someone thinks to manually re-run `upgrade.sh` afterward. `install.sh` has no way to know this host isn't actually fresh.

## Confirmed root cause of a real incident

Migration 96 (`scripts/upgrade.sh` — adds `CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT=0` to `config.env`, fixing issue #2142, merged in PR #2144 on 2026-08-02) never applied after the 2026-08-09 host reimage, because `config.env` was restored from an old backup and `install.sh` doesn't invoke migrations at all. This is distinct from #2142 itself (which was about the idle-timeout watchdog behavior and is already fixed) — #2142 fixed the *migration*, this issue is about the *migration never running* on a reimaged host.

## Fix

1. Make `install.sh` call the same migration runner `upgrade.sh` uses, unconditionally, at the end of every install run (fresh install or reimage-via-restore — same code path either way).
2. Add a lightweight startup self-check (extending `scripts/health-check-v3.sh`, which already does periodic checks) verifying a short list of load-bearing `config.env` keys are present and non-empty, starting with `CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT` — exactly the class of gap that caused the incident. Missing keys should surface via log + alert, not silently proceed.

Relates to #2142.

sayhar:
PR opened: #2211

- [x] Extract `run_migrations()` into `scripts/lib/migrations.sh` (shared, following the `scripts/lib/template.sh` convention)
- [x] `install.sh` sources it and calls `run_migrations()` unconditionally at the end of every run
- [x] `upgrade.sh` sources the same shared implementation instead of a local copy
[...]

PR #2247, opened 2026-08-20
fix(tests): stub crontab/sudo in test-migrations-lib.sh to stop real crontab writes

### LINKED ISSUE #2246: test-migrations-lib.sh writes real crontab entries during test runs
## Bug

`tests/test-migrations-lib.sh` (added in PR #2200) sandboxes migration tests by faking `LOBSTER_DIR`, `WORKSPACE_DIR`, `MESSAGES_DIR`, `LOBSTER_CONFIG_DIR`, etc. as paths under a `mktemp -d /tmp/lobster-test-migrations-XXXXXX` directory, then sources `scripts/lib/migrations.sh` and calls `run_migrations()`.

`run_migrations()` runs all ~99 migrations unconditionally (there's no way to target a single migration). Several of those migrations shell out directly to the real `crontab` binary — not sandboxed by any of the test's fake directory variables, since `crontab -l`/`crontab -` always reads/writes the real per-user system crontab regardless of `$LOBSTER_DIR` etc. Notably:

- Migration 28 (LOBSTER-LOG-EXPORT), `scripts/lib/migrations.sh` ~line 573-574
- Migration 52 (LOBSTER-GHOST-DETECTOR), `scripts/lib/migrations.sh` ~line 1022-1023
- Several others (health-check, self-check, nightly-consolidation, OOM-check, staleness-warn, etc. — see full `crontab` grep below)

So running `bash tests/test-migrations-lib.sh` on a real host mutates the real system crontab, inserting entries built from the fake temp path (e.g. `cd $TEST_TMPDIR/lobster && ...`). This already happened once and corrupted the live crontab on this host.

Broader grep of `scripts/lib/migrations.sh` for real system-mutating commands not sandboxed by the fake dir vars:
- `crontab -l` / `crontab -` — read/write the real per-user crontab (~25 call sites)
- `sudo systemctl enable/disable/start/stop/restart/daemon-reload` — several migrations (service rename cleanup, mcp-local install, transcription service, systemd timer migrations)
- `sudo usermod -aG crontab <user>` — migration 46, gated behind `sudo -n true` probe
- A `python3 -c` block (migration ~55, systemd timer migration from jobs.json) that shells out to `subprocess.run(["sudo", "tee", ...])` and `subprocess.run(["sudo", "systemctl", ...])` — currently gated behind `[ -f "$WORKSPACE_DIR/scheduled-jobs/jobs.json" ]`, which is false in the current test (fake dir has no such file), but nothing stops a future test setup from tripping it.

## Root cause

The test only fakes environment *variables* migrations.sh reads (paths), not the *commands* migrations.sh invokes. `crontab`, `sudo`, `systemctl` are real system binaries resolved via `$PATH`; faking `$LOBSTER_DIR` etc. does nothing to sandbox them.

## Fix

Stub these commands (via a fake-bin directory prepended to `$PATH`, so both direct shell calls and `subprocess.run(...)` calls from the embedded Python block are caught) inside `tests/test-migrations-lib.sh`, so `run_migrations()` can be exercised end-to-end without ever touching the real host crontab, systemd, or user groups. Verify by diffing `crontab -l` before/after the test run inside a throwaway Docker container.

Considered but out of scope for this fix: refactoring `run_migrations()` (currently one large sequential function with ~99 inline migration blocks) so tests can target a single migration by name. That would reduce blast radius further but is a much larger refactor; the command-stub fix fully closes the actual leak on its own.

sayhar:
## Problem

`run_migrations()` (`scripts/lib/migrations.sh`) runs all ~99 migrations unconditionally in a single pass. `tests/test-migrations-lib.sh` (PR #2200) sandboxes this by faking `$LOBSTER_DIR`, `$WORKSPACE_DIR`, `$MESSAGES_DIR`, `$LOBSTER_CONFIG_DIR`, etc. as paths under a throwaway `mktemp -d` directory — but several migrations shell out directly to the real `crontab` binary (migration 28 `LOBSTER-LOG-EXPORT`, migration 52 `LOBSTER-GHOST-DETECTOR`, and others) and to `sudo` (`systemctl`, `usermod`, `tee`). None of those are sandboxed by faking the directory variables, because `crontab -l`/`crontab -` always read and write the real per-user system crontab regardless of what `$LOBSTER_DIR` points at.

Running `bash tests/test-migrations-lib.sh` on a real host therefore wrote real crontab entries built from the test's throwaway temp path (e.g. `cd /tmp/lobster-test-migrations-XXXXXX/lobster && ...`) into the live system crontab. This already happened once on the maintainer's host.

## Fix

`tests/test-migrations-lib.sh` now prepends a fake-bin directory to `$PATH` before sourcing `scripts/lib/migrations.sh`, containing stub `crontab` and `sudo` executables:

- **`crontab`** — mimics `crontab -l` / `crontab -` / `crontab <file>` against a local state file instead of the real spool. Writes go through a temp-file-then-`mv` (atomic replace), matching how the real `crontab` binary behaves — see "Race condition found" below for why this matters.
- **`sudo`** — swallows every privileged call (`systemctl`, `usermod`, `cp`, `tee`, `ldconfig`, ...) as a no-op. `sudo -n true` (the passwordless-sudo probe migration 46 uses to decide whether to attempt `usermod`) returns success; `sudo tee` consumes stdin so pipelines don't block.

Stubbing at the `$PATH` resolution boundary (real executables shadowing the real binaries) rather than as bash functions was a deliberate choice: one migration (~line 1083, systemd-timer migration from `jobs.json`) shells out via `subprocess.run(["sudo", "tee", ...])` from an embedded Python block. A bash function override only intercepts direct shell invocations in `migrations.sh` itself — it would not have caught that `subprocess.run` call. `$PATH`-shadowing catches both.

**Race condition found and fixed along the way:** several migrations chain `crontab -l | grep -v MARKER | crontab -`. Bash starts every pipeline stage concurrently, so the read-side (`-l`) and write-side (`-`) stub invocations run at the same time. An initial version of the stub wrote via `cat > "$STATE"`, which truncates the file at open time — racing the concurrent read-side's `cat "$STATE"` and intermittently handing it a truncated/empty file. This silently dropped previously-written crontab entries (e.g. migration 28's `LOBSTER-LOG-EXPORT` entry vanished by the time migration 52 ran) without failing the test, since each migration only checks for its own marker. Fixed by writing to a temp file and `mv`-ing it into place — same technique the real `crontab` binary uses — so a concurrent reader always sees either the complete old content or the complete new content, never a partial file.

## What's new in the test suite

A new assertion group, "run_migrations() never touches the real system crontab":
- Captures the real system crontab's content via the real `crontab` binary's resolved path *before* the `$PATH` stub is installed, and re-checks it *after* `run_migrations()` runs (through the same resolved path) — asserts byte-for-byte equality. This is a permanent regression guard: if any future migration is added that bypasses the stub, this assertion fails instead of silently mutating the real host.
- Asserts the crontab-writing migrations (28, 52) actually exercised the stub — i.e. the isolation test is not a no-op because the crontab code paths were skipped for an unrelated reason.

## Verified

- Rebuilt `tests/docker/Dockerfile.test` (clean Debian container with a real `cron`/`crontab` binary and `sudo`).
- Ran the **unpatched** (pre-fix, `origin/main`) `test-migrations-lib.sh` inside the container: `crontab -l` before = `no crontab for testuser`; after = 3 real entries written, including `LOBSTER-LOG-EXPORT` and `LOBSTER-GHOST-DETECTOR` pointing at the throwaway `/tmp/lobster-test-migrations-*` path — reproducing the exact bug.
- Ran the **patched** test in the same container. See "Follow-up" below for current pass counts (updated after a second commit fixed an environment-dependent regression the first follow-up commit introduced).
- Also ran patched test on the dev host directly: real crontab (`crontab -l | md5sum`) identical before/after the run.

Out of scope, noted for awareness: on this dev host (not in the clean Docker container), two "Migration 96" assertions can fail depending on host shell state — see "Follow-up" below, they are not caused by this PR's changes.

## Follow-up: Migration 55 coverage, and a fix for an environment-dependent regression it introduced

A follow-up commit (`ed528595`) seeded a minimal `jobs.json` fixture so Migration 55 (`systemd-timer-from-jobs.json`, the migration whose embedded Python does the `subprocess.run(["sudo", "tee", ...])` call this suite's `sudo` stub is specifically designed to catch) actually runs during the test, and added a hard assertion that its `sudo tee` call landed in the stub's log. This closed a real gap: without it, the suite's crontab/sudo isolation coverage never exercised the one migration that goes through Python `subprocess.run` rather than a direct shell call.

That assertion, however, hard-failed in `tests/docker/Dockerfile.test` — the same container this PR's own "Verified" section uses for its safety claims — because Migration 55 is additionally gated in `scripts/lib/migrations.sh` on `pidof systemd >/dev/null 2>&1` succeeding, and `debian:bookworm-slim` in that container never runs systemd as PID 1, so the migration's body (and the `sudo tee` call inside it) was silently skipped, and the new hard assertion failed even though the crontab/sudo isolation itself was working correctly.

Fixed by adding a `pidof` stub to the same fake-bin directory the suite already uses for `crontab`/`sudo` (rather than guarding the assertion into a conditional skip): `pidof` always exits 0, so `pidof systemd` reports success regardless of whether the test runs on a real systemd host or in a systemd-less container. This makes Migration 55 actually execute and get exercised for real in every environment the suite runs in, rather than only skip-passing in containers — the more robust of the two options, since a skip-pass would leave the container run with zero real coverage of the `subprocess.run(["sudo", "tee", ...])` path.

Two other `pidof systemd` call sites in `migrations.sh` (mcp-local service install, transcription service enable/start) are also covered by this stub, but neither's migration body executes in this test's fake `$LOBSTER_DIR`/`$WORKSPACE_DIR` fixtures — both are gated behind `-f` checks on service files that don't exist there — so stubbing `pidof` globally doesn't change their behavior in this suite.

Separately (not caused by the `pidof` stub — reproduces identically on `787dcf4c`, the commit before either follow-up change): the two "Migration 96" assertions read `${CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT:-}` from the *inherited* shell environment rather than only from the fake `config.env` fixture. When this test is run from inside a Lobster agent shell (which itself has `CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT=0` exported, per issue #2142), Migration 96 sees the var already set and skips writing it to the fixture file, failing the assertion. This is pre-existing host-state leakage unrelated to the crontab/sudo/pidof isolation work in this PR — confirmed reproducing on `787dcf4c` via `git stash` — and does not occur in the Docker container (no such var exported there) or in a dev-host shell without that var inherited. Left out of scope here; flagging for a possible separate follow-up to isolate `CONFIG_FILE`-only sourcing in that assertion.

## Considered but out of scope

Refactoring `run_migrations()` (currently one large sequential function with ~99 inline migration blocks) so tests could target a single migration by name, reducing blast radius further. This is a much larger refactor of the shared migration runner used by both `install.sh` and `upgrade.sh`; the command-stub fix here fully closes the actual crontab-leak on its own without touching `scripts/lib/migrations.sh` at all.

## Functional patterns

The stubs are pure state-machine scripts (read/write a single local file, no hidden global mutation beyond that file) and the new assertions are declarative (`assert_file_contains`, string equality) rather than imperative control flow — consistent with the existing test file's style. No changes to `scripts/lib/migrations.sh` itself; this is entirely test-harness isolation.

## Tests run
- [x] `bash tests/test-migrations-lib.sh` (dev host, run from inside a Lobster agent shell with `CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT=0` inherited) — 10/12 pass; the 2 failures are the pre-existing "Migration 96" host-state leakage described above (present unmodified on `787dcf4c`, unrelated to this commit); Migration 55 assertion passes; real crontab confirmed byte-identical before/after via `crontab -l | md5sum`
- [x] `bash tests/test-migrations-lib.sh` (dev host, `env -u CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT`, i.e. without that host-state leak) — 12/12 pass
- [x] `bash tests/test-migrations-lib.sh` inside `tests/docker/Dockerfile.test` container (clean `debian:bookworm-slim`, no systemd as PID 1, real `crontab`/`cron`/`sudo`) — 12/12 pass, including the Migration 55 `sudo tee` assertion (now genuinely exercised via the `pidof` stub, not skip-passed)
- [x] Unpatched `origin/main` version of the same test, same container — reproduces the bug (writes real crontab entries); confirms this test would have caught the original issue

## Manual test
Required (this PR touches file I/O affecting a real system's crontab). See "Verified" above — this *is* the manual/integration test: real `crontab -l` diffed before/after inside a throwaway Docker container, for both the unpatched (bug reproduced) and patched (bug fixed) versions of the test file, plus a same-check run on the dev host itself.

## Definition of Done
- [x] Unit tests pass (see Tests run)
- [x] Integration/manual test run — see Manual test
- [x] User-visible outcome: N/A (developer-facing test-harness fix, no user-visible behavior change)
- [x] Side-effect audit: verified via Docker crontab diff that no real crontab writes occur; `sudo` calls are no-ops in the stub so no real systemctl/usermod/tee side effects either
- [x] External service calls: N/A — this PR stubs local system binaries (`crontab`, `sudo`), no network/external services involved

Closes #2246.

### DIFF
--- a/tests/test-migrations-lib.sh
+++ b/tests/test-migrations-lib.sh
@@ -212,6 +212,29 @@ exit 0
 EOF
 chmod +x "$FAKE_BIN_DIR/sudo"
 
+# Fake pidof stub (tests/test-migrations-lib.sh, issue #2246 follow-up) -
+# Migration 55 (systemd-timer-from-jobs.json) and a couple of other
+# migrations gate their body on `pidof systemd >/dev/null 2>&1` succeeding,
+# i.e. "are we actually running under systemd as PID 1". That's true on a
+# real Lobster host but false in this suite's Docker test container
+# (tests/docker/Dockerfile.test runs debian:bookworm-slim without systemd as
+# PID 1), so those migration branches - including the one this suite's
+# sudo-tee assertion below depends on - would otherwise be silently skipped
+# there, and the assertion would hard-fail even though the crontab/sudo
+# isolation itself works correctly. Stub `pidof` to always report systemd as
+# present so Migration 55 actually runs and gets exercised in every
+# environment this suite runs in, the same way `crontab`/`sudo` are stubbed
+# above rather than conditionally skipped.
+cat > "$FAKE_BIN_DIR/pidof" <<'EOF'
+#!/bin/bash
+# Fake pidof stub (tests/test-migrations-lib.sh, issue #2246 follow-up) -
+# always reports the queried process as running (exit 0), so
+# `pidof systemd >/dev/null 2>&1` succeeds regardless of whether this test
+# runs on a real systemd host or in a systemd-less container.
+exit 0
+EOF
+chmod +x "$FAKE_BIN_DIR/pidof"
+
 export PATH="$FAKE_BIN_DIR:$PATH"
 
 DRY_RUN=false
--- a/tests/test-migrations-lib.sh
+++ b/tests/test-migrations-lib.sh
@@ -12,6 +12,19 @@
 #      so a reimage-via-restore silently skipped every config.env migration.
 #   3. upgrade.sh still sources the shared lib and calls run_migrations().
 #
+# Also verifies (issue #2246): run_migrations() runs all ~99 migrations
+# unconditionally, and several of them shell out directly to the real
+# `crontab` binary and to `sudo` (systemctl, usermod, tee) - commands that
+# are NOT sandboxed by faking $LOBSTER_DIR/$WORKSPACE_DIR/etc, since
+# `crontab -l`/`crontab -` always read/write the real per-user system
+# crontab regardless of those variables. Running this test previously wrote
+# real entries (built from the fake temp path) into the actual system
+# crontab. This suite stubs `crontab` and `sudo` via a fake-bin directory
+# prepended to $PATH (catches both direct shell calls and the
+# subprocess.run(["sudo", ...]) calls made by the embedded Python migration
+# block) and asserts the real system crontab is byte-for-byte unchanged
+# after run_migrations() executes.
+#
 # Usage: bash tests/test-migrations-lib.sh
 #        (run from repo root or any directory)
 #===============================================================================
@@ -35,6 +48,17 @@ LIB="$REPO_ROOT/scripts/lib/migrations.sh"
 TEST_TMPDIR=$(mktemp -d /tmp/lobster-test-migrations-XXXXXX)
 cleanup() { rm -rf "$TEST_TMPDIR"; }
 trap cleanup EXIT
+
+# Resolve the real `crontab` binary's absolute path BEFORE any $PATH
+# stubbing happens below, and snapshot the real crontab's current contents.
+# After run_migrations() executes (with the fake-bin stub shadowing
+# `crontab` in $PATH), we re-snapshot via this same absolute path and
+# assert byte-for-byte equality - proving the stub, not the real binary,
+# absorbed every migration's crontab read/write (issue #2246).
+REAL_CRONTAB_BIN="$(command -v crontab || true)"
+if [ -n "$REAL_CRONTAB_BIN" ]; then
+    REAL_CRONTAB_BEFORE="$("$REAL_CRONTAB_BIN" -l 2>/dev/null || true)"
+fi
 
 pass() { PASS=$((PASS + 1)); TOTAL=$((TOTAL + 1)); echo -e "  ${GREEN}PASS${NC} $1"; }
 fail() { FAIL=$((FAIL + 1)); TOTAL=$((TOTAL + 1)); echo -e "  ${RED}FAIL${NC} $1"; echo -e "       ${YELLOW}$2${NC}"; }
@@ -93,6 +117,71 @@ step()    { :; }
 substep() { :; }
 log_to_file() { :; }
 
+# ---------------------------------------------------------------------------
+# Command isolation (issue #2246): several migrations shell out directly to
+# `crontab` and `sudo` (systemctl, usermod, tee) - real system binaries that
+# are NOT sandboxed by the fake $LOBSTER_DIR/$WORKSPACE_DIR/etc above. A
+# fake-bin directory prepended to $PATH intercepts these at the command
+# resolution boundary, so it catches both plain shell invocations in
+# migrations.sh AND the subprocess.run(["sudo", ...]) calls made by the
+# embedded Python migration block (a bash function override would only
+# catch the former).
+FAKE_BIN_DIR="$TEST_TMPDIR/fake-bin"
+FAKE_CRONTAB_STATE="$TEST_TMPDIR/fake-crontab-state"
+mkdir -p "$FAKE_BIN_DIR"
+: > "$FAKE_CRONTAB_STATE"
+
+cat > "$FAKE_BIN_DIR/crontab" <<EOF
+#!/bin/bash
+# Fake crontab stub (tests/test-migrations-lib.sh, issue #2246) - never
+# touches the real system crontab. Mimics the subset of \`crontab\` usage
+# migrations.sh relies on: \`crontab -l\`, \`crontab -\` (write stdin),
+# and \`crontab <file>\`.
+#
+# Several migrations chain \`crontab -l | grep ... | crontab -\` - bash
+# starts every pipeline stage concurrently, so the read-side (\`-l\`) and
+# write-side (\`-\`) invocations of this stub run at the same time. Writing
+# via a plain \`cat > "\$STATE"\` truncates the state file at open time,
+# racing the concurrent read-side's \`cat "\$STATE"\` and intermittently
+# handing it a truncated/empty file. Write to a temp file and \`mv\` it into
+# place instead (same technique the real crontab binary uses) so the state
+# file is always replaced atomically: a concurrent reader sees either the
+# complete old content or the complete new content, never a partial file.
+STATE="$FAKE_CRONTAB_STATE"
+case "\${1:-}" in
+    -l)
+        [ -s "\$STATE" ] && cat "\$STATE" || exit 1
+        ;;
+    -|"")
+        TMP="\$STATE.tmp.\$\$"
+        cat > "\$TMP" && mv "\$TMP" "\$STATE"
+        ;;
+    *)
+        [ -f "\$1" ] && cp "\$1" "\$STATE" || exit 1
+        ;;
+esac
+EOF
+chmod +x "$FAKE_BIN_DIR/crontab"
+
+cat > "$FAKE_BIN_DIR/sudo" <<'EOF'
+#!/bin/bash
+# Fake sudo stub (tests/test-migrations-lib.sh, issue #2246) - swallows all
+# privileged calls (systemctl, usermod, cp, tee, ldconfig, ...) so tests
+# never mutate the real host. `sudo -n true` (passwordless-sudo probe)
+# succeeds; `sudo tee ...` consumes stdin so pipelines don't block.
+if [ "${1:-}" = "-n" ] && [ "${2:-}" = "true" ]; then
+    exit 0
+fi
+if [ "${1:-}" = "tee" ]; then
+    cat >/dev/null
+    exit 0
+fi
+exit 0
+EOF
+chmod +x "$FAKE_BIN_DIR/sudo"
+
+export PATH="$FAKE_BIN_DIR:$PATH"
+
 DRY_RUN=false
 LOBSTER_DIR="$FAKE_LOBSTER_DIR"
 WORKSPACE_DIR="$FAKE_WORKSPACE_DIR"
@@ -113,6 +202,32 @@ assert_file_contains "Migration 96 applies CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT=0 t
 run_migrations
 occurrences=$(grep -c "^CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT=" "$CONFIG_FILE")
 assert_count "Migration 96 is idempotent (running twice does not duplicate the key)" "1" "$occurrences"
+
+#===============================================================================
+# Group 1b: crontab/sudo isolation (issue #2246)
+#===============================================================================
+echo ""
+echo "-- run_migrations() never touches the real system crontab --"
+
+# Sanity check: the crontab-writing migrations actually ran and exercised
+# the stub (proves this is a meaningful regression test, not a no-op because
+# the crontab code paths were skipped for some unrelated reason).
+assert_file_contains "Migration 28 (LOBSTER-LOG-EXPORT) wrote to the fake crontab stub" \
+    "$FAKE_CRONTAB_STATE" "LOBSTER-LOG-EXPORT"
+assert_file_contains "Migration 52 (LOBSTER-GHOST-DETECTOR) wrote to the fake crontab stub" \
+    "$FAKE_CRONTAB_STATE" "LOBSTER-GHOST-DETECTOR"
+
+if [ -n "$REAL_CRONTAB_BIN" ]; then
+    REAL_CRONTAB_AFTER="$("$REAL_CRONTAB_BIN" -l 2>/dev/null || true)"
+    if [ "$REAL_CRONTAB_BEFORE" = "$REAL_CRONTAB_AFTER" ]; then
+        pass "Real system crontab is byte-for-byte unchanged after run_migrations()"
+    else
+        fail "Real system crontab is byte-for-byte unchanged after run_migrations()" \
+            "real crontab changed during the test run - the crontab stub did not intercept every call"
+    fi
+else
+    pass "Real system crontab is byte-for-byte unchanged after run_migrations() (skipped: no crontab binary on this host)"
+fi
 
 #===============================================================================
 # Group 2: install.sh actually wires the runner in (the real bug fix)
--- a/tests/test-migrations-lib.sh
+++ b/tests/test-migrations-lib.sh
@@ -107,6 +107,27 @@ TELEGRAM_BOT_TOKEN=fake-token
 LOBSTER_ADMIN_CHAT_ID=12345
 EOF
 
+# Minimal jobs.json (issue #2246 follow-up): seeds a single enabled job with
+# an absolute `command` path and a `schedule`, matching the exact shape
+# Migration 55 (systemd-timer-from-jobs.json, scripts/lib/migrations.sh
+# ~line 1055) requires to actually run its migration branch rather than be
+# skipped because $WORKSPACE_DIR/scheduled-jobs/jobs.json doesn't exist. That
+# branch is the one embedding the Python subprocess.run(["sudo", "tee", ...])
+# call this suite's sudo stub needs to intercept and prove it caught.
+mkdir -p "$FAKE_WORKSPACE_DIR/scheduled-jobs"
+cat > "$FAKE_WORKSPACE_DIR/scheduled-jobs/jobs.json" <<'EOF'
+{
+  "jobs": {
+    "test-follow-up-job": {
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


sayhar (commented):
## Independent review

**Verdict: PASS**

### What I verified

- Checked out the branch and read the full patched `tests/test-migrations-lib.sh` (not just the diff).
- Confirmed the crontab isolation mechanism is real: a `fake-bin/` dir with stub `crontab` and `sudo` executables is prepended to `$PATH` before `scripts/lib/migrations.sh` is sourced. Grepped `migrations.sh` and confirmed there are ~20+ direct `crontab` call sites and multiple `sudo` call sites (systemctl, usermod, cp, ldconfig, tee) that are genuinely not sandboxed by the existing fake-directory variables — the described bug is real and matches what I independently found investigating tonight's crontab corruption.
- Ran `bash tests/test-migrations-lib.sh` on this host directly from the PR branch, with `crontab -l` snapshotted before and after: **byte-for-byte identical, confirmed via diff**. Result: 9/11 assertions pass. The 2 failures are the pre-existing "Migration 96" `cron-manage.sh`-not-found failures, unrelated to this fix (I confirmed these are host-state issues, not caused by this PR — matches what the PR body claims).
- Verified the atomicity fix for the pipeline race (`crontab -l | grep -v X | crontab -` runs both pipeline stages concurrently): the stub writes to a per-PID temp file and `mv`s it into place, which is a genuinely atomic replace on the same filesystem. This is a real fix for a real race (plain `cat > $STATE` would truncate-then-fill, and a concurrent reader could see a partial/empty file).
- Verified the "byte-for-byte crontab unchanged" assertion is a real check, not a tautology — it resolves and calls the *real* `crontab` binary via its absolute path (captured before the `$PATH` stub is installed), independent of the stub.
- Independently confirmed (outside the test file, in an isolated sandbox) that `subprocess.run(["sudo", ...])` from a Python child process does honor an exported `$PATH` override and gets intercepted by the stub — so the core justification for PATH-shadowing over bash-function-override (catching the embedded Python migration's `subprocess.run` calls) is technically sound.

### One gap worth flagging (not blocking)

The embedded Python migration block that calls `subprocess.run(["sudo", "tee", ...])` (~line 1069 in `migrations.sh`, the systemd-timer-from-jobs.json migration) is gated on `$WORKSPACE_DIR/scheduled-jobs/jobs.json` existing. The test's fake workspace never seeds this file, so **that specific migration branch is skipped in this test run** — the regression suite's "exercised the stub" sanity checks only cover migrations 28 and 52 (direct bash `crontab` calls), not the Python `subprocess.run(["sudo", ...])` path the PR description specifically calls out as the reason for choosing PATH-shadowing.

This isn't a functional bug — the `sudo`/`crontab` stubs are blanket `$PATH` interceptors, so any migration that shells out to those binaries is protected regardless of whether the test explicitly exercises that branch, and I confirmed independently that the Python subprocess path does get intercepted. But the test suite doesn't *prove* this for the specific code path the PR's design rationale leans on. Worth a follow-up: seed a minimal `jobs.json` in the fake workspace so that branch actually runs and the sudo-tee interception gets its own explicit assertion, the same way migrations 28/52 do for crontab.

### Summary

The described bug is real, the fix is real, and I reproduced the "real crontab untouched" result directly against this host's production crontab (diffed before/after, identical). The atomic-write race fix is correct. The one gap (untested Python subprocess.run sudo path) doesn't undermine the safety of the fix, just the completeness of its own regression coverage — recommend a fast-follow, not a blocker.

sayhar (commented):
Verdict: NEEDS-WORK

Independent fresh re-review of the current PR state (base `origin/main` + both commits, HEAD `ed528595`), following up on the prior PASS review that flagged the jobs.json/Migration-55 coverage gap.

## What I verified as correct

- Scope: confirmed only `tests/test-migrations-lib.sh` changed (149 insertions, 0 deletions) across the full PR diff — matches the PR's stated "additions: 149, deletions: 0".
- Read the full current file (not just the new-commit diff) — the jobs.json seeding and sudo-tee logging integrate cleanly with the pre-existing crontab stub/assertions; no interaction bugs, no accidental interference with migrations 28/52's assertions.
- The new `jobs.json` fixture (`command`/`schedule` shape) does correctly trigger Migration 55's branch — traced through `scripts/lib/migrations.sh` (`local JOBS_JSON=...`, `timer_path = SYSTEMD_DIR / f"{UNIT_PREFIX}{name}.timer"`) and confirmed the assertion's expected string (`/etc/systemd/system/lobster-test-follow-up-job.timer`) is exactly what the embedded Python computes for a job named `test-follow-up-job`. Not a tautology — I ran the test with the sudo stub logging temporarily disabled and confirmed the assertion fails as expected.
- Ran `bash tests/test-migrations-lib.sh` on a real host (systemd running): 10/12 pass, with the only failures being the pre-existing, unrelated "Migration 96" `cron-manage.sh`-not-found issue (confirmed present on unmodified `origin/main` too) — matches what the PR/prior review describe. All 3 new crontab/sudo-isolation assertions (28, 52, and the new 55 one) pass here.
- **Safety property directly verified, not just read**: captured `crontab -l` before and after the run on this host — byte-for-byte identical (`diff` clean). No file appeared under `/etc/systemd/system/lobster-test-follow-up-job.*` on the real host. The isolation genuinely works when it engages.

## Blocking issue

**The new Migration-55 assertion fails in the project's own Docker test container** (`tests/docker/Dockerfile.test`), the exact "clean Debian" environment the PR's "Verified" section relies on for its safety claims.

Reproduced directly: built `tests/docker/Dockerfile.test` and ran the test inside it.

- At the original commit (`787dcf4c`, pre-follow-up): **11/11 pass** in the container — matches the PR body's claim.
- At current HEAD (`ed528595`, with the follow-up commit): **11/12 pass** in that same container. The new assertion fails:
  ```
  FAIL Migration 55 (systemd-timer-from-jobs.json) exercised the sudo tee stub
       expected '.../fake-sudo-tee-log' to contain: /etc/systemd/system/lobster-test-follow-up-job.timer
  ```

Root cause: Migration 55 is gated in `migrations.sh` on `pidof systemd >/dev/null 2>&1` succeeding (in addition to `jobs.json` existing and `python3` being present). `tests/docker/Dockerfile.test`'s `debian:bookworm-slim` container never runs systemd as PID 1 and has no systemd process at all, so `pidof systemd` fails, the migration's whole body is skipped, `sudo tee` is never invoked, and the new hard `assert_file_contains` fails — even though the crontab/sudo isolation itself is working exactly as intended (it's simply never exercised for this specific migration in that environment).

This is a real, environment-dependent regression introduced by the follow-up commit, not a flake: it's 100% reproducible and it specifically breaks in the container the PR uses as its main "clean environment" validation vector. The fix is straightforward — mirror the pattern already used a few lines below for the real-crontab-unchanged check (`if [ -n "$REAL_CRONTAB_BIN" ] ... else pass ... (skipped)`): guard the new assertion (or better, guard/stub the `pidof systemd` check itself, e.g. add a fake `pidof` to the fake-bin dir the same way `crontab`/`sudo` are stubbed) so the assertion degrades to a skip-pass with a clear message when the migration's own preconditions aren't met, rather than hard-failing.

## Non-blocking nits

- The PR description / "Tests run" checklist should be updated once fixed — it currently says 11/11 in the Docker container, which was true for the original commit but is not true for current HEAD (11/12, one new failure).
- Given the above, worth double-checking whether `pidof` is even installed by default on other hosts CI might eventually run this on (procps isn't in the Dockerfile's package list at all here; `pidof` happened to be present via another dependency in this run, but that's incidental).

## Other checks

- No `list_decisions()`-style architectural decision mechanism found in this repo (only a `memory/canonical-templates/pending-decisions.md` template, not a queryable log) — noting as requested, nothing applicable found.
- No CI workflow currently invokes `test-migrations-lib.sh` (no `.github/workflows` reference), so this gap won't be caught by CI today — another reason to fix it now rather than defer, since the only safety net is exactly the manual Docker run this breaks.

Once the Migration 55 assertion is made environment-robust (skip-pass when the migration's preconditions, including `pidof systemd`, aren't met — same pattern as the existing real-crontab guard), this is ready to sign off. The core crontab/sudo isolation fix itself is sound and independently verified safe on a real host.

sayhar (commented):
Verdict: PASS

Fresh, independent third-pass review of the full current PR (all 3 commits, base `main` vs current HEAD `530c28cc`). I did not rely on the PR description's claims — I rebuilt the environment and re-ran everything myself.

**What I verified directly:**

1. **Dev host, inherited env** (`bash tests/test-migrations-lib.sh`): 10/12 pass. The 2 failures are the pre-existing Migration 96 host-state leak (`CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT` inherited from the agent shell) — confirmed unrelated to this PR's crontab/sudo/pidof work, out of scope as described.
2. **Dev host, clean env** (`env -u CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT bash tests/test-migrations-lib.sh`): **12/12 pass**, including the Migration 55 sudo-tee assertion.
3. **Docker container** (`tests/docker/Dockerfile.test`, `debian:bookworm-slim`, no systemd as PID 1 — this is the exact environment that broke on the second review pass): built the image and ran the test script directly inside it (the Dockerfile's own CMD runs a different pytest target, so I invoked the script manually). Confirmed `pidof systemd` genuinely fails in this container (`pidof_exit=1`) on the real binary — then ran the test suite: **12/12 pass**, including "Migration 55 (systemd-timer-from-jobs.json) exercised the sudo tee stub." This confirms the fix works as intended: the new `pidof` stub in the fake-bin `$PATH` directory makes Migration 55's `pidof systemd` gate succeed regardless of host, so the migration body actually executes and gets exercised for real — not skip-passed. This is the correct fix (real coverage > conditional skip) and it resolves the exact regression flagged in review #2.
4. **Real crontab safety, re-verified end-to-end**: captured `crontab -l | md5sum` on the dev host before any test runs, ran the test suite three times (twice on host, once via a nested container run), and confirmed the hash is byte-identical afterward. Also confirmed the containerized run shows `no crontab for testuser` both before and after. The suite's own internal regression guard (comparing `$REAL_CRONTAB_BIN -l` before/after via the pre-stub resolved path) also passed in both environments.
5. **Scope check**: `git diff --stat` from base to HEAD across all 3 commits touches exactly one file, `tests/test-migrations-lib.sh` (+172/-0). No other files touched.
6. **Root-cause verification of the pidof gate claim**: grepped `scripts/lib/migrations.sh` for `pidof systemd` — 3 call sites (lines 897, 1069, 1722). Line 1069 is Migration 55 (now genuinely exercised, confirmed above). The other two (mcp-local service install, transcription service enable/start) are additionally gated behind `-f` checks on service files (`$LOBSTER_DIR/services/lobster-transcription.service`, etc.) that don't exist in the test's fake `$LOBSTER_DIR` fixture, so stubbing `pidof` globally doesn't change their (no-op) behavior in this suite, matching the PR description's claim.
7. **PR description accuracy**: the "Follow-up" section's stated test results (10/12 host w/ inherited env, 12/12 clean env, 12/12 Docker) match what I independently reproduced. No stale claims found.

**Non-blocking observations (not new, already flagged/acknowledged in the PR description, no action needed here):**
- The Migration 96 host-state leak (reading `CLAUDE_CODE_MCP_TOOL_IDLE_TIMEOUT` from the inherited shell env rather than only the fixture) is real but predates this PR and is out of scope — correctly flagged as a possible separate follow-up.
- The blanket `pidof` stub (always exit 0) is a deliberate, reasonable simplification over per-call-site environment detection, and its blast radius is correctly scoped to only the migrations whose `-f` preconditions are actually met in this test's fixtures.

**Conclusion:** This PR is ready for sign-off. It fixes a real, previously-reproduced production incident (real crontab entries written from a test's throwaway temp path), the fix is correctly scoped to the test harness only, the regression from the second review pass (Docker/non-systemd environment) is genuinely resolved (verified by directly reproducing the failure condition and confirming the fix corrects it), and the real-crontab safety property holds byte-for-byte across every environment I tested.

PR #2253, opened 2026-08-23
fix: retry Telegram photo downloads on transient network timeouts

### LINKED ISSUE #2252: Photo message lost on Telegram getFile timeout, no retry
## What happened

At 2026-08-22 21:17:48 EDT, a user-sent photo failed to process. The bot replied "❌ Failed to process photo." and the message was never written to the inbox — total data loss, no way to recover it short of the user re-sending.

## Root cause

`handle_photo_message` in `src/bot/lobster_bot.py` (line ~1018) calls `context.bot.get_file(photo.file_id)` with no retry. This particular call hit an `httpx.ReadTimeout` → `telegram.error.TimedOut`, caught by the broad `except Exception` at line 1072, logged, and surfaced to the user as a terminal failure.

```
Traceback (most recent call last):
  ...
  File ".../src/bot/lobster_bot.py", line 1018, in handle_photo_message
    file = await context.bot.get_file(photo.file_id)
  ...
telegram.error.TimedOut: Timed out
```

Two contributing factors:

1. **Timeouts are too tight for file operations.** `Application.builder().token(BOT_TOKEN).write_timeout(30).build()` (line ~2182) only raises `write_timeout`; `read_timeout`/`connect_timeout` are left at the python-telegram-bot default (5s), which is short for `getFile` + download under any network jitter.
2. **No retry on transient failures.** A single `ReadTimeout` on `get_file` immediately drops the photo with no automatic retry, even though this is a classic transient-failure case.

This is a single occurrence so far (checked `telegram-bot.log` + `logs/archive/`, only one match for "Error handling photo message" ever), so it's rare — but the failure mode is 100% data loss with no recovery path, which is worth hardening against.

## Also affected

`_handle_media_group_photo` (line ~1088) has the identical `get_file`/`download_to_drive` pattern with no retry — and is actually worse, since on failure it just logs and `return`s with **no user-facing error at all**, so a dropped photo in an album is silent.

## Suggested fix

- Bump `connect_timeout`/`read_timeout` on the `Application.builder()` (e.g. to 15-20s) alongside the existing `write_timeout(30)`.
- Add a small retry helper (2-3 attempts, short backoff) around the `get_file` + `download_to_drive` calls in both `handle_photo_message` and `_handle_media_group_photo`, catching `telegram.error.TimedOut` / `NetworkError`.
- For the media-group path specifically, surface a user-facing failure message on exhausted retries instead of silently dropping the photo.

Filed after a user asked to "audit and fix" a photo-processing failure; direct edit was blocked by the system-file edit guard (`LOBSTER_DEBUG` gate), so routing through the normal branch/PR workflow instead.

sayhar:
## Problem

A user-sent photo was lost entirely: no inbox entry was written, and the bot replied "❌ Failed to process photo." `handle_photo_message`'s `context.bot.get_file(photo.file_id)` call hit a `telegram.error.TimedOut` (an `httpx.ReadTimeout`) with no retry, and the failure propagated to a terminal user-facing error. Two contributing factors:

1. `Application.builder()` only set `write_timeout(30)`; `read_timeout`/`connect_timeout` were left at python-telegram-bot's 5s default — tight for a `getFile` + download round trip.
2. There was no retry on this transient failure class.

`_handle_media_group_photo` (the media-group/album path) has the identical unretried `get_file`/`download_to_drive` pattern, and is actually worse: on failure it just logs and returns, with **no user-facing error at all** — a dropped photo in an album was completely silent.

This has only reproduced once in production (checked `telegram-bot.log` + `logs/archive/`), so it's rare, but the failure mode is 100% data loss with no recovery path.

## What changed

- **Timeout headroom:** `Application.builder()` now sets `connect_timeout(20)` and `read_timeout(20)` alongside the existing `write_timeout(30)`, instead of leaving both at PTB's 5s default.
- **Shared retry helper:** added `_download_telegram_file_with_retry()` — a pure function wrapping `get_file` + `download_to_drive` in a 3-attempt retry loop with linear backoff, catching `telegram.error.TimedOut` / `telegram.error.NetworkError`. Both `handle_photo_message` and `_handle_media_group_photo` now call this single helper instead of each duplicating the unretried `get_file`/`download_to_drive` pair.
- **Media-group failure now surfaces to the user:** `_handle_media_group_photo` sends "❌ Failed to process a photo in this album." when retries are exhausted, instead of silently dropping the photo with only a log line.

Functional patterns: the retry helper is a small pure function (inputs → retry policy → result-or-raise, no shared mutable state beyond the two module-level constants `FILE_DOWNLOAD_MAX_ATTEMPTS`/`FILE_DOWNLOAD_BACKOFF_SECONDS`), and it's composed into both call sites rather than duplicated.

This is scoped as a minimal, targeted fix — timeout bump + retry wrapper — not a broader refactor of the photo-handling pipeline, matching how rarely this failure mode has actually occurred.

## Out of scope

No changes to `handle_document_message`, voice/audio handling, the media-group flush/buffer logic, or the inbox message schema. Only the `get_file`/`download_to_drive` call sites and the `Application.builder()` timeouts were touched.

## Flow: media-group failure path

```
Before:
  get_file/download fails -> log.error() -> return   (photo silently dropped, no album entry, no user notice)

After:
  get_file/download retried up to 3x (TimedOut/NetworkError only)
    -> succeeds: unchanged behavior, photo buffered into the album as before
    -> exhausts retries: log.error() -> reply_text("Failed to process a photo in this album.") -> return
```

The single-photo path (`handle_photo_message`) already had a user-facing failure reply; its `get_file`/`download_to_drive` call is now routed through the same retry helper, so it benefits from the same retry budget before that reply fires.

## Tests run

- [x] `uv run pytest tests/unit/test_bot/test_photo_download_retry.py -v` — 6 passed, 0 failed. Covers: succeeds without retry, recovers after N transient `TimedOut`/`NetworkError` failures within budget, raises after exhausting all attempts, does not retry (fails fast) on a non-`NetworkError` exception (`Forbidden`), and the media-group path now sends a user-facing failure reply on exhausted retries (regression test for the previously-silent drop).
- [x] `uv run pytest tests/unit/test_bot/` — 248 passed, 20 failed. All 20 failures are pre-existing and unrelated to this change: `ModuleNotFoundError: No module named 'slack_bolt'`/`'slack_sdk'` (optional dependency not installed in this environment) and one pre-existing `test_slack_router_config.py` assertion failure. No photo/message-handling test regressed.
- [x] `uv run python -c "import ast; ast.parse(open('src/bot/lobster_bot.py').read())"` — syntax check passes.

## Manual test

No live Telegram API call was made — this environment has no test bot token, and the failure mode (a real `ReadTimeout` from Telegram's servers) isn't reproducible on demand without one. Verification was done by unit-testing the retry helper in isolation with a mocked `context.bot.get_file` that raises `telegram.error.TimedOut`/`NetworkError` N times before succeeding (or exhausts retries), per the issue's own suggested test approach. No systemd/cron/DB changes are involved — this is a pure Python code change to an existing long-running bot process, picked up on next bot restart/deploy like any other code change to this file.

## Definition of Done

- [x] Unit tests pass with proper mocking (no production Telegram calls in automated tests)
- [ ] Integration/manual test against live Telegram — not run; no test bot token available in this environment, and the original failure (`ReadTimeout`) is not reliably reproducible on demand. Recommend a maintainer either watches the next real occurrence resolve automatically, or does a manual photo-send test in production after merge if that reassurance is wanted before/after deploy.
- [x] User-visible outcome: the media-group failure message text change is Telegram-observable but not independently verified against a live chat (see above) — reasoning verified via unit test asserting `reply_text` is awaited with the expected text.
- [x] Side-effect audit: this only affects the failure path of photo downloads (retry + one new failure reply); no new writes to shared state, no risk of message floods (bounded to 3 attempts per photo), no change to successful-path behavior.
- [x] External service calls: `context.bot.get_file`/`download_to_drive` are mocked in all automated tests; no manual runs were made against the live Telegram API.

Closes #2252

### DIFF
--- a/src/bot/lobster_bot.py
+++ b/src/bot/lobster_bot.py
@@ -78,8 +78,14 @@ from dataclasses import dataclass, field
 from typing import Optional
 
 from telegram import Update, InlineKeyboardButton, InlineKeyboardMarkup, ReplyParameters
+from telegram.error import TimedOut, NetworkError
 from telegram.ext import Application, CommandHandler, MessageHandler, CallbackQueryHandler, ChatMemberHandler, MessageReactionHandler, filters, ContextTypes
 from collections import deque
+
+# Retry policy for downloading files from Telegram (get_file + download_to_drive).
+# A single transient network hiccup should not drop a user's photo permanently.
+FILE_DOWNLOAD_MAX_ATTEMPTS = 3
+FILE_DOWNLOAD_BACKOFF_SECONDS = 1.0
 
 
 # URLs longer than this are copy-paste targets (e.g. OAuth flows).  Telegram
@@ -998,6 +1004,41 @@ async def handle_callback_query(update: Update, context: ContextTypes.DEFAULT_TY
     log.info(f"Wrote callback message to inbox: {msg_id}")
 
 
+async def _download_telegram_file_with_retry(
+    context: ContextTypes.DEFAULT_TYPE,
+    file_id: str,
+    dest_path: Path,
+    *,
+    max_attempts: int = FILE_DOWNLOAD_MAX_ATTEMPTS,
+    backoff_seconds: float = FILE_DOWNLOAD_BACKOFF_SECONDS,
+) -> None:
+    """Download a Telegram file (get_file + download_to_drive) with retry.
+
+    Retries on telegram.error.TimedOut / telegram.error.NetworkError, which are
+    the transient-failure classes for a getFile + download round trip (e.g. a
+    ReadTimeout under network jitter). Note NetworkError is python-telegram-bot's
+    base class for several non-transient errors too (e.g. BadRequest), so those
+    are retried as well; only exceptions outside the NetworkError/TimedOut
+    hierarchy (e.g. Forbidden) fail fast. Uses linear backoff between attempts.
+    Re-raises the last exception if all attempts are exhausted.
+    """
+    last_exc: Optional[Exception] = None
+    for attempt in range(1, max_attempts + 1):
+        try:
+            file = await context.bot.get_file(file_id)
+            await file.download_to_drive(dest_path)
+            return
+        except (TimedOut, NetworkError) as e:
+            last_exc = e
+            log.warning(
+                f"Telegram file download attempt {attempt}/{max_attempts} failed "
+                f"for file_id={file_id}: {e}"
+            )
+            if attempt < max_attempts:
+                await asyncio.sleep(backoff_seconds * attempt)
+    raise last_exc
+
+
 async def handle_photo_message(update: Update, context: ContextTypes.DEFAULT_TYPE, msg_id: str):
     """Handle photo messages: download and save to inbox with metadata."""
     user = update.effective_user
@@ -1014,10 +1055,9 @@ async def handle_photo_message(update: Update, context: ContextTypes.DEFAULT_TYP
         # Get the largest photo size
         photo = message.photo[-1]
 
-        # Download the photo
-        file = await context.bot.get_file(photo.file_id)
+        # Download the photo (with retry on transient network/timeout errors)
         image_path = IMAGES_DIR / f"{msg_id}.jpg"
-        await file.download_to_drive(image_path)
+        await _download_telegram_file_with_retry(context, photo.file_id, image_path)
         log.info(f"Downloaded photo to: {image_path}")
 
         caption = message.caption or ""
@@ -1087,13 +1127,16 @@ async def _handle_media_group_photo(update: Update, context: ContextTypes.DEFAUL
 
     try:
         photo = message.photo[-1]
-        file = await context.bot.get_file(photo.file_id)
         # Use msg_id (which is unique per photo update) as the filename
         image_path = IMAGES_DIR / f"{msg_id}.jpg"
-        await file.download_to_drive(image_path)
+        await _download_telegram_file_with_retry(context, photo.file_id, image_path)
         log.info(f"Downloaded media group photo to: {image_path}")
     except Exception as e:
         log.error(f"Error downloading media group photo: {e}", exc_info=True)
+        try:
+            await message.reply_text("❌ Failed to process a photo in this album.")
+        except Exception:
+            log.error("Failed to send media-group photo failure notice", exc_info=True)
         return
 
     if group_id not in _media_group_buffers:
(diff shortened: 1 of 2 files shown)


sayhar (commented):
**Verdict: PASS**

Independent review — checked out the PR branch directly (`fix/issue-2252-photo-timeout-retry`) and ran the tests myself rather than trusting the PR description.

What I checked:
- Read the full diff and surrounding code in `src/bot/lobster_bot.py` (both `handle_photo_message` and `_handle_media_group_photo`, before and after).
- No `list_decisions()`-style architectural-decision mechanism found in this repo (searched for "decision"/"ADR" patterns) — noting as not applicable rather than skipped.
- Retry logic: 3 attempts, linear backoff (1s, 2s), catches `TimedOut`/`NetworkError` only. Verified via `python -c "... .__mro__"` that `BadRequest` is indeed a `NetworkError` subclass (so it gets retried, as the docstring honestly flags) while `RetryAfter` is *not* a `NetworkError`/`TimedOut` subclass, so a rate-limit response correctly fails fast instead of being retried against Telegram's own backoff hint — good, that's the behavior you want.
- No duplicate-download/side-effect risk: `download_to_drive` retries write to the same `dest_path` each attempt, which is an idempotent overwrite; no external side effects are duplicated (only a read from Telegram's API, no writes to Telegram).
- Confirmed the "no retry today" claim by diffing against `main`: prior code had bare `get_file`/`download_to_drive` calls in both call sites with no retry wrapper, and `_handle_media_group_photo` really did just `log.error()` + silent `return` on failure — the "silent drop" bug description checks out.
- Tests: ran `uv run pytest tests/unit/test_bot/test_photo_download_retry.py -v` myself on the PR branch — all 6 pass, covering first-attempt success, recovery after N transient failures, exhaustion/re-raise, fail-fast on non-network exception (`Forbidden`), and the media-group regression test for the new user-facing failure reply.
- Scope: `git diff --stat main...pr-branch` shows only `src/bot/lobster_bot.py` (+66/-6) and the new test file (+164) — matches the PR's stated scope exactly, no drive-by changes.

Non-blocking nits:
- `BadRequest` (permanent errors, e.g. invalid file_id) will be retried 3x with ~3s of cumulative backoff before failing, since it's a `NetworkError` subclass upstream. Not incorrect, just a minor UX delay on a genuinely non-transient error class — the docstring already calls this out explicitly, so it's a documented tradeoff rather than an oversight.
- `last_exc` is typed `Optional[Exception]` and the final `raise last_exc` would raise `None` if `max_attempts` were ever 0 — not reachable today since both call sites use the constant default (3), but a `assert last_exc is not None` or an explicit guard would make the function robust to future misuse.
- New test file imports `asyncio` but doesn't appear to use it directly (relies on `pytest.mark.asyncio` only) — harmless unused import.

No blocking issues.

Issue #2176, opened 2026-08-07
agent-monitor.py: dispatcher row can be misclassified GHOST_CONFIRMED via PID path (5th occurrence of dispatcher-exclusion gap)


aeschylus:
## Problem

Since PR #2152 merged (2026-08-05), the ghost-kill cron (`agent-monitor.py --alert --mark-failed`, every 30 min) and the `on-fresh-start.py` SessionStart hook (`agent-monitor.py --mark-failed`, on every dispatcher restart) have intermittently marked the **live dispatcher's own session row** as a "ghost agent" and dropped a false `agent_failed` notification. ~10 occurrences since the PR merged, 0 before it. Confirmed live in production: the current `agent_sessions.db` row for `id='lobster-dispatcher'` is sitting at `status='failed'` right now, despite the dispatcher being alive.

## Root cause (verified against code + git history, not just the prior summary)

PR #2152 added a new "PID ground truth" classification path to `scripts/agent-monitor.py`'s `classify()`:

```python
if pid_alive is not None:
    return "HEALTHY" if pid_alive else "GHOST_CONFIRMED"
```

This runs **before** the legacy age/output-file heuristic and does not go through `output_file`. Previously, the dispatcher's row (which always has `output_file=NULL`) could only ever land in `STALE_NO_FILE` — never `GHOST_CONFIRMED` — which is why the *existing* dispatcher guard in `mark_failed_all_ghosts()` only filters the `stale_no_file` list (by a static `agent_id == "lobster-dispatcher"` string match), not the `confirmed` (GHOST_CONFIRMED) list. That guard is also never applied at all in `send_alert()`.

`load_running_agents()` — the query that feeds `classify_agent()` — does not select `agent_type` and does not apply the shared `utils.agent_types.DISPATCHER_EXCLUSION_SQL` filter that `session_store.cleanup_stale_running_sessions()` already uses for the equivalent query. So the dispatcher's row is not excluded at the query boundary at all in this script.

[...]

PR #2254, opened 2026-08-24
fix: exclude dispatcher's own session row even when agent_type is NULL

### LINKED ISSUE #2226: periodic-self-check.sh fires spurious [1 agents pending] every 3 min when dispatcher's own session row has NULL agent_type
## Summary

Since the dispatcher restarted at ~04:24 UTC on 2026-08-13, `periodic-self-check.sh` (cron, every 3 min) has been injecting `status? (Self-check) [1 agents pending]` into the inbox **every single cycle, indefinitely** — 6+ consecutive firings as of 04:48 UTC, continuing at time of filing, with zero actual subagent work in flight.

## Root cause

`agent_sessions.db` currently contains exactly one `running` row for chat `6645894734`:

```
id=lobster-dispatcher | agent_type=NULL | status=running | chat_id=6645894734
  | spawned_at=2026-08-13T04:24:45 | completed_at=NULL
  | description="Lobster dispatcher main loop"
```

This is the dispatcher's own self-registration row (per the bootup instructions in `.claude/sys.dispatcher.bootup.md` step 0/1: `session_start(agent_type="dispatcher", agent_id="lobster-dispatcher", description="Lobster dispatcher main loop", chat_id=<ADMIN_CHAT_ID>)`). The `description` field matches that template exactly, confirming this row was written by the documented dispatcher self-registration call — but `agent_type` landed as `NULL` instead of `'dispatcher'`.

`periodic-self-check.sh` counts pending agents via:
```sql
SELECT COUNT(*) FROM agent_sessions
WHERE status IN ('running','starting') AND COALESCE(agent_type, '') != 'dispatcher'
```
(`DISPATCHER_EXCLUSION_SQL`, shared source of truth in `scripts/lib/agent_sessions.sh` / `src/utils/agent_types.py`, previously fixed at 5 call sites per issue #781 / PR #2099 / PR #2103 / issue #2176.)

Because `agent_type` is `NULL` (not the empty string, and not `'dispatcher'`), `COALESCE(agent_type, '') != 'dispatcher'` evaluates to `'' != 'dispatcher'` → **TRUE**, so this row is *not* excluded and is counted as 1 pending agent. Verified directly against the live DB:

```
$ sqlite3 agent_sessions.db "SELECT COUNT(*) FROM agent_sessions WHERE status IN ('running','starting') AND COALESCE(agent_type,'') != 'dispatcher';"
1
```

The exclusion mechanism itself (`DISPATCHER_EXCLUSION_SQL`) is correctly designed and well-tested — the bug is that the dispatcher's own `session_start` call sometimes lands with `agent_type` unset. `agent_type` is an **optional** parameter in the `session_start` MCP tool schema (`inputSchema.required = ["agent_id", "description", "chat_id"]`, `src/mcp/inbox_server.py` ~line 2934) — nothing enforces that a call using `agent_id="lobster-dispatcher"` must also carry `agent_type="dispatcher"`. If the dispatcher LLM's bootup turn omits the optional param (as apparently happened this restart), the row silently loses its exclusion tag for the rest of that dispatcher's lifetime — i.e. **every 3-minute cron cycle from restart until the next restart fires a spurious self-check**, forever.

### Dead safety net

There is a second, independent mechanism intended to guard exactly this case: `hooks/write-dispatcher-session-id.py`, a `SessionStart` hook that writes a `INSERT OR IGNORE` row tagged `agent_type='dispatcher'` (keyed by the real Claude session UUID, not the literal string `lobster-dispatcher`) whenever it detects the current session is the dispatcher. Its own docstring calls this "permanent insurance" against exactly this failure mode (issue #781).

This hook is **not registered** in `~/.claude/settings.json`'s `SessionStart` hooks list on this deployment — only `on-fresh-start.py` and `inject-bootup-context.py` are wired up. Confirmed via:
```
$ grep -rn "write-dispatcher-session-id" ~/.claude/settings.json ~/.claude/settings.local.json
(no matches)
```
So the belt-and-suspenders protection against a missing `agent_type` tag currently does nothing.

## Impact

Every dispatcher session, for its entire lifetime (until the next restart), generates one spurious `status? (Self-check) [1 agents pending]` message every 3 minutes — pure inbox noise and wasted dispatcher cycles reasoning about a "pending agent" that doesn't exist. Confirmed still firing at time of filing (04:30, 04:33, 04:36, 04:39, 04:42, 04:45, 04:48 UTC and counting).

## Suggested fix

Pick at least one (ideally both, for defense in depth):

1. **Make the exclusion robust to a missing tag on the known-static id**, not just `agent_type`: extend `DISPATCHER_EXCLUSION_SQL` (or add a second guard specifically in `periodic-self-check.sh`) to also exclude `id = 'lobster-dispatcher'` regardless of `agent_type`, mirroring the belt-and-suspenders pattern `agent-monitor.py`'s `_is_dispatcher_agent()` already uses (`is_dispatcher_agent_type(row.agent_type) or row.agent_id == _DISPATCHER_AGENT_ID`).
2. **Wire up `hooks/write-dispatcher-session-id.py`** in `~/.claude/settings.json`'s `SessionStart` hooks (per the install snippet already documented in its own docstring) so the "permanent insurance" row actually gets written, independent of whether the dispatcher's manual `session_start` call included `agent_type`.

Not opening a PR for this — filing for review since it's inbox noise, not urgent/breaking.

aeschylus:
## Problem

Since the dispatcher restarted at ~04:24 UTC on 2026-08-13, `periodic-self-check.sh` (cron, every 3 min) has been injecting `status? (Self-check) [1 agents pending]` into the inbox every single cycle, indefinitely, with zero real subagent work in flight — confirmed still firing live at the time this issue was filed.

`agent_type` is an *optional* parameter on the `session_start` MCP tool. When the dispatcher's own bootup `session_start(agent_id="lobster-dispatcher", ...)` call omits it, the row lands in `agent_sessions.db` with `agent_type=NULL`. `DISPATCHER_EXCLUSION_SQL`'s `COALESCE(agent_type, '') != 'dispatcher'` then evaluates to `'' != 'dispatcher'` → `TRUE`, so the row is **not** excluded, and gets counted as a pending agent on every 3-minute cycle.

This is the same "dispatcher's own row miscounted as a subagent" bug class documented in `docs/engineering-lessons-learned.md` ("Dispatcher-Exclusion Bug"), now confirmed as its **7th** occurrence (issue #781, PR #2099, PR #2103, PR #2152/issue #2176 ×2, and now this). `scripts/agent-monitor.py`'s `_is_dispatcher_agent()` had already closed this exact gap for its own call site via a belt-and-suspenders `agent_id` check (issue #2176) — but that fix lived only at that one call site.

## What changed

Applies the same belt-and-suspenders fallback **once, at the shared source of truth**, so every call site that already depends on it is protected simultaneously instead of patching `periodic-self-check.sh` alone and leaving the same gap open at the other five:

- `src/utils/agent_types.py` — adds `DISPATCHER_AGENT_ID = "lobster-dispatcher"`; `DISPATCHER_EXCLUSION_SQL` now also excludes `id = 'lobster-dispatcher'`; `is_dispatcher_row()` now falls back to the `id` check when `agent_type` is `None`
- `scripts/lib/agent_sessions.sh` — mirrors the same extension for the shell-side fragment (sourced directly by `periodic-self-check.sh` via raw `sqlite3`, bypassing the Python module entirely)
- `scripts/agent-monitor.py` — now imports `DISPATCHER_AGENT_ID` from `utils.agent_types` instead of hand-copying the literal `"lobster-dispatcher"`, removing one more duplicate definition of the same value

Because `DISPATCHER_EXCLUSION_SQL` / `is_dispatcher_row()` is the single shared definition consumed by `session_store.py` (x2), `inbox_server.py` (x2), `agent-monitor.py`, `health-check-v3.sh`, and `periodic-self-check.sh`, this one change closes the NULL-`agent_type` gap at all of them, not just the one that was reported firing live.

## Functional patterns

`is_dispatcher_row()` remains a pure predicate over an immutable session dict — no new mutation introduced. `DISPATCHER_EXCLUSION_SQL` stays a plain composed string constant (data, not control flow), so the belt-and-suspenders `id` clause is additive rather than a rewrite of the existing check. The `agent-monitor.py` change replaces a hand-copied constant with an import, reducing duplicate state rather than adding any.

## Out of scope

The issue also suggested (as an alternative, "ideally both") wiring `hooks/write-dispatcher-session-id.py` into `SessionStart` as defense-in-depth. While investigating I found `scripts/upgrade.sh`'s Migration 86 actively *removes* that hook from existing installs ("no longer needed... deleted entirely", issue #1908) while `install.sh` still wires it in on fresh installs — a pre-existing, unrelated contradiction in the repo's own hook-wiring logic, unconnected to the bug this PR fixes. Reconciling that is a separate investigation; the fix in this PR fully resolves the reported false-positive on its own, independent of that hook's wiring state.

No behavior change to anything else in `periodic-self-check.sh`, `health-check-v3.sh`, `session_store.py`, or `inbox_server.py` — only the shared exclusion predicate gained the `id` fallback.

## Tests run

- [x] `uv run pytest tests/unit/test_agent_types.py tests/unit/test_ghost_detector_dispatcher_exclusion.py tests/unit/test_mcp_server/test_ghost_session_fixes.py tests/unit/test_agent_monitor_pid_liveness.py` — 82 passed (new: 5 tests reproducing the NULL-`agent_type` false positive at the Python-predicate, raw-SQL-fragment, and `load_running_agents()` query-boundary levels, plus 1 negative case confirming a real subagent with `agent_type=None` is *not* swallowed by the `id` fallback)
- [x] `bash tests/test-periodic-self-check-dispatcher-null-agent-type.sh` — 6 passed (new file: sources the real shared shell fragment and runs periodic-self-check.sh's exact `PENDING_COUNT` query against a DB shaped like the live bug report, plus a regression case for the already-working `agent_type='dispatcher'` path and a syntax check)
- [x] `bash tests/test-health-check-count-active-subagents.sh` — 6 passed (added 1 case: `count_active_subagents()` — a second call site sharing the same fragment — also now excludes a NULL-`agent_type` dispatcher row)
- [x] `uv run pytest tests/unit/` (full suite) — 4379 passed, 45 failed, 32 skipped. All 45 failures are pre-existing and unrelated (migration/PII-scan/push-token-endpoint/context-monitor tests) — confirmed by running the same failing files against an unmodified `origin/main` checkout, which reproduces the identical 7 failures checked directly (`test_on_compact_write_claude_session_id.py`); the rest are the same class of environment-dependent tests untouched by this diff.
- [x] `uv run ruff check` on all touched Python files — 18 pre-existing style findings before and after this diff (byte-identical list except line-number shifts); zero new findings introduced.
- [x] `bash -n` syntax check on all touched/added shell files — clean.

**Falsifiability check:** stashed the three source changes (`src/utils/agent_types.py`, `scripts/lib/agent_sessions.sh`, `scripts/agent-monitor.py`) and reran the new tests — `test_agent_types.py` failed to import (`DISPATCHER_AGENT_ID` doesn't exist pre-fix), and `test-periodic-self-check-dispatcher-null-agent-type.sh` failed 2/6 with the exact wrong counts (`expected '0', got '1'`, `expected '1', got '2'`) — then restored the fix and reran everything clean.

## Manual test

This is a cron script (`periodic-self-check.sh`) that reads/writes real filesystem state (`agent_sessions.db`, the inbox directory), so automated tests alone aren't sufficient evidence — ran the actual script end-to-end, not just the extracted query:

1. Built a throwaway `agent_sessions.db` with the exact row shape from the live bug report (`id='lobster-dispatcher'`, `agent_type=NULL`, `status='running'`) in an isolated temp `LOBSTER_MESSAGES` dir, with `AGENT_TASKS_DIR` pointed at an empty directory to isolate the run from this instance's own real completed-task files.
2. **Pre-fix** (source changes stashed): ran `scripts/periodic-self-check.sh` against that DB for real — it wrote `{"text": "status? (Self-check) [1 agents pending]", ...}` into the temp inbox, reproducing the exact reported symptom byte-for-byte.
3. **Post-fix** (restored): ran the identical script against the identical DB — inbox stayed empty (0 files), confirming the fix closes the gap through the real script, not just the unit-tested fragment.
4. Note: `periodic-self-check.sh`'s rate-limit guard (`STATE_DIR="$LOBSTER_INSTALL_DIR/.state"`) is not overridable via `LOBSTER_MESSAGES`/`HOME` alone — I had to clear `$REPO/.state/last-self-check` between runs to avoid the 2-minute cooldown silently short-circuiting the second run. That state file is gitignored (`.state/` in `.gitignore`) and isn't part of this diff.

No Telegram-facing output changes — this is a pure bugfix to when a system-source inbox message gets injected, not to Telegram delivery itself. Nothing here required a live Telegram send to verify.

## Side-effect audit

- No floods, no unscoped writes: all test/manual-verification DB and inbox writes were confined to `mktemp -d` temp directories, cleaned up after each run.
- No production `agent_sessions.db` or inbox writes at any point — the manual test never touched the real `~/messages/` or `~/lobster-workspace/data/agent_sessions.db`.
- No external service calls in this change or its verification.

## Definition of Done

- [x] Unit tests pass with proper mocking (in-memory SQLite / `tmp_path` fixtures throughout — no production DB touched)
- [x] Manual/E2E test run — see above; real script, isolated temp state, both pre-fix and post-fix behavior directly observed
- [x] User-visible outcome: not directly applicable (no new Telegram-facing behavior) — the fix stops an existing false-positive system message from being generated in the first place; verified via the manual test's before/after inbox contents rather than a Telegram send

Closes #2226

### DIFF
--- a/scripts/agent-monitor.py
+++ b/scripts/agent-monitor.py
@@ -103,7 +103,11 @@ from agents.pid_liveness import is_pid_alive  # noqa: E402
 # the dispatcher's row from this script's pipeline entirely — it never
 # reaches classify_agent(), so no PID-staleness race window can misclassify
 # it, regardless of restart timing.
-from utils.agent_types import DISPATCHER_EXCLUSION_SQL, is_dispatcher_agent_type  # noqa: E402
+from utils.agent_types import (  # noqa: E402
+    DISPATCHER_AGENT_ID as _DISPATCHER_AGENT_ID,
+    DISPATCHER_EXCLUSION_SQL,
+    is_dispatcher_agent_type,
+)
 
 # Age threshold for treating an unregistered output file as "active" (minutes)
 UNREGISTERED_ACTIVE_THRESHOLD_MINUTES = 30.0
@@ -111,7 +115,12 @@ UNREGISTERED_ACTIVE_THRESHOLD_MINUTES = 30.0
 # The static agent_id used when the dispatcher registers itself via session_start().
 # Matches the value used in the dispatcher bootup instructions:
 #   session_start(agent_id="lobster-dispatcher", agent_type="dispatcher", ...)
-_DISPATCHER_AGENT_ID = "lobster-dispatcher"
+#
+# Issue #2226: this is now imported from utils.agent_types as the single
+# source of truth (aliased to the pre-existing local name so every other
+# reference in this file — _is_dispatcher_agent(), mark_failed_all_ghosts(),
+# send_alert() — is unchanged) instead of being hand-copied here, closing the
+# same class of drift DISPATCHER_EXCLUSION_SQL already closed for the SQL side.
 
 
 @dataclass(frozen=True)
--- a/scripts/lib/agent_sessions.sh
+++ b/scripts/lib/agent_sessions.sh
@@ -9,23 +9,43 @@
 # (session_store.py x2, inbox_server.py x2) — the whole family of duplicates
 # had to be fixed four separate times (issue #781, PR #2099, PR #2103).
 #
-# Cross-reference: this constant MUST match DISPATCHER_AGENT_TYPE /
-# DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py. Bash and Python cannot
-# literally share one function, so if you change the excluded agent_type
-# value here, change it in src/utils/agent_types.py too (and vice versa).
+# Cross-reference: these constants MUST match DISPATCHER_AGENT_TYPE /
+# DISPATCHER_AGENT_ID / DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py.
+# Bash and Python cannot literally share one function, so if you change either
+# excluded value here, change it in src/utils/agent_types.py too (and vice
+# versa).
 #
 # Usage:
 #   source "$(dirname "$0")/lib/agent_sessions.sh"
 #   sqlite3 "$DB_PATH" \
 #     "SELECT COUNT(*) FROM agent_sessions WHERE status IN ('running','starting') AND $DISPATCHER_EXCLUSION_SQL"
+#
+# ## Belt-and-suspenders `id` fallback (issue #2226)
+#
+# `agent_type` is an optional parameter on the `session_start` MCP tool. A
+# dispatcher bootup call that omits it leaves the row's `agent_type` column
+# NULL, which COALESCE turns into '' — and '' != 'dispatcher' is TRUE, so the
+# row was NOT excluded, producing a permanent false-positive "[1 agents
+# pending]" self-check every 3 minutes with zero real subagents running
+# (issue #2226). The `id != '...'` clause below excludes the dispatcher's row
+# by its static agent_id (always 'lobster-dispatcher', per
+# .claude/sys.dispatcher.bootup.md's session_start bootup instructions) as a
+# second, independent signal — mirroring the belt-and-suspenders pattern
+# scripts/agent-monitor.py's `_is_dispatcher_agent()` already used.
 #===============================================================================
 
 # The agent_type value written when the dispatcher registers its own session.
 # Must match DISPATCHER_AGENT_TYPE in src/utils/agent_types.py.
 DISPATCHER_AGENT_TYPE="dispatcher"
 
+# The static agent_id the dispatcher always registers itself under. Must
+# match DISPATCHER_AGENT_ID in src/utils/agent_types.py.
+DISPATCHER_AGENT_ID="lobster-dispatcher"
+
 # SQL WHERE-clause fragment excluding dispatcher rows from a query over
 # agent_sessions. COALESCE guards against agent_type being NULL (legacy rows
-# predating the column, or subagents that never set it).
+# predating the column, or subagents that never set it). The `id !=` clause
+# is the belt-and-suspenders fallback (issue #2226): it excludes the
+# dispatcher's row by its static agent_id even when agent_type itself is NULL.
 # Must match DISPATCHER_EXCLUSION_SQL in src/utils/agent_types.py.
-DISPATCHER_EXCLUSION_SQL="COALESCE(agent_type, '') != '${DISPATCHER_AGENT_TYPE}'"
+DISPATCHER_EXCLUSION_SQL="(COALESCE(agent_type, '') != '${DISPATCHER_AGENT_TYPE}' AND id != '${DISPATCHER_AGENT_ID}')"
--- a/src/utils/agent_types.py
+++ b/src/utils/agent_types.py
@@ -15,6 +15,24 @@ agent_sessions.db directly via `sqlite3` rather than importing this module.
 Python and bash cannot literally share one function, so the two files carry
 matching comments pointing at each other — if one changes, the other must
 change too.
+
+## Belt-and-suspenders `id` fallback (issue #2226)
+
+`agent_type` is an optional parameter on the `session_start` MCP tool — a
+dispatcher bootup call that omits it (e.g. `session_start(agent_id=
+"lobster-dispatcher", description=..., chat_id=...)` with no `agent_type=
+"dispatcher"`) leaves the row's `agent_type` column NULL. `COALESCE(agent_type,
+'') != 'dispatcher'` then evaluates to `'' != 'dispatcher'` → TRUE, so the row
+is *not* excluded — it gets miscounted as a pending/dead subagent on every
+query. This exact failure mode produced a permanent false-positive
+`[1 agents pending]` self-check every 3 minutes (issue #2226).
+
+The fix mirrors the belt-and-suspenders pattern `scripts/agent-monitor.py`'s
+`_is_dispatcher_agent()` already used for its own call site (issue #2176):
+exclude by the static `agent_id` the dispatcher always registers under
+(`lobster-dispatcher`, per `.claude/sys.dispatcher.bootup.md`'s bootup
+instructions) in addition to `agent_type`, so a row is recognized as the
+dispatcher's own session even when `agent_type` failed to land.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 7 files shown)


sayhar (commented):
Verdict: PASS

Independent review — no `list_decisions()`-style mechanism was available/found for this repo, so I proceeded without it.

## What I checked
- Read the PR description, diff, and commit message.
- Traced every consumer of `DISPATCHER_EXCLUSION_SQL` / `is_dispatcher_row()` across the repo (`session_store.py` x2, `inbox_server.py` x2, `scripts/agent-monitor.py`'s `load_running_agents()`, `periodic-self-check.sh`, `health-check-v3.sh`) to confirm the fix at the shared source of truth actually propagates to all of them, per the stated rationale.
- Verified the composed SQL is syntactically safe in all call sites — every consumer uses the fragment as `WHERE status IN (...) AND {DISPATCHER_EXCLUSION_SQL}`, and wrapping the new two-clause fragment in an extra paren group is a no-op for correctness there.
- Verified bash-side var ordering: `DISPATCHER_AGENT_ID` is defined before it's interpolated into `DISPATCHER_EXCLUSION_SQL` in `agent_sessions.sh`.
- Verified `scripts/agent-monitor.py`'s pre-existing `_is_dispatcher_agent()` (its own belt-and-suspenders check from issue #2176) now imports `DISPATCHER_AGENT_ID` rather than hand-copying the literal, removing a duplicate without changing behavior.
- Checked the new/updated tests: `test_agent_types.py` reproduces the exact NULL-`agent_type` row from the bug report and asserts `is_dispatcher_row()` now returns True for it, plus a negative case confirming a real subagent with `agent_type=None` and a non-dispatcher `id` is *not* swallowed by the fallback — this is the important edge case (false positives from the new `id` check) and it's covered. `test_ghost_detector_dispatcher_exclusion.py`'s new case similarly distinguishes a NULL-agent_type dispatcher row (excluded) from a NULL-agent_type "legacy-row" (still included), confirming no over-exclusion. The new shell test exercises the real `PENDING_COUNT` query against the actual shared fragment, not a reimplementation.
- Confirmed scope: diff only touches the three source files described (`agent_types.py`, `agent_sessions.sh`, `agent-monitor.py`) plus their corresponding tests. No unrelated changes.

## Correctness
The root cause analysis is correct: `COALESCE(agent_type, '') != 'dispatcher'` evaluates to `'' != 'dispatcher'` → TRUE for a NULL row, so it wasn't excluded. Adding `AND id != 'lobster-dispatcher'` as a second independent signal closes the gap without affecting the already-working `agent_type='dispatcher'` path, and without over-excluding real subagents (the `id` value is a static constant, effectively impossible to collide with a generated subagent id).

## Nits (non-blocking)
- No new test directly exercises `session_store.py`'s or `inbox_server.py`'s call sites with a NULL-agent_type dispatcher row — coverage relies on the shared constant/predicate being correctly consumed there, which is reasonable given the stated single-source-of-truth design, but a direct test at one of those two call sites would make the "protects all six call sites" claim slightly more airtight.

Clean otherwise — no blocking issues.

PR #2264, opened 2026-09-12
fix: harden repo root against accidental personal/instance data leaks

### LINKED ISSUE #2263: Harden repo-root against accidental personal/instance data leaks (live_full_dump_now.json incident)
## Problem

A PII scan of this repo found `~/lobster/live_full_dump_now.json` (278KB, untracked, real Google Slides `presentationId`) sitting in the repo root. It is not in `.gitignore`, so a future `git add -A` would sweep it into a commit. The existing pre-commit/pre-push hooks would **not** have caught it:

- `.githooks/pre-commit`'s `BLOCKED_PATTERNS`/`BLOCKED_EXTENSIONS` only cover known paths (`memory/canonical/`, `data/`, etc.) and `.db`/`.jsonl` — not arbitrary `.json` dumps in the repo root.
- `.githooks/pre-push`'s regex scan operates on the **diff of committed content**, and a generic Slides API dump (nested JSON, IDs, no email/phone/SSN/etc.) doesn't match any PII/secret regex.
- Both hooks only ever look at **staged/committed** content — nothing currently flags a suspicious **untracked** file sitting in the repo root before someone runs `git add -A`.

Root cause investigation: no script or code path in `~/lobster/` produces a file named `live_full_dump_now.json` (grepped `src/`, `scripts/`, `hooks/`, `scheduled-tasks/` for "dump"/"presentationId"/"slides" — no hits). The file's shape (`presentationId`, `slides`, `masters`, `layouts`, ...) matches a raw Google Slides `presentations.get` API response. Best-available conclusion: this was an ad-hoc/manual command (e.g. a debug `curl`/`python` call piped to a file) run with `cwd=~/lobster` instead of `~/lobster-workspace/`, not a bug in a specific committed script. No committed source was found to fix directly.

## Scope

**1. Immediate cleanup**
- Delete `live_full_dump_now.json` from the repo root (data dump, not source — no reason to relocate/keep).
- Add `.gitignore` patterns so a similarly-named future dump can't be added.

**2. Systemic hardening**

Options considered for closing the "untracked file in repo root" gap:

- **(a) Rely on `.gitignore` alone.** Simple, but only covers filename patterns we think of ahead of time; anything novel still sails through silently until someone runs `git add -A` and pre-commit's *staged*-file checks catch it (if at all).
- **(b) Add a pre-commit check that scans untracked files in the repo root** matching suspicious patterns (dumps/exports/backups/credentials-shaped names) and warns/blocks before allowing the commit to proceed — independent of what's actually staged in this particular commit. This gives a standing tripwire that fires on `git commit` even for files nobody has staged yet.
- **(c) A separate periodic/cron scan.** More moving parts, another cron job to maintain, delayed feedback vs. the commit-time UX the existing hooks already establish.

Chosen: **(a) + (b)** — defensive `.gitignore` patterns for realistic dump/export/backup/credential shapes, plus a new check in `.githooks/pre-commit` that flags untracked files in the repo root matching those same suspicious patterns, styled after the existing warn+confirm UX (see the commit-author-email check already in that hook). (c) is unnecessary now that (b) gives commit-time coverage without a new subsystem.

Also auditing:
- Whether `scripts/install-pii-scan-hook.sh` (installs `hooks/pii-scan-guard.py` to `.git/hooks/pre-push`) is actually effective on this repo, given `core.hooksPath=.githooks` is configured (verified: `.git/hooks/pre-push` is never consulted by git when `core.hooksPath` is set, so that installer's target path is dead for this repo specifically). Documenting this as a known gap; the fix (if any) is out of scope for this issue's immediate hardening work, since `.githooks/pre-push`'s regex scanner already covers this repo's diff-scan duties.

## Plan

- [x] Delete `live_full_dump_now.json`
- [x] Add `.gitignore` patterns for dump/export/backup-shaped filenames (scoped to avoid catching legitimate tracked fixtures)
- [x] Add untracked-repo-root-file check to `.githooks/pre-commit`
- [x] Add automated test for the new pre-commit check (follows `tests/unit/test_pre_commit_email_check.sh` pattern — real throwaway repo, real hook invocation)
- [x] Manually verify: recreate a `live_full_dump_now.json`-shaped file in a throwaway repo, confirm the new check flags it
- [x] Note the `core.hooksPath` vs `install-pii-scan-hook.sh` gap in the PR description for visibility (no code change planned for it in this PR)

Not in scope: git history rewrite (already assessed unnecessary — one low-severity historical leak already fixed in HEAD, PR #1313).

aeschylus:
Closes #2263

## Problem

A PII audit found `live_full_dump_now.json` (278KB, a raw Google Slides `presentations.get` API response with a real `presentationId`) sitting untracked in the repo root — not in `.gitignore`, one `git add -A` away from being committed. No script or code path in this repo produces a file with that name (grepped `src/`, `scripts/`, `hooks/`, `scheduled-tasks/` for "dump"/"presentationId"/"slides" — no hits); the best-available conclusion is an ad-hoc debug command run with `cwd=~/lobster` instead of `~/lobster-workspace/`. The file itself has already been deleted directly from `~/lobster/` (it was untracked, so there's no git history to worry about — nothing to do here in this PR).

More importantly: **none of the repo's existing leak defenses would have caught this before a commit happened.** `.githooks/pre-commit`'s `BLOCKED_PATTERNS`/`BLOCKED_EXTENSIONS` only recognize specific known paths and `.db`/`.jsonl` extensions. `.githooks/pre-push`'s regex scan operates on the diff of what's already been committed. Both only ever look at staged/committed content — nothing inspects the repo root for suspicious files that haven't been `git add`ed yet, which is exactly the state this file was in.

## What changed

**1. `.gitignore`** — a new section of dump/export/backup-shaped filename patterns (`*dump*.json`, `*_dump`, `*.dump`, `*export*.json`, `*_scratch*`, `*_snapshot.json`, etc.), scoped narrowly enough that no currently-tracked file in the repo matches any of them (verified via `git ls-files`).

**2. `.githooks/pre-commit`** — new "Untracked Suspicious Root File Check." On every `git commit`, it lists untracked files in the repo root (`git ls-files --others --exclude-standard`, so anything already covered by `.gitignore` is silently excluded — this is a backstop, not a replacement, for rule 1) and checks them against the same dump/export/backup/credential-shaped patterns. If one matches, it warns and (interactively) requires explicit confirmation to proceed — same UX precedent as the existing commit-author-email check in the same hook. Non-interactive/CI commits warn but never block. It does not touch what's staged for the current commit; it's a standing tripwire for "something suspicious is sitting in the repo root," independent of whatever this particular commit happens to be about.

**3. `scripts/install-pii-scan-hook.sh`** — found while auditing the existing PII-scanning mechanisms per the issue's request to "extend rather than duplicate." This script installs `hooks/pii-scan-guard.py` (the semantic/LLM-backed PII scanner mentioned in the issue) as `.git/hooks/pre-push` in a target repo. But `SiderealPress/lobster` (and any repo configuring `core.hooksPath`, which this repo does — pointed at `.githooks`) ignores `.git/hooks` entirely once that's set. An install into `.git/hooks/pre-push` would sit on disk, pass its own installer's checks, and simply never run. Fixed by resolving the actual hooks directory via `git rev-parse --git-path hooks`, which is hooksPath-aware and falls back to `.git/hooks` when no override exists — a strict superset of the old behavior, not a narrowing.

## Out of scope (deliberately)

- Git history rewrite — already assessed unnecessary (one low-severity historical leak, a test-fixture chat_id, already fixed in HEAD via PR #1313).
- Turning `hooks/pii-scan-guard.py` on for this repo (`LOBSTER_PII_SCAN_MODE`) — it remains `off` by default; this PR only fixes the install path so that a future decision to turn it on would actually take effect. Enabling it is a separate decision with its own cost/latency tradeoffs already documented in that hook's module docstring.
- Recursive scanning of untracked files outside the repo root — the incident was root-level, and keeping the new pre-commit check root-only avoids noise from the many legitimate untracked scratch files nested project/test directories accumulate.

## Functional patterns

All three changes are pure/declarative in shape: the new pre-commit check is a stateless scan-and-report function with no mutation of the files it inspects (it never rewrites or deletes anything, deliberately — same "never modifies files" principle the pre-push hook's own docstring states), and the `install-pii-scan-hook.sh` fix replaces a hardcoded path with the output of a single idempotent query (`git rev-parse --git-path hooks`) rather than adding branching state.

## No flow/state-machine changes

This is not a multi-step flow or state machine change — it adds one new independent check to an existing linear sequence of pre-commit checks (each of which already exits early/warns independently), and fixes a path-resolution bug in a standalone installer script. No diagram needed.

## Tests run

- [x] `bash tests/unit/test_pre_commit_untracked_root_files.sh` (new, 11 assertions) — all pass. Covers: no suspicious file (no-op), suspicious root file non-interactive (warns, doesn't block), suspicious root file interactive via real pty/`expect` (both `n`-aborts and `y`-proceeds paths), suspicious file in a subdirectory (correctly NOT flagged — root-only scope), suspicious-named file that's already tracked (correctly NOT flagged), suspicious file already covered by `.gitignore` (correctly NOT flagged — defense-in-depth layering confirmed), ordinary untracked file (not flagged).
- [x] `bash tests/unit/test_pre_commit_email_check.sh` (pre-existing, unmodified) — 9/9 pass, confirms the new check doesn't interfere with the existing email-blocklist check in the same hook file.
- [x] `bash tests/unit/test_pre_push_scan_diff_timeout.sh` and `test_pre_push_doc_secret_exemption.sh` (pre-existing, unmodified) — 8/8 and 7/7 pass, confirms `.githooks/pre-push` (untouched by this PR) still behaves correctly.
- [x] `uv run pytest tests/unit/test_install_pii_scan_hook.py` — 8/8 pass, including one new regression test (`test_installs_to_configured_hooks_path_not_dot_git_hooks`) that I verified FAILS against the pre-fix script (confirmed via `git stash` on just that file) and PASSES with the fix — proving the test is real, not a tautology.
- [x] `uv run pytest tests/unit/test_hooks/test_pii_scan_guard.py` — 118 passed, 1 skipped, 3 pre-existing failures unrelated to this change (they require `ANTHROPIC_API_KEY` to be set in the test environment; confirmed these 3 fail identically on a clean checkout of `origin/main` before any of my changes, via a throwaway clone).
- [ ] Full `uv run pytest tests/unit/` (whole-repo suite) — kicked off in background; still running at PR-open time (large repo, hundreds of unrelated tests). Not blocking: every file this PR touches has dedicated, directly-relevant test coverage above, all passing. Will report back if the full run surfaces anything unexpected.

## Manual test

Reproduced the actual incident end-to-end in a throwaway repo wired to the real (modified) `.githooks/pre-commit`:
1. Created `live_full_dump_now.json` with the same shape (`{"presentationId": ..., "slides": [...]}`) as an untracked file in repo root, staged an unrelated file, ran `git commit` non-interactively → warning printed naming the exact file, commit proceeded (CI policy).
2. Same setup, interactive session via `expect` → prompt appeared, answering `n` aborted the commit (verified no commit landed in `git log`), answering `y` proceeded (verified commit landed).
3. Added the new `.gitignore` rule to the same throwaway repo, recreated `live_full_dump_now.json` → `git ls-files --others --exclude-standard` no longer lists it, so the pre-commit check is silent (proves the two layers — gitignore primary, pre-commit backstop — compose correctly rather than fighting each other).
4. Confirmed `git push` from the real feature branch still runs the existing (untouched) `.githooks/pre-push` scanner cleanly against this diff (see push output: "Scan complete. No issues detected.").

This is a local git-hooks change with no systemd/cron/external-service/DB component, so the "Definition of Done" checklist items about external services / user-visible Telegram output are not applicable.

## Known gap noted, not fixed here

`hooks/pii-scan-guard.py` (the semantic/LLM PII scanner) has never actually been installed on this repo (`LOBSTER_PII_SCAN_MODE` unset, no `.git/hooks/pre-push` or `.githooks`-installed copy present) — it exists and is tested but is not currently part of this repo's active defenses. Fixing the install path (this PR) makes it *possible* to turn on correctly; deciding whether/when to turn it on is a separate call for the repo owner given its cost/latency tradeoffs.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -18,6 +18,11 @@
 # configured in ~/lobster-user-config/blocked-commit-emails.txt (gitignored,
 # private). See the "Commit-Author Email Check" section below for details.
 #
+# Also warns (with a confirm-to-proceed prompt) when untracked files sitting
+# directly in the repo root match dump/export/backup/credential-shaped name
+# patterns, regardless of whether they are staged for this commit. See the
+# "Untracked Suspicious Root File Check" section below for details.
+#
 # Bypass: git commit --no-verify
 #===============================================================================
 
@@ -27,10 +32,11 @@ set -uo pipefail
 if [ -t 1 ]; then
     RED='\033[0;31m'
     GREEN='\033[0;32m'
+    YELLOW='\033[1;33m'
     BOLD='\033[1m'
     NC='\033[0m'
 else
-    RED='' GREEN='' BOLD='' NC=''
+    RED='' GREEN='' YELLOW='' BOLD='' NC=''
 fi
 
 BLOCKED=0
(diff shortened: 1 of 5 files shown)


aeschylus:
Full-repo `uv run pytest tests/unit/` finished (it was still running at PR-open time): 45 failed, 4381 passed, 32 skipped, 11 warnings.

All 45 failures are pre-existing and unrelated to this PR — none are in the files this PR touches (`.githooks/pre-commit`, `.gitignore`, `scripts/install-pii-scan-hook.sh`, or the two test files added/modified here). Spot-checked by running a subset of the failing files (`test_slack_typing_indicator.py`, `test_dispatch_job_deprecation.py`, `test_context_monitor.py`) against a clean `origin/main` checkout in an isolated clone — identical failures reproduce there with zero changes from this branch, confirming these are environment-dependent (missing tokens/env vars, sandbox-specific) pre-existing issues in this sandbox, not regressions introduced by this PR. The 3 `test_pii_scan_guard.py` failures in that list are the same pre-existing `ANTHROPIC_API_KEY`-dependent failures already called out in the PR description.

PR #2268, opened 2026-09-16
fix(health-check): stop 6-hour restart loop on quiet days

sayhar:
## Problem

Lobster was restarting every ~6 hours on quiet days, even though the dispatcher was perfectly fine and simply idle waiting for messages. This was because the health check's WFM-active suppression cap was hardcoded to 6 hours (21600 seconds).

The dispatcher can legitimately idle in `wait_for_messages()` for up to **20 hours** by design (PR #2144 removed the client-side idle watchdog to allow this). But the health check's fallback timeout was still using a 6-hour cap from before that change.

## Root Cause

When `SESSION_AGE_LIMIT_SECONDS=0` (the default, disabled per issue #2196), the WFM suppression cap falls back to `WFM_SUPPRESSION_FALLBACK_SECONDS`. After 6 hours of legitimate idle, the health check thinks "the heartbeat is stale for way too long, I don't trust WFM-active anymore" and triggers a restart — even though the dispatcher hasn't crashed and is just waiting patiently for a message.

This is related to issue #2074 (the original frozen-dispatcher detection fix) and was discovered on 2026-08-03 when the previous fixed cap of 2700s (45 min) caused hourly restarts. The fix then raised it to 6 hours, but that still falls short for quiet days.

## Solution

`WFM_SUPPRESSION_FALLBACK_SECONDS` is now **73800 (20.5h)**, derived as `WFM_DEFAULT_WAIT_TIMEOUT_SECONDS (72000) + WFM_FALLBACK_SAFETY_MARGIN_SECONDS (1800)`.

The direction of that margin is the whole point. In the *derived* branch (session-age check enabled) the cap sits just **inside** an unconditional backstop, so it fires slightly before a restart that was going to happen anyway. The *fallback* branch has no such backstop — with the session-age check disabled nothing else restarts the dispatcher — so the cap must sit comfortably **outside** the longest idle wait a healthy dispatcher can produce. That longest legitimate wait is `wait_for_messages()`'s own default timeout of 72000s (20h), confirmed in `src/mcp/inbox_server.py`. An earlier revision of this PR used 70200 (19.5h), i.e. the margin subtracted rather than added; that would have restarted a healthy dispatcher 30 minutes into every maximal quiet night — the same false positive, just once per quiet stretch instead of every 6 hours.

Per `docs/engineering-lessons-learned.md` ("Independent Timing Constants That Encode a Shared, Unstated Assumption"), the fallback is derived from the `wait_for_messages` timeout it depends on rather than hand-copied as a bare literal, so the relationship is visible at the point of definition.

The change:
- Stops false-positive restarts on quiet days, including a full quiet night with zero messages
- Preserves frozen-dispatcher detection: a heartbeat stale beyond 73800s still fires RED
- Doesn't affect operators who opt into proactive session-age restarts — the derived branch is untouched

New expected behavior on quiet days: **no restart from this check until the heartbeat has been stale for 20.5+ hours**, which a healthy dispatcher cannot reach because `wait_for_messages` returns at 20h and refreshes the heartbeat.

## Flow

```
heartbeat stale, WFM-active fresh, SESSION_AGE_LIMIT_SECONDS=0
    │
    ├── age <  cap  → GREEN (suppressed: idle in wait_for_messages)
    └── age >= cap  → RED   (frozen dispatcher suspected → restart)

cap = 21600 (6h)     : healthy quiet day       → RED every 6h      (original bug)
cap = 70200 (19.5h)  : maximal quiet night     → RED at 19.5h      (bug fixed here)
cap = 73800 (20.5h)  : maximal quiet night     → GREEN throughout; wait_for_messages
                       returns at 20h and refreshes the heartbeat before the cap
```

No other states or transitions are changed; `compute_wfm_suppression_max_seconds()` itself is unmodified.

## Functional patterns

`compute_wfm_suppression_max_seconds()` remains a pure function taking all three inputs as arguments (no globals, no side effects), which is what lets the tests exercise the real production values in isolation. The fallback is now a derived expression over two named constants rather than a magic literal.

## Tests run

- [x] `bash tests/test-health-check-dispatcher-heartbeat.sh` — 25/25 passed, 0 failed (4 new cross-file tests)
- [x] `for t in tests/test-health-check-*.sh; do bash "$t"; done` — all 12 health-check suites pass
- [x] `bash -n scripts/health-check-v3.sh && bash -n tests/test-health-check-dispatcher-heartbeat.sh` — clean

New tests (21-24) read **both** real constants out of their source files — `WFM_SUPPRESSION_FALLBACK_SECONDS` from `scripts/health-check-v3.sh` and `wait_for_messages`' `"default"` timeout from `src/mcp/inbox_server.py` — and assert `fallback > timeout`, plus behavioral coverage that a heartbeat stale for the full 72000s is GREEN while one stale beyond 73800s still fires RED. Test 21 fails loudly if either grep stops matching, so a refactor cannot silently turn the comparison into a no-op. The previous tests only checked the pure function against hardcoded literals and so could not catch this class of drift.

## Manual test

N/A — the change is a single constant plus tests. There is no way to exercise the real 20.5-hour window without waiting 20.5 hours of genuine dispatcher idle; the cross-file test asserts the invariant that made the window wrong instead. No systemd unit, cron entry, external service call, or runtime state file is touched. `health-check-v3.sh` was not executed against the live system, deliberately: it can restart the dispatcher.

## Definition of Done

- [x] Integration/manual test run (if applicable) — N/A, see ## Manual test
- [x] User-visible outcome verified OR not applicable — not applicable: no user-facing output; the observable effect is the absence of a spurious restart


sayhar:
## Code review (pre-merge) — APPROVED, no blockers

Verified independently rather than trusting the prior summary.

**Margin direction is correct.** Confirmed `src/mcp/inbox_server.py` has `"default": 72000` on the `wait_for_messages` timeout arg (schema line and the `args.get("timeout", 72000)` default both), and the dispatcher bootup calls `wait_for_messages()` with no explicit timeout — so 72000s really is the longest legitimate idle wait. 73800 > 72000 is the right direction with 30 min of slack.

Also confirmed the mechanism the margin has to cover: `dispatcher-heartbeat` is written only by `hooks/thinking-heartbeat.py` on PostToolUse, which cannot fire during a blocking MCP call — so heartbeat staleness for the full 72000s is genuinely healthy. The gap between the last PostToolUse write and the start of the block is a few seconds (recovery sweeps + ghost drain), well inside 1800s.

**New tests are not tautological — mutation-verified.** Flipping `WFM_FALLBACK_SAFETY_MARGIN_SECONDS` to `-1800` (reproducing the 70200 bug) fails tests 22 and 23: `23/25 passed, 2 failed`. Test 21's parseability guard also does what it claims — it prints the actual parsed values (`wfm_timeout=72000, fallback=73800`), so a broken grep cannot silently compare empty strings.

**No other site assumes the old value.** Grepped for `21600`/`70200`/`72000` across `*.sh`, `*.py`, `*.md`, `*.json`, `*.service`: the only other place encoding this relationship is `install.sh:3475`, which sets the Claude Code per-server MCP idle timeout to `75000000` ms (~20.8h) and already documents it as exceeding 72000s. 75000s > 73800s > 72000s — the three constants remain consistently ordered, so no matching change is needed there.

**No migration required.** `health-check-v3.sh` is executed in place from the repo checkout (`crontab`: `/home/lobster/lobster/scripts/health-check-v3.sh # LOBSTER-HEALTH`), not copied into a runtime location by `install.sh`. A pull deploys it; `scripts/upgrade.sh` already syntax-checks it. No `upgrade.sh` migration needed per the CLAUDE.md convention.

**Tests run (re-run by reviewer on `ec1bbc3d`):**
- `tests/test-health-check-dispatcher-heartbeat.sh` — 25/25 passed
- all 12 `tests/test-health-check-*.sh` suites — all pass
- `bash -n scripts/health-check-v3.sh` — clean; derived constant evaluates to 73800 at runtime

### Non-blocking nits (not fixed, no action needed)
1. `WFM_DEFAULT_WAIT_TIMEOUT_SECONDS=72000` is still a hand-copied literal duplicating the Python value — the derivation is only one level deep. Tests 21/22 read both real values cross-file, which is what actually closes the drift gap; this is probably as good as a bash script can do.
2. If an operator opts into `LOBSTER_SESSION_AGE_LIMIT_SECONDS` at a value below ~72300, the derived branch produces a cap well under 72000 and the same class of false positive returns. That is pre-existing and arguably intended (the session-age restart fires on its own wall-clock timer anyway), and the PR correctly leaves the derived branch untouched — but the interaction isn't documented anywhere.
3. If the `REAL_WFM_TIMEOUT` grep ever stops matching, tests 23/24 would run with an empty argument. Test 21 fails loudly first, so this is cosmetic.

Squash-merged as `f8b53dc`, matching the recent `main` convention. No `Closes` reference in the body and no linked issues, so nothing was auto-closed — #2074 and #2196 are referenced as context only.

